PR #4, opened 2019-04-04
Use DNF for managing RPM content

Conan-Kudo:
This pull request includes an RFC to move from Zypper to DNF for managing RPM content in Uyuni.

To summarize, this is being proposed for two major reasons:

* Restoring support for running Uyuni in a supportable way on RHEL/CentOS 7+ and Fedora
* Enabling proper support for RH/Fedora modules (which is core to RHEL 8 and increasingly important in Fedora).

I'm also willing to help with implementation as I can.

This relates to https://github.com/uyuni-project/uyuni/issues/804
This relates to https://github.com/uyuni-project/uyuni/issues/805

(cc: @dmach @MichaelMraka)

### DIFF
--- a/accepted/00054-use-dnf-for-rpm-repos.md
+++ b/accepted/00054-use-dnf-for-rpm-repos.md
@@ -1,0 +1,187 @@
+- Feature Name: Use DNF for managing the RPM content
+- Start Date: 04-04-2019
+- RFC PR: https://github.com/uyuni-project/uyuni-rfc/pull/4
+
+# Summary
+[summary]: #summary
+
+Uyuni contains components that support importing external repositories;
+collecting and retrieving information on packages, products, modules, and other content;
+and hosting repositories to be consumed by Spacewalk clients. For RPM repositories,
+it currently uses Zypper and libsolv (and previously used YUM), per
+[a previously implemented RFC](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00047-yum-replacement.md).
+
+However, this makes it difficult to run Uyuni on Red Hat Enterprise Linux and CentOS servers
+without replacing base packages (violating subscription support terms in the RHEL case),
+and makes it even more challenging to properly support newer extensions to RPM repositories,
+such as RH/Fedora modules introduced in Fedora 30 and RHEL 8.
+
+This RFC proposes to redo the ["YUM replacement" RFC](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00047-yum-replacement.md)
+to add another implementation using DNF instead of Zypper.
+
+# Motivation
+[motivation]: #motivation
+
+The work done to switch to Zypper broke the ability for the Uyuni server to run properly
+on CentOS/RHEL servers because Zypper is unavailable.
+
+Previously, Uyuni used YUM (through the `yum` Python library module) for these pieces,
+which is available on CentOS/RHEL 7. However, in order to move to Python 3, Uyuni cannot
+use `yum` anymore.
+
+For expediency for the downstream SUSE Manager product delivery schedule, Uyuni is using
+Zypper for now. However, the intent for Uyuni is to be an open upstream project that
+supports more host environments than SUSE Manager does.
+
+Moreover, neither YUM nor Zypper have any reasonable awareness of how to properly
+handle [RH/Fedora modules](https://docs.pagure.org/modularity/). It's currently
+only implemented in DNF. Using DNF opens the door to developing features specific
+to handling modules in the future.
+
+As of today, DNF is available across RHEL/CentOS 7 and newer, Fedora, and
+(open)SUSE Linux 15.0 and newer (though PackageHub for SLE 15). Starting with
+openSUSE Leap 15.1, DNF supports RH/Fedora modules. So migrating to DNF will
+restore the capability to run Uyuni on CentOS and Fedora servers, while
+preserving the ability to run on openSUSE and SUSE Linux Enterprise 15.
+
+# Detailed design
+[design]: #detailed-design
+
+The below design information is derived from detailed design information from the "YUM replacement" RFC:
+
+### The `spacewalk-repo-sync` architecture
+
+![spacewalk-repo-sync diagram](images/reposync-diagram.png)
+
+### The purpose of a "ContentSource" plugin:
+
+Since there are multiple types of content sources (i.a. repositories types),
+`spacewalk-repo-sync` delegates the handling of each type to plugins. These "ContentSource" plugins must:
+
+- Get repository information and metadata from a repository (ContentSource).
+- Provide a list of packages, products and errata coming from that ContentSource. (ContentPackage)
+- Provide contact & download parameters to the Downloader Thread for downloading the packages.
+- Resolve package dependencies. Since `spacewalk-repo-sync` allows filtering (include/exclude) packages when syncing, package dependencies need to be calculated when running on this mode.
+
+With this information collected by the plugin, the `DownloaderThread` from `reposync`
+is able to perform the actual download of the necessary package files when `spacewalk-repo-sync` is called.
+
+Currently, `spacewalk-repo-sync` comes with the following plugins:
+- `yum_src`: RPM based repositories
+- `deb_src`: DEB based repositories
+- `uln_src`: Oracle ULN repositories (a wrapper of `yum_src` with custom authentication)
+
+### The "ContentSource" plugin API
+
+Mandatory methods:
+- `get_md_checksum_type`: Return the checksum_type of primary.xml
+- `get_products`: Return products metadata if any
+- `get_susedata`: Return suse metadata if any
+- `get_updates`: Return "updateinfo" / "patches" info
+- `get_groups`: Return repo groups if any
+- `get_file`:
+- `get_modules`: Return module metadata if any
+- `raw_list_packages`: Return the raw list of packages available after filtering.
+- `list_packages`: Return a list of `ContentPackage` rhn class for the available packages after filtering.
+- `clear_cache`: Clear cache files
+- `set_download_parameters`: Prepare the downloader params dictionary
+
+Yum specific methods we also need to cover:
+- `repomd_up_to_date`: Check if repomd.xml checksum is up-to-date.
+- `get_metadata_paths`: Simply load primary and updateinfo path from repomd.
+
+#### The "ContentPackage" class
+A "ContentSource" plugin needs to provide a list of "ContentPackage" objects to the reposync. This RHN class represents the package like by:
+
+- NEVRA
+- unique ID
+- checksum
+- checksum type
+
+### Adding a "DNF" plugin like the custom "Zypper" + "libsolv" plugin.
+
+The idea here would be the following:
+
+- The repository metadata downloading is delegated to DNF (which uses [librepo](http://rpm-software-management.github.io/librepo/) for the actual fetching)
+- The downloaded XML metadata can be manually read (packages, updates, products, etc).
+- Authentication with ULN repositories can be done by porting the YUM plugin to DNF
+- Legacy metadata not normally handled by DNF could be fetched by [librepo](http://rpm-software-management.github.io/librepo/) or urlgrabber through a plugin module like the ULN one
+
+The usage of this plugin will be gated by a conditional in the `spacewalk-backend`
+package spec file to use Zypper on SUSE Linux Enterprise to ensure that the
+SUSE Manager downstream product continues to use Zypper.
+
+#### How to calculate the dependencies?
+To access this information, we would use [the DNF API](https://dnf.readthedocs.io/en/latest/api.html) (OBS packages at [openSUSE:Factory](https://build.opensuse.org/package/show/openSUSE:Factory/dnf) - [openSUSE:Backports:SLE-15-SP1](https://build.opensuse.org/package/show/openSUSE:Backports:SLE-15-SP1/dnf)).
+
+#### Requirements for this approach:
+- We would need to use `python3-dnf` library ([OBS link](https://build.opensuse.org/package/show/openSUSE:Backports:SLE-15-SP1/dnf)) to do this.
+- For fetching metadata not referenced by `repomd.xml`, we could use the `python3-librepo` library ([OBS link](https://build.opensuse.org/package/show/openSUSE:Backports:SLE-15-SP1/librepo)).
+  + Alternatively, the `python3-urlgrabber` library ([OBS link](https://build.opensuse.org/package/show/openSUSE:Factory/python-urlgrabber)) could be used.
+
+#### DNF Python code examples
+- Some DNF use cases from Python: https://dnf.readthedocs.io/en/latest/use_cases.html
+- Working with DNF: http://abregman.com/2016/11/29/python-working-with-rpms/
+- DNF `reposync` subcommand code: https://github.com/rpm-software-management/dnf-plugins-core/blob/master/plugins/reposync.py
+  + This implementation will download all metadata referenced by `repomd.xml`, even stuff not currently handled by DNF itself.
+- Example of resolving dependencies in a repo with DNF: https://pagure.io/releng/blob/master/f/scripts/spam-o-matic
+- Some Python examples with librepo: https://github.com/rpm-software-management/librepo/tree/master/examples/python
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The main drawback of this would be imposing the DNF stack as a dependency for the server.
+
+As it currently stands, this is not much of a drawback, as we are currently going to keep using
+the Zypper plugin for the SUSE Manager product and with SUSE Linux Enterprise for Uyuni.
+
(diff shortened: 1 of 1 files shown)


accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,175 @@
...
+#### The "ContentPackage" class
+A "ContentSource" plugin needs to provide a list of "ContentPackage" objects to the reposync. This RHN class represents the package like by:
+
+- NEVRA
+- unique ID
+- checksum
+- checksum type
+
+### Replacing the custom "Zypper" + "libsolv" plugin with a "DNF" plugin.
+
```
mcalmer: What about implementing a dnf module in parallel to yum.
Is there a reason why we would need to replace the existing one?

Conan-Kudo: I would imagine that for simplicity, it's easier to have one component that manages the rpm source stuff. It also makes the behaviors of the server consistent across whatever it is hosted on.

Maybe it makes sense to have an extra module for fetching legacy metadata (SLE 10 and older?), but it might be easy enough to handle everything through DNF.

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,175 @@
...
+Finally, Zypper does not implement support for handling RH/Fedora modules, so it will be very hard
+to properly support handling RH/Fedora modules without using DNF anyway (most of the sorting, filtering,
+and solution management logic for modules is implemented in DNF itself right now). This becomes a
+major issue with Red Hat Enterprise Linux 8 as managing it does not work in large part due to this.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Will we need to extend libdnf/DNF API to pull arbitrary repodata types mentioned in `repomd.xml`?
```
Conan-Kudo: We don't need to do anything here. Anything referenced in `repomd.xml` will be fetched per the `dnf reposync` code listed in example.

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,175 @@
...
+- Would we need to extend the librepo API to pull extra or legacy repodata information?
```
Conan-Kudo: If we need to pull stuff not referenced in `repomd.xml`, we probably need to add it to librepo unless we want to just call urlgrabber to fetch it manually...

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,177 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+The main drawback of this would be imposing the DNF stack as a dependency for the server.
+This is only slightly an issue in SUSE Linux Enterprise, as it's maintained in PackageHub.
+For the downstream SUSE Manager product, it is likely that the DNF stack packages would need
+to be imported to be maintained exclusively for being used with SUSE Manager.
+
+However, this is unlikely to be much of a maintenance burden as the community packaging
+is actively maintained and well-structured. The packager (who is also the author of this RFC)
```
brejoc: That is indeed a very big plus in my eyes!

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,177 @@
...
+
+Moreover, [YUM has been retired in Fedora 31](https://fedoraproject.org/wiki/Changes/Retire_YUM_3)
+and "YUM" in Red Hat Enterprise Linux 8 is just a CLI shim to DNF, and only the DNF Python interface
+and libraries are available. Thus, reverting to the "YUM" version would not work.
+
+## Package and port Zypper to other distributions for Uyuni
+
+Another alternative is to package Zypper for distributions that lack it. But this is not appropriate
+for CentOS/RHEL as doing so requires replacing and upgrading packages in the distribution base to
+build it. Notably, the `libsolv` package is too old in CentOS/RHEL, and even if it was not, it is
```
brejoc: Ah, okay. I think this answers my question from above. So libsolv is the main problem here.

accepted/00000-use-dnf-for-rpm-repos.md
brejoc: Having one code path that fails would indeed be easier. 😉  I know that there was a special case for <SLE10, but I don't know if this would/could be handled by DNF. @mcalmer, I think you mentioned that somewhere. Do you still remember what that was?

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,177 @@
...
+is actively maintained and well-structured. The packager (who is also the author of this RFC)
+is responsive to Uyuni/SUSE Manager needs and the upstream DNF developers are interested
+and responsive to community consumers' needs.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Revert back to using YUM
+
+One alternative to resolve this issue is to roll back to using YUM. However, this is a bad idea.
```
brejoc: This is indeed something we can't do! Everything > openSUSE/SLE15 is Python3, so we can't revert to a Python2 solution.

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,177 @@
...
+[summary]: #summary
+
+Uyuni contains components that support importing external repositories;
+collecting and retrieving information on packages, products, modules, and other content;
+and hosting repositories to be consumed by Spacewalk clients. For RPM repositories,
+it currently uses Zypper and libsolv (and previously used YUM), per
+[a previously implemented RFC](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00047-yum-replacement.md).
+
+However, this makes it difficult to run Uyuni on Red Hat Enterprise Linux and CentOS servers
+without replacing base packages (violating subscription support terms in the RHEL case),
```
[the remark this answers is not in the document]
Conan-Kudo: Yeah. The actual problem is that zypper requires enabling HELIXREPO and SUSEREPO libsolv features. They're not enabled in RHEL/CentOS and they don't want to enable it (I've asked as the Fedora zypper maintainer before).

Zypper team has no interest in fixing it so that those aren't required, since the former is used for the solver testcases, and the latter is the legacy (<SLE10) YaST2 metadata stuff.

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,177 @@
...
+such as RH/Fedora modules introduced in Fedora 30 and RHEL 8.
+
+This RFC proposes to redo the ["YUM replacement" RFC](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00047-yum-replacement.md)
+and select to implement it by using DNF instead of Zypper.
+
+# Motivation
+[motivation]: #motivation
+
+The work done to switch to Zypper broke the ability for the Uyuni server to run properly
+on CentOS/RHEL servers because Zypper is unavailable.
```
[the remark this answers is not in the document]
Conan-Kudo: It clashes with the subscription support things because we need to overwrite base packages to make it work, as mentioned earlier.

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,177 @@
...
+use `yum` anymore.
+
+For expediency for the downstream SUSE Manager product delivery schedule, Uyuni is using
+Zypper for now. However, the intent for Uyuni is to be an open upstream project that
+supports more host environments than SUSE Manager does.
+
+Moreover, neither YUM nor Zypper have any reasonable awareness of how to properly
+handle [RH/Fedora modules](https://docs.pagure.org/modularity/). It's currently
+only implemented in DNF. Using DNF opens the door to developing features specific
+to handling modules in the future.
```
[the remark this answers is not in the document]
Conan-Kudo: Transactional servers are not special. The `transactional-update` mechanism is merely a shell script that runs zypper commands on top. If you _really_ wanted me to write a port of _that_ interface for DNF, I could. But from the Uyuni point of view, it doesn't matter, since there's nothing that it cares about beyond orchestrating command runs and reboots, which is not part of this feature.

Conan-Kudo:
> Will the modularity concept fully be handled in DNF or will there be something on top? Then we've got the transactional server on the openSUSE side. 

@brejoc It is intended that DNF (and libdnf beneath it) will fully handle all the modularity interactions. It is fully integrated into the DNF CLI and API interfaces. @dmach's team has been working on making it a seamless part of the integration from the beginning.

accepted/00000-use-dnf-for-rpm-repos.md
Conan-Kudo: @brejoc librepo has a stub for dealing with legacy susetags data (< SLE 10) which could be filled out, but maybe some custom urlgrabber code is sufficient here?

Conan-Kudo:
>  Then we've got the transactional server on the openSUSE side. 

From the Uyuni point of view, there's nothing special about transactional-server. You query the rpmdb the way you always have, the only real change is that the clients need to shell out to `transactional-update` and immediately force a reboot afterwards. This feature has no interaction with that.

Conan-Kudo:
@brejoc Also, is there really a reason to support SLE 10 metadata anymore? SUSE Linux Enterprise 10 has been fully dead since 2016 (no LTSS support even!).

mbrookhuis:
What happens when customers have still SLES10 channels in the current SUSE Manager and that needs to be migrated?

Conan-Kudo:
@mbrookhuis Well, Zypper in SLE 10 does support rpm-md repositories, so we could just generate those and have it point to that...

But otherwise, I guess they'd be blocked and no longer be able to sync from SCC if it relied on legacy metadata formats...

SLE10 isn't receiving any updates anyway...

mcalmer:
@Conan-Kudo - this is not correct. For SUSE Manager we shipped a new zypp stack for SLE10 so it can read and understand repomd.
To keep SLE10 support in SUSE Manager is a business decission of SUSE.

I think it is possible to find a way to implement new things in uyuni without making us the life to complicated to keep legacy code running for SUSE Manager.

Conan-Kudo:
@mcalmer If SLE 10 zypp stack supports repomd, then there's no issues. Everything should Just Work(TM) with what I'm proposing.

Syncing through the legacy repodata shouldn't be terribly difficult if needed. We can either fetch it through urlgrabber (as we did before), or implement support for it in librepo so that we can use librepo's Python API to fetch it.

mcalmer:
I am not against a new module which use a different lib. I am also open to make it the default module sooner or later.
My only request is, to develop it next to the existing module. Then everybody can chose and switch to what he like most or what is working best.

Conan-Kudo:
@mcalmer I think that it makes sense to support SLE10 legacy ones as an auxiliary module, similar to how ULN support is done. That way it's optional and unneeded in most deployments, and can be activated for those that need it (which can be detected and the user can be requested to install it, or it can be installed by default in SUSE Manager deployments).

accepted/00000-use-dnf-for-rpm-repos.md
meaksh: I have the same feeling than @mcalmer here. Why not to implement this `dnf` plugin as an alternative plugin for managing RPM repositories?

Even if `dnf` is already packaged in PackageHub, bringing it as the default plugin for RPM means we would need also to support the DNF stack in our enterprise products.

Therefore, implementing this `dnf_src` plugin not as a replacement but as an alternative would be more interesting for us.

I'll elaborate this more in the main review comment.

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,177 @@
...
+- The repository metadata downloading is delegated to DNF (which uses [librepo](http://rpm-software-management.github.io/librepo/) for the actual fetching)
+- The downloaded XML metadata can be manually read (packages, updates, products, etc).
+- Authentication with ULN repositories can be done by porting the YUM plugin to DNF
+- Legacy metadata not normally handled by DNF could be fetched by [librepo](http://rpm-software-management.github.io/librepo/) or urlgrabber through a plugin module like the ULN one
+
+#### How to calculate the dependencies?
+To access this information, we would use [the DNF API](https://dnf.readthedocs.io/en/latest/api.html) (OBS packages at [openSUSE:Factory](https://build.opensuse.org/package/show/openSUSE:Factory/dnf) - [openSUSE:Backports:SLE-15-SP1](https://build.opensuse.org/package/show/openSUSE:Backports:SLE-15-SP1/dnf)).
+
+#### Requirements for this approach:
+- We would need to use `python3-dnf` library ([OBS link](https://build.opensuse.org/package/show/openSUSE:Backports:SLE-15-SP1/dnf)) to do this.
```
meaksh: At least at this particular moment, we wouldn't like to add a new stack to maintain and support in our Enterprise products since we already have an alternative based on supported components.

accepted/00000-use-dnf-for-rpm-repos.md
meaksh: Of course, switching back to yum makes no sense here.

accepted/00000-use-dnf-for-rpm-repos.md
meaksh: Ok, it's clear that bringing Zypper to CentOS/RHEL is not possible at the moment without breaking terms of supportability on the RHEL side

accepted/00000-use-dnf-for-rpm-repos.md
```diff
@@ -0,0 +1,177 @@
...
+Uyuni contains components that support importing external repositories;
+collecting and retrieving information on packages, products, modules, and other content;
+and hosting repositories to be consumed by Spacewalk clients. For RPM repositories,
+it currently uses Zypper and libsolv (and previously used YUM), per
+[a previously implemented RFC](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00047-yum-replacement.md).
+
+However, this makes it difficult to run Uyuni on Red Hat Enterprise Linux and CentOS servers
+without replacing base packages (violating subscription support terms in the RHEL case),
+and makes it even more challenging to properly support newer extensions to RPM repositories,
+such as RH/Fedora modules introduced in Fedora 30 and RHEL 8.
```
meaksh: Supporting and managing the new app streams from RH/Fedora apparently requires more changes on the codebase apart from the reposync plugin. So this is something we will be investigating outside of the context of the reposync plugins.

The downloading part and metadata reading is also feasible to do with our current zypper plugin.

Conan-Kudo:
> I agree with you that this adds more complexity but having both plugins would allow us to not necessary support the DNF stack in our Enterprise products and at the same time would allow Uyuni to run on RHEL/CentOS and Fedora. Of course, this may change in the future, and we could adopt the "dnf" plugin as the default if needed.

@meaksh, based on the options here, would you be okay with a spec file based control to trigger which plugin is installed? That's the simplest for implementation and testability, and at this point is more or less matching the XOR relationship we're going to have for now.

> In any case, also note that classes like `UpdateNotice` and others that are implemented on the current Zypper-based plugin can be reused in this new DNF module, so it would be great to share them properly instead of duplicating code.

If it looks like I can make it share code, I can try.

> We want to give us some time using the new Zypper plugin, also because of the Enterprise implications, so taking the decision of implementing this as an alternative module seems to be the more convenient way to proceed here at this point. Again, as already said, maybe this changes on the future and we would finally adopt the new DNF as the default for both Uyuni and SUSE Manager.

I don't understand this argument, is this over the legacy susetags/yastmd reposyncing?

meaksh:
> @meaksh, based on the options here, would you be okay with a spec file based control to trigger which plugin is installed? That's the simplest for implementation and testability, and at this point is more or less matching the XOR relationship we're going to have for now.

Yes, I would agree on selecting the default plugin in the spec file :+1:

> > We want to give us some time using the new Zypper plugin, also because of the Enterprise implications, so taking the decision of implementing this as an alternative module seems to be the more convenient way to proceed here at this point. Again, as already said, maybe this changes on the future and we would finally adopt the new DNF as the default for both Uyuni and SUSE Manager.
> 
> I don't understand this argument, is this over the legacy susetags/yastmd reposyncing?

Nothing special here. We are currently working on SUSE Manager 4.0 GMC release which is currently based on the Zypper plugin and without the DNF stack. We don't think switching to DNF here is feasible at the moment on the Enterprise context (also because legal implications). Of course, we still support old SLE10 repos, susetags, yastmd on the Enterprise product, and that should also work in the future at least from now.

Conan-Kudo:
@meaksh I've implemented the wording changes to indicate that both will remain in parallel and how they'll be selected.

Conan-Kudo:
@meaksh Shouldn't I get a number for the file? It shouldn't be `00000` before merging...

meaksh:
@Conan-Kudo - Yes, that's the way to proceed :+1:

That's the last step on https://github.com/uyuni-project/uyuni-rfc#tldr

Conan-Kudo:
@meaksh I've set the number now.

PR #5, opened 2019-04-29
New package predownloading (staging) iteration

moio:
View: https://github.com/uyuni-project/uyuni-rfc/blob/new-package-predownloading/accepted/00052-new-package-predownloading.md
Comment: https://github.com/uyuni-project/uyuni-rfc/pull/5/files


moio:
@joesusecom, @mcalmer: I'd like to have your input about the general approach here.

@mbologna: your feedback would be welcome as the co-author of the original feature.

@ncounter: please forgive me for the brutality of the UI part here. Would love to find a nicer way to introduce the chain taking the chance to make this look and feel a bit better, input welcome.

@dincamihai, @aaannz: mainly FYI, although of course input is also welcome.

joesusecom:
I'm sure this approach can be useful for _some_ customers. Especially as we make sure that the option is also available from the API and thus can be automated. After all, it technically allows for a workflow that makes sure the packages are downloaded first.
But I really had something more automated in mind, where the download and the installation can be scheduled in the same step.
That said, for now I see no downside of implementing this first and letting customers give us more feedback later.

moio:
> But I really had something more automated in mind, where the download and the installation can be scheduled in the same step.

@joesusecom that would be middle-way when it comes to implementation difficulty: UI-wise we would have a textbox to enter a number of advance hours instead of a checkbox, API-wise it's an extra number instead of a boolean.

At Action scheduling time in the backend, we would have to schedule two Actions instead of just one.

I think this could add some 50% in development time, it's probably not a crazy amount of extra complexity. I can change the RFC in that sense if you feel like investing a bit more here (changing this later will be more complex because we would have to support both schemes).

What is your thought about that?

accepted/00052-new-package-predownloading.md
```diff
@@ -0,0 +1,85 @@
...
+  - [pkgdownload.sls](https://github.com/uyuni-project/uyuni/blob/c8ffe6b9425392f5235864ad070646bb8ebc2ecb/susemanager-utils/susemanager-sls/salt/packages/pkgdownload.sls) instead of [pkginstall.sls](https://github.com/uyuni-project/uyuni/blob/c8ffe6b9425392f5235864ad070646bb8ebc2ecb/susemanager-utils/susemanager-sls/salt/packages/pkginstall.sls) if set
+  - [patchdownload.sls](https://github.com/uyuni-project/uyuni/blob/c8ffe6b9425392f5235864ad070646bb8ebc2ecb/susemanager-utils/susemanager-sls/salt/packages/patchdownload.sls) instead of [patchinstall.sls](https://github.com/uyuni-project/uyuni/blob/c8ffe6b9425392f5235864ad070646bb8ebc2ecb/susemanager-utils/susemanager-sls/salt/packages/patchinstall.sls) if set
+  - alternatively, those files might be unified and use `pkg.downloaded` instead of `pkg.installed` conditionally (resp. `pkg.patch_downloaded` instead of `pkg.patch_installed`)
+- existing configuration options are to be deprecated (in SUSE Manager supported versions) or removed
+  - existing code handling the configuration options can be removed from Uyuni
+
+# Drawbacks
```
mbologna: Another drawback to be added: from the user point of view, we will have two different options in two different locations to enable content staging for traditional and minions.

accepted/00052-new-package-predownloading.md
```diff
@@ -0,0 +1,85 @@
...
+ - this is limited to Salt minions. Traditional clients retain their existing mechanism and are not affected
+ - as the `pkg.downloaded` state [is not yet supported in Salt for Debian systems](https://docs.saltstack.com/en/2018.3/ref/states/all/salt.states.pkg.html#salt.states.pkg.downloaded), this remains an exclusive for `yum` and `zypper` (this RFC does not change this aspect)
+ - current users of the existing feature will have to adapt their workflows
+
+# Alternatives
+- add UI pages and API calls to define "download time windows". Allow users to Actions to "download time windows" at scheduling time
```
mbologna: Why not uniform the name? "Content staging for minions"

mbologna:
Apart from the user perspective "confusion" (see comments) I do not foresee any problem in changing the staging mechanism :+1:

accepted/00052-new-package-predownloading.md
moio: That also used to be the case before - the difference is now the two mechanisms are also conceptually different. Clarified.

accepted/00052-new-package-predownloading.md
moio: Fixed.

ncounter:
@moio if the solution is to add a simple `flag` parameter that is just an additional value to inject in the current workflow behavior, I'm fine with the concept but I'd shape it more like what we already have in the `Apply Highstate` page where you can add the simple `Test mode` flag. Below you can see it:
Screenshot from 2019-05-03 13-46-07

abhinav-puri:
@moio This is indeed useful on its own, in its current form. I also agree with @joesusecom  that it would be good to provide the ability to schedule both the download and the installation at the same step.

@moio - I would however not consider changing the RFC at this stage. Let us implement the approach in its current form. It is useful, gives us the ability to bring something to the customers early, and initiate the next steps based on the requisite feedback.

accepted/00052-new-package-predownloading.md
```diff
@@ -0,0 +1,86 @@
...
+  - alternatively, those files might be unified and use `pkg.downloaded` instead of `pkg.installed` conditionally (resp. `pkg.patch_downloaded` instead of `pkg.patch_installed`)
+- existing configuration options are to be deprecated (in SUSE Manager supported versions) or removed
+  - existing code handling the configuration options can be removed from Uyuni
+
+# Drawbacks
+ - this is limited to Salt minions. Traditional clients retain their existing mechanism and are not affected
+ - as the `pkg.downloaded` state [is not yet supported in Salt for Debian systems](https://docs.saltstack.com/en/2018.3/ref/states/all/salt.states.pkg.html#salt.states.pkg.downloaded), this remains an exclusive for `yum` and `zypper` (this RFC does not change this aspect)
+ - current users of the existing feature will have to adapt their workflows
+ - proposed mechanism is conceptually different from the one in place for traditional clients
```
mcalmer: On big drawback was not named here, but in the alternative approach:

- with the UI, users have to go through the package/patch selection twice

The big problem I see is, that when a user made a very individual selection he is nearly unable to reproduce it a second time for the real action execution.

The result might be, that he choose a different selection then for the download and additional packages needs to be downloaded when the real action will be executed.

mcalmer: @moio - I think you suggestion in the comments about using a text field instead of a checkbox where a user can enter the number of hours a download should happen before the real action is executed is a good idea.

We hear already complains for the SP Migration UI where we have "Dry Run" vs. "Real execution" and there are the options for spection selections much smaller then here.

I think we should take the extra effort.

ncounter:
(as discussed offline with @moio ) I am all for https://github.com/uyuni-project/uyuni-rfc/pull/5#issuecomment-487592263 and I agree we should take the extra effort to achieve it.

There is a small additional corner case: it was mentioned [in this comment](https://github.com/uyuni-project/uyuni-rfc/pull/5#discussion_r281137396) about the very individual selection packages and patches, but it can even happen that a very individual selection is made for minions as well, and for any reason that selection (persisted in SSM) is deleted in the meanwhile.

I see the most usable solution to be **saving both actions at the same time**. The workflow will not change at all for the normal use case for both users who want to use the pre-download feature and those who don't. And it will be only a matter of setting one single value that is the "offset of the pre-downloading schedule time desired" during the normal usage.

As we are talking about hours, a simple drop-down with numbers (e.g. visible only in supported cases, from 1 to 24 hours and with an `empy selected value` by default) will be an easy, robust and complete graphical object to address the need: no need of extra checkbox to enable the feature, no need to check invalid input values, etc. From the API POV instead, this can still be enabled as a **two-detached-steps** at all.

moio:
I changed the PR addressing all feedback so far, and changing the screenshots. Will merge before end of the week unless more points are brought up.

Thanks everybody for the constructive feedback so far.

accepted/00052-new-package-predownloading.md
```diff
@@ -0,0 +1,114 @@
...
+![Screenshot of the new option form](images/00052-earliest.png)
+
+*Note: the new form controls are highlighted in yellow.*
+
+Clicking on the selection element will bring up possible options:
+
+![Screenshot of the new option, selected](images/00052-selected.png)
+
+
+If an Action Chain is selected as the scheduling option, the new form elements will be disabled:
```
ncounter: This comment is more an implementation detail for later probably, but I think disabling the option is not really necessary if we design it clearly that the "Download packages" drop-down field belongs to the "Earliest" option only.
It can also be set to disabled in that case, I agree and I would keep it, but I see this as a *plus*.
Below I hacked your picture to give you a rough idea (it would needs some refinement anyway):
pre-download

moio: We will surely involve you guys when the front-end writing time comes :wink:

PR #6, opened 2019-04-29
Optionally disallow patching if packages have not been predownloaded yet

moio:
View: https://github.com/uyuni-project/uyuni-rfc/blob/disallow-downloading-when-patching/accepted/00053-disallow-downloading-when-patching.md
Comment: https://github.com/uyuni-project/uyuni-rfc/pull/6/files

### DIFF
--- a/accepted/00053-disallow-downloading-when-patching.md
+++ b/accepted/00053-disallow-downloading-when-patching.md
@@ -1,0 +1,61 @@
+- Feature Name: disallow-downloading-when-patching
+- Start Date: 29-03-2919
+- RFC PR: https://github.com/uyuni-project/uyuni-rfc/pull/6
+
+# Summary
+Optionally fail package installations and patch applications if they require any downloading.
+
+This is limited to minions on SUSE-based distros.
+
+# Motivation
+Minions with limited connectivity options might have little bandwidth available or bandwidth usage could come with a high cost (eg. in remote areas). In those cases we want users to be fully in control of when downloading happens, eg., to schedule that during off-peak hours (see https://github.com/uyuni-project/uyuni-rfc/pull/6). This RFC is about giving users the possibility to prevent package managers' default behavior of downloading at installation time.
+
+Package installation and patch application Actions should thus fail in case packages were not previously downloaded (if so configured).
+
+# Detailed design
+- two new functions are implemented in [zypper's execution module](https://github.com/openSUSE/salt/blob/openSUSE-2018.3.0/salt/modules/zypper.py). Given a list of packages or patches, return `True` if any downloading is required
+
+Those functions can leverage `zypper`'s estimation of downloaded bytes in dry-run mode to determine if any downloading is needed:
+
+```
+minion:~ # zypper --non-interactive --xmlout install --dry-run zsh
+<?xml version='1.0'?>
+<stream>
+
+...
+
+<install-summary download-size="0" space-usage-diff="7389792" packages-to-change="1">
+
+...
+</stream>
+```
+
+Those functions should accept a subset of parameters from the respective original state modules ([pkg.installed](https://docs.saltstack.com/en/2018.3/ref/states/all/salt.states.pkg.html#salt.states.pkg.installed) and [pkg.patch_installed](https://docs.saltstack.com/en/2018.3/ref/states/all/salt.states.pkg.html#salt.states.pkg.patch_installed)).
+
+
+- The Server's [package installation](https://github.com/uyuni-project/uyuni/blob/c8ffe6b9425392f5235864ad070646bb8ebc2ecb/susemanager-utils/susemanager-sls/salt/packages/pkginstall.sls) and [patch application](https://github.com/uyuni-project/uyuni/blob/c8ffe6b9425392f5235864ad070646bb8ebc2ecb/susemanager-utils/susemanager-sls/salt/packages/patchinstall.sls) modules will call those functions and conditionally skip actual installation/application via `unless` clauses
+
+- This should only happen if the minion's `mgr_fail_installation_if_download_is_required` Pillar is set
+
+- Pillars controlling this feature can be set at a global, per-minion or per System Group basis ([similarly to what we do to set the package download endpoint override](https://github.com/SUSE/doc-susemanager/pull/366/files#diff-8bf787463eb899039a59878b8b2ce800R171)).
+
+
+# Drawbacks
+ - this is limited to Salt minions. Traditional clients have no such feature
+ - only covers Zypper at this point
+ - no UI
+
+# Alternatives
+- use Action Chains
+  - pro: more visible to users
+  - con: could have scalability issues, more difficult to implement
+- implement custom modules instead of adding new functions to Salt's core
+  - pro: does not need a pull request against Salt
+  - con: the functionality might be useful in other contexts
+- add this feature directly to Zypper via a commandline switch
+  - pro: arguably cleaner design, does not need two calls to Zypper
+  - con: updated Zypper would be needed, more difficult to implement
+
+# Unresolved questions
+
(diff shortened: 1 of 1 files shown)


moio:
@joesusecom, @abhinav-puri, @mcalmer: I'd like to have your input about the general approach here.

@dincamihai: I would like to hear from you here, as all code changes here would be in the Salt part (either in the core or in SUSE Manager-shipped states). I can't think of that many alternatives, maybe you have ideas.

@aaannz: mainly FYI, although of course extra input is always welcome.

Conan-Kudo:
Why is this separate from #5? And why is this SUSE-only? This functionality should also work with salt minions working with yum or dnf as well.

accepted/00053-disallow-downloading-when-patching.md
```diff
@@ -0,0 +1,61 @@
...
+- RFC PR: https://github.com/uyuni-project/uyuni-rfc/pull/6
+
+# Summary
+Optionally fail package installations and patch applications if they require any downloading.
+
+This is limited to minions on SUSE-based distros.
+
+# Motivation
+Minions with limited connectivity options might have little bandwidth available or bandwidth usage could come with a high cost (eg. in remote areas). In those cases we want users to be fully in control of when downloading happens, eg., to schedule that during off-peak hours (see https://github.com/uyuni-project/uyuni-rfc/pull/6). This RFC is about giving users the possibility to prevent package managers' default behavior of downloading at installation time.
+
```
Conan-Kudo: Don't you mean to refer to #5 instead of this one?

accepted/00053-disallow-downloading-when-patching.md
```diff
@@ -0,0 +1,61 @@
...
+Package installation and patch application Actions should thus fail in case packages were not previously downloaded (if so configured).
+
+# Detailed design
+- two new functions are implemented in [zypper's execution module](https://github.com/openSUSE/salt/blob/openSUSE-2018.3.0/salt/modules/zypper.py). Given a list of packages or patches, return `True` if any downloading is required
+
+Those functions can leverage `zypper`'s estimation of downloaded bytes in dry-run mode to determine if any downloading is needed:
```
Conan-Kudo: This can also be implemented in `yum` or `dnf` by attempting to run in test-only made with the `--cacheonly` flag to see if it fails due to requesting something to download.

moio:
@Conan-Kudo this is separate because the two features, although make most sense to be used in conjunction, can also be used separately, and implementations are also expected to be independent.

The reason to limit this to `zypper` clients is just out of ease of (first iteration) implementation - remember we will have to contribute the [Salt core bits](https://github.com/uyuni-project/uyuni-rfc/blob/disallow-downloading-when-patching/accepted/00053-disallow-downloading-when-patching.md#detailed-design), so that means two more functions for `yum`, two for `dnf`, two for `apt` and so on.

It is currently my assumption that our users are mostly interested in SUSE distros, to be acknowledged by people mentioned above from the SUSE Manager side and from you and the user community from the Uyuni side.

We can expand the RFC in case I am wrong, or we can just implement SUSE distros as a first step and then support other package managers as additions. We would of course more than welcome any community effort to cover more platforms.

accepted/00053-disallow-downloading-when-patching.md
```diff
@@ -0,0 +1,61 @@
...
+- implement custom modules instead of adding new functions to Salt's core
+  - pro: does not need a pull request against Salt
+  - con: the functionality might be useful in other contexts
+- add this feature directly to Zypper via a commandline switch
+  - pro: arguably cleaner design, does not need two calls to Zypper
+  - con: updated Zypper would be needed, more difficult to implement
+
+# Unresolved questions
+
+None known.
```
Conan-Kudo: Why is this a separate RFC from #5? I would expect this to be part of that as well.

Conan-Kudo:
> It is currently my assumption that our users are mostly interested in SUSE distros, to be acknowledged by people mentioned above from the SUSE Manager side and from you and the user community from the Uyuni side.

I think it's a fair assumption that SUSE Manager is used often enough in mixed Red Hat/SUSE environments that it should be addressed together. It's certainly part of the marketing, so it must be important enough...

accepted/00053-disallow-downloading-when-patching.md
moio: Yes, thanks for spotting the typo. Corrected.

accepted/00053-disallow-downloading-when-patching.md
moio: I integrated this bit of information into the RFC. Thanks for pointing this out!

accepted/00053-disallow-downloading-when-patching.md
moio: This is separate because the two features, although make most sense to be used in conjunction, can also be used separately, and implementations are also expected to be independent.

joesusecom:
I agree with @Conan-Kudo that we shouldn't make this too SUSE-specific. Let's start implementation with the SUSE zypper stack, but also think about covering all the other OSes that Uyuni is supposed to support equally well.

Conan-Kudo:
@joesusecom @moio I don't think it'd be that easy to support for `apt` _et al_, but at least `yum`/`dnf` should be pretty easy, since it already will throw an error if it finds out it needs to download something when it is told to operate in cache only mode. My understanding is that this is more or less how you'll do it with zypper anyway (based on this RFC), so I think tacking on `yum`/`dnf` to the initial implementation of the RFC should not be much extra work and will let you support the feature out of the gate for SUSE Manager for RH/CentOS systems.

abhinav-puri:
@moio From a retail perspective, the first priority is definitely SUSE distros. So I would focus on that first in any case. But I would definitely not disregard other OSes as well. CentOS and Ubuntu, in that order. This would align with @Conan-Kudo 's recommendation that if its not much effort, we should consider including Salt minions working with yum/dnf within this RFC. 

As a next step, we should consider apt et al based systems.

accepted/00053-disallow-downloading-when-patching.md
```diff
@@ -0,0 +1,61 @@
...
+# Alternatives
+- use Action Chains
+  - pro: more visible to users
+  - con: could have scalability issues, more difficult to implement
+- implement custom modules instead of adding new functions to Salt's core
+  - pro: does not need a pull request against Salt
+  - con: the functionality might be useful in other contexts
+- add this feature directly to Zypper via a commandline switch (analoguous to [--cacheonly](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/6/html/deployment_guide/sec-working_with_yum_cache#idm140707649391200) in `yum` and `dnf`)
+  - pro: arguably cleaner design, does not need two calls to Zypper
+  - con: updated Zypper would be needed, more difficult to implement
```
mcalmer: I see this alternative as a goal for the future - as a second step. It has some more "pros" then listed.
In case there is a reposync between the download-only and the install/update action, it might happen that a newer package is available. This would cause a failed action as zypper would try to download the newest package.

With something like `--cacheonly`, zypper would just calculate based on the cache and install what it has cached.
It would also prevent a download of metadata, which could slow down the execution time.

But as we might need this feature also for very old clients (SLE11) where we will not get a zypper update anymore, we should implement this RFC as it is now as a first step.

But we should open a Feature request for `zypper` to implement `--cacheonly` and as soon as some distributions have it, we should switch the implementation to this alternative.

moio:
All open points have been addressed in the new version I have just pushed. Importantly, I changed the design so that supporting `yum`/`dnf` should be significantly easier (either as a stretch goal or in future) and also leaves the door open for `apt`.

I am merging this PR at the end of the weeks if no more remarks emerge.

Thanks for the inputs so far!

PR #7, opened 2019-06-12
Support for additional fields in the metadata of Deb repos

mateiw:
RFC for https://github.com/uyuni-project/uyuni/issues/1030.


0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,207 @@
...
+http://mirror.example.com/ubuntu/dists/bionic-updates/main/binary-amd64/
+```
+or in case of a flat repository to the directory containing both `Release*` and `Packages*` files:
+```
+https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Master:/Ubuntu1804-Uyuni-Client-Tools/xUbuntu_18.04/
+```
+
+### GPG keys management
+
+The keys used to verify the channel metadata will be stored in the keyring configured in `/etc/rhn/signing.conf` (`GNUPGHOME`)
```
mcalmer: I think this is wrong. The signing.conf is only used for signing the outgoing channel metadata.

The keyring used for checking metadata during repo-sync is maintained here: `/var/lib/spacewalk/gpgdir` .
We provide a package `uyuni-build-keys` or `susemanager-build-keys` which contains a default keyring .

There is a default to find GPG keys. When we configure a repository URL we have a checkbox "Has signed Metadata". If this is checked, we expect that the main file has also a `.asc` and a `.key` variant containing the signature and the key.

spacewalk-repo-sync try to check the signature. If the key is not in the keyring it check if the `.key` file is available and ask the user if it is trusted and should be imported.

If the answer is "yes" the key will be imported and the signature could be checked.

It seems that Ubuntu main repos do not have the key at that place.
So we can think about to ship it with our uyuni package. 

Using the fields for Channels maybe tricky as they describe the "outgoing" data, but we need something for the "incomming" data.

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,207 @@
...
+Homepage: http://httpd.apache.org/
+Description: Apache HTTP Server (common files)
+Task: lamp-server
+Description-md5: 9f2fab36019a61312dec627d1cd80365
+Supported: 5y
+```
+
+The database schema needs to be extended with a new table to store the additional fields:
+```
+rhnPackageDebFields
```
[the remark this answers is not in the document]
mateiw: I've changed the names to be more generic. However I'm not sure if they're a good fit for RPM tags. According to the [RPM specification](https://docs.fedoraproject.org/ro/Fedora_Draft_Documentation/0.1/html/RPM_Guide/ch-package-structure.html#id674921) the tags are strongly typed. I'll have to check if and how these tags are exported to xml metadata.

0000-deb-repo-additional-headers.md
mateiw: Good point. I've change the approach to ship Ubuntu (and maybe Debian) keys. For other keys used by custom repos the user can specify the GPG Keyerver (or url) and Key ID at the level of the repository. The key will be imported into `/var/lib/spacewalk/gpgdir/pubring.gpg` which is also used to check Yum repos.

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,227 @@
...
+Homepage: http://httpd.apache.org/
+Description: Apache HTTP Server (common files)
+Task: lamp-server
+Description-md5: 9f2fab36019a61312dec627d1cd80365
+Supported: 5y
+```
+
+The database schema needs to be extended with new tables to store the additional fields not covered by the existing `rhnPackage*` tables:
+```
+rhnPackageExtraTags
```
cbbayburt: I think it will be easier to maintain if we just add the values to the same table unless there's a specific need for the values table. There should be just one row per field entry.

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,227 @@
...
+====================
+content_source_id   number foreign key rhnContentSource(id)
+key_url             string not null
+key_id              string not null
+
+```
+
+The class `com.redhat.rhn.taskomatic.task.repomd.DebPackageWriter` must be changed to write any additional fields stored in the database into the generated `Packages.gz` file.
+
+_Note_: The same structure could be used to store additional RPM headers not yet stored.
```
mbologna: In the case of RPMs, what would be the associated class?

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,227 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Support only a very limited set of additional fields like `Multi-Arch` discarding all others.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should all the fields be stored ? Should only the fields defined in [1] be stored ? If yes what happens with user degined fields [2]
```
mbologna: How the end-user would be affected if we just store a limited subset (e.g. 'Multi-Arch')?

0000-deb-repo-additional-headers.md
mateiw: For RPMs it's `com.redhat.rhn.taskomatic.task.repomd.PrimaryXmlWriter`.

0000-deb-repo-additional-headers.md
mateiw: Agree, I'll change the RFC to use only one table for both tag names and values: `rhnPackageExtraTags`.

0000-deb-repo-additional-headers.md
mateiw: Other fields might be important for the user as well, although for now `Multi-Arch` is critical for now. For start we could just store all fields and add some kind of filtering later if really needed.

mateiw:
@mcalmer @cbbayburt @mbologna I'd like to finalize the RFC. Apart from the things already discussed please let me know if you see any other outstanding issues.

cbbayburt:
@isbm FYI

mateiw:
@mcalmer @cbbayburt @mbologna I've changed a bit the way the GPG verification is done to use the the data from the existing `rhnCryptoKey`. Please let me know your opinion.

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,222 @@
...
+
+The field must be shown as a selector. The values used to populate the selector come from the existing table `rhnCryptoKey`. Only keys for current organization must be shown.
+
+An additional option `Default keys` must shown in the selector as default. If this option is selected, the validation will be done using the GPG keys from package `uyuni-build-keys`.
+
+When saving the repository data, in case there was a GPG key selected other then `Default keys` the assignment will be saved into the new table `rhnContentSourceGpg`.
+
+### Key usage in the API
+
+The API method `channel.software.create_repo()` must be extended with one optional arguments to allow specifying the `GPG key ID`.
```
cbbayburt: Do we have an API method to get a list of the GPG keys?

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,222 @@
...
+gpgv --keyring /var/lib/spacewalk/gpgdir/pubring.gpg Release.gpg Release
+```
+
+Otherwise the custom GPG key associated with the repo must be used:
+```
+gpgv --keyring </path/to/temp/keyring/file.gpg> InRelease
+```
+or
+```
+gpgv --keyring </path/to/temp/keyring/file.gpg> Release.gpg Release
```
cbbayburt: Is the temp keyring file created and disposed of during this process?

mateiw: Yes.

0000-deb-repo-additional-headers.md
mateiw: Yes, `kickstart.keys.listAllKeys`

mateiw:
GPG verification moved to separate RFC.

@cbbayburt @mbologna @mcalmer please let me know if this can be merged.

isbm:
I would still clarify what we do with the broken repos. Error is shown but is that info is stored? Is the repo is still synced but marked as broken? Other suggestions?

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,107 @@
...
+Description-md5: 9f2fab36019a61312dec627d1cd80365
+Supported: 5y
+```
+
+The database schema needs to be extended with a new table to store the additional fields not covered by the existing `rhnPackage*` tables:
+```
+rhnPackageExtraTags
+====================
+id          number not null
+name        string no null
```
mcalmer: typo
```suggestion
name        string not null
```

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,107 @@
...
+value       string not null
```
mcalmer: `string` is wrong. Should it be `VARCHAR(XXX)` or better `TEXT` ?

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,107 @@
...
+created     timestamp
+modified    timestamp
+
+primary key (id)
+```
```
mcalmer: General question: should be continue with the maximal normalized DB schema?

I think we will will store a lot of data in this tabel and for the "key" (name) we will have tons of duplicates.
So we might want to have an extra trable `rhnPackageExtraTagsKeys` which just define `id` and `name` and
use `key_id` instead of `name` here in rhnPackageExtraTags.

Could save a lot of space

mateiw: Agree. Will add a new table `rhnPackageExtraTagKey`.

0000-deb-repo-additional-headers.md
mateiw: Changed. to VARCHAR.

0000-deb-repo-additional-headers.md
```diff
@@ -88,6 +88,50 @@ The integrity of the `deb` binary packages must be verified using the checksums
 
 This will be handled in a separate RFC.
 
+## Failure Registry
+
+Packages might have sync issues, such as wrong GPG signature or
+checksum mismatch or downloading issues etc. However, broken packages
+still can be assigned to the channels/repos and thus potentially
+invaliadate them. In some circumstances this might have serious impact
```
mcalmer: I think a broken package is / should not be imported. But such an error means that the repo is "incomplete" which
can cause dependency problems when somebody try to use it,

0000-deb-repo-additional-headers.md
```diff
@@ -88,6 +88,50 @@ The integrity of the `deb` binary packages must be verified using the checksums
...
+on the infrastructure update performance. To avoid that, we need to
+have a mechanism that will verify if a particular channel is "clean",
```
mcalmer: I would replace "clean" with "complete".

0000-deb-repo-additional-headers.md
```diff
@@ -88,6 +88,50 @@ The integrity of the `deb` binary packages must be verified using the checksums
...
+### Solution
+
+The solution is simply add the following table:
+
+```
+rhnPackageErrors
+================
+package_id  NUMERIC NOT NULL
+error       VARCHAR NOT NULL
+```
```
mcalmer: We could also go for "channel errors" and not for individual packages. Need some investigation what is easier/better to implement and review by admins when they get the errors.

0000-deb-repo-additional-headers.md
isbm: We don't have to though. But I would say, better we do, so we can also examine it later at "low level" using tools like `spacewalk-data-fsck` and sort of things. Also such error in general will cause deps problems _anyway_, regardless we have those "leftovers" from the package in place on disk or not. In any case, we still need a list of those broken files to be shown in UX somehow, hence the DB should at least know about the file and a cause why it was failed.

For that I would actually import it and mark as broken.

0000-deb-repo-additional-headers.md
renner: We currently send one UI notification per channel on successful or erroneous `repo-sync`. Now if I understand correctly, on download or verification errors of single packages the `repo-sync` process would still complete successfully for this channel? In this case I guess there could be a third state like "`repo-sync` completed with errors", but why don't we just call it failed and discard the whole attempt?

0000-deb-repo-additional-headers.md
```diff
@@ -0,0 +1,166 @@
...
+The integrity of the `deb` binary packages must be verified using the checksums from the `Packages` file.
+
+### GPG signature checking and checksum verification
+
+This will be handled in a separate RFC.
+
+## Failure Registry
+
+Packages might have sync issues, such as wrong GPG signature or
+checksum mismatch or downloading issues etc. However, broken packages
```
renner: As GPG signature checking and checksum verification is handled in a [separate RFC](https://github.com/uyuni-project/uyuni-rfc/pull/12), wouldn't the corresponding error handling better be described there as well?

isbm: If reposync would be a 15 minutes deal, I would rather indeed discarding the entire batch. But reposync can hang there for long hours. So you want to try to fix few files (and if that still failed, e.g. remote service temporarily out of serivice, hence the broken last two packages or so) you just may come back later and fix only those two packages.

If you want to delete the whole thing, because you gave up on that broken repo — also fine, because we've got the whole list of those files and we know where they are located on the disk.

isbm: Personally I would feel better that UX provides some sort of "try again" fix button for only those broken files alongside "Delete repo". So then as an admin, I can still poke it couple of times and decide I want to keep it broken, wait for repo to be fixed or delete it entirely. If we discard it on error, it is fine for us, but then the UX experience a bit annoying: you never know if the other side will sync or not.

Basically, I am only proposing here to build here `rsync` analogy that resumes broken download (and we do know it is broken!)  instead of `scp` that restarts all over again, even only two files left to resume out of thousand.

@paususe @mateiw @mcalmer thoughts?

paususe: > Personally I would feel better that UX provides some sort of "try again" fix button for only those broken files alongside "Delete repo". So then as an admin, I can still poke it couple of times and decide I want to keep it broken, wait for repo to be fixed or delete it entirely. If we discard it on error, it is fine for us, but then the UX experience a bit annoying: you never know if the other side will sync or not, yet again wait for hours it complete (or, again, not).
> 
> Basically, I am only proposing here to build here `rsync` analogy that resumes broken download (and we do know it is broken!) instead of `scp` that restarts all over again, even only two files left to resume out of thousand.
> 
> @paususe @mateiw @mcalmer thoughts?

Agreed. 

@joesusecom even mentioned (can't remember where) that we should automatically retry, at least once, missing/broken packages after the reposync has ended.

isbm: Well, retry doesn't necessarily helps. It might if the network or storage glitch is very short or local. But that is very rare, apparently. More often happens that the mirror storage (or mounted NFS glitched somewhere) causing broken package, then just another retry won't help here. We need that retry delayed for much longer period.

isbm: Plus another (sorry for writings) reason to _keep_ the broken repo: say, `vim` package is broken. But you want to install `emacs`. That mean you still can perfectly use that broken channel, even so we know that it is broken.

paususe:
What's missing here?

mateiw:
> What's missing here?

Mainly to finalize the discussion about how to handle packages with bad checksum and to update the RFC with that.

isbm:
> > What's missing here?
> 
> Mainly to finalize the discussion about how to handle packages with bad checksum and to update the RFC with that.

Sounds like we are actually [agreed on general outline](https://github.com/uyuni-project/uyuni-rfc/pull/7/commits/626e6293c3ef90459d0019ceeafbccf0955e5f26#r325060061)? I guess, details will remain details. I would suggest to round this up and indeed merge.

0000-deb-repo-additional-headers.md
mcalmer: Thinking again about this I have to say that this cannot work. If a package has a problem during download and the checksum is not the expected one, the package is not imported at all. 
This means there is no package ID where we can store the error.

I would go for a channel error and compile a text with all failed downloads (packages and patches)

mateiw:
@mcalmer @isbm I've update the section `Failure Registry` according to the latest discussion.

0000-deb-repo-additional-headers.md
```diff
@@ -169,30 +169,32 @@ of dependencies are involved, but still admin needs to be aware of it.
...
-downloaded".
+In case package cannot be synchronized at all due to various reasons,
+the information about the failure needs to appear in the database.
+So then tools, like `spacewalk-data-fsck` can make use of this information.
 
 This information can be reused across all the channels, UI and CLI
 tools to display warning/status of a specified channel.
 
+The repo-sync process will not stop if a broken is encountered. An error will be inserted in the `rhnChannelError` table and the process will continue.
+No entry in `rhnPackage` will be created for broken packages (this is the current behavior) and no package will be saved to disk.
```
mcalmer: Packages are saved on disk in a temporary area. I am not sure about the exact current status of the code, but they may stay there already until the next sync overwrite them.

isbm: I've added [further steps in notes](https://github.com/uyuni-project/uyuni-rfc/pull/7/commits/183cb0e758445cf886fe1e0a30e0875b6b962656) about this.

mateiw:
Thanks for the feedback. Merging.

PR #8, opened 2019-06-21
Feat: add Remote Command Auditing

mbologna:
See https://github.com/uyuni-project/uyuni-rfc/pull/8/files


mcalmer:
I would like to know how the current state is. Who can currently use the this UI? What permissions are needed? What minions can he access? What would this RFC change?

mbologna:
> I would like to know how the current state is. Who can currently use the this UI? What permissions are needed? What minions can he access? What would this RFC change?

OK, I agree. I pushed some more information about the as-is and the to-be.

accepted/00000-remote-command-auditing.md
```diff
@@ -0,0 +1,106 @@
...
+Every system outside the organization of the user issuing the command is not impacted nor is it visible by the user.
+
+The motivations behind the changes introduced in this RFC are _security_ and _auditing_. Any command issues via the `Remote Command` will be run with `root` privileges on the target system(s). Additionally, the user that issued the command is not logged along with the command issues and its output.
+Many customers claim to require this feature before migrating from traditional clients to Salt minions.
+
+Reference: https://fate.suse.com/325624 https://fate.suse.com/327158 
+
+In order to overcome the above limitations, in this RFC we are going to introduce:
+
+- Role-based restriction for `Remote Command`: a new role will be introduced. Every user associated with this role (in addition to one of the `Administrative Role`s) can access the `Remote Command` feature and issue commands. Every other user will not see the `Remote Command` tab nor the `Salt > Remote Commands` menu item.
```
mcalmer: `(in addition to one of the 'Administrative Roles')`

Does it mean beeing an Org Admin would not be enough to execute remote commands? 

This is not typicall. An Org Admin (or SUSE Manager Administrator) are like "root". They should be allowed to do everything. Indirectly all roles are assigned to them.

As I know people requested this extra role I wonder if they do not want to achieve something different.
Maybe they want to allow user who are not Org Admin to also access Remote Command page for a limited number of systems.
I think we should ask @mbrookhuis and Don again to explain what customers really want.

accepted/00000-remote-command-auditing.md
```diff
@@ -0,0 +1,106 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+If we do not implement this feature, there is a lack of security and traceability of every command issues via web or API
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Right now, if a user does not have an `Administrative Role`, cannot access any system (he/she does not even see them in the `Systems list`). The new `Remote Command Administrator` should follow the same principle and be an addition to the `Administrative Role` or a new `Administrative Role` by itself?
```
mcalmer: Another Question: What makes a `Remote Command Administrator` Role different from the `Configuration Administrator` ?

AFAIK as a `Configuration Administrator` you are allowed to create salt states (and execute them?). You can write `cmd.run` in the state and have the same result as using remote command. This needs to be discussed also with the Champions.

accepted/00000-remote-command-auditing.md
mbologna: > `(in addition to one of the 'Administrative Roles')`
> 
> Does it mean beeing an Org Admin would not be enough to execute remote commands?

It is in the current situation (as-is). But in the future, for new users, you will have to manually set the new role `Remote Command Administrator` to that user. Existing users will behave with the old behavior, as requested by champions.
I pushed a clarification on this regard.

> As I know people requested this extra role I wonder if they do not want to achieve something different.
> Maybe they want to allow user who are not Org Admin to also access Remote Command page for a limited number of systems.

In this case the `Remote Command Administrator` role should be then an `Administrative Role`, equal to the  `SUSE Manager Administrator` or `Organization Administrator roles` in order to target every system within the same organization as the user issuing the command.

accepted/00000-remote-command-auditing.md
mbologna: If a user only has the `Configuration Administrator` role then you can create states. But you cannot target any system registered to the product (you will have an empty list), neither with `Remote Command` tab (because you don't see any system) neither with `Salt > Remote Commands` with `*` (`no target systems`).
Having a user with just `Remote Command Administrator` is the same as the case above.
By noticing all of the above, the RFC has the unresolved question: is `Remote Command Administrator` something we want in addition to an `Administrative Role` or an `Administrative Role` by itself?

mbologna:
> To my understanding, "administrative roles" are the select-all roles with superuser privileges which have total access in the whole product, in contrast to the normal "roles", which gives some permissions for a specific feature to a normal user who otherwise does not have any administrative privileges.

For the sake of clarity, I just created a user with the following privileges:

SUSE Manager Administrator [ ]
Organization Administrator [ ]

Activation Key Administrator [x]
Configuration Administrator [x]
Image Administrator [x]
Channel Administrator [x]
System Group Administrator [x] 

Even with registered systems to the same organization, the user in question cannot see any systems in "System Overview".

> Following this logic, I think the remote cmd role suits better for a "normal role". Moreover, administrative roles (Org admin, SUMA admin) should automatically cover all of the privileges of the normal roles, including the remote cmd role.

Agreed.
 
> So the remote cmd role should only come into play for normal users:
> 
>     * If a user has this role, they should be able to run remote commands on any system that he has control over (needs to be the admin for that system(s)).
> 
>     * If a user does not have this role, they should never be able to run remote commands on any system even if they are the admin for a system (a migration procedure should add this role to all normal users to keep the existing behavior).

Sales Engineers requesting for this feature asked to have this feature enabled by default for the existing users, while it has to be purposefully enabled for newly created users.
 
> On the other hand, I think @mcalmer's comment about the "configuration admin" role is totally valid. Assuming my model above is the way to go, then a user would be able to create a state with `cmd.run` commands and run it on a system even if they don't have a remote cmd role. In this case, putting remote command features inside of the existing "configuration admin" instead of creating a "remote cmd" role might be a simple alternative (possibly renaming the role name to cover both features).

Valid point as well, I agree.

accepted/00000-remote-command-auditing.md
```diff
@@ -0,0 +1,106 @@
...
+In order to overcome the above limitations, in this RFC we are going to introduce:
+
+- Role-based restriction for `Remote Command`: a new role will be introduced. Every user associated with this role (in addition to either `SUSE Manager Administrator` or `Organization Administrator`) can access the `Remote Command` feature and issue commands. Every other user will not see the `Remote Command` tab nor the `Salt > Remote Commands` menu item.
+- Auditing for `Remote Command` and `Salt > Remote Commands`: the product will log the user that has issued the command, the command itself and its result.
+
+# Detailed design
+[design]: #detailed-design
+
+The first step is to introduce a new role in the roles list: `Remote Command Administrator` (`RhnUserGroupType` table).
+The role is a normal role (comparable to `Config Administrator`, `Image Administrator`, etc) and not an `Administrative Role` (`SUSE Manager Administrator` or `Organization Administrator`). If a user has a `Remote Command Administrator` role but not one of the `Administrative Role`s, then the user cannot see any system and thus should behave like the other normal roles.
```
meaksh: Keep in mind that even if a user has no administrative role, it might be assigned to a "System Group". Then the user is allow to manage that particular group according to the normal roles the user has.

So, I would say we need to refactor this because it's not true that "... If a user has a `Remote Command Administrator` role but not one of the `Administrative Role`s, then the user cannot see any system ...".

I also agree with @mcalmer that maybe it's enough to use the existing "Configuration Administration" which currently allows the user to manage the "Configuration States", since running Salt states is basically the same (in the context of permissions and accessibility) than running an execution module of Salt.

Why someone could think on allowing "Remote Execution Commands" and not "Configuration States" for a system? Both roles do allow to basically run whatever command as "root" on the targeted minion.

cbbayburt:
If we all agree on using the "configuration admin", the implementation would be as simple as adding the ACL checks to the remote commands page.

cbbayburt:
@mbrookhuis, @dvosburg, any thoughts, comments?

accepted/00000-remote-command-auditing.md
```diff
@@ -0,0 +1,106 @@
...
+
+Each user can only target the systems within the same organization that the systems are registered to, provided that the user has a particular role:
+- Every user that has `SUSE Manager Administrator` or `Organization Administrator` role can access the tab and run a remote command on the selected system or a group of systems
+- The API is always exposed but checks for the role of the user supplying the command after the API is called
+- For every user registered to SUSE Manager, independent of the role, the menu item `Salt > Remote Commands` is accessible. Only users with the `SUSE Manager Administrator` or `Organization Administrator` role, though, can impact systems using this feature, whereas all other users cannot target any system (`"no systems found"`).
+
+Every system outside the organization of the user issuing the command is not impacted nor is it visible by the user.
+
+The motivations behind the changes introduced in this RFC are _security_ and _auditing_. Any command issues via the `Remote Command` will be run with `root` privileges on the target system(s). Additionally, the user that issued the command is not logged along with the command issues and its output.
```
moio: There is one extra twist worth noting if we frame this from the POV of security (and not just auditing): as soon as an admin can install an package, they can also run whatever root can via post scripts.

So if an admin can run commands, they can also install software. And if they can install software, they can also run commands.

Essentially, from a security point of view, it is either "you are root on system X" or "you are not" - we are not actually able to differentiate between "install" and "run command" (and, as @meaksh mentioned, even "configure"). 

Users have been vocal about auditing of the Remote Commands page just because the introduction of that page made it more visible, but from a security perspective the threat model hasn't really changed.

accepted/00000-remote-command-auditing.md
```diff
@@ -0,0 +1,106 @@
...
+"target-pattern" + '*'
+""
+nastyuser
+192.168.18.81
+```
+
+The final format of the log could be formatted in a JSON fashion to be easily parseable.
+The log file should be rotated and archived as usual product logs.
+
+The `Remote Command Administrator` role must be explicitly set. By default, the `Remote Command Administrator` role is not assigned to a new user.
```
mcalmer: Current behavior of all roles is:

- nothing is selected by default
- when you select "Org Admin" all Roles are auto-selected.
- Same for "SUSE Manager Admin"

We should not break this concept. It maybe deeply integrated in the code.

accepted/00000-remote-command-auditing.md
mcalmer: Just showed: you can be a normal user and become admin of a system. You do this by adding a system to a system group and make a user "admin" of that system group.

In this case, the user can perform all tasks on that system. Install Software, Call Highstate and run Remote Commands.

accepted/00000-remote-command-auditing.md
mbologna: You're right, adding it to spec.

accepted/00000-remote-command-auditing.md
mbologna: Yes, I agree, I will add a paragraph describing the potential surface of attack when allowing every user to install packages.

mbologna:
> If we all agree on using the "configuration admin", the implementation would be as simple as adding the ACL checks to the remote commands page.

Not sure: a configuration admin can also apply Salt states that can install packages, and ad-hoc crafted packages installed can execute everything on the system.

Probably it could also be worth adding a "Configuration admin" for these purposes.

mbologna:
> If we all agree on using the "configuration admin", the implementation would be as simple as adding the ACL checks to the remote commands page.

Do you mean adding the Remote Command(s) pages to the same ACL as "Configuration Administrator"?

cbbayburt:
> Do you mean adding the Remote Command(s) pages to the same ACL as "Configuration Administrator"?

Yes, but then again, I'm not sure what the customer specifically expects at this point.

cbbayburt:
Joe's comment from the issue:

> This should disable both the Salt tab's remote command feature and the one that is per system.
>
> My suggestion is that (if technically possible) the Salt one should only be completely disabled if the user has no systems to run commands on. Otherwise it should just filter out the systems he/she has no remote command permission for.

This is already the case for "normal" users. So I imagine what he meant here is to somehow restrict the "Admin" users, which I also don't agree.

mbologna:
> For the logging part, we decided to implement it only for Salt Remote Commands page, since the other places already add the actions to the event history.
> 
> I've also noticed a few minor differences with the actual implementation: [uyuni-project/uyuni#690](https://github.com/uyuni-project/uyuni/pull/690) (log file name, etc.)

You're right, I was not aware of this development! I removed all the part pertaining to the logging.

accepted/00000-remote-command-auditing.md
mbologna: Added a clarifying section: `Administrative Roles` will not be impacted. They will remain `root`.

accepted/00000-remote-command-auditing.md
mbologna: > AFAIK as a `Configuration Administrator` you are allowed to create salt states (and execute them?). You can write `cmd.run` in the state and have the same result as using remote command. This needs to be discussed also with the Champions.

An user who is not a `Configuration Administrator` cannot apply any state.

cbbayburt: I think this indicates a deeper problem with the current role structure.

Users are probably using "Administrative" excessively because of some gap in the roles spectrum.

Let's think about it like this:
> If I don't want the user to run remote commands, that means I don't want the user to be an "admin/root". So what other chance do I have (what checkboxes should I check) to have a user with all the permissions equivalent to an admin *minus* remote command execution?

I think the problem here is that only Administrative roles can play around with "all" systems. What's really missing is some collection of roles which would result in a user with access to "all" systems, but with limited capabilities.

accepted/00000-remote-command-auditing.md
```diff
@@ -0,0 +1,106 @@
...
+Every system outside the organization of the user issuing the command is not impacted nor is it visible by the user.
+
+The motivations behind the changes introduced in this RFC are _security_ and _auditing_. Any command issues via the `Remote Command` will be run with `root` privileges on the target system(s). Additionally, the user that issued the command is not logged along with the command issues and its output.
+Many customers claim to require this feature before migrating from traditional clients to Salt minions.
+
+Reference: https://fate.suse.com/325624 https://fate.suse.com/327158 
+
+In order to overcome the above limitations, in this RFC we are going to introduce:
+
+- Role-based restriction for `Remote Command`: a new role will be introduced. Every user associated with this role (in addition to either `SUSE Manager Administrator` or `Organization Administrator`) can access the `Remote Command` feature and issue commands. Every other user will not see the `Remote Command` tab nor the `Salt > Remote Commands` menu item.
```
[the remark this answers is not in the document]
mbologna: OK, I was working under the assumption of what I read in FATE: customers want to hide `Remote Command(s)` from Administrative Roles.

I reworded the whole structure. `Administrative Roles` (`SUSE Manager Administrator or Organization Administrator`) will not be impacted by this change.

accepted/00000-remote-command-auditing.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Caveats
+
+From a security point of view, if a user does not have any role associated to him/her the surface of the attack is still not negligible: the user can install and remove packages using SUSE Manager on the systems he/she is assigned to. Those packages can be ad-hoc crafted to contain {pre, post}-install script to execute what an attacker wants.
+
+### Comparison with `Configuration Administrator`
+
+A user having the `Configuration Administrator` role can create and apply states to the systems he/she is entitled to. This offers the same surface of attack, as an attacker could simply drop a `cmd.run` into the desired state.
+
+Should `Remote Command Administrator` just another addition to the access control list of the `Configuration Administrator`?
```
moio: I am not sure I get the meaning of this question. Are you suggesting that Configuration Administrator implies Remote Command Administrator?

accepted/00000-remote-command-auditing.md
```diff
@@ -0,0 +1,116 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Users are probably using `Administrative Roles` excessively because of some gap in the roles spectrum. This change might not be what a user wants: a stripped-down `Administrative Role` with no access to `Remote Commands` (see next section).
+
+# Alternatives
+[alternatives]: #alternatives
+
+By reading the two fate entries linked it seems that customers are looking for a solution to disable Remote Commands for users within the Administrative Roles (SUSE Manager Administrator and Org Administrator). Those users are assimilated as root` so they cannot be changed.
```
moio:
```suggestion
By reading the two fate entries linked it seems that customers are looking for a solution to disable Remote Commands for users within the Administrative Roles (SUSE Manager Administrator and Org Administrator). Those users are assimilated as `root` so they cannot be changed.
```

accepted/00000-remote-command-auditing.md
mbrookhuis: I think these should be 2 different roles. A "configuration administrator" should only be able to create of change configuration channels. It could mean that he has no access to the system and should not be able to run any command.

mbologna: > I am not sure I get the meaning of this question. Are you suggesting that Configuration Administrator implies Remote Command Administrator?

Yes, given that a `Configuration Administrator` can apply states and execute them, why not assimilate `Remote Command Administrator` into the `Configuration Administrator`?

mbologna: > I think these should be 2 different roles. A "configuration administrator" should only be able to create of change configuration channels. It could mean that he has no access to the system and should not be able to run any command.

This could be another possibility. In other words: a `Configuration Administrator` only writes states, but cannot apply them. A `Remote Command Administrator` can apply states and issue commands. Did I get it right?

mbologna:
> I think that we need 2 solutions.
> 
>     1. We need an extra role to control who can execute remote commands

Related to normal (non-`Administrative Role`s) users?

>     2. a setting to enable/disable the execution of a remote command for the whole system

Do you mean enable/disable the `Remote Command(s)` feature for _every_ user (including the `Administrative Role`s ones)?

accepted/00000-remote-command-auditing.md
mcalmer: I fear we are currently mixing here the current situation with future expectations.

I think we understood the current permission system totally wrong. All currently existing administrator roles are **not** for Systems. They are only for management operation inside of SUSE Manager.

This means the **Configuration Administrator** can only create and modify config channels and states, but he cannot call state.apply. We should prove this. We all have not enough played around with the permission system to know this for sure.

Another "guess" we have is: Any normal user can become administrator of a system and with this permissions he is allowed to do everything with that system. Such a user is "root".
There is no special "Role" for this.

It seems that people want to change this. But this become much harder then expected.
This means we need to split the functionallity of "root" into pieces which makes sense.

In such a case we need to create table of new roles with functionalities in the systems page.

mbrookhuis: I agree with MC.

To your comment regarding remote command administrator. 
A system administrator can execute states or perform a highstate. Only when he also has the role remote command administrator he can execute remote commands for the systems he is administrator.

mbrookhuis:
> > I think that we need 2 solutions.
> > ```
> > 1. We need an extra role to control who can execute remote commands
> > ```
> 
> Related to normal (non-`Administrative Role`s) users?
correct
> 
> > ```
> > 2. a setting to enable/disable the execution of a remote command for the whole system
> > ```
> 
> Do you mean enable/disable the `Remote Command(s)` feature for _every_ user (including the `Administrative Role`s ones)?
correct

cbbayburt:
@mbrookhuis your suggested would have the following backdoor:
If a system admin does not have remote commands permission, he can write the commands into cmd.run states and still execute them.

Is this acceptable?

mbologna:
After asking on internal the mailing lists, it seems that the consensus is to introduce `Remote Command Administrator`: normal users (non-administrative users) without that role cannot launch any remote command on the managed systems.

accepted/00055-remote-command-auditing.md
```diff
@@ -0,0 +1,90 @@
...
+## A new role `Remote Command Administrator`
+
+The first step is to introduce a new role in the roles list: `Remote Command Administrator` (`RhnUserGroupType` table).
+The role is normal (comparable to `Config Administrator`, `Image Administrator`, etc).
+
+This translates to:
+
+  1. For non-`Administrative Role` users that can manage the selected systems who do not have the `Remote Command Administrator Role`, `Remote Command` tab would be unavailable.
+
+  2. `Salt > Remote Commands` is unavailable if the user is not associated with any `Administrative Role` or has not any assigned System Group.
```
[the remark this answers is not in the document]
mbologna: I rewrote it by simplifying: only Remote Command Administrator role would be required in this case.

accepted/00055-remote-command-auditing.md
```diff
@@ -0,0 +1,90 @@
...
+
+In the Java backend:
+
+  1. Traditional systems and Salt systems: `path="/systems/details/SystemRemoteCommand"`, `path="/systems/ssm/provisioning/RemoteCommand"` and `SystemRemoteCommandAction` and `ProvisioningRemoteCommand` Java classes
+  2. Salt systems: `path="/manager/systems/cmd"` and `RemoteMinionCommands` Java class.
+
+The role `Remote Command Administrator` will be enabled for all non-`Administrative` existing users. It will be disabled by default for new users: it must be manually enabled by the Administrator that creates the user.
+
+## A new config option `java.disable_remote_commands`
```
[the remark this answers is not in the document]
mbologna: Agreed. I will point out it is enabled by default.

accepted/00055-remote-command-auditing.md
```diff
@@ -0,0 +1,90 @@
...
+
+A new `rhn.conf` will be introduced that disables the `Remote Command` tab and the `Salt > Remote Commands` feature. Codepaths are the same as the previous section, but it will be enforced for every user.
+
+## Future Developments
+
+If needed, the `Remote Command Administrator` role can be customized with a system group granularity level: every user can access the `Remote Command` tab on a system if he/she is entitled to do via ACL.
+Additional tables to hold user and system groups associated with the role must be introduced and the associated backend code must be implemented.
```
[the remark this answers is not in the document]
mbologna: Agreed on easier implementation. Let's leave the door open for further improvements. I reworded a bit adding your comments.

PR #9, opened 2019-07-29
Accept and manage gpg keys with salt

mcalmer:
RFC about managing GPG repository keys with Uyuni.

### DIFF
--- a/accepted/00000-accept-and-manage-gpg-keys-with-salt.md
+++ b/accepted/00000-accept-and-manage-gpg-keys-with-salt.md
@@ -1,0 +1,99 @@
+- Feature Name: Accept and Manage GPG Keys with Salt
+- Start Date: 2019-07-19
+- RFC PR: (leave this empty)
+
+# Summary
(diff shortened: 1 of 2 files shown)


mbrookhuis:
MC had a look at the proposed solution and I think this is an excellent way to get this done. Would be great if we could set them via a formula. Will have a look at that. 

We should also give some examples in the documentation how to get the keys. Often they are stored in the repodate of the external repository with then name repomd.xml.key

mcalmer:
@mbrookhuis What do you think? Do we need to trust the GPG key before the system is registered?
Would it be enough to have it after the registration? This is the most important question we currently have.

mcalmer:
Feedback from @mbrookhuis via phone:

"Step 1" can be omitted as the GPG keys are mostly needed after the registration when the channels are assigned and packages from the activation key should be installed.

If needed it can be implemented at a later point in time. For a first step the salt states and a formula would be sufficient.

PR #11, opened 2019-08-14
RFC: Subscription Matching in Public Clouds

meaksh:
Read this RFC [HERE](https://github.com/uyuni-project/uyuni-rfc/blob/master-subscription-matching-in-public-clouds/accepted/0000-subscription-matching-in-public-clouds.md)


accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,121 @@
...
+TODO: Describe generic module
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Gathering the UUID for the virtual instances
+
+For AWS and GCE, the instance id that is returned from the API is not actually the real "uuid" from the instance. In order to get it, it would be previously needed to store that "uuid" on the metadata/tags for each instance.
+
+TODO: Evaluate if the "uuid" could be stored from within the instance into the "instanceMetadata" / "instanceTags" using the internal API available for the instances. That way, an automate action could be executed on the registered instances to store this information and then have it available when running `virtual-host-gatherer`.
```
mcalmer: AFAIK this means we need to try to change the registration process (salt and traditional) to put a different "string" into the uuid field. We just need to be able to get the identifier from inside of the vm.
I think there was no need in SUSE Manager that it must really look like an uuid. We may put any string into it which uniquely identify the instance.

meaksh: You're right, and I think using directly the provider "instance id" instead a real smbios "uuid" is the smoother way here. A minor change is needed on the Java side to allow providing "uuid" which are not real "uuid" but this is also covered on the RFC.

I've pushed some new changes to reflect this.

meaksh:
I've removed the WIP tag. I think this is ready now for a first round of review.

@mcalmer @paususe

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,194 @@
+- Feature Name: Subscription matching in public clouds
+- Start Date: 2019-08-14
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC set the guides for supporting "Subscription Matching" on virtual instances created on Public Clouds like AWS, Azure or Google Compute Engine. Exploring and gatehering virtual instances from those Public Clouds enables "Subscription Matching" to properly match subscriptions for "1-2 Virtual Machines".
```
paususe: What about unlimited VMs subscriptions? If they are disallowed by T&C, let's state that clearly

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,194 @@
...
+
+Most of the implementation effort needed is about coding each one of the needed plugins for `virtual-host-gatherer`. On the Java side, we would need to adapt `VirtualHostManagerProcessor.java` and most probably also `VirtualInstanceManager.java` in a similar way that we do when processing a KUBERNETES virtual host, since we don't always know about amount of RAM, number of CPUs, etc from the Virtualization Host.
+
+
+## 1-2 Virtual Machine Licensing
+From [SUSE Terms & Conditions](https://www.suse.com/products/terms_and_conditions.pdf):
+
+> Up to 2 Virtual Machines running on the same Virtualization Host or Virtualization Environment or within the same Private Cloudaccount or **Public Cloud zone can be deployed with one "1-2 Sockets or 1-2 Virtual Machines" Subscription Offering**.
+
+This means, all virtual instances that belong to the same "zone" (not region) will match with the "1-2 Virtual Machines" subscription.
```
paususe: If zone=! region, then please define zone

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,194 @@
...
+                "instance5": "i-fffcb2bd24b7b11"
+            }
+        }
+    }
+}
+```
+
+All the attributes on the above JSON example are currently required by SUMA to properly create the entries on the DB. Since we don't know the real hardware used to virtualize the instances, those values like `ramMb`, `cpuMhz`, `totalCpuCores`, `totalCpuSockets` have been faked to 0. As mentioned before, with some adaptations in the Java side we could not require those values (like for KUBERNETES).
+
+NOTE: All the "vms" that belongs to the same "Virtual Host Manager", are considered in the same "Virtualization group" for subscription matching. In the case of Public Clouds, this means a different "Virtual Host Manager" needs to be added for each one of the "Zones" of the particular Public Cloud provider where we want gather virtual instances.
```
paususe: Added by who? Based on what criteria? Let's be very explicit and specific

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,194 @@
...
+instances = list_instances(compute, project, zone)
+for i in instances:
+    print(i)
+```
+
+The response contains the "instance name" as well as the "instance id".
+
+### A generic JSON-file module:
+
+Currently, there is already a "File based" plugin for `virtual-host-gatherer` which allows to import instances from a custom provided JSON file. According to the existing documentation, the aim of this plugin is to import VMware instances when there is no access from the SUSE Manager server to the VMware.
```
paususe: Please link to the relevant docs. Is the JSON format identical to the one proposed at the beginning of this RFC?

accepted/0000-subscription-matching-in-public-clouds.md
mcalmer: Currently 1-2 Subscriptions and per instance Subscriptions are the only which can be used in Public Cloud.
Unlimited Virtualization cannot be used. We can add this to the RFC.

accepted/0000-subscription-matching-in-public-clouds.md
mcalmer: What scares me: `Private Cloudaccount` **or** `Public Cloud zone`. Everything which is gathered with 1 account can be grouped together? @joesusecom ?

And I would also like to know what `zone` is.

accepted/0000-subscription-matching-in-public-clouds.md
mcalmer: Hmm, we are working on a technical implementation and not on the Terms and Conditions.
I think we just need to make it possible what the Terms and Conditions describe. Explaining the Terms and Condition is not really the scope if the RFC. Only if it describes the implementation.

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,194 @@
...
+- A module for gathering virtual instances from Azure Cloud.
+- A module for gathering virtual instances from AWS.
+- A module for gathering virtual instances from Google Compute Engine.
+
+This way, the Administrators will be able to define and/or gather their virtualization environments in the public clouds and make the subscription matcher do a succesfull matching of the "1-2 Virtual Machine" subscriptions.
+
+# Detailed design
+[design]: #detailed-design
+
+Most of the implementation effort needed is about coding each one of the needed plugins for `virtual-host-gatherer`. On the Java side, we would need to adapt `VirtualHostManagerProcessor.java` and most probably also `VirtualInstanceManager.java` in a similar way that we do when processing a KUBERNETES virtual host, since we don't always know about amount of RAM, number of CPUs, etc from the Virtualization Host.
```
mcalmer: If I get it right, we only need to adapt the Java side to create a "foreign host" which stand for "the Cloud".
Would be good to say this more explicit.

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,194 @@
...
+    }
+}
+```
+
+All the attributes on the above JSON example are currently required by SUMA to properly create the entries on the DB. Since we don't know the real hardware used to virtualize the instances, those values like `ramMb`, `cpuMhz`, `totalCpuCores`, `totalCpuSockets` have been faked to 0. As mentioned before, with some adaptations in the Java side we could not require those values (like for KUBERNETES).
+
+NOTE: All the "vms" that belongs to the same "Virtual Host Manager", are considered in the same "Virtualization group" for subscription matching. In the case of Public Clouds, this means a different "Virtual Host Manager" needs to be added for each one of the "Zones" of the particular Public Cloud provider where we want gather virtual instances.
+
+## Add new Virtual Instance Types to the DB
+New virtual instance types would be needed on the `rhnVirtualInstanceType` database table, these would be `azure`, `aws`, `gce` and `generic` to indicate the type of the instance.
```
mcalmer: As you say below, there is already the file based type which should be used as `generic`. That type was always meant to be generic, but until now we only have VMware so the example data reflect VMware. 

I would remove `generic` here and explain more on top, that the existing file based type can be used for all types of virtual host managers and that we only need to adapt the docs.

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,194 @@
...
+
+Also, it's necessary to add new types of CPU Arch ("rhnCpuArch" DB table) and "Server Arch" ("rhnServerArch") to allow those new types of systems. We could adding one type per public cloud provider, i.a. aws, azure, etc, or maybe go with a generic type called `public_cloud`.
```
mcalmer: If somehow possible we should not introduce more of these "archs". We have already too much and nobody understand anymore why package, server, cpu and other archs needs different values and tables.

AFAIK we did not add new archs for VMware, Mainframe (s390x) now Kubernetes. Let's try to make sense of the current existing values.

Let's try to find out which arch the cloud support (currently mostly x86_64, but maybe aarch64 or others will come in future). Does the APIs have a function to get such a value?

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,194 @@
...
+
+An easy approach here would be to use the "Instance ID" (instead of smbios uuid) when registering a system which is a public cloud virtual instance. Salt currently does not provide the instance id as part of the grains but it would be really easy to provide a custom grain at the time of registration that would expose the "instance id" as part of the grains only when the system is an EC2, GCE or Azure instance. [Example here](https://gist.github.com/meaksh/1ed58ece0f26ce27a8445985de9ad6a2)
+
+This way, doing some minor fixes on the Java side ([example here](https://github.com/meaksh/uyuni/commit/03d88550dd87d22f3fabd25cebd7c23432285a3c)), we could easily use the "instance-id" as "UUID" for the registered system and automatically match it with the data provided by the `virtual-host-gatherer` plugin (which does not include "uuid" but instance id).
+
+In case of systems that are already registered in SUSE Manager using a smbios "uuid", if the new "instance_id" grain is there, it should be enough with scheduling a "Hardware Refresh" action to reflect the new "instance_id" grain value as the "uuid" for that system.
```
mcalmer: Please put this in the "Detailed design" section. It is exactly what I had in mind. It is not a drawback, it the way how we should implement it.

Another question: what about "traditional" clients? Should we implement that as well, or can we say that this is only supported with salt (@paususe) ?

Traditional  clients and salt is both python. So we could use the same code. But we would need to enhance the traditional registration protocol and the server side of that is in python, not in java.

@meaksh that could be a "unresolved question" and we can discuss it there.

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,194 @@
...
+### Use special "System Grouping" and do not use `virtual-host-gatherer`:
+This alternative approach will be based on implementing an special "System Groups" where Administrators could define their virtualization enviroments and set the systems that belong to those groups.
+
+Pros:
+- All virtual environments are defining via grouping. No need to actually reach the Public Clouds, no external APIs.
+
+Cons:
+- These virtualization groups need to be defined manually. No way of automation.
+- Refreshing the state of a virtualization group also needs to be done manually.
+- Require new DB tables and UI for managing those groups.
```
mcalmer: This is all correct. A note from my side: My feeling is, that no customer would "maintain" these data.

accepted/0000-subscription-matching-in-public-clouds.md
meaksh: I will clarify differences between zone and region on the RFC text.

AFAIU `Private Cloud account` refers i.a. to a private VMware cloud (like we already do in our vCenter example). So you define an Account, and all VMs gatherer from VMware are grouped together.

For the Public Cloud, it explicitly says "Public Cloud zone" (and not account), so this would mean, we cannot group by account but instead by "zone". Therefore, a customer using i.a. AWS will need to create as many "Virtual Host Manager" on SUSE Manager as "zones" they want to gather. Then each one of those "Virtual Host Manager" are considered as independent virtualization groups for "1-2 subscriptions" matching.

accepted/0000-subscription-matching-in-public-clouds.md
paususe: It's in our best interest as a company to detect as many subscriptions as possible.

How difficult is the traditional client case, in addition to salt? I'd do salt clients first, then traditional.

accepted/0000-subscription-matching-in-public-clouds.md
meaksh: AFAIK from the API we can only gather the "arch" from the instance, but not from the virtualization node, which is the one that we require here. So it might happen that you have instances that belong to the same zone but they have different CPU architecture.

But, let's keep in mind that we only need the "arch" for the "virtualization node" (foreign system) that is created and not for the VM instance itself.

accepted/0000-subscription-matching-in-public-clouds.md
meaksh: Yes, the JSON example would currently work (except for the reference to `aws` virtualization instance type) for feeding a File-based Virtual Host Manager on SUSE Manager.

accepted/0000-subscription-matching-in-public-clouds.md
meaksh: I pushed some changes that hopefully clarify the steps better.

accepted/0000-subscription-matching-in-public-clouds.md
meaksh: I guess we would need to do some changes on the Traditional Client code (registration & hardware refresh) to provide the "uuid" based on the "instance id" (instead of smbios) if we detect the instance is running in a Public Cloud environment. So introducing the same code that we have as example of a custom "instance_id" grain on Salt.

meaksh:
@mcalmer @paususe - I've pushed some changes on the RFC to address your feedback on the review. Take a look when you have some time.

Thanks!

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,204 @@
...
+
+- AWS EC2: i-1234567890abcdef0
+- GCE: 152986662232938449
+- Azure: 13f56399-bd52-4150-9748-7190aae1ff21
+
+Since the `virtual-host-gatherer` would only deal with the public APIs, each virtual instance will be identified by its "instance id" on the JSON output from the `virtual-host-gatherer` execution. In order to properly match this "instance id" with the "uuid" of a registered system, this will need some ajustments as describer later on this RFC.
+
+### Azure module:
+
+This is an example about how to deal with Azure Public API to collect the virtual instances, using Azure Python SDK:
```
paususe: Should we use each cloud's own API, or rather use something like cloud-init, which can provide information for a lot of clouds already?
https://cloudinit.readthedocs.io/en/latest/topics/instancedata.html
https://cloudinit.readthedocs.io/en/latest/topics/datasources.html#datasource-documentati

accepted/0000-subscription-matching-in-public-clouds.md
mcalmer: Ok, but we do not manage the foreign node. So we could use a dummy arch or "noarch" if we do not know. The value has no meaning so we should not intoroduce something new for it.

accepted/0000-subscription-matching-in-public-clouds.md
meaksh: AFAIU, `cloud-init` is a tool that runs inside the virtual instance basically for provisioning the cloud instance at startup time. It seems it gathers common virtual instance information from the "metadata" service (internal API) like the custom grain example I made.

But, we need some tool on the SUSE Manager server side that communicates with the Public Cloud provider using credentials and the external API to gather the instances (registered or not), and, at a first glance, it seems `cloud-init` does not do that.

accepted/0000-subscription-matching-in-public-clouds.md
mcalmer: I see that all available ServerArch values are very specific. But I think 1 new entry should be enough.
Remember: if we add a new value, it may be visible in the UI and might be selected. Not sure if this is the case for "ServerArch".

accepted/0000-subscription-matching-in-public-clouds.md
paususe: > AFAIU, `cloud-init` is a tool that runs inside the virtual instance basically for provisioning the cloud instance at startup time. It seems it gathers common virtual instance information from the "metadata" service (internal API) like the custom grain example I made.
> 
> But, we need some tool on the SUSE Manager server side that communicates with the Public Cloud provider using credentials and the external API to gather the instances (registered or not), and, at a first glance, it seems `cloud-init` does not do that.

Darn, you are right. I saw the cloud-init docs with exactly the information we need and was too happy. And I cannot find how to get the instance id in libcloud:
https://libcloud.apache.org/

:-(

meaksh: Actually, we could use `libcloud` to deal with AWS, Azure and Google Compute. The `list_nodes` method should return the list of instances including the `id` of the instance:

- AWS: https://libcloud.readthedocs.io/en/latest/compute/drivers/ec2.html
- Azure: https://libcloud.readthedocs.io/en/latest/compute/drivers/azure.html
- Google Compute: https://libcloud.readthedocs.io/en/latest/compute/drivers/gce.html

This way, we would only require `python3-apache-libcloud` as required dependency for dealing all the Public Clouds. I've created an example here: https://gist.github.com/meaksh/7153eba4ae6d98c9c6133e6790c7542d

If you're ok, I would change the RFC examples to make use of `libcloud`

paususe: I'm OK with whatever makes our life easier and gives the customer the best experience :-)

Will the use of libcloud give us magical support for the 50+ clouds libcloud supports, or will that require additional work on our side? How much work?
https://libcloud.readthedocs.io/en/latest/supported_providers.html

accepted/0000-subscription-matching-in-public-clouds.md
```diff
@@ -0,0 +1,287 @@
...
+
+This RFC defines the procedures to enabling "Subscriptions Matching" on different Public Clouds providers.
+
+Also provides implementation details for supporting the following modules for `virtual-host-gatherer`:
+
+- A generic (JSON-file based) module that enables Administrators to provide its own custom virtual instances definition.
+- A module for gathering virtual instances from Azure Cloud (based on libcloud).
+- A module for gathering virtual instances from AWS (based on libcloud).
+- A module for gathering virtual instances from Google Compute Engine (based on libcloud).
+
```
paususe: Alibaba?

meaksh: I added "Alibaba Cloud" to the list other potential providers to support (since metadata API is available): https://github.com/uyuni-project/uyuni-rfc/blame/master-subscription-matching-in-public-clouds/accepted/0000-subscription-matching-in-public-clouds.md#L168

I understood we're not implementing support on "Alibaba Cloud" in the first iteration of the implementation, that's why I didn't add a particular example for it. But, as discussed, implementing it using the "libcloud" approach should be trivial.

PR #12, opened 2019-08-30
GPG Signature verification for Debian repos

mateiw:
Split from https://github.com/uyuni-project/uyuni-rfc/pull/7.


0000-deb-repo-sync-gpg-check.md
```diff
@@ -0,0 +1,178 @@
...
+- What is the expected outcome?
+
+Uyuni can synchronize Debian repositories but currently it doesn't verify the GPG signature of the repo metadata.
+In order to import the metadata in a secure way Uyuni must be able to verify the GPG signature of that metadata.
+
+# Detailed design
+
+## Database changes
+
+A new table is needed to store the association of a GPG key to a repository:
```
mcalmer: Not sure why we need this. We do already GPG metadata signature checking for SUSE repos and we do not need such an extra table. Why should we need it for debian?

0000-deb-repo-sync-gpg-check.md
```diff
@@ -0,0 +1,178 @@
...
+content_source_id   number not null foreign key rhnContentSource(id)
+gpg_key_id          number not null foreign key rhnCryptoKey(id)
+created             timestamp
+modified            timestamp
+
+```
+
+## Repo signature checking
+
+The GPG verification should be optional but a warning should be shown to the the user in case he's not providing information about the GPG key used to verify the metadata.
```
mcalmer: I would not say "is optional", instead I would say, "can be disabled". The default should always be "check it".
Making it insecure is an option an admin has but we do not recommend it.

0000-deb-repo-sync-gpg-check.md
```diff
@@ -0,0 +1,178 @@
...
+https://download.opensuse.org/repositories/systemsmanagement:/Uyuni:/Master:/Ubuntu1804-Uyuni-Client-Tools/xUbuntu_18.04/
+```
+
+_Note #2:_
+
+Debian repos don't have a standard location for the public GPG keys used to verify the metadata. Sometimes the key is located in the root of a flat repo but the location is not standardized.
+
+### GPG keys management
+
+Uyuni ships the SUSE and OpenSUSE keys in the package `uyuni-build-keys`. The keys for Ubuntu and Debian should be shipped in this package too.
```
mcalmer: What if a user do not want to trust such a key? I would suggest to keep the automatically trusted keys very limited.
An option could be to create sub-packages and the admin can decide which keys he want to trust by install them or not.

0000-deb-repo-sync-gpg-check.md
```diff
@@ -0,0 +1,178 @@
...
+
+Any additional keys that need to be used (e.g. for Ubuntu derivatives distros or for custom repos) must be created manually using the existing UI (`Systems -> Autoinstallation -> GPG and SSL keys`)
```
mcalmer: spacewalk-repo-sync ask to trust the key and if the user answer with "yes" it imports it.
For Debian/Ubuntu the key may not exist at a specific place so we have a problem to download / guess it.
But maybe we could ask a keyserver?

`Systems -> Autoinstallation -> GPG and SSL keys` is a place where Keys gets stored for autoinstallation.
Sometimes they are used also for different use cases, but it is not very intuitive .

0000-deb-repo-sync-gpg-check.md
```diff
@@ -0,0 +1,178 @@
...
+
+
+### GPG Key usage in the UI
+
+The UI for creating/updating a repository (`Software -> Manage -> Repositories`) must be extended with one additional field:
+- GPG key
+
+This field must be visible only if the selected repository type is `deb` and the checkbox `Has Signed Metadata` is checked. For `deb` repositories the `Has Signed Metadata` checkbox must be enabled by default.
```
mcalmer: This suggest now a different way compared to the way a section before.
And it does it only for debian. Why not for RPM as well?

But this is new UI and different key handling as we have it now.
Do we really need to go that deep for a first step? Or is this already step 2?

0000-deb-repo-sync-gpg-check.md
```diff
@@ -0,0 +1,178 @@
...
+```
+gpgv --keyring </path/to/temp/keyring/file.gpg> InRelease
+```
+or
+```
+gpgv --keyring </path/to/temp/keyring/file.gpg> Release.gpg Release
+```
+
+The signature verification should be done during repo syncing when the metadata is parsed.
+Verification must be done only if the user has configured the GPG signing of the repos to be synced. Otherwise a warning must be printed in the logs about the lack of GPG configuration.
```
mcalmer: I would turn around this sentence as "verify" should be the default. 


```suggestion
If the user has turned off GPG signature verification, a warning should be printed into the logs.
```

0000-deb-repo-sync-gpg-check.md
```diff
@@ -0,0 +1,178 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+No drawbacks.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Support only a very limited set of additional fields like `Multi-Arch` while discarding all others.
```
mcalmer: this is an alternative for a different RFC.

0000-deb-repo-sync-gpg-check.md
```diff
@@ -0,0 +1,178 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should all the fields be stored ? Should only the fields defined in [1] be stored ? If yes what happens with user defined fields [2]
```
mcalmer: this is a question of a different RFC.

0000-deb-repo-sync-gpg-check.md
```diff
@@ -0,0 +1,178 @@
...
+- In case no GPG keys are configured for a repo (i.e. no GPG verification needed) should the `Release` file still be looked up and the checksums verified ?
```
mcalmer: I would suggest "yes". It would prevent using inconsistent repositories.

0000-deb-repo-sync-gpg-check.md
Conan-Kudo: Note that [`distribution-gpg-keys`](https://github.com/xsuchy/distribution-gpg-keys) package in Fedora and openSUSE ships many of the RPM distribution GPG keys.

In Fedora, we ship the Debian keys in `debian-keyring` package. The Ubuntu keys are shipped in the `ubu-keyring` package.

We may want to standardize on this approach and unify all the keys we ship this way...

0000-deb-repo-sync-gpg-check.md
mateiw: Ok, I made verification mandatory unless explicitly disabled.

0000-deb-repo-sync-gpg-check.md
Conan-Kudo: `Release` files are going away from Debian repositories. We need to start generating and validating `InRelease` files instead.

0000-deb-repo-sync-gpg-check.md
mateiw: The idea would be to let user add GPG keys for third-party repos.

Under `Software -> Manage -> Repositories` UI there's the option to select SSL certificates for checking and/or authenticating the HTTPS URL or the repo. The SSL certs are defined under `Systems -> Autoinstallation -> GPG and SSL keys` (indeed not very intuitive). This allows uploading SSL certs or GPG keys but there's currently no way to link the GPG key to a repo (only SSL certs can be linked to a repo), hence this new table.

0000-deb-repo-sync-gpg-check.md
mateiw: Changed to ship Debian and Ubuntu keys in separate packages.

0000-deb-repo-sync-gpg-check.md
mateiw: Please see also my previous [comment](https://github.com/uyuni-project/uyuni-rfc/pull/12#discussion_r322756074). 

I think there could be several options for adding new GPG keys for third party repos:

1) The UI (`Systems -> Autoinstallation -> GPG and SSL keys`). Not very intuitive. Maybe we could move it to some other place in the menu.

2) Let the user fiddle directly with `/var/lib/spacewalk/gpgdir/pubring.gpg`. I think this is risk.

3) Write a small shell script to let the user import or remove third party gpg keys into `/var/lib/spacewalk/gpgdir/pubring.gpg`.

0000-deb-repo-sync-gpg-check.md
mateiw: As a first step we could manage with shipping Ubuntu and Debian official keys in packages. I guess that's the main use case anyways and leave custom repos for a second phase.

0000-deb-repo-sync-gpg-check.md
mateiw: Removed.

0000-deb-repo-sync-gpg-check.md
mateiw: removed

0000-deb-repo-sync-gpg-check.md
mcalmer: Oh, these are used to access RedHat repos. They do client certificiate authentication.
repo-sync access these values and expect a X509 SSL certificate.

mateiw:
@mcalmer I've updated the RFC according to our last discussion.

0000-deb-repo-sync-gpg-check.md
Conan-Kudo: There are definitely Debian repos that guard with X.509 certificates. Many of the ones I have to deal with internally are set up that way.

mcalmer: Do you have an example repo which is not signed with GPG but with X509 ?

Conan-Kudo: Not one you can publicly access, unfortunately.

mateiw:
Thanks for feedback. Merging.

PR #13, opened 2019-09-23
RFC - Improve and normalize timezone handling

lneves12:
First draft of the RFC for the improvement and normalization of timezones handling


accepted/0000-timezones.md
```diff
@@ -0,0 +1,98 @@
...
+
+The user should be able to easily understand in which timezone a date is displayed and aware if there is any difference from the timezone
+he is accessing from.
+
+# Glossary
+[glossary]: #glossary
+
+When talking about uyuni there are 3 different Timezones contexts that are important to understand and distinguish:
+
+* *Server timezone:* the configured timezone in the server where uyuni is deployed
```
mcalmer: I this this is the timezone of the Uyuni Server. 
I think we should also add *managed system timezone* here. Even if they play no role in this rfc.

lneves12:
@paususe @ncounter @mcalmer a first draft version of the timezones normalization RFC is concluded. 

I think most of the main subjects are already covered here. Please let me know if you think there is any big topic I overlooked here, so I can take care of it.

Thanks

mcalmer:
One important point is missing in the RFC:

When a user schedule an action, for which timezone is it?
I think it is the timezone the user configured in his profile.
The RFC should say this explicitly.

accepted/0000-timezones.md
```diff
@@ -0,0 +1,102 @@
...
+
+There are 3 topics that we need to have in consideration here:
+
+## Display of dates
+
+The display of dates should always take in consideration the 3 user preferences settings: timezone, date-format and time-format.
+
+**For example:** '2019-09-10 13:00:00 **CEST**'
+
+At the moment only the setting timezone is implemented. The property date-format and time-format could be added on a second iteration,
```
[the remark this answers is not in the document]
lneves12: nice to know!

hum since we don't know yet when we will add the localization (language) again, it seems more feasible to add these new configuration props instead of inferring them from the location country. What do you think?

Nevertheless, the new implementation could take the old comment code as an inspiration :)

lneves12:
> One important point is missing in the RFC:
> 
> When a user schedule an action, for which timezone is it?
> I think it is the timezone the user configured in his profile.
> The RFC should say this explicitly.

Yes, it assumes the user-configured timezone, I will explicitly mention that! I would even risk saying that everything related with UI always should assume the user-configured timezone :)

accepted/0000-timezones.md
```diff
@@ -0,0 +1,103 @@
...
+[glossary]: #glossary
+
+When talking about uyuni there are 4 different Timezones contexts that are important to understand and distinguish:
+
+* **Managed systems timezone:** the timezone of each minion/client. This doesn't have any impact on this RFC, but it's important to mention.
+
+* **Uyuni Server timezone:** the configured timezone in the server where uyuni is deployed 
+
+* **User preferences timezone:** the defined timezone on the user preferences settings. Here we should be able to choose the timezone, date format (yyyyddmm) and the time format (12/24 hours).
+All the shown dates on the UI and operations on them should always assume these user settings.
```
paususe: Does this include the API? I can see we have several API calls where datetime is involved in ISO 8601 format but we only talk about YYYYMMDD HHMM, no timezone. E. g. scheduleChain, scheduleImageBuild, listAllPackages, etc.

accepted/0000-timezones.md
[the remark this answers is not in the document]
paususe: > nice to know!
> 
> hum since we don't know yet when we will add the localization (language) again, it seems more feasible to add these new configuration props instead of inferring them from the location country. What do you think?
> 
> Nevertheless, the new implementation could take the old commented code as an inspiration :)

We will add localization for 4.1, so please, take this into account.

accepted/0000-timezones.md
```diff
@@ -0,0 +1,103 @@
...
+There are 3 topics that we need to have in consideration here:
+
+## Display of dates
+
+The display of dates should always take in consideration the 3 user preferences settings: timezone, date-format and time-format.
+
+**For example:** '2019-09-10 13:00:00 **CEST**'
+
+At the moment only the setting timezone is implemented. The property date-format and time-format could be added on a second iteration, 
+but the code should be prepared to easily support them.
```
paususe: Can we write in this RFC how the three (timezone, date-format and time-format) would be implemented, and the implement in three steps? Are there any dependencies between those three?

accepted/0000-timezones.md
```diff
@@ -0,0 +1,103 @@
...
+
+Example:
+
+![timezones](images/timezones2.png)
+
+Ideally, these dates should be formatted server side using the already existing java methods:
```
paususe: What about Python (for the API, Salt, etc) ? Is it impacted?

accepted/0000-timezones.md
```diff
@@ -0,0 +1,103 @@
...
+
+```java
+LocalizationService.getInstance().formatCustomDate(date)
+```
+
+For the new stack (spark+react) if you want to convert a JSON to use the correct format, use the following GSON date adapter `UserLocalizationDateAdapter`
```
paususe: What about Struts? Any changes required?

accepted/0000-timezones.md
```diff
@@ -0,0 +1,103 @@
...
+
+For the new stack (spark+react) if you want to convert a JSON to use the correct format, use the following GSON date adapter `UserLocalizationDateAdapter` 
+that will convert your class dates to the user defined format.
+
+The previous gson adapter 'ECMAScriptDateAdapter' that converts dates to UTC should be avoided.
+
+## Manipulation of dates on the browser
+
+Since the browser Date implementation doesn't support timezones, if we ever need to do any date manipulation more than displaying a date
+ we should never use the object Date() directly but the `moment-timezone` library.
```
paususe: Will we package it or vendorize it?

accepted/0000-timezones.md
```diff
@@ -0,0 +1,103 @@
...
+
+This will require a refactor to the component react component `DateTimePicker` to properly manage timezones using the library `moment-timezone`
```
paususe: Should we rather use [react-moment](https://www.npmjs.com/package/react-moment#timezone-support)?

accepted/0000-timezones.md
```diff
@@ -0,0 +1,103 @@
...
+instead of manually calculating the diff between the user preference and browser timezone.
+
+## Create awareness to the user regarding the used timezones
+
+The user should always be aware in which timezone his account is configured and from where he is accessing from.
+
+To create context to the user regarding timezones, 2 clocks should be added to the top. One with the user configured timezone and
+the other one with the browser timezone.
```
paususe: What about the Proxy? Does the Proxy timezone have any impact? Can the Proxy timezone be different than the Server timezone?

accepted/0000-timezones.md
```diff
@@ -0,0 +1,103 @@
...
+
+To create context to the user regarding timezones, 2 clocks should be added to the top. One with the user configured timezone and
+the other one with the browser timezone. 
+
+![timezones](images/timezones.png)
+
+
+In cases we are dealing with user operations, for instance, scheduling new actions, more context information should be added.
+
+**Example:** `Applying the highstate has been scheduled at '2019-09-10 13:00:00 CEST', 14 hours from now`
```
paususe: Where should that be shown? Logs? UI?

accepted/0000-timezones.md
```diff
@@ -0,0 +1,102 @@
...
+ we should never use the object Date() directly but the `moment-timezone` library.
+
+This will require a refactor to the component react component `DateTimePicker` to properly manage timezones using the library `moment-timezone`
+instead of manually calculating the diff between the user preference and browser timezone.
+
+## Create awareness to the user regarding the used timezones
+
+The user should always be aware in which timezone his account is configured and from where he is accessing from.
+
+To create context to the user regarding timezones, 2 clocks should be added to the top. One with the user configured timezone and
```
ncounter: While I like the idea of the two different clocks, I am not sure I like the content of them. At the moment we could have three different timezone values set:
- browser
- user
- server

What has the most important impact is the difference between the client side and the server side: this difference will be reflected in scheduling actions. I'd better present the 2 clocks only if the server timezone is different that the one that is shown in the UI. And the one in the UI should always be in line with the user preference setting. To me the browser one should be completely overridden (and described clearly in docs). Unless, I'd be even fine with showing three different clocks: browser time, user time, server time. UI design can be really non intrusive adding small icons and showing the clocks only on a popup box. By this, user is aware of everything.

accepted/0000-timezones.md
```diff
@@ -0,0 +1,102 @@
...
+
+**Example:** `Applying the highstate has been scheduled at '2019-09-10 13:00:00 CEST', 14 hours from now`
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Making sure that all the legacy stack respect the 3 user preference properties timezone, date-format and time-format.
+It might not be trivial.
+- The use of the native browser Date implementation is forbidden, since it only supports the browser timezone.
```
ncounter: Reminder: we will need to spread the knowledge and the *rule* in a very highlighted way. Maybe a checkstyle rule or a linting rule will help us forbidding it.

accepted/0000-timezones.md
lneves12: We will vendorize it in the same we already do for all javascript libraries. 

We download all the libraries in jenkins (yarn install) -> generate a zip and licenses -> upload it to OBS -> generate js bundles in OBS -> susemager-web-libs.rpm (sub module of spacewalk-web)

There is already an open PR to add this library: https://github.com/uyuni-project/uyuni/pull/1408

accepted/0000-timezones.md
lneves12: react-moment is just a wrapper for moment to help to format dates, this is more implementation detail.

If we format all the dates server-side we don't need that. If there are cases where it makes sense, I would rather create our own wrapper for formatting dates.

accepted/0000-timezones.md
lneves12: On the UI. We already show "Applying the highstate has been scheduled". My suggestion would be to add just a bit more info

accepted/0000-timezones.md
lneves12: It shouldn't have any impact. This RFC shouldn't affect anything there, the communication between the server and proxy should stay the same.. as long you always keep the timezone information when handling dates it doesn't really matter on each timezone the server and proxy are.. 

Maybe @mcalmer has more insight here :)

accepted/0000-timezones.md
lneves12: humm I am not sure I understood everything here. 

The important time here is the "user settings" one and not the server one. Scheduling actions always assume the user settings one, the server time is irrelevant. So, the only important clocks differences are the browser (local user timezone) and user settings (the one defined in the account settings)

> I'd better present the 2 clocks only if the server timezone is different than the one that is shown in the UI

Agreed, I will add this to the rfc

accepted/0000-timezones.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+# Motivation
+[motivation]: #motivation
+
+The goal of this RFC is to define those rules and serve as a reference for the future.
+
+These rules should define a consistent way of display dates on the UI and make this information as clear as possible for the user.
+
+The user should be able to easily understand in which timezone a date is displayed and aware if there is any difference from the timezone
```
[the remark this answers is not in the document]
lneves12: humm we could remove the timezone information there, but I would not overcomplicate, that's not an easy thing to implement, because most of the time dates are being formatted server-side. IMO it's better to always show it to avoid misunderstandings.

accepted/0000-timezones.md
lneves12: Struts should stay the same. I think in most of the places the user settings timezone is already being considered. If not, it can just use the previous function: `LocalizationService.getInstance().formatCustomDate(date)`

accepted/0000-timezones.md
lneves12: To be honest my knowledge about the XMLRPC api is a bit limited. I was focusing the RFC more on the UI part. 

Do we have any problem with the API regarding timezones?

For me, it would make sense if we return the timezone information on the API as well, so no context is lost. Although I am not sure if this can be a breaking change.

Is Dario answer enough here?

> XMLRPC API endpoints should use and return datetime based on the user settings the one who is executing the API function. For the Salt command line AFAIK it can be used only on the server as root so, it datetime should be the one that the server has.

accepted/0000-timezones.md
lneves12: humm well, the timezone formatting already exists. The other two it's just a matter of adding more information to the date formaters not to use the default date/time-format but the user-configured instead. I don't know what we could add more to this RFC.

The image is an example of a WordPress admin panel.

accepted/0000-timezones.md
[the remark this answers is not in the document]
joesusecom: Internationalization (as in being ready for adding localization files that come from our translators has been a must-have feature in 4.0! The team doesn't have to care about translations, but the complete UI needs to be _enabled_ for localization. I'm not sure where the disconnect was, but this was always clearly communicated by product management and also part of the engineering interlock.

paususe:
@ncounter 
Is anyone looking into this RFC? It'd be nice to have this for 4.1

ncounter:
> @ncounter
> Is anyone looking into this RFC? It'd be nice to have this for 4.1

@paususe It's on my list on overtakeing Luis tasks, but I'm not yet there. There is an [EPIC](https://github.com/SUSE/spacewalk/issues/9049) related to this.
Do you directly want me to jump on it as a high priority or you want to rediscuss it during planning time? I will take a look anyway as a warm-up/research task to figure out what were Luis's plans and the expected amount of tasks. Sounds good?

moio:
@ncounter does it make sense to still keep this RFC PR around?

One possibility could be to add [WIP] in the title, or close it to reopen fresh when the time comes.

ncounter:
> @ncounter does it make sense to still keep this RFC PR around?
> 
> One possibility could be to add [WIP] in the title, or close it to reopen fresh when the time comes.

16 months without any chance to work on it, prefixing as a [WIP] would not reflect reality.
In the end either it is not perceived as important or there is always something more important.
Let's close it for now, we will revive when the right time will come.

PR #1, opened 2019-02-12
Add RFC template and README file

cavalheiro:
### What does this PR do?

This PR adds a template and README file for the Uyuni RFC process proposal. The "RFC" (request for comments) process is intended to provide a consistent and controlled path for new features to enter the project.

### Tests written?
Not applicable

PR #14, opened 2019-09-27
MVP CaaS Platform version 4 support (iteration #1)

mbologna:
First deliverable for CaaSP integration

### DIFF
--- a/accepted/0000-caasp-integration.md
+++ b/accepted/0000-caasp-integration.md
@@ -1,0 +1,74 @@
+- Feature Name: CaaS Platform support
+- Start Date: 2019-09-24
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Initial integration for managing a CaaS Platform cluster from Uyuni/SUSE Manager.
+
+# Motivation
+[motivation]: #motivation
+
+Uyuni/SUSE Manager will offer the possibility to manage the nodes that comprise a CaaS Platform cluster.
+Uyuni/SUSE Manager will register and show the status of each node of the cluster.
+
+# Detailed design
+[design]: #detailed-design
+
+## Node status reporting
+
+All the nodes in the cluster can be registered to Uyuni/SUSE Manager using bootstrapping by the already known methods.
+The added benefit will be that the user can see from Uyuni/SUSE Manager if a node has packages updates and patch available, along with all the information that a Salt minion is presenting when registered to Uyuni/SUSE Manager (e.g. CVE auditing).
+Another benefit is that the user can associate staged channels to the nodes registered (along with an activation key) to associate channels to the nodes of the cluster.
+The agent `skuba-update`, running on each node of the cluster, automatically picks up the packages from the associated channels and proceed to automatically update the nodes.
+NOTE: `skuba-update` is **NOT** invoked by Uyuni/SUSE Manager and it is running autonomously on the nodes of the CaaS Platform cluster.
+
+For example: an organization could have a "staging" cluster with a set of staging channels. The operator would cherry-pick Kubernetes updates from the SCC channels into the staging ones and then wait for skuba-update to roll them out into the staging cluster. At a later time, once testing is green inside of the staging cluster, the packages can be promoted to the "production channels" and be rolled out to the production cluster.
+
+If minion blackout is feasible in an easy way, then we can think to introduce it.
+
+It is paramount to state in the documentation that this information is presented in a read-only fashion.
+If a user reboots the node, installs a patch or installs a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) the Kubernetes may break.
+
+## Changes in the Kubernetes integration
+
+Under Virtual Host Manager, there is an already Kubernetes integration that supports CaaS Platform v3.
+The current implementation breaks with CaaS Platform v4: the Salt runner needs to take into account that CaaS Platform v4 switched to CRI-O. This needs to be fixed.
+
+Into the Virtual Host Manager, when a Kubernetes cluster is selected, the properties and the names of the nodes are shown, together with CPU Arch, CPU Sockets and RAM of the node.
+
+An addition to the already existing process would be:
+
+- when a user registers a new cluster, we introduce a new workflow to bootstrap (using Systems > Bootstrapping) each node of the cluster
+- when every cluster node is registered, a system group with the same name of the cluster is created and all registered systems are moved into this system group
+- each node presented the Nodes information is a link to the system overview page of that node
+
+## Documentation
+
+In addition to all the above, we are enriching our documentation on the CaaS Platform deployment. In the following sections, we are going to describe what we should introduce in the documentation.
+
+### AutoYaST Deployment documentation
+
+CaaS Platform provides AutoYaST templates for deploying a standard cluster (in the `SUSE-CaaSP-Management` package) comprising of:
+
+- 2 Load Balancers
+- 3 Masters
+- 3 Workers
+
+In the Uyuni/SUSE Manager documentation, we need to provide a step-by-step guide on how to customize the AutoYaST template, import it in Uyuni/SUSE Manager and deploy it to bare-metal machines.
+
+### Container registry documentation
+
+As an extra, in the documentation, we can write a guide on how to set up a container registry (using a standard container registry or Portus) and mirror SUSE container registry.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+If a user does not read the documentation, a Kubernetes cluster can be broken by simply rebooting a node from Uyuni/SUSE Manager.
+
+# Alternatives
+[alternatives]: #alternatives
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
--- a/accepted/0000-caasp-integration.md
+++ b/accepted/0000-caasp-integration.md
@@ -35,7 +35,7 @@ NOTE: `skuba-update` is **NOT** invoked by Uyuni/SUSE Manager and it is running 
 
 For example, an organization could have a "staging" cluster with a set of staging channels. The operator would cherry-pick Kubernetes updates from the SCC channels into the staging ones and then wait for skuba-update to roll them out into the staging cluster. At a later time, once testing is green inside of the staging cluster, the packages can be promoted to the "production channels" and be rolled out to the production cluster.
 
-It is paramount to state in the user interface and the documentation that the information presented for the minion is in read-only mode.
+It is paramount to state in the documentation that the information presented for the minion is in read-only mode. In the initial iteration, Uyuni/SUSE Manager must point out in the UI that executing actions (package install, upgrade, reboot) might break the cluster if the action are run on a node of the cluster.
 If a user reboots the node, installs a patch or installs a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) the Kubernetes may break.
 
 NOTE: both of the above actions will not cluster-coordinated, so it is entirely possible that a user assigns different channels, install packages or reboots nodes in a subset of nodes belonging to the same cluster.
@@ -87,7 +87,6 @@ In the documentation, [we can link to the CaaS Platform documentation to explain
 ### Management node
 
 As part of the support for the CaaS Platform version 4, SUSE Manager can elect (via the Add-On System Type) a node as "Management Node" for the cluster. This node must run SUSE Linux Enterprise Server and have `skuba` (a tool packaged by CaaS Platform) to deploy and manage the systems comprising the cluster.
-Alternatively, we can choose to run `skuba` on the SUSE Manager server, we should add it to the packaging pattern of SUSE Manager server. The package is Go-vendored, no other dependency is required.
 
 `skuba` is not packaged for OpenSUSE at the time of writing: Uyuni support will be implemented but enabled later, as soon as the package is released for OpenSUSE.
 
@@ -98,6 +97,10 @@ If a user does not read the documentation, a Kubernetes cluster can be broken by
 
 # Alternatives
 [alternatives]: #alternatives
+We can choose to run `skuba` on the SUSE Manager server, we should add it to the packaging pattern of SUSE Manager server. The package is Go-vendored, no other dependency is required.
+The problem of this approach is that the CaaS Platform admin needs shell access to the Uyuni/SUSE Manager server in order to access and run `skuba`.
+However, shell access to the Uyuni/SUSE Manager server could also be used for other (malicious) things beyond cluster management. And it breaks separation of concerns.
+For the reasons above, this approach was not followed.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/accepted/0000-caasp-integration.md
+++ b/accepted/0000-caasp-integration.md
@@ -25,18 +25,22 @@ In this regard, this RFC will cover the following aspects:
 ## Node status reporting
 
 All the nodes in the cluster can be registered to Uyuni/SUSE Manager using [bootstrapping methods](https://opensource.suse.com/doc-susemanager/suse-manager/client-configuration/registration-overview.html).
-This brings two major benefits:
+The code already present at the time of writing in Uyuni/SUSE Manager can support bootstrapping the nodes to Uyuni/SUSE Manager. We are not expecting any major code change - just testing is needed.
+
+Why node status reporting? This brings two major benefits:
 
 - The user can see from Uyuni/SUSE Manager if a node has packages updates and patch available, along with all the information that a Salt minion is presenting when registered to Uyuni/SUSE Manager (e.g. CVE auditing)
 - Another benefit is that the user can associate staged channels to the nodes registered (along with an activation key) to associate channels to the nodes of the cluster.
 
-The agent `skuba-update`, running on each node of the cluster, automatically picks up the packages from the associated channels and proceed to automatically update the nodes.
+The agent `skuba-update`, running on each node of the cluster, automatically picks up the packages from the associated channels and proceed to automatically patch the nodes.
 NOTE: `skuba-update` is **NOT** invoked by Uyuni/SUSE Manager and it is running autonomously on the nodes of the CaaS Platform cluster.
 
 For example, an organization could have a "staging" cluster with a set of staging channels. The operator would cherry-pick Kubernetes updates from the SCC channels into the staging ones and then wait for skuba-update to roll them out into the staging cluster. At a later time, once testing is green inside of the staging cluster, the packages can be promoted to the "production channels" and be rolled out to the production cluster.
 
+### Documentation
+
 It is paramount to state in the documentation that the information presented for the minion is in read-only mode. In the initial iteration, Uyuni/SUSE Manager must point out in the UI that executing actions (package install, upgrade, reboot) might break the cluster if the action are run on a node of the cluster.
-If a user reboots the node, installs a patch or installs a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) the Kubernetes may break.
+While rebooting a node might result in a Kubernetes cluster unavailable, installing a patch or a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) might result in an unpredictable non-working cluster situation.
 
 NOTE: both of the above actions will not cluster-coordinated, so it is entirely possible that a user assigns different channels, install packages or reboots nodes in a subset of nodes belonging to the same cluster.
 
@@ -49,9 +53,9 @@ The Kubernetes implementation that this RFC is targeting is the one in CaaS Plat
 
 Into the Virtual Host Manager, when a Kubernetes cluster is selected, the properties and the names of the nodes are shown, together with CPU Arch, CPU Sockets and RAM of the node.
 
-After the user has imported the `kubeconfig`, obtained from CaaS Platform, the process would be:
+Assuming that cluster bootstrap has already been done, after the user has imported the `kubeconfig`, obtained from CaaS Platform, the process would be:
 
-- When a user registers a new cluster, we introduce a new workflow to bootstrap (using Systems > Bootstrapping) each node of the cluster
+- When a user defines a new cluster, we introduce a new workflow to bootstrap (using Systems > Bootstrapping) each node of the cluster to Uyuni/SUSE Manager
 - When every cluster node is registered, a system group with the same name of the cluster is created and all registered systems are moved into this system group
 - Each node presented the Nodes information is a link to the system overview page of that node
 
@@ -93,7 +97,7 @@ As part of the support for the CaaS Platform version 4, SUSE Manager can elect (
 # Drawbacks
 [drawbacks]: #drawbacks
 
-If a user does not read the documentation, a Kubernetes cluster can be broken by simply rebooting a node from Uyuni/SUSE Manager.
+If a user does not read the documentation, a Kubernetes cluster can be made temporarily unavailable (until the machine is back) by simply rebooting a node from Uyuni/SUSE Manager or in an unpredictable non-working state by simply installing a patch or a package related to Kubernetes.
 
 # Alternatives
 [alternatives]: #alternatives
--- a/accepted/0000-caasp-integration.md
+++ b/accepted/0000-caasp-integration.md
@@ -10,39 +10,50 @@ Initial integration for managing a CaaS Platform cluster from Uyuni/SUSE Manager
 # Motivation
 [motivation]: #motivation
 
-Uyuni/SUSE Manager will offer the possibility to manage the nodes that comprise a CaaS Platform cluster.
-Uyuni/SUSE Manager will register and show the status of each node of the cluster.
+Uyuni/SUSE Manager is the systems management tool and must offer to the end-user the possibility to manage other SUSE products like CaaS Platform.
+Uyuni/SUSE Manager already supports Kubernetes cluster integration. It works with CaaS Platform version 3 (container engine: Docker).
+
+In this regard, this RFC will cover the following aspects:
+
+- Node status reporting
+- Update the current Kubernetes integration of Uyuni/SUSE Manager to work with CaaS Platform version 4 while still supporting CaaS Platform version 3
+- Document how to deploy CaaS Platform using AutoYaST
 
 # Detailed design
 [design]: #detailed-design
 
 ## Node status reporting
 
-All the nodes in the cluster can be registered to Uyuni/SUSE Manager using bootstrapping by the already known methods.
-The added benefit will be that the user can see from Uyuni/SUSE Manager if a node has packages updates and patch available, along with all the information that a Salt minion is presenting when registered to Uyuni/SUSE Manager (e.g. CVE auditing).
-Another benefit is that the user can associate staged channels to the nodes registered (along with an activation key) to associate channels to the nodes of the cluster.
+All the nodes in the cluster can be registered to Uyuni/SUSE Manager using [bootstrapping methods](https://opensource.suse.com/doc-susemanager/suse-manager/client-configuration/registration-overview.html).
+This brings two major benefits:
+
+- The user can see from Uyuni/SUSE Manager if a node has packages updates and patch available, along with all the information that a Salt minion is presenting when registered to Uyuni/SUSE Manager (e.g. CVE auditing)
+- Another benefit is that the user can associate staged channels to the nodes registered (along with an activation key) to associate channels to the nodes of the cluster.
+
 The agent `skuba-update`, running on each node of the cluster, automatically picks up the packages from the associated channels and proceed to automatically update the nodes.
 NOTE: `skuba-update` is **NOT** invoked by Uyuni/SUSE Manager and it is running autonomously on the nodes of the CaaS Platform cluster.
 
-For example: an organization could have a "staging" cluster with a set of staging channels. The operator would cherry-pick Kubernetes updates from the SCC channels into the staging ones and then wait for skuba-update to roll them out into the staging cluster. At a later time, once testing is green inside of the staging cluster, the packages can be promoted to the "production channels" and be rolled out to the production cluster.
+For example, an organization could have a "staging" cluster with a set of staging channels. The operator would cherry-pick Kubernetes updates from the SCC channels into the staging ones and then wait for skuba-update to roll them out into the staging cluster. At a later time, once testing is green inside of the staging cluster, the packages can be promoted to the "production channels" and be rolled out to the production cluster.
 
-If minion blackout is feasible in an easy way, then we can think to introduce it.
-
-It is paramount to state in the documentation that this information is presented in a read-only fashion.
+It is paramount to state in the user interface and the documentation that the information presented for the minion is in read-only mode.
 If a user reboots the node, installs a patch or installs a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) the Kubernetes may break.
 
-## Changes in the Kubernetes integration
+NOTE: both of the above actions will not cluster-coordinated, so it is entirely possible that a user assigns different channels, install packages or reboots nodes in a subset of nodes belonging to the same cluster.
 
-Under Virtual Host Manager, there is an already Kubernetes integration that supports CaaS Platform v3.
-The current implementation breaks with CaaS Platform v4: the Salt runner needs to take into account that CaaS Platform v4 switched to CRI-O. This needs to be fixed.
+## Update the current Kubernetes integration
+
+Under Virtual Host Manager, there is an already Kubernetes integration that supports CaaS Platform version 3.
+The current implementation breaks with CaaS Platform version 4: the Salt runner needs to take into account that CaaS Platform version 4 switched from Docker to CRI-O.
+Uyuni/SUSE Manager must check if the underlying container engine is running Docker (and use the existing Kubernetes integration) or CRI-O (and adapt the current implementation to support CRI-O).
+The Kubernetes implementation that this RFC is targeting is the one in CaaS Platform version 4: 1.15.2. Other Kubernetes distributions will not be targeted, at least initially.
 
 Into the Virtual Host Manager, when a Kubernetes cluster is selected, the properties and the names of the nodes are shown, together with CPU Arch, CPU Sockets and RAM of the node.
 
-An addition to the already existing process would be:
+After the user has imported the `kubeconfig`, obtained from CaaS Platform, the process would be:
 
-- when a user registers a new cluster, we introduce a new workflow to bootstrap (using Systems > Bootstrapping) each node of the cluster
-- when every cluster node is registered, a system group with the same name of the cluster is created and all registered systems are moved into this system group
-- each node presented the Nodes information is a link to the system overview page of that node
+- When a user registers a new cluster, we introduce a new workflow to bootstrap (using Systems > Bootstrapping) each node of the cluster
+- When every cluster node is registered, a system group with the same name of the cluster is created and all registered systems are moved into this system group
+- Each node presented the Nodes information is a link to the system overview page of that node
 
 ## Documentation
 
@@ -58,9 +69,24 @@ CaaS Platform provides AutoYaST templates for deploying a standard cluster (in t
 
 In the Uyuni/SUSE Manager documentation, we need to provide a step-by-step guide on how to customize the AutoYaST template, import it in Uyuni/SUSE Manager and deploy it to bare-metal machines.
 
+## Next steps
+
+Possible next steps are presented as additional "stretch goals": if time permits, those goals can be achieved in the scope of the initial integration.
+
+### Minion blackout configuration
+
+If the [minion blackout](https://docs.saltstack.com/en/latest/topics/blackout/) is feasible, we can set to blackout every minion registered to Uyuni/SUSE Manager that is part of a Kubernetes or CaaS Platform cluster. Every minion that is in blackout mode will reject all incoming commands.
+
 ### Container registry documentation
 
-As an extra, in the documentation, we can write a guide on how to set up a container registry (using a standard container registry or Portus) and mirror SUSE container registry.
+In the documentation, we can write a guide on how to set up a container registry (using a standard container registry or Portus) and mirror SUSE container registry.
+
+### Management node
+
+As part of the support for the CaaS Platform version 4, SUSE Manager can elect (via the Add-On System Type) a node as "Management Node" for the cluster. This node must run SUSE Linux Enterprise Server and have `skuba` (a tool packaged by CaaS Platform) to deploy and manage the node.
+Alternatively, we can choose to run `skuba` on the SUSE Manager server, we should add it to the packaging pattern of SUSE Manager server. The package is Go-vendored, no other dependency is required.
+
+`skuba` is not packaged for OpenSUSE at the time of writing: Uyuni support will be implemented but enabled later, as soon as the package is released for OpenSUSE.
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -72,3 +98,5 @@ If a user does not read the documentation, a Kubernetes cluster can be broken by
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
+
+- CaaS Platform: for Container registry do you suggest a standard container registry or Portus?
--- a/accepted/0000-caasp-integration.md
+++ b/accepted/0000-caasp-integration.md
@@ -45,7 +45,7 @@ NOTE: both of the above actions will not cluster-coordinated, so it is entirely 
 Under Virtual Host Manager, there is an already Kubernetes integration that supports CaaS Platform version 3.
 The current implementation breaks with CaaS Platform version 4: the Salt runner needs to take into account that CaaS Platform version 4 switched from Docker to CRI-O.
 Uyuni/SUSE Manager must check if the underlying container engine is running Docker (and use the existing Kubernetes integration) or CRI-O (and adapt the current implementation to support CRI-O).
-The Kubernetes implementation that this RFC is targeting is the one in CaaS Platform version 4: 1.15.2. Other Kubernetes distributions will not be targeted, at least initially.
+The Kubernetes implementation that this RFC is targeting is the one in CaaS Platform version 4: 1.16. Other Kubernetes distributions will not be targeted, at least initially.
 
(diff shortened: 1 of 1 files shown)


accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+Another benefit is that the user can associate staged channels to the nodes registered (along with an activation key) to associate channels to the nodes of the cluster.
+The agent `skuba-update`, running on each node of the cluster, automatically picks up the packages from the associated channels and proceed to automatically update the nodes.
+NOTE: `skuba-update` is **NOT** invoked by Uyuni/SUSE Manager and it is running autonomously on the nodes of the CaaS Platform cluster.
+
+For example: an organization could have a "staging" cluster with a set of staging channels. The operator would cherry-pick Kubernetes updates from the SCC channels into the staging ones and then wait for skuba-update to roll them out into the staging cluster. At a later time, once testing is green inside of the staging cluster, the packages can be promoted to the "production channels" and be rolled out to the production cluster.
+
+If minion blackout is feasible in an easy way, then we can think to introduce it.
```
paususe: Please explain what is minion blackout, what would be get from it, link to Salt docs, explain if that would be for all kind of Kubernetes clusters or only CaaSP4 clusters, etc

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+
+It is paramount to state in the documentation that this information is presented in a read-only fashion.
```
paususe: I'd also add a UI hint

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+If a user reboots the node, installs a patch or installs a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) the Kubernetes may break.
+
+## Changes in the Kubernetes integration
+
+Under Virtual Host Manager, there is an already Kubernetes integration that supports CaaS Platform v3.
+The current implementation breaks with CaaS Platform v4: the Salt runner needs to take into account that CaaS Platform v4 switched to CRI-O. This needs to be fixed.
```
paususe: What does "fixed" mean? That we will only support CRI-O? That we will support CRIO-O and whatever CaaSP3 was using? Please be explicit, that changes scope a lot :-)

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+For example: an organization could have a "staging" cluster with a set of staging channels. The operator would cherry-pick Kubernetes updates from the SCC channels into the staging ones and then wait for skuba-update to roll them out into the staging cluster. At a later time, once testing is green inside of the staging cluster, the packages can be promoted to the "production channels" and be rolled out to the production cluster.
+
+If minion blackout is feasible in an easy way, then we can think to introduce it.
+
+It is paramount to state in the documentation that this information is presented in a read-only fashion.
+If a user reboots the node, installs a patch or installs a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) the Kubernetes may break.
+
+## Changes in the Kubernetes integration
+
+Under Virtual Host Manager, there is an already Kubernetes integration that supports CaaS Platform v3.
```
paususe: Maybe I'd move this to earlier, below "Motivation" and start with: "_SUSE Manager 4.0 offers Kubernetes cluster integration. It works with CaaSP 3 (based on docker), with the following limitations:_
_- Cluster is presented as a system group_
_- Nodes are presented under Virtual Host Manager
_- Individual nodes can be managed and thus the cluster broken__
_- ..._

_This RFC is about extending that to support CaaSP4 and still supporting CaaSP3._"

If case this may work with some other Kubernetes distribution that we know about, I'd mention it, especially if it's a FLOSS one (for Uyuni). If not, I'd also mention it: "this RFC does not try to integrate other third-party Kubernetes distributions". Let's be very clear with the scope.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+The current implementation breaks with CaaS Platform v4: the Salt runner needs to take into account that CaaS Platform v4 switched to CRI-O. This needs to be fixed.
+
+Into the Virtual Host Manager, when a Kubernetes cluster is selected, the properties and the names of the nodes are shown, together with CPU Arch, CPU Sockets and RAM of the node.
+
+An addition to the already existing process would be:
```
paususe: Maybe briefly explain what that process is, so that it's easier to appreciate the new stuff?

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+
+### AutoYaST Deployment documentation
+
+CaaS Platform provides AutoYaST templates for deploying a standard cluster (in the `SUSE-CaaSP-Management` package) comprising of:
+
+- 2 Load Balancers
+- 3 Masters
+- 3 Workers
+
+In the Uyuni/SUSE Manager documentation, we need to provide a step-by-step guide on how to customize the AutoYaST template, import it in Uyuni/SUSE Manager and deploy it to bare-metal machines.
```
paususe: Maybe the CaaSP4 guys are already writing that?

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Container registry documentation
+
+As an extra, in the documentation, we can write a guide on how to set up a container registry (using a standard container registry or Portus) and mirror SUSE container registry.
```
paususe: Which one would be best (easier/more complete)? Do we have any idea?

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+If a user does not read the documentation, a Kubernetes cluster can be broken by simply rebooting a node from Uyuni/SUSE Manager.
```
paususe: Not worse than what happens now :-)

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+
+Uyuni/SUSE Manager will offer the possibility to manage the nodes that comprise a CaaS Platform cluster.
+Uyuni/SUSE Manager will register and show the status of each node of the cluster.
+
+# Detailed design
+[design]: #detailed-design
+
+## Node status reporting
+
+All the nodes in the cluster can be registered to Uyuni/SUSE Manager using bootstrapping by the already known methods.
```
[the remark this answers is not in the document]
mbologna: No, we are talking about two different things:
- AutoYaST will be used for **deploying** CaaSP
- in the RFC we are talking about **registering** systems to Uyuni/SUSE Manager. The already known methods are the ones outlined in the manual: https://opensource.suse.com/doc-susemanager/suse-manager/client-configuration/registration-overview.html

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Summary
+[summary]: #summary
+
+Initial integration for managing a CaaS Platform cluster from Uyuni/SUSE Manager.
+
+# Motivation
+[motivation]: #motivation
+
+Uyuni/SUSE Manager will offer the possibility to manage the nodes that comprise a CaaS Platform cluster.
```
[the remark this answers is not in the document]
mbologna: > Even with the reduced scope for the next patch release, will Uyuni be able to do that for CaaSP? Don't we depend on skuba, which cannot be installed on openSUSE?

For everything described in this RFC we are *not* going to use `skuba`: Uyuni can manage the CaaSP nodes.

> Also, I would mention somewhere (probably between Motivation and Detailed Design) what we want to achieve now and how, i. e:
> 
>     * Update our current Kubernetes integration to work with CaaSP 4
> 
>     * Document how to deploy CaaSP using AutoYaST
> 
>     * Report node status (salt-minion)
> 
>     * Stretch goal: salt-blackout
> 
>     * Stretch goal: document how to set up a container registry and mirror SUSE’s?

OK I will add this.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,74 @@
...
+Uyuni/SUSE Manager will register and show the status of each node of the cluster.
+
+# Detailed design
+[design]: #detailed-design
+
+## Node status reporting
+
+All the nodes in the cluster can be registered to Uyuni/SUSE Manager using bootstrapping by the already known methods.
+The added benefit will be that the user can see from Uyuni/SUSE Manager if a node has packages updates and patch available, along with all the information that a Salt minion is presenting when registered to Uyuni/SUSE Manager (e.g. CVE auditing).
+Another benefit is that the user can associate staged channels to the nodes registered (along with an activation key) to associate channels to the nodes of the cluster.
```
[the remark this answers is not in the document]
mbologna: Makes sense, I will add it.

accepted/0000-caasp-integration.md
mbologna: OK

accepted/0000-caasp-integration.md
mbologna: OK

accepted/0000-caasp-integration.md
mbologna: OK

accepted/0000-caasp-integration.md
mbologna: OK

accepted/0000-caasp-integration.md
mbologna: OK

accepted/0000-caasp-integration.md
mbologna: No, the CaaSP does not provide any documentation on how to deal with AutoYaST in SUSE Manager: we should provide it.

accepted/0000-caasp-integration.md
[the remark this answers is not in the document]
mbologna: OK

mbologna:
PR open: pinging @flavio @kkaempf @ereslibre : please review it!

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+[motivation]: #motivation
+
+Uyuni/SUSE Manager is the systems management tool and must offer to the end-user the possibility to manage other SUSE products like CaaS Platform.
+Uyuni/SUSE Manager already supports Kubernetes cluster integration. It works with CaaS Platform version 3 (container engine: Docker).
+
+In this regard, this RFC will cover the following aspects:
+
+- Node status reporting
+- Update the current Kubernetes integration of Uyuni/SUSE Manager to work with CaaS Platform version 4 while still supporting CaaS Platform version 3
+- Document how to deploy CaaS Platform using AutoYaST
```
kkaempf: This should be three different steps/iterations instead of one. :-/

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+- Another benefit is that the user can associate staged channels to the nodes registered (along with an activation key) to associate channels to the nodes of the cluster.
+
+The agent `skuba-update`, running on each node of the cluster, automatically picks up the packages from the associated channels and proceed to automatically update the nodes.
+NOTE: `skuba-update` is **NOT** invoked by Uyuni/SUSE Manager and it is running autonomously on the nodes of the CaaS Platform cluster.
+
+For example, an organization could have a "staging" cluster with a set of staging channels. The operator would cherry-pick Kubernetes updates from the SCC channels into the staging ones and then wait for skuba-update to roll them out into the staging cluster. At a later time, once testing is green inside of the staging cluster, the packages can be promoted to the "production channels" and be rolled out to the production cluster.
+
+It is paramount to state in the user interface and the documentation that the information presented for the minion is in read-only mode.
```
kkaempf: "read-only mode" sounds like a simple by-product while I'd guess it's a major effort to mark individual systems as read-only in Uyuni ?!

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+If a user reboots the node, installs a patch or installs a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) the Kubernetes may break.
```
kkaempf: How does Uyuni 'prevent stupid' here ?

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+
+NOTE: both of the above actions will not cluster-coordinated, so it is entirely possible that a user assigns different channels, install packages or reboots nodes in a subset of nodes belonging to the same cluster.
+
+## Update the current Kubernetes integration
+
+Under Virtual Host Manager, there is an already Kubernetes integration that supports CaaS Platform version 3.
+The current implementation breaks with CaaS Platform version 4: the Salt runner needs to take into account that CaaS Platform version 4 switched from Docker to CRI-O.
+Uyuni/SUSE Manager must check if the underlying container engine is running Docker (and use the existing Kubernetes integration) or CRI-O (and adapt the current implementation to support CRI-O).
+The Kubernetes implementation that this RFC is targeting is the one in CaaS Platform version 4: 1.15.2. Other Kubernetes distributions will not be targeted, at least initially.
```
kkaempf: Please target k8s 1.16. CaaSP will likely update to 1.16 before this RFC get implemented.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Documentation
+
+In addition to all the above, we are enriching our documentation on the CaaS Platform deployment. In the following sections, we are going to describe what we should introduce in the documentation.
+
+### AutoYaST Deployment documentation
+
+CaaS Platform provides AutoYaST templates for deploying a standard cluster (in the `SUSE-CaaSP-Management` package) comprising of:
+
+- 2 Load Balancers
```
kkaempf: Mhh, you might want to leave out load balancers for the initial mvp. caasp does not ship load balancers.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+- 3 Masters
+- 3 Workers
```
kkaempf: The [minimum supportable cluster](https://github.com/SUSE/doc-caasp/blob/master/adoc/deployment-sysreqs.adoc) is 3 masters and 2 workers.

accepted/0000-caasp-integration.md
kkaempf: CaaSP does provide an AutoYaST template/sample for masters and workers.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+
+In the Uyuni/SUSE Manager documentation, we need to provide a step-by-step guide on how to customize the AutoYaST template, import it in Uyuni/SUSE Manager and deploy it to bare-metal machines.
+
+## Next steps
+
+Possible next steps are presented as additional "stretch goals": if time permits, those goals can be achieved in the scope of the initial integration.
+
+### Minion blackout configuration
+
+If the [minion blackout](https://docs.saltstack.com/en/latest/topics/blackout/) is feasible, we can set to blackout every minion registered to Uyuni/SUSE Manager that is part of a Kubernetes or CaaS Platform cluster. Every minion that is in blackout mode will reject all incoming commands.
```
kkaempf: "[all remote execution commands are disabled](https://docs.saltstack.com/en/latest/topics/blackout/)" does not sound like a great idea. I would certainly want to run check or debug tools.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Container registry documentation
+
+In the documentation, we can write a guide on how to set up a container registry (using a standard container registry or Portus) and mirror SUSE container registry.
```
kkaempf: That should be covered in CaaSP docs ?!.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Management node
+
+As part of the support for the CaaS Platform version 4, SUSE Manager can elect (via the Add-On System Type) a node as "Management Node" for the cluster. This node must run SUSE Linux Enterprise Server and have `skuba` (a tool packaged by CaaS Platform) to deploy and manage the node.
```
kkaempf: Don't use `node` here, better use `client` or `system`. A `node` is part of the cluster (master or worker). But in CaaSP v4 the 'admin' system is outside of the cluster.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+Alternatively, we can choose to run `skuba` on the SUSE Manager server, we should add it to the packaging pattern of SUSE Manager server. The package is Go-vendored, no other dependency is required.
```
kkaempf: Please check back with consultants if shell access to SUSE Manager for k8s cluster admins is feasible.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,102 @@
...
+
+If a user does not read the documentation, a Kubernetes cluster can be broken by simply rebooting a node from Uyuni/SUSE Manager.
+
+# Alternatives
+[alternatives]: #alternatives
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- CaaS Platform: for Container registry do you suggest a standard container registry or Portus?
```
kkaempf: Portus _is_ the standard container registry for SLE ;-)

accepted/0000-caasp-integration.md
paususe: > This should be three different steps/iterations instead of one. :-/

Do you mean shipping those 3 features in 3 different MUs? Why?

Or implementing this in 3 steps, that may or may not ship in the same MU?

accepted/0000-caasp-integration.md
mbologna: Read-only for the user: IOW if you do execute an action on a minion that is part of a node cluster, you are going to break it. Making them _really_ read-only can be done via minion blackout or other meanings.

accepted/0000-caasp-integration.md
mbologna: For the initial iteration, we thought of a UI message and documentation. In the next iterations, we might want to consider an approach like minion blackout.

accepted/0000-caasp-integration.md
mbologna: Fixed

accepted/0000-caasp-integration.md
mbologna: Fixed

accepted/0000-caasp-integration.md
mbologna: Fixed (this might need to be corrected on the CaaSP side https://documentation.suse.com/suse-caasp/4/single-html/caasp-deployment/#_provisioning_the_cluster_nodes)

accepted/0000-caasp-integration.md
mbologna: We can either follow two approaches in this case:
1. Remove the blackout when we need to debug
2. Whitelist the additional functions that will be allowed during blackout for debugging.

Adding this note.

accepted/0000-caasp-integration.md
mbologna: Is this what you are referring to? https://github.com/SUSE/doc-caasp/blob/master/adoc/admin-crio-registries.adoc#mirror

accepted/0000-caasp-integration.md
mbologna: Fixed

accepted/0000-caasp-integration.md
mbologna: @paususe do you have any champion consultant to suggest here?

accepted/0000-caasp-integration.md
mbologna: Fixed

accepted/0000-caasp-integration.md
paususe: @kkaempf 

I'm not sure I understand the requirement. 

Do you mean we should create a local account in the SUSE Manager Server for each k8s cluster admin? With what rights (e. g. groups)? Should they also have shell access to the Proxies?

accepted/0000-caasp-integration.md
kkaempf: Implement iteratively, ship early, ship often. Each step above represent a product value increment which you can get customer feedback about.

accepted/0000-caasp-integration.md
kkaempf: imho it's sufficient to make cluster nodes _read only_ on the UI. We can't prevent someone to ssh (as root) into a node, and we shouldn't over-engineer here.
But Uyuni should certainly "know" that package installs, removals, or updates cannot be scheduled on individual cluster nodes.

accepted/0000-caasp-integration.md
kkaempf: :+1: for ui message and documentation. I'm not convinced that minion blackout is the right thing.

accepted/0000-caasp-integration.md
kkaempf: If you have `skuba` on a Uyuni server, the caasp admin needs shell access to the Uyuni server in order to access and run `skuba`.
However, shell access to the Uyuni server could also be used for other (malicious) things beyond cluster management. And it breaks separation of concerns.

mbologna: Yes, I agree: the Management Node must be the server with `skuba` instead of the Uyuni/SUSE Manager one.

accepted/0000-caasp-integration.md
mbologna: I clarified the sentence.

accepted/0000-caasp-integration.md
mbologna: Added the above to the alternative section, with the reason of why we don't want `skuba` to run on the SUSE Manager server.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,106 @@
...
+- Node status reporting
+- Update the current Kubernetes integration of Uyuni/SUSE Manager to work with CaaS Platform version 4 while still supporting CaaS Platform version 3
+- Document how to deploy CaaS Platform using AutoYaST
+
+# Detailed design
+[design]: #detailed-design
+
+## Node status reporting
+
+All the nodes in the cluster can be registered to Uyuni/SUSE Manager using [bootstrapping methods](https://opensource.suse.com/doc-susemanager/suse-manager/client-configuration/registration-overview.html).
```
moio: Maybe I am missing context here, are any code changes expected on the Manager side to support this, at all?

mbologna: We are not expecting any major code change but we are not sure yet: what if the Salt registration commands install software that conflicts with `kubectl` and related? Some buffer is required here.

kkaempf: @flavio successfully bootstrapped CaaSP v4 nodes with SUSE Manager. Besides that, `kubectl` (and other k8s code) is written in `go` (aka "statically linked binaries"). The odds of conflict are pretty low.

kkaempf: > Maybe I am missing context here, are any code changes expected on the Manager side to support this, at all?

Not for the initial iteration ('mvp'). Later I would hope that Manager implements "read-only" clients, to prevent changes of individual nodes.

joesusecom: I'm with Klaus here. The initial scope should really be to make sure all the templates and resources needed (e.g. AutoYaST XML, repos) work with SUSE Manager and are available for consumption and that the process is properly documented.
We need that part ASAP (December release) and should then start working on the improvements that require code changes in Uyuni/Manager 4.1 alpha/beta in iterative steps.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,106 @@
...
+If a user reboots the node, installs a patch or installs a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) the Kubernetes may break.
+
+NOTE: both of the above actions will not cluster-coordinated, so it is entirely possible that a user assigns different channels, install packages or reboots nodes in a subset of nodes belonging to the same cluster.
+
+## Update the current Kubernetes integration
+
+Under Virtual Host Manager, there is an already Kubernetes integration that supports CaaS Platform version 3.
+The current implementation breaks with CaaS Platform version 4: the Salt runner needs to take into account that CaaS Platform version 4 switched from Docker to CRI-O.
+Uyuni/SUSE Manager must check if the underlying container engine is running Docker (and use the existing Kubernetes integration) or CRI-O (and adapt the current implementation to support CRI-O).
+The Kubernetes implementation that this RFC is targeting is the one in CaaS Platform version 4: 1.16. Other Kubernetes distributions will not be targeted, at least initially.
```
flavio: It's probably out of the scope of this RFC, but I would be interested to know why you're going so "low level" on the kubernetes cluster to talk straight with the container runtime. I'm pretty sure you can operate at a higher level.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,106 @@
...
+
+Into the Virtual Host Manager, when a Kubernetes cluster is selected, the properties and the names of the nodes are shown, together with CPU Arch, CPU Sockets and RAM of the node.
+
+After the user has imported the `kubeconfig`, obtained from CaaS Platform, the process would be:
+
+- When a user registers a new cluster, we introduce a new workflow to bootstrap (using Systems > Bootstrapping) each node of the cluster
```
flavio: What do you mean with "register a new cluster"? I thought that was the result of bootstraping each node of the cluster

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,106 @@
...
+[drawbacks]: #drawbacks
+
+If a user does not read the documentation, a Kubernetes cluster can be broken by simply rebooting a node from Uyuni/SUSE Manager.
+
+# Alternatives
+[alternatives]: #alternatives
+We can choose to run `skuba` on the SUSE Manager server, we should add it to the packaging pattern of SUSE Manager server. The package is Go-vendored, no other dependency is required.
+The problem of this approach is that the CaaS Platform admin needs shell access to the Uyuni/SUSE Manager server in order to access and run `skuba`.
+However, shell access to the Uyuni/SUSE Manager server could also be used for other (malicious) things beyond cluster management. And it breaks separation of concerns.
+For the reasons above, this approach was not followed.
```
flavio: I see, I would still allow SUMA node to be selected by the user as "CaaSP management node" though.

accepted/0000-caasp-integration.md
moio: > We are not expecting any major code change but we are not sure yet: what if the Salt registration commands install software that conflicts with kubectl and related? Some buffer is required here.

I understand that.

My POV is that any RFC should describe an implementation plan with the main goal being to guide the developer who will receive it through implementation. (A second, not-less-important objective is to present explored alternatives with their pros and cons).

So my suggestion in this particular case would be to limit the main message about this section to something like "only testing is needed" - and to reframe that all other (important!) information here is as context.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Node status reporting
+
+All the nodes in the cluster can be registered to Uyuni/SUSE Manager using [bootstrapping methods](https://opensource.suse.com/doc-susemanager/suse-manager/client-configuration/registration-overview.html).
+This brings two major benefits:
+
+- The user can see from Uyuni/SUSE Manager if a node has packages updates and patch available, along with all the information that a Salt minion is presenting when registered to Uyuni/SUSE Manager (e.g. CVE auditing)
+- Another benefit is that the user can associate staged channels to the nodes registered (along with an activation key) to associate channels to the nodes of the cluster.
+
+The agent `skuba-update`, running on each node of the cluster, automatically picks up the packages from the associated channels and proceed to automatically update the nodes.
```
ereslibre:
```suggestion
The agent `skuba-update`, running on each node of the cluster, automatically picks up the packages from the associated channels and proceed to automatically patch the nodes.
```

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,106 @@
...
+NOTE: `skuba-update` is **NOT** invoked by Uyuni/SUSE Manager and it is running autonomously on the nodes of the CaaS Platform cluster.
+
+For example, an organization could have a "staging" cluster with a set of staging channels. The operator would cherry-pick Kubernetes updates from the SCC channels into the staging ones and then wait for skuba-update to roll them out into the staging cluster. At a later time, once testing is green inside of the staging cluster, the packages can be promoted to the "production channels" and be rolled out to the production cluster.
+
+It is paramount to state in the documentation that the information presented for the minion is in read-only mode. In the initial iteration, Uyuni/SUSE Manager must point out in the UI that executing actions (package install, upgrade, reboot) might break the cluster if the action are run on a node of the cluster.
+If a user reboots the node, installs a patch or installs a package related to Kubernetes (`kubernetes-kubeadm kubernetes-kubelet kubernetes-client cri-o cni-plugins`) the Kubernetes may break.
```
ereslibre: Rebooting the nodes is a safe operation; Kubernetes won't break (although it can be temporarily unavailable depending on what and how many nodes are being rebooted simultaneously). Installing or upgrading packages on their own is a different story.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,106 @@
...
+Uyuni/SUSE Manager must check if the underlying container engine is running Docker (and use the existing Kubernetes integration) or CRI-O (and adapt the current implementation to support CRI-O).
+The Kubernetes implementation that this RFC is targeting is the one in CaaS Platform version 4: 1.16. Other Kubernetes distributions will not be targeted, at least initially.
+
+Into the Virtual Host Manager, when a Kubernetes cluster is selected, the properties and the names of the nodes are shown, together with CPU Arch, CPU Sockets and RAM of the node.
+
+After the user has imported the `kubeconfig`, obtained from CaaS Platform, the process would be:
+
+- When a user registers a new cluster, we introduce a new workflow to bootstrap (using Systems > Bootstrapping) each node of the cluster
+- When every cluster node is registered, a system group with the same name of the cluster is created and all registered systems are moved into this system group
+- Each node presented the Nodes information is a link to the system overview page of that node
```
ereslibre: I don't understand this process, looks like first you import a `kubeconfig` and then bootstrap? The `kubeconfig` is a result of bootstrapping itself, and either you could import an existing `kubeconfig` file, so you can talk to a Kubernetes cluster, or you bootstrap a new cluster and then download the `kubeconfig`.

accepted/0000-caasp-integration.md
```diff
@@ -0,0 +1,106 @@
...
+## Documentation
+
+In addition to all the above, we are enriching our documentation on the CaaS Platform deployment. In the following sections, we are going to describe what we should introduce in the documentation.
+
+### AutoYaST Deployment documentation
+
+CaaS Platform provides AutoYaST templates for deploying a standard cluster (in the `SUSE-CaaSP-Management` package) comprising of:
+
+- 3 Masters
+- 2 Workers
```
ereslibre: Are we specifying this?

accepted/0000-caasp-integration.md
ereslibre: Broken seems definitive to me, it can be temporarily unavailable (until the machine is back), depending on how many machines get rebooted and what their role is.

accepted/0000-caasp-integration.md
mbologna: What are the tools that we can use at a higher level that we can use?

accepted/0000-caasp-integration.md
mbologna: I `s/register/defines` - would that clarify?

accepted/0000-caasp-integration.md
mbologna: Fixed

accepted/0000-caasp-integration.md
mbologna: I clarified the sentence: what we call "Boostrapping" is not related to Kubernetes but is rather the action to register a system to Uyuni/SUSE Manager. We assume that the bootstrap of the Kubernetes cluster is already done by an external actor.

accepted/0000-caasp-integration.md
mbologna: We can remove it, why do you think it's unnecessary to specify it?

ereslibre: Oh the wording led me to think that on CaaSP we provide AutoYaST templates for a 3+2, but I believe we are agnostic on the cluster size (was wondering where the cluster size was coming from).

mbologna: 3+2 comes from https://github.com/SUSE/doc-caasp/blob/master/adoc/deployment-sysreqs.adoc - does it make sense to remove it in light of this?

accepted/0000-caasp-integration.md
flavio: You should just talk with the kubernetes API server using one of the [client libraries](https://kubernetes.io/docs/reference/using-api/client-libraries/).

This will allow you to see all the pods running, look at the containers that are part of it and do any kind of interesting operation. There's no need to talk with the docker daemon running on each node.

accepted/0000-caasp-integration.md
flavio: Yep, that would clarify. So the steps would be:

1. User logs into SUMA, defines a new cluster by clicking some button
1. User creates the nodes using autoyast/terraform. Nodes trigger somehow the bootstrap script from SUMA
1. Profit!

Is that correct?

accepted/0000-caasp-integration.md
flavio: Yes, I think in some cases it would still make sense to have SUMA be also the node from which skuba has to be run. There are circumstances where the operator of SUMA is also the operator of the k8s cluster.

accepted/0000-caasp-integration.md
mbologna: Now I understand what you are referring to. What I am referring to is that, when Uyuni/SUSE Manager interfaces with the Kubernetes engine, a Salt runner queries the running containers on the cluster: https://bugzilla.suse.com/show_bug.cgi?id=1149741#c0

What I was suggesting to is to check: is the Salt runner returning `docker-pullable` containers? Then we are dealing with a CaaS Platform running Docker. Otherwise, CRI-O is running there.

The sentence is rephrased to clarify the concept. Are we on the same boat on this matter?

accepted/0000-caasp-integration.md
mbologna: The point 2 is not correct: the user creates the nodes using AutoYaST in Uyuni/SUSE Manager (following the documentation, see "AutoYaST Deployment documentation" in this RFC) or using Terraform (outside the scope of Uyuni/SUSE Manager).
When the nodes of the cluster have been bootstrapped (in the CaaSP domain), then the user imports the `kubeconfig` into SUSE Manager: the new workflow will bootstrap (in the Uyuni/SUSE Manager domain, read: register) the nodes into Uyuni/SUSE Manager.

mbologna: Pushed clarification in this sentence.

accepted/0000-caasp-integration.md
mbologna: IMHO auto-registering Uyuni/SUSE Manager to itself (installing and registering salt-minion on the same machine running the salt-master) might not probably be a good solution.
In these special cases, you are mentioning we can maybe add a note in the documentation saying that to fulfill these special needs users can install `skuba` on the SUSE Manager itself, provided they have shell access. What do you think?

joesusecom: Actually, most non-SUSE Manager Salt setups will do exactly that: Make the master a minion, too. I really want to see this going forward. Off topic for this particular RfC. But as long as we are careful with a few things (e.g. not let the Manager Server patch itself from its own repos), that would allow some simplifications that would make a lot of sense. E.g., currently we need to call salt-call on the master to use Salt for setting up the "self-monitoring" exporters.

accepted/0000-caasp-integration.md
kkaempf: Please start with the most simple case: An exisiting CaaSP cluster is added to SUSE Manager in order to control updates/patches visible to the nodes.
Deployment of a cluster (or new nodes) should be the second iteration of this feature.

mbologna:
Are we all on the same boat regarding the current status of the RFC? Can we have some green lights here, please?

kkaempf:
> Are we all on the same boat regarding the current status of the RFC? Can we have some green lights here, please?

What will happen if I green-light this. Will SUMA team then start boiling the ocean and try to implement the whole RFC ?

mbologna:
> > Are we all on the same boat regarding the current status of the RFC? Can we have some green lights here, please?
> 
> What will happen if I green-light this ? Will SUMA^wUyuni team then start boiling the ocean and try to implement the whole RFC ?

As soon as it's accepted, we will start the development of the features described:

1. Register the cluster systems to Uyuni/SUSE Manager (no read-only implementation!)
2. Update the current Kubernetes integration to work with CaaS Platform version 4
3. Document how to deploy CaaS Platform using AutoYaST

kkaempf:
>     1. Register the cluster systems to Uyuni/SUSE Manager (no read-only implementation!)

Oops, I don't think leaving out read-only is a good idea. (I can already see the L3s "I broke my cluster with SUSE Manager, help !")

> 
>     2. Update the current Kubernetes integration to work with CaaS Platform version 4
This should be no. 1.

> 
>     3. Document how to deploy CaaS Platform using AutoYaST
:+1:

mbologna:
> > ```
> > 1. Register the cluster systems to Uyuni/SUSE Manager (no read-only implementation!)
> > ```
> 
> Oops, I don't think leaving out read-only is a good idea. (I can already see the L3s "I broke my cluster with SUSE Manager, help !")

For the initial iteration, we thought just to state in the documentation and via UI that the user must not install packages/patches/reboot the node:

https://github.com/uyuni-project/uyuni-rfc/pull/14#discussion_r330735784

In the case of the L3 you are imagining: "I broke my cluster" -> "documentation/instructions not followed"

As a stretch-goal, the read-only solution might be implemented.

kkaempf:
> 
> As a stretch-goal, the read-only solution might be implemented.

Reach out to consulting, if they think doing the read-only as a strech goal is a good idea.

paususe:
> > As a stretch-goal, the read-only solution might be implemented.
> 
> Reach out to consulting, if they think doing the read-only as a strech goal is a good idea.

Our current Kubernetes support does not do read-only and we've been fine enough. 

We will surely do read-only as soon as possible but I don't think we should go "all or nothing": we'll go 1, 2, 3, 4, 5 as I said in https://github.com/uyuni-project/uyuni-rfc/pull/14#discussion_r329585060

I'm 99% sure we will not have all the 5 items for the next release. Maybe we'll get 1 and 2 in for the next release, then 3, 4, 5 for the release after that; maybe we can get 1, 2, 3 for the next release and 4, 5 for the release after that. We'll see. But we have to get the RFC approved and start working.

kkaempf:
> 
>     1. Update our current Kubernetes integration to work with CaaSP 4
> 
>     2. Document how to deploy CaaSP using AutoYaST
> 
>     3. Report node status (salt-minion)
> 
>     4. Stretch goal: salt-blackout
> 
>     5. Stretch goal: document how to set up a container registry and mirror SUSE’s?
> 

Well, this doesn't match [my priorities](https://github.com/uyuni-project/uyuni-rfc/pull/14#pullrequestreview-296151925) but it's your product.

mbologna:
> > Our current Kubernetes support does not do read-only and we've been fine enough.
> 
> I'm pretty sure it actually _is_ read-only. 😉

Just to give some context: the current Kubernetes integration is read-only. What we refer to when saying “read-only” is the registration type of the clients in Uyuni/SUSE Manager. At the time of writing, read-only client registration is not implemented neither with Kubernetes integration nor outside of it.

joesusecom:
Honestly, I think we've now spent more time discussing priorities in the RfC than it would have taken to actually get the (already working) AutoYaST deployment documented. That's really all I have in scope for the December release. :-)

Looking at Klaus' 2nd priority

> Ability to add an existing CaaSP cluster and get updates from SUSE Manager

Is there any code change required for that?

If yes, what do you have in mind?

I'd think this would be as simple as making sure you on-board all physical/virtual nodes of the CaaSP cluster to SUSE Manager, maybe with an activation key that makes sure they end up in the same system group?

mbologna:
> > Ability to add an existing CaaSP cluster and get updates from SUSE Manager
> 
> Is there any code change required for that?
> 
> If yes, what do you have in mind?
> 
> I'd think this would be as simple as making sure you on-board all physical/virtual nodes of the CaaSP cluster to SUSE Manager, maybe with an activation key that makes sure they end up in the same system group?

Yes, we are not expecting any code change and we plan to follow that you have just described - just testing is needed: https://github.com/uyuni-project/uyuni-rfc/pull/14/files#diff-3c7eb70251ad500b3e3ff5e2b6bd0db4R28

accepted/0000-caasp-integration.md
ereslibre: That's referring to the minimum number of nodes to have a failure tolerant system, not that we are providing templates for a cluster of this specific size, but it's a wording issue on the RFC I think, nothing else.

accepted/0000-caasp-integration.md
flavio: I looked into the bug. CRI-O (and podman) are more explicit about the source of the image compared to docker.

Docker assumes that, when no registry is specified, the image must come from the Docker Hub (`docker.io`). CRI-O is more explicit about that.

My question to you is: what is SUSE Manager expecting from the output of this salt runner?

Something tells me everything can be solved by changing the code that parses the runner's response.

mbologna: > I looked into the bug. CRI-O (and podman) are more explicit about the source of the image compared to docker.
> 
> Docker assumes that, when no registry is specified, the image must come from the Docker Hub (`docker.io`). CRI-O is more explicit about that.
> 
> My question to you is: what is SUSE Manager expecting from the output of this salt runner?
> 
> Something tells me everything can be solved by changing the code that parses the runner's response.

Yes, the solution would be to modify the Salt runner and add support for images that do not have the `docker-pullable` heading string. I do not think that would be a big deal.

mbologna:
I am really trying to make everyone happy and I decided to remove everything from this RFC except for the top priority: fix existing Kubernetes integration. Other RFCs with the previous content activities will follow.

This means that this document is just an RFC to fix the Bugzilla bug regarding the broken Kubernetes integration with CaaSP 4.
Can I have consensus on this RFC and see some greens, please?
@kkaempf @flavio @ereslibre @paususe

mbologna:
Consensus & reviews collected: opened https://github.com/SUSE/spacewalk/issues/9818 to start the development of the first iteration.
Thanks all for the valuable input!

PR #15, opened 2019-10-01
Content Lifecycles: Restore functionality

hustodemon:
Initial version of the restore functionality.

The picture might needs some update too.

### DIFF
--- a/accepted/00039-content-lifecycle-management.md
+++ b/accepted/00039-content-lifecycle-management.md
@@ -29,6 +29,8 @@ This RFC was inspired by best practices, Customer feedback and Katello.
 
 * *Content Build (CB)*: A snapshot of *Content* - optionally filtered - in an *Environment*.
 
+* *Content Snapshot* (*Snapshot*): Data containing a version and a description of a *Content Build* (e.g. channels, packages, errata, activation keys). Can be used for re-creating a *Content Build* later in the future (used for the "restore" functionality).
+
 * *Filter*: A filter to deny or allow different kind of *Content*.
 
 * *Environment*: AKA *Stage* or *Landscape* (`DEV`, `TEST`, `PRODUCTION`). Contains a version of a *Content Build*.
@@ -39,6 +41,7 @@ This RFC was inspired by best practices, Customer feedback and Katello.
 
 * *Promote*: *Content Build*s are copied along the *Lifecycle* to the different *Environments* defined in the *Content Project* (e.g. `DEV` to `TEST` to `PRODUCTION`). During *promotion* no filtering happens.
 
+* *Restore*: Restore *Content* of given *Environment* to the state described by given *Content Snapshot*. Again, no filtering happens.
 
 # Detailed design
 [design]: #detailed-design
@@ -53,8 +56,7 @@ Optionally *Filters* can be added to globally filter *Sources* content. *Filters
 Filters for Software Channels can be denylisting a package by name or version, or denylisting patches by date or type.
 The implementation should be generic to add other filters in future versions.
 
-The *Content Project* also defines the *Lifecycle*. Minimal one Environment must be specified. It should be possible to add as much *Environments* as needed
-in a row.
+The *Content Project* also defines the *Lifecycle*. Minimal one Environment must be specified. It should be possible to add as many *Environments* as needed in a row.
 
 *Sources* to attach could be Software Channels, Configuration Channels, Formulas, etc.
 For Software Channels *one* base channel and all or some of its children can be added to the *Content Project*.
@@ -76,7 +78,9 @@ Promote *Content Builds* is just copying it from one *Environment* to the next. 
 *Filter*s are applied only when *building* the *Content Build* for the first *Environment*.
 
 As the *Content* in *Sources* changes and also *Filters* can be added and removed, new *building* compiles a new version of the *Content Build*.
-A new version **overwrites** the older version. Only the latest version of a *Content View* in an *Environment* materializes on disk and can be used.
+A new version **overwrites** the older version. Only the latest version of a *Content Build* in an *Environment* materializes on disk and can be used.
+
+The user still has option to restore an *Environment* to a previous state(s) using the *restore* functionality and *Content Snapshot*s (see the [Restore functionality](#restore-functionality) section).
 
 If an *Environment* does not have a successor it should be possible to append a new one to the end of the *Lifecycle*.
 In case there is a successor, adding an *Environment* would insert it in between the current *Environment* and the current successor.
@@ -171,18 +175,44 @@ similar to the schedule definition of a repository sync.
 Adding a timer is optional.
 
 
-## Support restore
+## Restore functionality
 
-As the current design does not allow a rollback of an *Environment* to previous versions of the *Content Build* the only chance to restore an *Environment* is to use the one from one stage above (restore).
+Support limited *Content* version tracking and restoring (or rolling back).
 
-Example scenario:
-* 3 *Environment*s: `DEV`, `TEST`, `PRODUCTION`
-* Promoting *Content Build* from `DEV` to `TEST`
-* Testing the patches on systems in `TEST` fails, now we want to bring back the previous state:
+The most important entity is the *Snapshot*. It has a version and it contains all the information needed for re-creating a *Build* in particular state in a future:
+* software channels reference (also the pointer which channel is the base (or "leader) in the *Build*
+* packages per channel reference
+* errata per channel
+* others (after we implement support for them): activation key, configuration channels, ...
 
-The solution can be:
- * 'restore' from `PRODUCTION` to `TEST`
+### Create Content Snapshot (backup)
+*Snapshot* is created after a successful *Project* *build*. It is populated with the references to channels and their packages/errata based on the *build*. Note that the software channels references point to the **source** channels of a project, not to the built ones.
 
+### Restore Content Snapshot
+Restoring a *Snapshot* to a *Environment*.
+
+The data from the *Snapshot* is evaluated and the *Content* is re-created from it (e.g. the software channel tree is re-created in given environment and populated by packages and errata from the *Snapshot*).
+
+Technical note: For this, the adapted code for promoting *Environment*s could be used
+
+### Possible improvement: Using restore functionality for promoting
+The already existing *promote* functionality could be replaced by the *restore* code since it does the same.
+
+### Version management
+Keeping all versions of *Snapshot*s throughout the *Project* existence would be impractical for 2 reasons:
+* DB pollution (growing table for *Snapshot*)
+* integrity issues (it is likely that some entities referenced from very old *Snapshot*s will be deleted).
+
+The proposal is to keep only selected subset of the *Snapshot* versions per each *Environment*:
+* for the last *Environment*  in the *Path* (typically a "production" environment), keep 3 *Snapshot*s,
+* for other *Environment*s, keep 1 *Snapshot*.
+
+These numbers can be configurable via a SUSE Manager global option (alternatively, this could be overriden on a *Project* level).
+
+Some kind of "garbage collection" has to be run on *Project* operations to remove old *Snapshot* versions. The algorithm makes sure that the existing *Snapshot*s conform to the conditions above. It should be run on *build*, *promote* and *restore* actions.
+
+### Limitations
+The *Snapshot* only contains pointers to other entities (software channels, packages, errata, ...). When a referenced entity is removed by the user, the *restore* function will not work as expected.
 
 ## Removing an Environment
 
@@ -233,14 +263,6 @@ The *Content Project* wizard could be enhanced with template support.
 Enhance the managed *Content* with Configuration Channels. This will include also *Custom States* when they are implemented as Configuration Channels.
 
 Add more *Filter*s.
-
-
-# Drawbacks
-[drawbacks]: #drawbacks
-
-* No full history of content snapshots available. Only the current Environments exists.
-* No rollback to previous versions
-
 
 # Alternatives
 [alternatives]: #alternatives
--- a/accepted/00039-content-lifecycle-management.md
+++ b/accepted/00039-content-lifecycle-management.md
@@ -31,7 +31,7 @@ This RFC was inspired by best practices, Customer feedback and Katello.
 
 * *Content Snapshot* (*Snapshot*): Data containing a version and a description of a *Content Build* (e.g. channels, packages, errata, activation keys). Can be used for re-creating a *Content Build* later in the future (used for the "restore" functionality).
 
-* *Filter*: A filter to deny or allow different kind of *Content*.
+* *Filter*: 2 types of filters (allow and deny) to filter different kind of *Content*. The purpose of the "allow" filters is to override "deny" filters: content filtered out by a "deny" filter can be brought back using an "allow" filter.
 
 * *Environment*: AKA *Stage* or *Landscape* (`DEV`, `TEST`, `PRODUCTION`). Contains a version of a *Content Build*.
 
@@ -181,7 +181,7 @@ Support limited *Content* version tracking and restoring (or rolling back).
 
 Note: for Software Channels, the restore functionality does not roll back installed packages. It only affects contents of the channels.
 
-The most important entity is the *Snapshot*. It has a version and it contains all the information needed for re-creating a *Build* in particular state in a future:
+The most important entity is the *Snapshot*. It has a version and it contains all the information needed for re-creating a *Build* in particular state in the future:
 * software channels reference (also the pointer which channel is the base (or "leader) in the *Build*
 * packages per channel reference
 * errata per channel
--- a/accepted/00039-content-lifecycle-management.md
+++ b/accepted/00039-content-lifecycle-management.md
@@ -367,6 +367,8 @@ Using System Groups as Environments:
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
+* Q: Decide on how many *Snapshot*s per *Environment* do we want. See
+  the [Version Management](#version-management) section for more detail.
 
 # Resolved questions
 [resolved]: #resolved-questions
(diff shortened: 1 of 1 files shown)


accepted/00039-content-lifecycle-management.md
```diff
@@ -171,18 +175,44 @@ similar to the schedule definition of a repository sync.
...
+Support limited *Content* version tracking and restoring (or rolling back).
+
+The most important entity is the *Snapshot*. It has a version and it contains all the information needed for re-creating a *Build* in particular state in a future:
+* software channels reference (also the pointer which channel is the base (or "leader) in the *Build*
+* packages per channel reference
+* errata per channel
+* others (after we implement support for them): activation key, configuration channels, ...
+
+### Create Content Snapshot (backup)
+*Snapshot* is created after a successful *Project* *build*. It is populated with the references to channels and their packages/errata based on the *build*. Note that the software channels references point to the **source** channels of a project, not to the built ones.
```
ncounter: > Note that the software channels references point to the **source** channels of a project, not to the built ones.

While reading it, I feel I am missing a clarification about it, where *source* means *original*, and *built* means *generated by the current project for the built*. Or something like that.

accepted/00039-content-lifecycle-management.md
```diff
@@ -171,18 +175,44 @@ similar to the schedule definition of a repository sync.
...
 
-The solution can be:
- * 'restore' from `PRODUCTION` to `TEST`
+### Version management
+Keeping all versions of *Snapshot*s throughout the *Project* existence would be impractical for 2 reasons:
+* DB pollution (growing table for *Snapshot*)
+* integrity issues (it is likely that some entities referenced from very old *Snapshot*s will be deleted).
 
+The proposal is to keep only selected subset of the *Snapshot* versions per each *Environment*:
+* for the last *Environment*  in the *Path* (typically a "production" environment), keep 3 *Snapshot*s,
```
ncounter: Drawback of this nice differentiation of the *production* environment: what happen if a user already provides himself to create an additional environment at the end of the *Path* like an *archived* or *post-production* environment? I am afraid users would end up asking the possibility to *flag* somehow a certain environment as a *production* environment and keep 3 *Snapshot*s for it instead. I am more for keeping 1 for each environment, and adding a documentation best-practice of creating always an additional environment at the end of the *Path*. Combining the two I see more flexibility and we don't need to distinguish any environment in anyway.

accepted/00039-content-lifecycle-management.md
```diff
@@ -171,18 +175,44 @@ similar to the schedule definition of a repository sync.
...
+* for other *Environment*s, keep 1 *Snapshot*.
+
+These numbers can be configurable via a SUSE Manager global option (alternatively, this could be overriden on a *Project* level).
+
+Some kind of "garbage collection" has to be run on *Project* operations to remove old *Snapshot* versions. The algorithm makes sure that the existing *Snapshot*s conform to the conditions above. It should be run on *build*, *promote* and *restore* actions.
+
+### Limitations
+The *Snapshot* only contains pointers to other entities (software channels, packages, errata, ...). When a referenced entity is removed by the user, the *restore* function will not work as expected.
```
ncounter: Just as a reminder: this will need to be **huge bold and clear** in the documentation.

accepted/00039-content-lifecycle-management.md
hustodemon: I added an alternative proposal with your suggestion!

accepted/00039-content-lifecycle-management.md
hustodemon: Sure. We could also enhance at least the web UI with some warnings on channel deletion. Added.

joesusecom:
I've basically reviewed the resulting complete design document from scratch. So there might be comments that refer to older text.

**Motivation**

- We should mention rollback there as well.
- What's missing in the motivation paragraph is the idea that a stage should (in an ideal world) reflect _all_ aspects of a system's configuration, not just package versions. And that's why we want to add configuration channels etc. at a later point.

**Glossary**

_Content_: I think the statement about VMs is misleading. Yes, VMs are built from content stages, but at the same time it would certainly make sense to attach the same versioning to VMs in a VM repo that the stage it was built from has. And maybe even have "auto-triggers" like in OBS that let you determine that a new image should be built if there's a new version of the stage.

So, operationally, when you want to update a VM from version 1 to 2, you can either apply the latest updates/highstate or switch the base image to the matching new image (especially for "throw-away" workloads that are used like containers).

_Content Snapshot_: What about activation keys in this context? Are we going to version-control the contents of the key as well? If not, things could become messy.

**Managing Content**

The feedback I've got several times now is that the most important rollback/restore is from last working production to current production. We're currently missing that because once you've promoted to `PROD` , your previous version in `PROD` is gone.

One way of handling this is to add another stage/environment _after_ `PROD`, e.g. `ARCHIVE`, that stores at least one previous version and is only updated from version **x** to version **x+1** once `PROD` is updated from **x+1** to **x+2**, so at any point in time you can still roll back production to the last archived (and likely known to work) version.

I'm sure some customers may even want more comprehensive archives that allow them to go back in time. Not necessarily for production environments, but if you want to replicate an older setup for the labs.

What's not so clear form the whole picture (even though it's clarified later in the text) is that currently the versions, while still counted, actually get lost once you've promoted all environments to the next one. The way the boxes are drawn, one could assume that those boxes are all kept.

_Filters_: This text hasn't been updated to reflect the **whitelists** on top of blacklists.

> It is not possible to change the content between the Environments.

**_Is this technically true?_** While not encouraged, I could actually remove a package in, let's say, my current `QA` stage and then promote (clone) it to `PROD`. As `PROD` will be an actual clone of `QA`, the package will be missing in the resulting stage, right? (I've verified this for a package removal).

**Using Environments**

As of today, we don't have the activation key feature nor the auto-generated system groups. And the only way of finding out which systems are attached to a content project is to go to the _old_ channel listing and check which systems are subscribed to the channel.

**Content Profiles**

We don't have any of this yet. Does that concept even make sense? Or should we rather look into how we could use one project as a template for another one, and maybe allow syncing some actions (like adding a filter to the original project would ask you if you also want to update the filters for the copied project)?

We're covering some of that in:

**Releases vs. Staging**

Typo: "_Buildss_"

**Fast Track**

We should re-visit this. Yes, being able to manually clone individual patches to the fast track channels works, but isn't very _visible_ and could lead to missing certain servers.

**SP migration**

I have very little experience with how customers actually do SP migrations. But I think the problem starts earlier. Once there is a new SP out, we should allow for one of two things (not sure which one is better):
- make it really easy to basically create a full clone of a content project, just with the latest SP vendor channels
- instead, allow switching an _existing_ content project to a new SP, starting with the first stage. In that case, we'd have to somehow block "normal" updates for systems that are attached to the already migrated update channels until an SP migration has been run.

Especially if we choose to _not_ do the in-place update of an existing content project, we certainly should investigate allowing to re-apply an activation key (or rather using the activation key as the method to change the system's configuration consistently its the new "home"). This feature has long been overdue and would help with lots of other issues (e.g. losing all context when a system is moved to a different organization).

**Automatic channel archiving**

In the context of the "How to restore the `PROD` stage?" problem, wouldn't it potentially make more sense to put archives _into_ the _Content Project_? We'd just keep an archive around for every snapshot you want to keep (as "hidden" pseudo-stages). That way one could always clone back (restore) from any of the archived snapshots.

**Restore**

I think it should say "the future" (not "a future"). Let's hope there's only one. :-)

I'm a bit confused about the concept of a snapshot. Aren't we re-inventing a concept here that already exists? A tree of channels (with its attached packages and erratas) already _is_ basically a snapshot of the channel it was cloned from. What's missing is a "wrapper" around the channel that also tracks the non-RPM stuff and keeps it all consistent (e.g., has references to both the matching RPM channels and the matching configuration state channels). But maybe that's exactly what you mean?

I'm not sure I understand this part:

> Note that the software channels references point to the source channels of a project, not to the built ones.

The source channels change regularly when you re-sync from SCC. How can you use them as the base for staging?

**Restore Content Snapshot**

Again pretty confusing. How does this make more sense than just keeping a clone as an archive and re-cloning it back to an environment? Same with the next part. Why replace the _promote_ code with the new _restore_ code, not just _used_ the _promote_ code for restoring? As long as we're using "old-style" channels behind the scenes anyway, isn't _restore_ just a _promote_ where the input and output channels have different names? E.g. from "archive 1" to "prod" instead of from "qa" to "prod"?

**Version management**

Combine this with the requirement of keeping archive channels (see above), and you'll see that if the customer wants to have exactly that, you'll _have_ to keep those old ones and accept "database pollution" as part of normal operation.
I agree that customers should be able to define cleanup procedures, like how many archives to keep.
How would you run into integrity issues? My understanding is that the DB would never forget any RPM package that's still used in a channel. Not sure how the RPM store on disk handles this. Re-phrasing: If there's no working reference counting of sorts _today_, we should fix it and then rely on it. Because this could hurt us even outside the proposed restore workflows.
Next one is again confusing. Why would you need snapshots _per environment_? As long as you have _one_ complete channel clone that's on, let's say, your version 14, you can always use that to restore/back-promote it to _any_ stage/environment, right?
There's only one caveat, but that's also solveable: As long as we don't technically prevent customers from manually messing with the lifecycle-managed cloned channels, one could have inconsistencies between what's in the archive and what was in the environment before it was overwritten with a newer version.
This could be fixed by disabling manual channel modifications by default for lifecycle-managed channels.
As for how many versions/archives are kept: I think the "important" flag is more confusing then helpful. Doesn't make much sense from a workflow perspective.
I like a global default that can be overridden per _Project_. See above why I think there's no reason why you need per _Environment_ snapshots in the first place.
**Example scenario**
Again, I really don't get why that whole complexity with per-environment snapshots. Shouldn't it look more like this?

- When you initially build, you create a channel tree `my_project_V1`, _keep_ it, and then clone it to `my_project_DEV`.
- You later clone `DEV' to 'TEST' to `PROD` just as usual.
- Now you create a new build. Same procedure: First create a `my_project_V2`, then clone it to `DEV` etc.
- Because you still have the `V1` archive, you can always restore _any_ of the environments back to the V1 level.

Once you've reached the limit, you start deleting the oldest snapshots (which are really just named channel trees) when doing a new build.

Note that those archives wouldn't be just channel trees, but really behave like an extra stage/environment with its additional metadata (i.e. there would be a whole cloned channel tree per archive, but also the equivalent for other things we'd start staging, e.g. configuration channels).

**TBD**: What would happen if you manually modify one of the environment channels (and we haven't disabled that in the first place)? Should we trigger another archive clone, so that you can later go back to that modified version rather than the "clean" `Vx` one?

This setup also wouldn't have any problems with _colliding version numbers_.

Alternative approach: You could _not_ do the initial `V1`, `V2` etc. clones, but only do one at the end, based on your last environment in the project (usually `PROD`). Should work the same as long as channels don't get manually modified.

Now, I see why you'd _potentially_ want per-environment snapshots: If you really want to be able to go back in the history of one particular environment, not the "build" versions. But what's the use case for that? And wouldn't just "logging" a history of when you promoted what be fully sufficient, because as long as you have the required archives not garbage-collected yet, you could always look up a version in the history log and restore it back.

**Garbage Collection**

With the approach above (first one, not the alternative), you always handle cleanups at build time only. And the algorithm is very simple: Once you've reached the maximum number of archived versions, start removing the oldest when you create a new one during the build step.

If customers change the settings later, you can do a one-time extra cleanup of course.

**Limitations**

Ah, we should really not allow manually removing any stuff that's used in a snapshot/archive.

**Removing an Environment**

The sane solution is to require all systems to be migrated off (e.g. by using an activation key) to a different environment first.

accepted/00039-content-lifecycle-management.md
```diff
@@ -171,18 +175,113 @@ similar to the schedule definition of a repository sync.
...
-* Promoting *Content Build* from `DEV` to `TEST`
-* Testing the patches on systems in `TEST` fails, now we want to bring back the previous state:
+An alternative proposals:
+- allow users flagging *Environment*s with `important` flag, which would mean 3 snapshots (instead of 1) would be tracked for those, or
+- make the number of kept *Snapshot*s configurable per *Environment*.
 
-The solution can be:
- * 'restore' from `PRODUCTION` to `TEST`
+#### Garbage collection
+Some kind of "garbage collection" has to be run on *Project* operations to remove old *Snapshot* versions. The algorithm makes sure that the existing *Snapshot*s conform to the conditions above (it makes sure that for each *Environment* the wanted number of *Snapshot*s is kept and rest is removed). It should be run on *build*, *promote* and *restore* actions. For this, we need to track the *Snapshot*s used by each *Environment* (M-N relationship).
```
hustodemon: @mcalmer I think storing only the latest *Environment* in *Snapshot* as we discussed could be problematic since it lacks the information about order of the snapshots in given environment. We need to be able to reconstruct the order of the snapshots per environment, but the snapshots versions do not need to be strictly in descending order (see the PROD environment in the example below), so we can't rely on those. 
We need such order for 
1. garbage collection, and for
2. displaying the history to the user, so that can revert to "snapshot that was there before").
Keeping track of this with the solution of simple Snapshot -> Environment pointers (N-1) would make this hard.

We could store the information either in an M-N relationship (with explicit order parameter in the linking table) or using some kind of explicit "snapshot path" ("linked list" fashion).

joesusecom: See my earlier long comment. At the moment I'm not sure whether you aren't over-engineering things.
But I have a hard time distinguishing between data tables you may need for just keeping history and full versioned environments. The proposal is a bit confusing in that regard.
To repeat my assumption: I don't see an obvious reason why we'd need snapshots per environment in the first place. Snapshots _are_ like environments. At least when you look at the current scope (basically a channel tree with a naming convention plus a database entry that keeps track of the version). As long as you have, let's say, version 3 of your staging, stored _once_, you can use it to get _any_ environment restored to that version.
Let's discuss in person next week. This is hard to get right in GitHub comments. :-)

hustodemon: @joesusecom , there are 2 things to differentiate here:
- snapshots
- pointers to snapshots

Environments would only store *pointers* to snapshots, but the actual snapshots and their content won't be duplicated.

joesusecom: Ah, that clarifies things. Let's make sure that distinction (storing metadata/pointers vs. duplicating the actual snapshots) becomes clear.

hustodemon:
Pushed latest changes.

* About the using channels-as-snapshots: this was evaluated, discussed and declined with other devels. Channel tree snapshot will be modeled as a separate entity (with separate db table(s)).

* Some of the functionality suggested in the comments and in the email from 17th of October is related to this RFC, but I'd prefer not to put it here, but create new RFCs instead. The RFC is too bloated already. I tried to squeeze in only the stuff related to the restore functionality into this PR.

* The snapshot "garbage collection" topic: I tried to describe this according to my notes from the meeting, please review this. Honestly, the original method with keeping environments-snapshots references appears to me to be easier to understand than having bunch of important/unimportant snapshots. The main confusion I see with this is the lack of history of snapshots that was used in an environment (the used snapshot versions don't need to be in ascending order (if you juggle with restoring and promoting)).

hustodemon:
Still waiting for the review of the garbage collection part.

accepted/00039-content-lifecycle-management.md
```diff
@@ -171,18 +175,57 @@ similar to the schedule definition of a repository sync.
...
-The solution can be:
- * 'restore' from `PRODUCTION` to `TEST`
+### Version management
+Keeping all versions of *Snapshot*s throughout the *Project* existence would be impractical for 2 reasons:
+* DB pollution (growing table for *Snapshot*)
+* integrity issues (it is likely that some entities referenced from very old *Snapshot*s will be deleted).
 
+The proposal is to keep only selected subset of the *Snapshot*s. There are 2 kinds of snapshots: impotant ones and other (unimportant).
+
+By default, history of 3 important *Snapshot*s and 1 unimportant one is kept (these numbers can be overriden in Uyuni global settings). *Snapshot* of current state of *Environment*s are kept by default.
```
mcalmer: Keeping "3 important" is easy, as we take the overall number and remove the "oldest" or lowest version if we have more than 3.

Keeping "1 unimportant" is not so clear and we should specify it a bit more how the algorithm is.
The reason is, that we will have more than 1 unimportant snapshot as we need to keep all which are materialized in Environments.

I think we should say we keep snapshots in addition to the materialized snapshots in the environments.

This would mean for the algorithm:

- take all unimportant snapshots
- remove the versions which are build in one of the environments
- from the leftovers remove the oldest/lowest versions until we reach the configured number (e.g. 1)

(Same for the important snapshots.)

```suggestion
By default, history of 3 important *Snapshot*s and 1 unimportant one is kept in addition to the materialized versions (these numbers can be overriden in Uyuni global settings). *Snapshot* of current state of *Environment*s are kept by default.
```

accepted/00039-content-lifecycle-management.md
```diff
@@ -171,18 +175,57 @@ similar to the schedule definition of a repository sync.
...
+
+A *Snapshot* gains the *important* property if a corresponding version is promoted to the last *Environment*.
+
+A *Snapshot* is removed after a successful *build*/*promote* operation if the conditions for its existence don't hold true anymore (the number of kept important/unimportant *Snapshot* is exceeded).
```
mcalmer:
```suggestion
A *Snapshot* is removed after a successful *build*/*promote* operation if the conditions for its existence don't hold true anymore (the number of kept important/unimportant *Snapshot* is exceeded).

The algorithm to identify snapshots which should be removed should work like this.

For unimportant snapshots:
- make a list of all unimportant snapshots
- remove the versions which are build in one of the environments
- from the leftovers remove the all the snapshorts until we reach the configured number starting from the lowest version number.

For important snapshots:
- make a list of all important snapshots
- remove the version which is build in the last environment
- from the leftovers remove the all the snapshorts until we reach the configured number starting from the lowest version number.
```

hustodemon:
Thanks @mcalmer for the additional notes. I'm merging the PR finally.

PR #16, opened 2019-10-08
MVP for the integration of Yomi in Uyuni

aplanas:
Early RFC proposing the integration of Yomi in Uyuni.

Lets explore the options.


accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+
+The core of the proposal is to add Yomi into the catalog of formulas
+available in Uyuni.
+
+As Yomi is indeed a Salt formula, this integration can be done
+directly, installing the `yomi-formula` package into the Manager.
+
+The user can later decide to add the Yomi state into the `top.sls` for
+this system, fill the pillar data using the SaltForm provided in the
+package, and launch the installer by applying the highstate.
```
aaannz: I assume by `SaltForm` you mean `form.yml` or Formula form, right?

Note that as soon as formula is enabled for the registered system it will be automatically applied in the highstate (see `mgr_master_tops.py`) so no need for user to manually modify `top.sls`.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+some systemd overwrites that will help during the initial
+configuration, and the CLI tools required when a new installation is
+performed. You can find more information here:
+https://github.com/openSUSE/yomi#the-iso-image
+
+The current Yomi states are aware of differentiating when the system
+is booting from an external image (like the Yomi image), from when the
+boot comes from the local device. This is used to decide when the
+installation of the node makes sense or not when a highstate is
+re-applied.
```
aaannz: Saltboot has similar thing - when minion is running within saltboot inird, there is extra grain (`saltboot_initrd`) set to true. Is it done similarly for Yomi? If so, how about also changing the Uyuni registration handler to automatically apply yomi state to trigger the installation on yomi startup from initrd? (see [RegisterMinionEventMessageAction.java#L359](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java#L359) )

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+one:
+
+1) Boot the new node via the Yomi ISO image. The local DHCP server
+   assign a FQDN to the system, and this is used automatically as a
+   minion ID.
+
+   (note: the master address was indicated via a Grub command line
+   that an overwrite in the image can read and interpret, how the new
+   minion will automatically discover the master is an open question
+   for me yet)
```
aaannz: In Retail it is done that the saltboot initrd minion asks for `salt` CNAME resolution. Then this FQDN is used as master. This place a requirement on customers DNS to have a proper CNAME for `salt` configured. See [saltboot.sh#L111](https://github.com/uyuni-project/retail/blob/master/kiwi-desc-saltboot/saltboot/suse-SLES12/root/saltboot.sh#L111)

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+   the system.
+
+3) Assign to the new node all the Yomi states (Installer group),
+   fulfill the SaltForms and apply the highstate to the new node.
+
+Once the node reboot, the highstate can be re-applied without
+consequences, as the state detect early that the root filesystem
+device is already mounted.
+
+Some minor bugs in Uyuni (bsc#1150586, bsc#115089, bsc#115090,
```
aaannz: You are missing 5 in last two bug numbers 115089 -> 1150589 and 115090 -> 1150590

aaannz: I'll take a look at those bugs because it is interesting that I didn't notice behaviour you are describing there in our formulas. And somehow I was not aware of these bugs.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+bsc#1152673, bsc#1152767) makes this process less direct, but for all
+except bsc#1152673 Yomi is providing a workaround. The remaining bug
```
aaannz: For the bsc#1152673 we have a workaround in the [RegisterMinionEventMessageAction.java#L505](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java#L505)

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+
+(2) also exists for saltboot. But as of today it's just an API call,
+no UI version yet. Trivial to add. What's missing is applying an
+activation key to such a system as a template. As of today you'd have
+to inject the activation key grain on the minion side before
+registering the system to Manager.
+
+--<--<--
+
+## Disabling automatic highstate
```
aaannz: See https://github.com/SUSE/spacewalk/issues/2714

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+To potentially fix this issue, I propose in this RFC to have a grain
+emitted by the just-discovered system that will disable any automatic
+highstate application, if present. Only the highstate for this system
+can be manually applied:
+
+```yaml
+  automatic-highstate: no
+```
+
+## High-level forms?
```
aaannz: This will hopefully be fixed once I finish my work on https://github.com/SUSE/spacewalk/issues/9001 and https://github.com/SUSE/spacewalk/issues/8367

accepted/0000-yomi-initial-integration.md
aplanas: > I assume by `SaltForm` you mean `form.yml` or Formula form, right?

Correct. Is there better name to reference this feature? I can update the RFC to use the correct names.

> Note that as soon as formula is enabled for the registered system it will be automatically applied in the highstate (see `mgr_master_tops.py`) so no need for user to manually modify `top.sls`.

I see. I think that this problem is documented later. Maybe makes sense to have a different workflow for this case, and if the grains that is mentioned later is present, skip the automatic `state.highstate`?

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+2) Accept the Salt key in the manager node (Uyuni 4.0.2), and register
+   the system.
+
+3) Assign to the new node all the Yomi states (Installer group),
+   fulfill the SaltForms and apply the highstate to the new node.
+
+Once the node reboot, the highstate can be re-applied without
+consequences, as the state detect early that the root filesystem
+device is already mounted.
+
```
joesusecom: See above. In a _perfectly_ idempotent world, applying the high state after changing something in the installer states _would_ re-run those states. But in the current design, the Salt minion and its tools would probably not survive if you re-partition the disk, right? So if we want to allow triggering the re-installation without having to PXE boot again, we'll probably always have to either use something like MicroOS (where we can apply the new OS to the snapshot) or boot/kexec into a RAM disk before the installer states are kicked off.

accepted/0000-yomi-initial-integration.md
aplanas: Good question.

I try to make Yomi as much idempotent as possible, so I put in place a mechanism in the Yomi state that will detect from where it is booting. So, for example, if the root file system device (indicated in the pillar) is already mounted as the current root_fs, Yomi have nothing to do and no installation will be done during the highstate.

But if the roof_fs is not mounted and is available in the system, Yomi can apply the rest of the states.

So there is no grain to signalize if an installation must be done, as Yomi have all the information.

But interestingly enough I propose a new grain later, that will be use for Yomi to signal to the manager to not apply the highstare automatically.

accepted/0000-yomi-initial-integration.md
aplanas: Yes, as is described the last option is to use `master` as a valid address form the master, I can move it to `salt` instead, but I wonder if there is a better option, like adding some metadata during the DHCP resolution, that a service can read and parse, and configure the minion accordingly.

accepted/0000-yomi-initial-integration.md
aaannz: Absolutely. How this works is that when formula is enabled for the minion, then is is written to `/srv/susemanager/formula_data/minion_formulas.json` under minion id name. This file is that parsed by external pillar [suma_minion.py#L147](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/modules/pillar/suma_minion.py#L147) into the minion pillar. Manager tops module provide access to `/usr/share/susemanager/formulas/states/` directory where resides [formulas.sls](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/formulas/states/formulas.sls) which include all the formulas from this pillar. So anywhere in this chain can be added a jinja processing to selectively disable formula from highstate.
Another question however is if we wanted to do that from overall formulas strategy. For our image sync formula we solved that that our image sync state by default does nothing, only if enabled in formula it will include and apply actual state (see [image-synchronize/init.sls](https://github.com/uyuni-project/retail/blob/master/image-sync-formula/image-synchronize/init.sls))

accepted/0000-yomi-initial-integration.md
aplanas: Fixed! Thanks!

accepted/0000-yomi-initial-integration.md
aplanas: IIUC the highstate will succeed but the SLS rendering will continue to fail?

This somehow invalidate the manual application of the hs from the user, something that can be required later if the grains proposal is approved.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+big form, without defaults.
+
+To potentially fix this issue, I propose in this RFC to have a grain
+emitted by the just-discovered system that will disable any automatic
+highstate application, if present. Only the highstate for this system
+can be manually applied:
+
+```yaml
+  automatic-highstate: no
+```
```
joesusecom: I don't particularly like this part. Actually, I'd _expect_ the complete high state to be applied during installation (if that's possible in one go without reboot). The problem isn't _partially_ filled in pillar data. It's that without magic a newly added system has _no_ states and pillars assigned, unless you add global defaults somewhere (e.g. in a manually edited top.sls). This can be fixed by either pre-creating the exact system in the database and assigning all the required pillar data via Formulas with Forms (see above) and software channels (not possible yet?) or preparing defaults via system groups and then using an activation code to assign systems to their "templates". Let's double-check with the retail team how they are doing it. I guess they have hardware-specific groups for the defaults and do some magic with a reactor at install time to add the system to the right group, but I haven't checked yet.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+subvolumes for a BtrFS file system, use or not of snapper, partition
+layout, etc. You can express MicroOS or classical installations with
+those parameters, or create new mixtures of them tailored for specific
+use cases that are not directly supported by SUSE.
+
+A, maybe, better approach can be to provide some UI where the user can
+select the typical installation profile, similar to the skelcd-control
+from YaST. Based on that and the hardware information, a proposal of a
+pillar can be generated for this system.
+
```
joesusecom: You can use system groups for that.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+Is not clear to me how we can integrate high-level and low-level forms
+inside Uyuni, without making very confusing the usage of the
+interface. For example, if we decide to register the repositories from
+SLE via a high-level form, maybe is a good idea to provide a mechanism
+that will disable the low-level version of the form where the
+repositories and packages can be registered, to avoid possible
+conflicts.
+
```
joesusecom: Formulas assigned to an individual system can already override the defaults from the group. But we know that finding out where things are coming from is a bit messy and needs some re-work. Not just for this use case.

accepted/0000-yomi-initial-integration.md
aplanas: Yeah I know, that is what I replaced it (to avoid trivial hijacking of minions). I agree to revert to `salt` instead of `master` in the systemd overwrite that is executed before `salt-minion.service`, but I still think that there is a better solution that avoid default FQDN.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+  for different nodes. (IS THIS TRUE?)
+
+- For [open]SUSE based systems we can upload an AutoYaST file, but
+  this is usually designed for a single type on system (IS THIS TRUE?)
+
+- We can leverage Cobbler manage the PXE-Boot server and automate the
+  installation of new systems (via Kickstart, AutoYaST and Cheetah or
+  Jinja2 template engine)
+
+- SUMA Retailer allows the installation of new systems dumping a
```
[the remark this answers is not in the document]
aplanas: Sorry. Fixed.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+  partition image (generated by KIWI, for example) into the local root
```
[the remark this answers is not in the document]
aplanas: Fixed.

I was confused here, as Yomi can do a similar thing now, but the image can be generated manually too, with different filesystems and no template is required (I guess that the template is required because the SUMA expectations)

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+
+## Integrating Yomi as a Salt state
+
+The core of the proposal is to add Yomi into the catalog of formulas
+available in Uyuni.
+
+As Yomi is indeed a Salt formula, this integration can be done
+directly, installing the `yomi-formula` package into the Manager.
+
+The user can later decide to add the Yomi state into the `top.sls` for
```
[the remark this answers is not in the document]
aplanas: Fixed.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+Currently in Uyuni / SUMA we have several ways for installing a new
+system:
+
+- For RedHat based systems we can use the Web UI to create, update and
+  manage Kickstart files. Those files can be designed to be applied
+  for different nodes. (IS THIS TRUE?)
+
+- For [open]SUSE based systems we can upload an AutoYaST file, but
+  this is usually designed for a single type on system (IS THIS TRUE?)
+
```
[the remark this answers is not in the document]
aplanas: Right. I re-formulated the text.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+this system, fill the pillar data using the SaltForm provided in the
+package, and launch the installer by applying the highstate.
+
+## SaltForms and sub-forms
+
+Uyuni provides a mechanism, SaltForms, that allows the creation of the
+pillar data that suit the formula. The current relation between a
+SaltForm and a Salt state is 1:1, but using some empty `init.sls` we
+have the option of splitting a big SaltForm into multiple smaller
+SaltForms (subforms), that optionally belongs to the same group.
```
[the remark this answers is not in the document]
aplanas: Right. Added a note.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,361 @@
...
+   assign a FQDN to the system, and this is used automatically as a
+   minion ID.
+
+   (note: the master address was indicated via a Grub command line
+   that an overwrite in the image can read and interpret, how the new
+   minion will automatically discover the master is an open question
+   for me yet)
+
+2) Accept the Salt key in the manager node (Uyuni 4.0.2), and register
+   the system.
```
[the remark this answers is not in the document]
aplanas: As far I tested, it did not see any problem with the installed node. I compared the behavior from a pre-installed node (and later registered in SUMA), with the one that I installed from SUMA via Yomi.

I can be missing something.

accepted/0000-yomi-initial-integration.md
aplanas: > I don't particularly like this part. Actually, I'd _expect_ the complete high state to be applied during installation (if that's possible in one go without reboot). 

Actually I expect the same. I want the high state be applied during the installation. This is about a different issue: I do not want SUMA to schedule and automatic highstate for this node that I booted via PXE boot the Yomi image, and I want to install from scratch.

Also a full highstate have other consequences that are missing in the RFC (I will add those later): the main problem is that Yomi is talking about the to-be-chrooted environment, and the other states are talking about the system in place (that first is the Yomi image, and later will be the already installed system).

For example, if we take a new node, boot it via the Yomi image, register the salt-minion key in the manager, and register the bind and the yomi formulas, the bind one is talking about the Yomi image (the system that is in the memory), meanwhile the yomi formula is talking about the system that will be chrooted.

> The problem isn't _partially_ filled in pillar data. It's that without magic a newly added system has _no_ states and pillars assigned, unless you add global defaults somewhere (e.g. in a manually edited top.sls).

This confuses me. When I boot a system from the Yomi image, and I accept the minion key, I am able to assign some formulas from the catalog.

Not having any state and pillars pre-assigned is good (IMHO), as give the opportunity to manually assign the yomi formulas to bootstrap the installation.

But if the new system have some state and some pillar pre-assigned and one of them is Yomi with some generic (and incorrect) pillar data, this can be a problem.

In the default case the pillar data after the discovery is empty, so if meanwhile a highstate is automatically applied, I expect that the Yomi states will fail (but nothing bad will happen)

As meanwhile I have my pillar data in multiple forms, I also was expecting that meanwhile I am adding the next subform, an automatic highstate is applied in my system, maybe causing an undesired reboot.

This is why I propose a new grain: to avoid any automatic state.highstate.

If this mental model is not correct, I need to fix this ASAP.

> This can be fixed by either pre-creating the exact system in the database and assigning all the required pillar data via Formulas with Forms (see above) and software channels (not possible yet?) or preparing defaults via system groups and then using an activation code to assign systems to their "templates". Let's double-check with the retail team how they are doing it. I guess they have hardware-specific groups for the defaults and do some magic with a reactor at install time to add the system to the right group, but I haven't checked yet.

What is the difference between starting with no state and no pillar, assigning yomi from the formula catalog and filling the pillars, from pre-assigning the yomi formula and the adapted pillar data and later registering the system?

Sorry if I am a bit obtuse in this point, I am clearly missing some data.

accepted/0000-yomi-initial-integration.md
joesusecom: I'd always try to put logic onto the minion, not the master. If the Yomi state can figure things out, it should.

accepted/0000-yomi-initial-integration.md
joesusecom: To make things secure, we should ideally inject the master's public key into the RAM image and also enable strict checking. The image can also have the master's FQDN hardcoded. For non-PXE non-DNS/DHCP landscapes that's the only choice anyway.
Once we have to have managed DNS/DHCP anyway, I think just resolving the "salt" CNAME like saltboot is doing it isn't a bad approach.
We could of course also use SLP (Manager should still have that feature, which we never used) to find the Manager server on the network. But this doesn't sound more elegant than DNS.

accepted/0000-yomi-initial-integration.md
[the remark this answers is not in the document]
joesusecom: Oh, maybe it just works.

accepted/0000-yomi-initial-integration.md
joesusecom: As for not triggering another automated highstate: Those are all triggered from Java based on registration events if I understand things correctly. We can add logic there to determine whether we are in the installation phase or not.

As for assigning formulas to a to-be-deployed system: The thing is, that you may want to assign all configuration _before_ you install the system, not boot/register, then manually add formulas to the already registered system.

Of course you can _still_ just generically boot a node and then assign it a configuration. That's actually _exactly_ my first scenario, where the minion will automatically stay in a "wait" state, with the Salt minion running from the RAM disk, as long as the installation highstate isn't triggered. I just think that the more common case would be to plan things ahead (including host names, roles etc.) and pre-create those systems in the database. OR use an activation key grain. That way you can define the complete "template" server-side, but still allow injecting something on the client side that lets you choose the role.

Don't worry, we don't _usually_ apply automated regular. high states anywhere, only have that as an option during the initial onboarding. You don't need a grain to suppress it.

Some customers want scheduled highstates like in Puppet, but once we implement that, we can always deactivate it until the installation phase is over (or, as mentioned in my notes, introduce a separate state for installation).

The difference between starting with no state/pillar and starting with a pre-assigned configuration is just two different ways how customers work. We've had the same two patterns in SLEPOS as well: Some customers want to have all their systems pre-defined before the rollout (e.g. every single cash register, with host name defined by exact position in the store, matching to an exact MAC address). Others will want to just bring in a new box and determine last minute (e.g. with an activation key) what role it should get.

We can easily support both, as we already can create "empty" systems (at least via the API).

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+The prototype provide a set of subforms that cover all the pillar data
+that Yomi can understand. Maybe there is the option of hidding those
+forms and provide a simplified version, based of a single option like:
+(1) what kind of OS want to be installed, (2) simplified partition
+definition.
+
+The list of packages, repositories, service configuration and other
+elements will be described and configured by current parts of Uyuni,
+after the first boot.
+
```
joesusecom: The current UI already allows adding/removing forms, even as groups. With just a little bit of dependency handling we could offer _alternative_ versions like "simple installation" and "custom installation".

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+## Workflow
+
+In the 'Detailed design' section we pointed a problem with the naive
+workflow, and two alternatives were provided. Which one to choose, and
+how to implement it, is something that needs to be clarified before
+the approval of this RFC.
+
```
joesusecom: I don't think the two workflows are mutually exclusive. Just think of pre-creating the system in the UI and booting/registering an unknown system as two alternative first steps to tell SUSE Manager about a new node. The second step is then always the same: Run the installation state.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+## Overlap with different pillar and formula
+
+The pillar data for Yomi requires the definition of the language,
+keyboard layout to use, timezone, list of packages, etc. This can
+overlap with other formulas like 'locate', or some other components of
+Uyuni like 'Sofware'
+
+How to resolve this overlap is still an open question.
```
joesusecom: We should approach this like YaST modules are doing it. We can re-use the same formulas with forms for installation and later configuration changes, as long as there is logic in the states to determine whether they have to be applied to the changeroot or not.
Software selection is a beast we need to refactor completely in SUSE Manager. For example, we also have package states (which no-one seems to be using) and a field in the activation key form to add packages during activation. For now, a simple form will do.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+
+## Be ready for Redfish integration
+
+In the future maybe we will want to integrate more different ways of
+installation, like Redfish, where Yomi can have a role on
+it. Decisions done in this stage can influence the possibilities that
+we will have in the future.
```
joesusecom: In the first iteration, I don't expect too many complications with Redfish. Unless I'm missing something, Redfish will mainly allow for three things that could be useful for us in the Yomi scenario:

1. http(s) boot (technically not a Redfish feature, but all servers that support Redfish also have an http(s) boot enabled UEFI firmware, and you can use Redfish to put the server into http boot mode). In a nutshell, http boot is really just PXE boot with http instead of tftp. This will allow to directly boot the full image from http(s), without having to chain-load it via TFTP => iPXE. Larger images are allowed, and the transmission can be secured with TLS (provided that the admin correctly deploys the certificates via Redfish as well).
2. Set certain parameters on the system that can later be read from the RAM disk image (e.g. we could use that to deploy an activation key grain to the firmware I guess, or set a hostname that the image can pick up).
3. At least in some vendor extensions, you can directly point Redfish to a URL with the boot media, no need for PXE or http boot.

Finally, you can also switch the system on via Redfish and do firmware updates or configure RAID controllers.

If we really want to go all in with my idea of having a fully idempotent "deployment template" from firmware/RAID to boot to Yomi to "normal" high state, yes, we'll need to do some work (mostly in the Salt stack itself).

accepted/0000-yomi-initial-integration.md
aplanas: > To make things secure, we should ideally inject the master's public key into the RAM image and also enable strict checking.

I like the idea. I added it in the section "How the minion can find the master automatically"

> The image can also have the master's FQDN hardcoded. For non-PXE non-DNS/DHCP landscapes that's the only choice anyway.

In this case you can add the parameter 'master=master-address' in Grub, as is already in place.

> Once we have to have managed DNS/DHCP anyway, I think just resolving the "salt" CNAME like saltboot is doing it isn't a bad approach.

Yes, the image now is using 'salt' instead of master.

accepted/0000-yomi-initial-integration.md
aplanas: I added this information in the RFC

accepted/0000-yomi-initial-integration.md
aplanas: Very nice. I think that I understand how this can help. I added a note about this in the RFC.

accepted/0000-yomi-initial-integration.md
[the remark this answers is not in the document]
aplanas: > What happens if we mix "installation states" (e.g. partitioning) and "normal states" (e.g. adding and configuring an Apache web server) on the same system?

Uhmmm ... Do you mean mixing Yomi with other formulas? I write something about this in a different comment (https://github.com/uyuni-project/uyuni-rfc/pull/16#discussion_r332596880)

As is currently designed Yomi this is what will be happening:

* the yomi states knows about the chroot, and will provision the node and will install the packages described in the yomi pillar data inside the chroot, and will enable the services that are named inside the pillar data

* the apache-formula will install the rpm and enable the service on top on the yomi image, in the OS that is in RAM, not inside the chroot

> IMHO, the _desired_ behavior should be that you can idempotently call the high state at any point in time, and that all states in the high state are checked (and executed if required).

That is definitively desired. But I wonder how can we achieve this.

The big problem is that this 'normal state' will require to know about the chroot target, and point to it when is targeting some actions, like Yomi is doing when the software is installed:

https://github.com/openSUSE/yomi/blob/master/salt/yomi/software/pkgmanager.sls#L37

We have a module in place that help doing this transparently, that is also used by Yomi here:

https://github.com/openSUSE/yomi/blob/master/salt/yomi/users.sls#L31

Somehow, this apache formula needs to be aware that there is a chroot, and there is living.

And the relation between the apache formula and the Yomi states needs to be more deep than that. The chroot is only available during a period of time (not before the software stage and not after the post-installation state), so the apache formula needs to inject the states inside those phases.

> The only downside is that this could lead to accidental data loss, e.g. if someone changes a minor thing in a partition state (e.g. the label name of a partition) and triggers a complete re-installation by accident.
> An alternative would be to introduce the concept of a "super state" that runs "installation state" and "high state" (or even preceded by a "hardware prep" state). In that case, the "normal" high state would not include the Yomi states.

Ah yes. I think that is something like what this comment is proposing: https://github.com/uyuni-project/uyuni-rfc/pull/16#discussion_r332558510

As how the prototype is working, the pillar data and the yomi states can live during the full live cycle of the system. Yomi can detect if there is something for him or not. But you are right, in the current shape, this check enable or fully disable yomi. Internally the states are self indempotent, but if the is not rootfs to mount (because is already mounted), all the yomi states are discarded.

accepted/0000-yomi-initial-integration.md
aplanas: > See above. In a _perfectly_ idempotent world, applying the high state after changing something in the installer states _would_ re-run those states. But in the current design, the Salt minion and its tools would probably not survive if you re-partition the disk, right?

Uhm. If you change the partition schema, but you do not boot using the PXE boot image or the USB image from Yomi, nothing will happen.

But if you boot from the Yomi image, the current implementation of the partitioner will detect a mismatch between the pillar and the current layout, and will refuse to remove anything.

The plan is to add a resize / wipefs option in the pillar in yomi, that will clean the devices or try to resize the partitions and the file system if this mismatch is present.

The wipefs state was implemented some time ago, but outside of Yomi to lower the risk of loosing data:

https://github.com/openSUSE/yomi/blob/master/README.md#cleaning-the-disks

 So if we want to allow triggering the re-installation without having to PXE boot again, we'll probably always have to either use something like MicroOS (where we can apply the new OS to the snapshot) or boot/kexec into a RAM disk before the installer states are kicked off.

accepted/0000-yomi-initial-integration.md
aplanas: > As for assigning formulas to a to-be-deployed system: The thing is, that you may want to assign all configuration _before_ you install the system, not boot/register, then manually add formulas to the already registered system.

I understand. We want to define the full highstate (yomi and what is not yomi), and apply it in a single step during the provisioning. 

To achieve this we need to make the states that are not part of yomi aware of the chroot when this is present, and unaware of it when it is not present.

I added some new comments about this in the RFC, under the "Unresolved questions" section (Is Yomi part of the high state?). This is a big topic that was missing.

accepted/0000-yomi-initial-integration.md
aplanas: I added a note about saltclass. I think that can cover this case too.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+One of the drawback that I can identify of this proposal is the
+integration of yet one more installation mechanism : )
+
+From the user perspective can be confusing the existence of the
+Autoinstallation option, the interactions with AutoYaST, Cobbler and
+the different alternatives that Uyuni provides.
+
```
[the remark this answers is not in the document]
aplanas: Indeed. No replacement is proposed here.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+of code will set a minion ID for the system based on some rules:
+
+- Use the ID injected via the bootloader parameter 'minion='
+
+- If a FQDN is provided via DHCP, this name will be used as the ID
+
+- If no parameter or of the hostname is localhost, the MAC address
+  will be used as minion ID
+
+The question is to validate those rules for the general use case.
```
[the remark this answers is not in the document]
aplanas: Fixed!

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+In a similar way, an systemd overlay will take care of configuring
+where to find the salt-master node.
+
+- If the bootloader parameter 'master=' is provided, a new file
+  'master.conf' will be added inside '/etc/salt/minion.d'
+
+- If not, the default 'master' name will be used
+
+Is there a way to use DHCP to inject the correct address for the
+salt-master node?
```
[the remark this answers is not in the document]
aplanas: I updated the Yomi image and the documentation to follow this process.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+
+The current PXE Boot and ISO image contains a root user than can be
+accessed without password. The problem is that as the SSH service is
+enabled, you can remote login into the new system as root without
+password.
+
+One solution is to disable password-less login via SSH, and inject via
+DHCP some parameters that can enable the login, or inject authorized
+certificates.
+
```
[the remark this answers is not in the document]
aplanas: I also removed the DHCP suggestion. Is not a good idea.

I was thinking something like the metadata injection from cloud-init and ignition, where an agent living in the early stages of the boot process can read (from an URL, a TXT entry in a DNS, or some other mechanism) a JSON document that contains dynamic information related with users, certificates, the proper address of the master, and stuff like that.

I guess that is better to move it outside of the scope of the proposal.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+conflicts.
+
+## Better integration
+
+Another aspect to consider is that this proposal is moving the
+installation into the Salt Formula / State realm, and moving it out
+from the Systems / Autoinstallation section. Maybe future work can be
+done to integrate both sections, providing access to the pillar data
+that change the Yomi behavior inside this section.
+
```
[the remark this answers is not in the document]
aplanas: I added this as a comment in the 'Workflow' section

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,363 @@
...
+and integrate it into the Kickstart / AutoYaST panel.
+
+Internally will still be a Salt state driven by Salt, but we will
+provide a Java API similar to those other mechanisms already in place,
+that will be used to define the pillars that will be applied when a
+new autoinstallation / reinstallation is commanded.
+
+This will imply that the Yomi state is outside the highstate, and that
+this new API needs to be designed.
+
```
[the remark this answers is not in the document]
aplanas: Added!

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,420 @@
...
+system:
+
+- For RedHat based systems we can use the Web UI to create, update and
+  manage Kickstart files. Those files can be designed to be applied
+  for different nodes. (IS THIS TRUE?)
+
+- For [open]SUSE based systems we can upload an AutoYaST file designed
+  for a single type on system.
+
+- We can leverage Cobbler manage the PXE-Boot server and automate the
```
aaannz: Nitpick, I think you are missing 'to'? Leverage Cobble **to** manage

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,427 @@
...
+  installation of new systems (via Kickstart, AutoYaST and Cheetah or
+  Jinja2 template engine)
+
+- SUMA Manager for Retail allows the installation of new systems
```
aaannz: SUSE Manager, not SUMA Manager

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,427 @@
...
+  dumping a partition image (generated by KIWI from a given template)
+  into the local root partition. For this case the PXE boot stack will
+  be different that the one provided by Cobbler
```
aaannz: I don't know why the future tense. PXE boot stack is different for saltboot. The functionality was already shipped with SUMA 3.2 last year.
Maybe add a note that for it PXE is managed by other formulas provided with SUMA Retail.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,427 @@
...
+## Integrating Yomi as a Salt state
+
+The core of the proposal is to add Yomi into the catalog of formulas
+available in Uyuni.
+
+As Yomi is indeed a Salt formula, this integration can be done
+directly, installing the `yomi-formula` package into the Manager.
+
+The user can later decide to add the Yomi state into the
+auto-generated top file for this system, fill the pillar data using
```
aaannz: The wording seems strange to me. What does it means to add Yomi to the auto-generated top file? I know what you mean, but why not write 'The user can later decide to enable Yomi formula for registered system, ...'

accepted/0000-yomi-initial-integration.md
aplanas: This needs to be done also for forms that are living in a different formula, as today we are splitting them. Or it is working already? I mean, do the value on one form affect if another form is displayed (subform)

accepted/0000-yomi-initial-integration.md
aplanas: Updated!

accepted/0000-yomi-initial-integration.md
aaannz: Unfortunately no. Forms as currently implemented are independent regarding displaying data in UI. They can influence each other when their pillar data are merged ofc.

accepted/0000-yomi-initial-integration.md
aplanas: Fixed!

accepted/0000-yomi-initial-integration.md
aplanas: Ouch. Done!

accepted/0000-yomi-initial-integration.md
aplanas: Fixed.

accepted/0000-yomi-initial-integration.md
aplanas: Yes, your proposal is more clear. Replaced!

accepted/0000-yomi-initial-integration.md
aplanas: I integrated this description in the RFC.

> If we really want to go all in with my idea of having a fully idempotent "deployment template" from firmware/RAID to boot to Yomi to "normal" high state, yes, we'll need to do some work (mostly in the Salt stack itself).

I think that I get this idea, but do you have this written in any other place? I think that will help this RFC, and can impact a lot into Yomi

accepted/0000-yomi-initial-integration.md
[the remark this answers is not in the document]
joesusecom: I was waiting for a comment that mentions cloud-init or ignition. :-)

The thing is: With Salt you have a much more powerful system to start with. Using cloud-init, ignition, or even ssh to _configure_ Salt is a bit like using the bad tool to set up the good tool. :-)

So I strongly prefer a simple mechanism like plain DNS for bringing up the Salt minion-master connection early on and then using Salt for all the rest.

Of course it's a different story if you're using public cloud and cloud-init is the default for injecting stuff. But then you're most likely not going to use Yomi, right? You'll start a minion on the node and use "plain" Salt.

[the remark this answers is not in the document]
aplanas: > The thing is: With Salt you have a much more powerful system to start with. Using cloud-init, ignition, or even ssh to _configure_ Salt is a bit like using the bad tool to set up the good tool. :-)

Only for metadata that runs after the minion is connected, and we are talking about the metadata that is needed before the connection (where is the master, certificates, transient users that will be used for ssh, etc). Salt will not help here : (

> So I strongly prefer a simple mechanism like plain DNS for bringing up the Salt minion-master connection early on and then using Salt for all the rest.

Actually I agree. Is a good idea to move this topic out. I write a recommendation in the RFC to drop the topic.

accepted/0000-yomi-initial-integration.md
joesusecom: Not sure I understand your last comment. If we're using different naming, we'll obviously have to make it consistent. I'd put it differently: We shouldn't have a separate section on a Yomi form asking for language/timezone/keyboard if there _is_ a form for that already.

accepted/0000-yomi-initial-integration.md
joesusecom: I have prepared a few slides for a partner, trying to explain the concept. Will send them to you. But this idea certainly needs more work. Just to give you an idea: I have no clue whether that "master template" should actually run a highstate or whether it would be more useful to introduce an "ultra-highstate" (name TBD) that runs all three of (hardware prep, Yomi, normal highstate) as required, and you could still use them separately as well.

accepted/0000-yomi-initial-integration.md
aplanas: (Resolve this conversation, as some time ago I changed the mechanics of detection: https://github.com/openSUSE/yomi/commit/ccc1e406e066b6331adf3a512372903d59019f19, so not need of any external grain)

accepted/0000-yomi-initial-integration.md
aplanas: I will resolve this conversation, as Yomi is able to skip itself if the main device of the system is mounted, and the chroot cannot be done.

This skip the problem of breaking the system as a side effect of re-applying the highstate.

accepted/0000-yomi-initial-integration.md
aplanas: As this last bug was confirmed, we can wait for the resolution before editing this part of the proposal.

accepted/0000-yomi-initial-integration.md
aplanas: >  I'd put it differently: We shouldn't have a separate section on a Yomi form asking for language/timezone/keyboard if there is a form for that already.

No question about this.

aplanas:
I added a new commit were there is a TL:DR that try to summarize some of the actions that can be done.

accepted/0000-yomi-initial-integration.md
```diff
@@ -0,0 +1,488 @@
...
+from Uyuni.
+
+The big problem for this workflow has been pointed by Joachim Werner,
+and is about the default values for the pillar data. IIUC the problem
+is that if Yomi provides some default pillar configuration, the new
+discovered node can wrongly apply a highstate (that AFAIK are applied
+automatically or scheduled by the user) with non-optimal data (disk
+configuration, packages, repositories, etc)
+
+There are two proposals, also from Joe, to address this issue (copied
```
moio: As the spirit of an RFC is to submit a design proposal to the team in order to gather feedback and ultimately agree on a text that can be used as a guideline for the next implementation step.

I would thus suggest taking a decision here (and in all other cases where multiple alternatives are possible) and propose just one direction. The pros/cons wrt the other known possibilities should be described separately in the "Alternatives" section below.

Also note that an RFC doesn't necessarily need to be an all-encompassing design document over a big solutions with many unknowns still to discover - setting a general design direction and specifying "just enough" so that the first iteration can be implemented is sufficient. What I often see is a separate "Next Steps" paragraph, which is informative but not really binding to the acceptance of the RFC, or even subsequent RFCs (once the first "chunk" is implemented/feedback is gathered/next step is clearer).

aplanas: This is what I try do to in the areas that I have the knowledge to make an informed decisions. The places with open questions are to motivate the discussion.

But I will take a guess in the RFC and go for a single proposal.

aplanas: Done. I moved this part to the unresolved question, and I moved to a workflow that I think is more similar to the first one proposed.

meaksh:
This is ready to be merged. I'll proceed merging it tomorrow if there are no more requests here.

PR #18, opened 2019-10-10
Hub ServerGroups/Orgs/Users RFC

moio:
This RFC contains details on mechanisms around Users, Orgs, SystemGroups and permissions in the context of the Hub project.

### DIFF
--- a/accepted/00061-hub-users.md
+++ b/accepted/00061-hub-users.md
@@ -91,10 +91,11 @@ The Hub _gateway_ XMLRPC API ([RFC here](accepted/00062-hub-xmlrpc-api.md)) give
 - Access to the gateway API itself requires Hub-User credentials. Access to individual Server APIs to manage clients requires Server-User credentials
   - in worst case, N+1 pairs of credentials will be needed (one for the Hub, one per N Servers)
   - as a special convenience option, the Hub gateway API can use Hub-User credentials to authenticate against Servers. So if a certain username/password was uniformly configured across several Servers (this is easy to implement with the states above), Hub gateway API usage becomes simpler: only one log in is needed, with its Hub-User credentials. After that a list of Servers to work with will have to be specified. This authentication mode can be optionally activated via a parameter (eg. `relayAuthentication = True`) on the `login` method
+    - note: this only makes sense if it is deemed appropriate to use the same credentials of Server sysadmin users (Hub-Users) to also act on Clients (Server-Users). This might not always be the case
+    - note: if Server sysadmin credentials must be separate from Client sysadmin credentials are required to be separate, it is still possible to use this option by creating a Hub-Org with no Servers. Any Hub-User in this Hub-Org will be able to connect to the Hub gateway API, and have permissions on Clients (assuming Servers have Server-Users with same name and passwords, again possibly configured via the above states). Org trusts will be needed to be set up if those users also need to perform content management
   - as an additional special convenience option, the Hub gateway API can automatically select the list of Servers a Hub-User works with, based on the list of Servers that same Hub-User has permissions on. Once again this can be optionally activated via a parameter (eg. `autoConnect = True`)
-    - in this case only a single login with a pair of Hub-User credentials is needed, and then calls can be made directly
-    - Note that this only makes sense if the customer is OK with the idea of using Server sysadmin users to act on Clients
-    - If they prefer to manage Clients via users that have no permissions to manage Servers, they can create a Hub-Org with no Servers. Any Hub-User in this Hub-Org will be able to connect to the Hub gateway API, and possibly have permissions on Clients (if Servers have Server-Users with same name and passwords, again possibly configured via the above states). `autoConnect` won't be possible in those cases, at least initially
+    - note: in this case only a single login with a pair of Hub-User credentials is needed, and then calls can be made directly
+    - note: this only makes sense if it is deemed appropriate to use the same credentials of Server sysadmin users to also act on Clients. This might not always be the case
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/accepted/00061-hub-users.md
+++ b/accepted/00061-hub-users.md
@@ -121,6 +121,7 @@ All code would be new in new modules and formulas, so no change to existing comp
 
 # Next steps
 [Next steps]: #next-steps
+ - add an execution module to call a generic XMLRPC call, to make it possible to cover (potentially) any other area outside of this RFC
  - lift Limitations by adding new functionality to introduce UIs/APIs to centrally manage Users, Orgs, SystemGroups, permissions
 
 # Unresolved questions
--- a/accepted/00061-hub-users.md
+++ b/accepted/00061-hub-users.md
@@ -122,6 +122,7 @@ All code would be new in new modules and formulas, so no change to existing comp
 # Next steps
 [Next steps]: #next-steps
  - add an execution module to call a generic XMLRPC call, to make it possible to cover (potentially) any other area outside of this RFC
+ - add Form support to Formulas
  - lift Limitations by adding new functionality to introduce UIs/APIs to centrally manage Users, Orgs, SystemGroups, permissions
  - create states automatically from an existing Server
  
--- a/accepted/00061-hub-users.md
+++ b/accepted/00061-hub-users.md
@@ -123,7 +123,8 @@ All code would be new in new modules and formulas, so no change to existing comp
 [Next steps]: #next-steps
  - add an execution module to call a generic XMLRPC call, to make it possible to cover (potentially) any other area outside of this RFC
  - lift Limitations by adding new functionality to introduce UIs/APIs to centrally manage Users, Orgs, SystemGroups, permissions
-
+ - create states automatically from an existing Server
+ 
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
--- a/accepted/00061-hub-users.md
+++ b/accepted/00061-hub-users.md
@@ -1,0 +1,128 @@
+- Feature Name: hub_users
+- Start Date: 2019-10-10
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Uyuni Server has the concepts of System Group, Org, User and permissions. All of those assume there is one central management point for the whole infrastructure.
+
+With the Hub architecture, there will be multiple Servers - this RFC details how those concepts will work in that environment.
+
+See the [Hub general RFC](accepted/00060-hub-general.md) for an introduction to the Hub project.
+
+# Motivation
+[motivation]: #motivation
+
+End users might want to define security policies for the whole Uyuni infrastructure (Hub and Servers).
+
+Additionally, from the requirements:
+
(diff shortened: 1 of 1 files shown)


accepted/00061-hub-users.md
```diff
@@ -0,0 +1,128 @@
...
+  - as an additional special convenience option, the Hub gateway API can automatically select the list of Servers a Hub-User works with, based on the list of Servers that same Hub-User has permissions on. Once again this can be optionally activated via a parameter (eg. `autoConnect = True`)
+    - in this case only a single login with a pair of Hub-User credentials is needed, and then calls can be made directly
+    - Note that this only makes sense if the customer is OK with the idea of using Server sysadmin users to act on Clients
+    - If they prefer to manage Clients via users that have no permissions to manage Servers, they can create a Hub-Org with no Servers. Any Hub-User in this Hub-Org will be able to connect to the Hub gateway API, and possibly have permissions on Clients (if Servers have Server-Users with same name and passwords, again possibly configured via the above states). `autoConnect` won't be possible in those cases, at least initially
```
rjmateus: If we use this solution how can users manage Software channels for example, if they will be part of a different organization of the clients?

moio:
> While not part of the scope of this RfC (user/groups management), we should gradually extend this to cover more or less the complete XML-RPC API as relevant for pre-populating Uyuni (i.e. activation codes, system entries for to-be-deployed systems, assigning and filling in Formulas with Forms, ...).

More on this is likely come in an upcoming Hub RFC ("Server of Servers").

> Just like in the runner you've referenced, it would be possible to provide a generic API function (execution module only, but can be used from states) to catch features we don't have state module code for yet.

Agreed, added as "Next Step".

> One minor caveat that you should be aware of: Due to the way Uyuni handles unique ids, this will always only be "pseudo-idempotent": If you run your example state, then run a state that removes all those items (e.g. user.absent), then re-apply your state, you'll get all the users back, but with different unique DB ids!

We will keep that in mind as an implementation detail and make sure we don't re-create entities from scratch if not needed.

> I'd recommend in the docs that users should normally not apply advanced changes to their "slave" Uyuni servers, but rather wipe/redeploy from the template instead. Smaller (and incremental) changes like adding users/groups should be safe though.

We will see how well the mechanism works post-implementation and add recommendations to docs accordingly.

> This feature will be really helpful for setting up singleton Uyuni servers as well (including sumaform of course), as long as we either run the state with salt-call locally or (finally) allow the server to be registered against itself as a minion.

That's also in the area of the upcoming "Server of Servers" RFC.

> What would make this even more attractive is an export feature (shouldn't be _terribly_ hard to do) that queries the server via the XML-RPC API and creates the state file for you. So one could set up one server by example and then clone its setup (potentially slightly modifying or Jinja-ifying the template) for mass deployments.

Added as next step.

> I'm sure you've already thought about that, but the logical next step is to now take those state modules and build an Uyuni Formula that can be driven entirely from Pillar data, either scripted or even using Formulas with Forms in the UI.

Added as next step (right, we did think about that but forgotten in the final text).

Can you give your approval to this version of the text?

joesusecom:
I saw the comment on working with channels. Actually, one problem we haven't really looked into is how hub (and ISS in general) should handle staged channels from lifecycle management projects.
My _assumption_ would be that in most cases the actual content staging would happen centrally, but especially once we have configuration staging as well, there may be scenarios where you'd like to do local refinements on top of the contents synced from the central.
Not an issue as of today because the current staged channels just behave like any other channels ISS can sync. But something to keep in mind.

moio:
> I saw the comment on working with channels. Actually, one problem we haven't really looked into is how hub (and ISS in general) should handle staged channels from lifecycle management projects.

More on that in an upcoming Hub RFC (around ISS). Stay tuned.

mbrookhuis:
+1 :)

moio:
I am merging this now, discussion can be reopened anytime if necessary.

Thanks everyone for the cooperation!

paususe:
How would I remove a user?

moio:
We could implement an `uyuni.user.absent` state.

That's actually a very small (and obvious) "Next step", I'll just add it to the text now.

paususe:
> > For how long do I need to keep the "user creation state"? For as long as the users should exist? Or is this a one-shop operation and after the users are created/removed, I can remove the state?
> 
> Well, this follows standard Salt semantics, so the answer is "it depends on what the user wants to do".
> 
> If the user just wants a one-time initialization, then any state files can be discarded after it's done. This is what happens, for example, in sumaform (we leave the files in place "just in case", but they are generally not used after the deployment).
> 
> If the user intends to schedule a cron job to make sure a certain state is does not deviate from a sanctioned configuration, he has to keep state files around (commonly versioned in a Git repo) and apply them every night, or hour, or whatever else makes sense as an enforcement policy in the particular environment. Salt's logic is idempotent by design, so the end result will be always the same no matter how drifted away the actual minion state is.

My point is:
* I wouldn't feel very comfortable having admin passwords in plain text in a Salt state forever (in case this state needs to remain there for as long as users should be able to access)
* If this state can be removed, how does someone get a list of users so that they can be removed by `uyuni.user.absent`? I guess using the API

BTW: I really like that we are using "uyuni.user..." and not "rhn.user...", "spacewalk.user...", etc :-)

moio:
> I wouldn't feel very comfortable having admin passwords in plain text in a Salt state forever (in case this state needs to remain there for as long as users should be able to access)

Sure, they absolutely shouldn't be in Salt state files in any serious setting - there is Pillars for that, another Salt concept that is specifically designed for secure data:

https://docs.saltstack.com/en/latest/topics/tutorials/pillar.html

Among other things, one can [GPG encrypt pillars](https://gijs.io/2017/02/28/secure-pillar-data-in-saltstack-with-gpg/) and that's a pattern we see pretty frequently as well.

So here we are just leveraging existing Salt concepts and patterns to allow users to set up a secure environment. As with any Salt use, we have to keep in mind its limitations, by the way, the main one security wise being that the mere act of having a Salt minion daemon always means to give `root` permissions to its Master. That's something that's built in the basic assumptions in Salt - if someone takes over the Master, game is over anyway.

> If this state can be removed, how does someone get a list of users so that they can be removed by uyuni.user.absent? I guess using the API

Well, one could use the execution modules that the state modules are built upon. Execution modules implement functions that can be "run on any minion", while state modules implement idempotent state application by using those execution modules.

Practically that could look like the following:

```bash
hub$ salt '*' uyuni.user.list
server-1.suse.com:
    - User1
    - User2
    - User3
server-2.suse.com:
    - User1
    - User2
    - User3
    - ForgottenUser4

hub$ salt 'server-2.suse.com' state.apply uyuni_users

# server-2 is aligned to others
```

paususe:
> Among other things, one can [GPG encrypt pillars](https://gijs.io/2017/02/28/secure-pillar-data-in-saltstack-with-gpg/) and that's a pattern we see pretty frequently as well.

I'm probably missing something here. 

IUC the hub server would store both the encrypted pillar data and the private GPG key (because you need to decrypt the pillar data in order to do something with it on the slave servers).

I can't see how
encrypted credentials file (pillar) + GPG key to decrypt
is any more secure than
plaintext credentials file

¿?
 
> > If this state can be removed, how does someone get a list of users so that they can be removed by uyuni.user.absent? I guess using the API
> 
> Well, one could use the execution modules that the state modules are built upon. Execution modules implement functions that can be "run on any minion", while state modules implement idempotent state application by using those execution modules.
> 
> Practically that could look like the following:
> 
> ```shell
> hub$ salt '*' uyuni.user.list
> server-1.suse.com:
>     - User1
>     - User2
>     - User3
> server-2.suse.com:
>     - User1
>     - User2
>     - User3
>     - ForgottenUser4
> 
> hub$ salt 'server-2.suse.com' state.apply uyuni_users
> 
> # server-2 is aligned to others
> ```

If I think of the use case of the hub, it's for organizations with tens of thousands of clients. 

Those organizations will 100% sure some sort of LDAP-compatible directory available (Active Directory, eDirectory, Google GSuite LDAP or something alike).

In my mind, there is a simpler alternative:
1. Set a hard requirement for the hub: "if you want to run a hub server, then you will need LDAP".
2. Define a set of groups, roles or whatever for the Server and for the Hub. Those will be the global admins, org admins, regular users or whatever we want to define.
3. Use Bo's LDAP sync tool (https://github.com/uyuni-project/uyuni-rfc/pull/23) to sync and map users and groups from LDAP to Uyuni Hub and Uyuni Server

I think that will provide us with essentially the same outcome but faster and easier (less code to be developed). Also, as hub-consuming organizations will already have an LDAP, the fact that Uyuni Hub and Uyuni Server now use the very same users (instead of  silo) will probably look appealing to them.

Does that make sense?

moio:
> > Among other things, one can [GPG encrypt pillars](https://gijs.io/2017/02/28/secure-pillar-data-in-saltstack-with-gpg/) and that's a pattern we see pretty frequently as well.
> 
> I'm probably missing something here.

And I guess I picked a wrong example article. Pillars do not need to come from a local file, they are actually a pluggable data source that can come from whatever:

https://docs.saltstack.com/en/latest/ref/pillar/all/index.html

[An LDAP pillar module](https://docs.saltstack.com/en/latest/ref/pillar/all/salt.pillar.pillar_ldap.html) exists as well.




> In my mind, there is a simpler alternative:
>     1. Set a hard requirement for the hub: "if you want to run a hub server, then you will need LDAP".
>     2. Define a set of groups, roles or whatever for the Server and for the Hub. Those will be the global admins, org admins, regular users or whatever we want to define.
>     3. Use Bo's LDAP sync tool (#23) to sync and map users and groups from LDAP to Uyuni Hub and Uyuni Server
> 
> I think that will provide us with essentially the same outcome but faster, easier (less code to be developed) and more secure (no need to manage passwords). Also, as hub-consuming organizations will already have an LDAP, the fact that Uyuni Hub and Uyuni Server now use the very same users (instead of silo) will probably look appealing to them.
> 
> Does that make sense?

It does make some sense, but I am not completely convinced.

The part that seems missing is how to define permissions in additions to users with this module. I guess we should come up with some way to encode in LDAP user Alice can access ServerGroups A and B, while Bob manages channels C and D, and so on - and the LDAP mapping tool will have to recognize these attributes and translate them into XMLRPC calls to configure Servers accordingly. I think that is actually the bulk of the complexity, and that will have to be coded in both cases.

Additionally: diven that Servers are minions already, and will be configured with normal Salt states for any other aspect anyway, why not using Salt to also configure users and have a uniform "UI", and reuse a lot of what we already have for Salt including versioning, Formulas, Forms, etc.?

moio:
As clarified in our meeting, the bulk of the complexity actually lies in the mapping: how can a certain piece of data in LDAP be transformed into a User - not just username and password, but also permissions to certain entities (channels, systems...) and other configurable parameters.

With the solution proposed here, users can implement any kind of mapping as they have a full programming language at their disposal. Moreover they are not limited to LDAP, and can decide when the mapping is enforced (eg. via `salt '*' state.apply` in a `cron` job).

By using Salt local mode, this can also be used in non-Hub use cases.

By using Formulas with Forms this can be easily given a UI, which is also useful in Public Cloud use cases.

As a result of these considerations, I am not presently changing the RFC text, although as usual an additional PR can always be opened at a later point.

PR #19, opened 2019-10-10
Hub XMLRPC API RFC

moio:
This RFC contains details on an XMLRPC API endpoint for the Hub.

### DIFF
--- a/accepted/00062-hub-xmlrpc-api.md
+++ b/accepted/00062-hub-xmlrpc-api.md
@@ -28,11 +28,17 @@ From the requirements:
 [design]: #detailed-design
 
 ### Basic funtionality
-The Hub is a Server, so as far as it and its minions (Servers) are concerned, the XMLRPC API work as of today.
+The Hub is a Server, so as far as it and its minions (Servers) are concerned, the Hub XMLRPC API work as of today.
 Content management APIs will work as of today. Inter-Server Synchronization (ISS) will be used to push managed content from Hub to Servers.
 
+### New service and endpoint
+A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
+  * Implementation will be in Python based on top of Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library
+  * all I/O should be handled asynchronously via Tornado. Calls to several Servers should happen in parallel, with a maximum timeout value for unreachable/not responding Servers
+  * there will be no backing database. All new methods will delegate to existing XMLRPC APIs (either the Hub's or individual Servers')
+
 ### Topology exposure functionality
-New API endpoints to be created to expose how Servers, clients and Hubs are connected, in particular methods will be needed to:
+Gateway methods will be created to expose how Servers, clients and Hubs are connected, in particular methods will be needed to:
   * Get a list of `serverIds`
   * Get a list of clients registered to a Server given its `serverId`
   * Get a list of all clients, with the `serverIds` they belong to
@@ -43,18 +49,19 @@ New API endpoints to be created to expose how Servers, clients and Hubs are conn
 Idea is to expose existing Server XMLRPC endpoints/methods as they are on the Hub. The Hub will simply route calls to the right Server(s) and relay back results, rather than reimplementing any functionality. Areas are detailed below.
 
 * Authorization and authentication
-  * Hub API will be secured via `sessionKey` tokens like Server's API
-  * Consuming a Server's API requires a valid `sessionKey` token
+  * Gateway API will be secured via `hubSessionKey` tokens analoguously to Server's API
+    * Actually, Gateway will delegate authentication to the Hub API
+  * Consuming a Server's API from the Gateway continues to require a valid Server `sessionKey` token
   * There are three ways to attach Server sessions to the Hub session
     * Manual mode: programmer "attaches" Server sessions explicitly to his Hub session. Attachment lasts until logout
       * `hub.login(username, password)` → `hubSessionKey`
       * `hub.attachToServer(hubSessionKey, serverId, username, password)` → `serverSessionKey`
       * `hub.serverMethod(serverSessionKey, parameters)` → `output`
       * N+1 pairs of credentials will be needed (one for the Hub, one per each of the N Servers)
-     * Authentication relay mode: as an additional convenience option, which can be activated with a flag, Hub-User credentials can be re-used to authenticate against Servers. So if a certain username/password is uniformly configured on the Hub and across several Servers, Hub gateway API usage becomes simpler: only one log in is needed, with its Hub-User credentials
+     * Authentication relay mode: as an additional convenience option, which can be activated with a flag, Hub-User credentials can be re-used to authenticate against Servers. So if a certain username/password is uniformly configured on the Hub and across several Servers, Gateway usage becomes simpler: only one log in is needed, with its Hub-User credentials
      * Automatic connect mode: as an additional convenience option, which can be activated with a flag, the Hub gateway API can automatically select the list of Servers a Hub-User works with, based on the list of Servers that same Hub-User has permissions on
        - note: that this only makes sense if users are OK with the idea of using Server sysadmin users to act on Clients
-       - If they prefer to manage Clients via users that have no permissions to manage Servers, they can create a Hub-Org with no Servers. Any Hub-User in this Hub-Org will be able to connect to the Hub gateway API, and possibly have permissions on Clients (if Servers have Server-Users with same name and passwords)
+       - If they prefer to manage Clients via users that have no permissions to manage Servers, they can create a Hub-Org with no Servers. Any Hub-User in this Hub-Org will be able to connect to the Gateway, and possibly have permissions on Clients (if Servers have Server-Users with same name and passwords)
 
 * Multicasting: call an XMLRPC method on _N_ Servers at once
   * method name stays the same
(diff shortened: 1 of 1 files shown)


accepted/00062-hub-xmlrpc-api.md
```diff
@@ -0,0 +1,129 @@
...
+       - note: that this only makes sense if users are OK with the idea of using Server sysadmin users to act on Clients
+       - If they prefer to manage Clients via users that have no permissions to manage Servers, they can create a Hub-Org with no Servers. Any Hub-User in this Hub-Org will be able to connect to the Gateway, and possibly have permissions on Clients (if Servers have Server-Users with same name and passwords)
+
+* Multicasting: call an XMLRPC method on _N_ Servers at once
+  * method name stays the same
+  * Assuming the method on the Server accepts _M_ parameters, the same method on the Hub will accept an array of _N_ parameters, each of which contains as elements the _M_ parameters for a given Server
+    * eg. `actionchain.listChains(sessionKey)` → `hub.actionchain.listChains([serverSessionKey1, serverSessionKey2, ...])`
+    * eg. `channel.org.enableAccess(sessionKey, channelLabel, orgId)` → `hub.channel.org.enableAccess([serverSessionKey1, channelLabel1, orgId1], [serverSessionKey2, channelLabel2, orgId2], ...)`
+  * return types would be arrays of _N_ elements
```
joesusecom: So the server's ids aren't required in this scenario? Are they automatically detected by identifying the matching `serverSessionKey`?
Just asking: Should we also add a "convenience feature" for calls that are exactly the same for all servers, where I only provide the dict of serverSessionKey's and the rest of the parameters only once?
Not a big deal because one certainly can expand those in the calling (e.g. Python) code as well.

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -0,0 +1,129 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+We are not adding any Hub-centered functionality basically, this is a little more than a router.
+
+## Limitations
+
+- implementation of API methods currently not implemented on the Server, eg. downstream [#2361](https://github.com/SUSE/spacewalk/issues/2361) [#2362](https://github.com/SUSE/spacewalk/issues/2362), is explicitly not covered by this RFC
+
```
joesusecom: You may need some minimal error handling code in case the API call that was multicast isn't available on all nodes.

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -0,0 +1,129 @@
...
+  * Alternative idea 1: a configuration file on the Hub contains a list of credentials and `serverIds`. Hub would automatically maintain a pool of Server `sessionKey`s from those credentials
+    * `hub.login(username, password)` → `hubSessionKey`
+    * `hub.serverMethod(hubSessionKey, serverId, parameters)` → `output`
+  * Alternative idea 2: god mode. We would assume Hub is root on Servers, so somehow automatically gets full administrative API access to all anyway
+
+* Multicasting
+* Alternative idea 2: each of the _M_ parameters is replaced by an array of _N_ elements
+  * eg. `actionchain.listChains(sessionKey)` → `hub.actionchain.listChains([serverSessionKey1, serverSessionKey2, ...])`
+  * eg. `channel.org.enableAccess(sessionKey, channelLabel, orgId)` → `hub.channel.org.enableAccess([serverSessionKey1, serverSessionKey2, ...], [channelLabel1, channelLabel2, ...], [orgId1, orgId2, ...])`
+  * return types would be arrays of _N_ elements
```
joesusecom: Potential risk with that: What if the parameters have changed and one of the servers is older/newer than the others? Very unlikely though. In general, I like the non-alternative approach better.

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -0,0 +1,129 @@
...
+### Basic funtionality
+The Hub is a Server, so as far as it and its minions (Servers) are concerned, the Hub XMLRPC API work as of today.
+Content management APIs will work as of today. Inter-Server Synchronization (ISS) will be used to push managed content from Hub to Servers.
+
+### New service and endpoint
+A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
+  * Implementation will be in Python based on top of Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library
+  * all I/O should be handled asynchronously via Tornado. Calls to several Servers should happen in parallel, with a maximum timeout value for unreachable/not responding Servers
+  * there will be no backing database. All new methods will delegate to existing XMLRPC APIs (either the Hub's or individual Servers')
+
```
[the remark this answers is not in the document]
moio: Right. That might be needed at a later stage. I'll keep that as an implementation detail.

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -0,0 +1,129 @@
...
+### Topology exposure functionality
+Gateway methods will be created to expose how Servers, clients and Hubs are connected, in particular methods will be needed to:
+  * Get a list of `serverIds`
+  * Get a list of clients registered to a Server given its `serverId`
+  * Get a list of all clients, with the `serverIds` they belong to
+  * Get the `serverId` for a given client (based on `minion_id`, FQDN or other unique identifier. Normal client IDs will not work, as they will overlap between Servers)
```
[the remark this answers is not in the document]
moio: I understand and agree we have this problem, in general, especially about client-addressed routing (CAR), described in "Next steps" below.

For the mere topology exposure functionality, it will be OK if this particular method returns more than one `serverId`, adjusting text in this sense.

accepted/00062-hub-xmlrpc-api.md
moio: > So the server's ids aren't required in this scenario? Are they automatically detected by identifying the matching serverSessionKey?

That's the idea in this example, yes.

>  Should we also add a "convenience feature" for calls that are exactly the same for all servers, where I only provide the dict of serverSessionKey's and the rest of the parameters only once?

Yeah, why not. Added in the next steps.

accepted/00062-hub-xmlrpc-api.md
moio: Right, I did not state the assumption here is that all Servers are on the same version of Manager, at least major, where we already guarantee no breaking changes. Added now.

accepted/00062-hub-xmlrpc-api.md
moio: Agreed, that I also clarified in limitations anyway.

moio:
@joesusecom thanks for the review, several gaps closed now! Can you approve this?

moio:
I am merging this now, discussion can be reopened anytime if necessary.

Thanks everyone for the cooperation!

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -0,0 +1,136 @@
...
+# Detailed design
+[design]: #detailed-design
+
+### Basic funtionality
+The Hub is a Server, so as far as it and its minions (Servers) are concerned, the Hub XMLRPC API work as of today.
+Content management APIs will work as of today. Inter-Server Synchronization (ISS) will be used to push managed content from Hub to Servers.
+
+### New service and endpoint
+A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
+  * Implementation will be in Python based on top of Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library
```
paususe: Why implement from scratch with Tornado instead of using an existing API gateway?
E. g. [Graviteer](https://gravitee.io/products/apim/) (first page in Google) already implements [caching](https://docs.gravitee.io/apim_policies_cache.html), [rate limiting](https://docs.gravitee.io/apim_policies_rate_limiting.html), can [serve XMLRPC as REST](https://docs.gravitee.io/apim_policies_rest2soap.html) APIs, does [RBAC](https://docs.gravitee.io/apim_policies_role_based_access_control.html), [API keys](https://docs.gravitee.io/apim_policies_apikey.html) to limit what API calls are available to what users, analytics, load balancing, etc.

Other API gateways may provide similar or better features: [WSO2 API Management](https://wso2.com/api-management/), [Tyk](https://tyk.io/features/api-gateway/), [APIman](http://www.apiman.io/), [KrakenD CE](https://github.com/devopsfaith/krakend-ce), etc.

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -0,0 +1,136 @@
...
+* Multicasting: call an XMLRPC method on _N_ Servers at once
+  * method name stays the same
+  * Assuming the method on the Server accepts _M_ parameters, the same method on the Hub will accept an array of _N_ parameters, each of which contains as elements the _M_ parameters for a given Server
+    * eg. `actionchain.listChains(sessionKey)` → `hub.actionchain.listChains([serverSessionKey1, serverSessionKey2, ...])`
+    * eg. `channel.org.enableAccess(sessionKey, channelLabel, orgId)` → `hub.channel.org.enableAccess([serverSessionKey1, channelLabel1, orgId1], [serverSessionKey2, channelLabel2, orgId2], ...)`
+  * return types would be arrays of _N_ elements
+    * special elements would be needed to signal Exceptions
+
+* Availability
+  * if a Server is down at the time a Gateway call targeting it is made, call should fail after a configurable timeout. If multiple Servers are targeted, only that call fails and others continue
```
paususe: IMO this should be managed differently: Server API calls should be classified in several categories according to the required response time and the hub API should do the retries.

Use case: it's 10 AM in the morning and I am scheduling an update cycle for tonight at 11 PM. I have 10 Uyuni Servers and 1 of them is down at 10:00:56. Instead of returning failure, Hub should keep retrying until e. g. 10 PM => different API calls have different timeout needs.

E. g. if I want to run a command on all my clients, I want the response to be immediate (i. e. not more than a very short timeout, maybe 1 minute) => different API calls have different timeout needs

accepted/00062-hub-xmlrpc-api.md
moio: As per our last meeting, I will be following up on this possibility through an add-on RFC.

accepted/00062-hub-xmlrpc-api.md
moio: One thing that could be done is to make it configurable per-call. I will be amending the text in this sense in a follow-up RFC, thanks for pointing out.

PR #20, opened 2019-10-15
Saltclass as external pillar mechanism

aaannz:
Introducing SaltClass as an external pillar mechanism RFC

Link to [rendered document](https://github.com/uyuni-project/uyuni-rfc/blob/38ba806c2c7408c79c4677d6932598e2af873384/accepted/00000-saltclass-external-pillar.md)

### DIFF
--- a/accepted/00000-saltclass-external-pillar.md
+++ b/accepted/00000-saltclass-external-pillar.md
@@ -1,0 +1,147 @@
+- Feature Name: Saltclass as external pillar mechanism
+- Start Date: 2019-08-20
+- RFC PR:
+
+# Summary
+[summary]: #summary
+
+Enhance/replace current custom external pillar mechanism with upstream provided Saltclass module.
+
+# Motivation
+[motivation]: #motivation
+
+SUSE Manager needs a way how to export data about systems, groups, services etc. to the salt configuration management. In current implementation this is done by custom python script [suma_minion.py](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/modules/pillar/suma_minion.py) used by salt master leveraging its external pillar mechanism. This worked so far very well, even allowing Uyuni some extra tricks like integration of Formulas with Forms mechanism. However the implementation has also some limitations preventing further growth.
+
+This RFC proposes to, at first, enable yet another external pillar mechanism in the salt, but this time not a custom script, but salt's own [saltclass](https://docs.saltstack.com/en/latest/ref/pillar/all/salt.pillar.saltclass.html) mechanism.
+
+Examples of limitations of our external pillar implementation:
+- allows only static pillars (no jinja preparsing, no pillar dependencies on other pillars)
+- problematic data merging
+- made integration of formulas easy, but causes duplication of code where i.e. form.yml is parsed on both java and python level
+
+# Detailed design
+[design]: #detailed-design
+
+There are three choices how to address the introduction of saltclass to the SUSE Manager:
+
+A) Enable concurrent usage of both custom `suma_minion.py` and `saltclass` external pillars
+B) Enable concurrent usage of both custom `suma_minion.py` and `saltclass` external pillars, but with the target to gradually migrate from `suma_minion.py` to `saltclass`.
+C) Do the one time migration and switch from custom `suma_minion.py` to `saltclass`
+
+## What means migrating from suma_minion.py to saltclass?
+
+Even in its relative simplicity `suma_minion.py` does more then just loading `yaml` files and exporting to the `salt` ecosystem. `suma_minion.py` is also integral part of the Formulas with Forms enablement in the SUSE Manager/Uyuni.
+
+`suma_minion.py` responsibilities regarding formulas are:
+- read what formulas are enabled for given minion and prepare `formulas` grain which is then used by `mgr_tops`, respectively `formulas.sls` loader to apply formulas as part of highstate
+- read formulas `json` data and `form.yml` and construct complete pillar information
+
+Complete migration from `suma_minoni.py` to `saltclass` then means:
+- implement Formulas with Forms enablement in other way then in external pillar, preferably in java code itself as the current implementation causes triplication of the formulas enablement code
+- make current minion pillars accessible to `saltclass`
+- make images pillars accessible to `saltclass`
+
+## Are conflicts between existing suma_minion pillars and saltclass pillars possible?
+
+Our external pillar implementation fortunately looks only for files with `pillar_` prefix and following minion id with `.yml` extension:
+```python
+# Including generated pillar data for this minion
+data_filename = os.path.join(MANAGER_PILLAR_DATA_PATH, 'pillar_{minion_id}.yml'.format(minion_id=minion_id))
+```
+Saltclass on the other hand looks (recursively) for files with names matching minion id with `.yml` extension.
+
+These two formats are thus living in separate namespaces and conflicts are then improbable for minions.
+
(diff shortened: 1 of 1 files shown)


accepted/00000-saltclass-external-pillar.md
```diff
@@ -0,0 +1,147 @@
...
+## What means migrating from suma_minion.py to saltclass?
+
+Even in its relative simplicity `suma_minion.py` does more then just loading `yaml` files and exporting to the `salt` ecosystem. `suma_minion.py` is also integral part of the Formulas with Forms enablement in the SUSE Manager/Uyuni.
+
+`suma_minion.py` responsibilities regarding formulas are:
+- read what formulas are enabled for given minion and prepare `formulas` grain which is then used by `mgr_tops`, respectively `formulas.sls` loader to apply formulas as part of highstate
+- read formulas `json` data and `form.yml` and construct complete pillar information
+
+Complete migration from `suma_minoni.py` to `saltclass` then means:
+- implement Formulas with Forms enablement in other way then in external pillar, preferably in java code itself as the current implementation causes triplication of the formulas enablement code
```
joesusecom: I don't quite get the comment about "triplication" of the enablement code. Look at how it's done with custom/custom_groups/custom_org, the code used for Salt-based configuration channels. It isn't any more elegant.
But yes, those two need to be unified as they're basically doing the same thing (adding states to a highstate).

accepted/00000-saltclass-external-pillar.md
```diff
@@ -0,0 +1,147 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+There are three choices how to address the introduction of saltclass to the SUSE Manager:
+
+A) Enable concurrent usage of both custom `suma_minion.py` and `saltclass` external pillars
+B) Enable concurrent usage of both custom `suma_minion.py` and `saltclass` external pillars, but with the target to gradually migrate from `suma_minion.py` to `saltclass`.
+C) Do the one time migration and switch from custom `suma_minion.py` to `saltclass`
+
```
joesusecom: I'd strongly prefer C).

The current code is convoluted enough. It usually takes me a few minutes to find out how, for example, the pillar data indirectly drives formula inclusion via the master_tops file etc.

Adding yet another concept without removing this mess would make it even harder for customers to extend Manager in a meaningful way.

accepted/00000-saltclass-external-pillar.md
```diff
@@ -0,0 +1,147 @@
...
+The major motivator in these cases is the jinja preparsing of the pillar data.
+
+### Changes in Formulas with Forms implementation
+
+1) Make FormulaFactory.java to store not just raw `json` data, but complete pillar structure as `saltclass pillar data`
+   This formula pillar will be stored as a `saltclass` `class` data and then included from `saltclass pillar data `of minion (this step may conflict with below mentioned Step 2b about migrating `existing pillar data`)
+2) If `saltclass` is also enabled as a tops module, minion `saltclass pillar data` can contain a list of state names to be included in the highstate. If `saltclass` is not enabled as a tops module then I recommend to keep this functionality in the `suma_minion.py` for this round of implementation.
+
+> **NOTE**
+>Choosing `saltclass` to be also enabled as a tops module and using it to load formulas can make implementation of https://github.com/SUSE/spacewalk/issues/9490 and https://github.com/SUSE/spacewalk/issues/2714 considerably harder.
```
joesusecom: Why exactly?

accepted/00000-saltclass-external-pillar.md
```diff
@@ -0,0 +1,147 @@
...
+[drawbacks]: #drawbacks
+
+## Big refactoring underway, but stalled
+
+There is an upstream pull request about major `saltclass` refactoring - https://github.com/saltstack/salt/pull/52407 - however this PR seems to be stalled. This PR meant to fix several issues and fulfil couple of feature requests.
+The unknown is if the PR will be ever finished or if we can invest and try to finalize it ourselves. It is possible that the work stopped only temporarily due to changes in salt releases, or that the author being external contributor chose to abandon the effort.
+
+## Performance impact
+
+Including another pillar engine can have a negative performance impact. (#TODO measure)
```
joesusecom: That's why I'd go all in with saltclass or not at all. Of course we can take a detour during development in Uyuni, but once we hit a major release milestone we should have just one of the two pillar plugins.
Performance-wise, one thing that could potentially hit us hard is that JSON parsing is supposedly _much_ faster than YAML parsing, especially if there's Jinja involved as well. We need to measure that with very large data sets and also look at peak memory usage etc. when thousands of minions ask for their pillars.
If necessary, this can be partially fixed by some clever caching though.

accepted/00000-saltclass-external-pillar.md
```diff
@@ -0,0 +1,147 @@
...
+## Upstream breakages, abandonment
+
+There is always some risk that upstream may decide to abandon SaltClass. So far maintaining SaltClass should be cheap, due to upstream involvement, but if the upstream decides for any reason to abandon SaltClass the cost of maintenance can increase dramatically (up to impossible if upstream decides to abandon it for some deep design reasons). As a mitigation there is a possibility to use [reclass](https://reclass.pantsfullofunix.net/) upon which SaltClass is based and shares storage details (see below).
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Implement custom jinja processing of pillars
+
+The major driver for `saltclass` is the ability to preparse pillars with jinja and also to reference pillars between each other. This can be implemented within existing `suma_minion.py` external pillar as well. Using `saltclass` saves us the development time in exchange for development time on integration, but given `saltclass` is an upstream project it in theory can help with maintenance.
```
joesusecom: I'd really like to see valid example for why I'd need the Jinja preparsing for pillars. As mentioned before, this could have severe performance impacts. It also makes it harder to use third party providers for pillars (e.g. AD or LDAP) when customers integrate with other backend systems, as such complex pillar structures are very hard to get right.
Personally I've always seen the main advantage of saltclass/reclass in its ability to inherit and overload defaults, thus reducing the number of pillars I have to duplicate.

accepted/00000-saltclass-external-pillar.md
```diff
@@ -0,0 +1,147 @@
...
+## Use original reclass instead
+
+Instead of salt native implementation there is an option to use the original [reclass](https://reclass.pantsfullofunix.net/) with the [reclass](https://docs.saltstack.com/en/latest/ref/pillar/all/salt.pillar.reclass_adapter.html) adapter. This can be used again as just external pillar or external tops as well.
+Drawback is need of another python package and that `reclass` itself seems to be abandoned (last commit is from 2015). There is newer fork in salt-formula git repository - https://github.com/salt-formulas/reclass (last commit from January, 2019), however from the issue list this one seems to have a problem with Python 3.6 and newer.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* performance impact
+* status of the upstream refactoring
```
joesusecom: I see two areas where this RfC may cause problems elsewhere:

1) Git integration. Many of our customers rely on Git for their pillar and state files when they work on their own Salt content. We need to look into this! One pointer from the web: https://thatremindsme.woodwose.net/computing/2018/08/30/Using-saltclass-from-a-gitrepository.html

2) YAML vs. JSON: Stratos seems to now be able to display JSON files with JSON schemas from Kubernetes and Cloud Foundry in a conceptually very similar way that Formulas with Forms do. I see this potentially as a V2 (or at least alternative format) for the current forms.yml approach. Also, should we decide to store the SUSE Manager pillars in PostgreSQL at some point in the future, JSON would make things easier as well as there is a native JSON type for PostgreSQL.

rjmateus:
closed. We are considering external pillar loaded from database. Upstream is buggy

PR #21, opened 2019-10-21
Hub Inter-Server Sync RFC

moio:
This RFC contains details on an ISS overhaul for the Hub.

### DIFF
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -171,7 +171,7 @@ The first implementation of this RFC will:
   - output of the ISS Master tool will be a directory with files and a SQL script
   - the ISS Slave tool will mainly copy files to correct locations and execute the SQL script
   - triggering of the ISS Slave tool might happen via Salt, this is left as an implementation detail
-  - transfer proper could be done via `rsync`, which provides transparent compression and delta encoding
+  - transfer proper could be done via `rsync` or `rclone`, which provide transparent compression and delta encoding
 - all new code will be typed Python 3, connecting to Postgres via psycopg2 and no dependencies to the old Python stack
 - inspection of the schema can be implemented via `SELECT` queries on Postgres's internal tables
   - current and new implementations would thus be completely independent, so could coexist for an initial period of time, as the new implementation matures
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -197,7 +197,7 @@ With the proposal in this RFC, it is possible to implement protection from case 
 
 pl/SQL could still be used, as a last resort, to apply different commands to different ISS Slave versions, but that would be hardly maintainable, so it would only be suggested to handle critical situations temporarily.
 
-General recommendation would be to upgrade an ISS Master, then upgrade ISS Slaves before attempting any new Inter-Server Synchronization.
+General recommendation would be to upgrade an ISS Master, then upgrade ISS Slaves before attempting any new Inter-Server Synchronization. Another RFC will detail how Hub could make such upgrades easier.
 
 
 # Alternatives
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -13,7 +13,7 @@ With the Hub architecture, there will be multiple Servers that receive content m
 
 This RFC describes an overhaul of the ISS implementation in order to cover more use cases (in particular, synchronization of content different than channel data) and with better performance to attain higher scalability.
 
-See the [Hub general RFC](accepted/00060-hub-general.md) for an introduction to the Hub project.
+See the [Hub general RFC](00060-hub-general.md) for an introduction to the Hub project.
 
 # Motivation
 [motivation]: #motivation
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -227,6 +227,7 @@ General recommendation would be to upgrade an ISS Master, then upgrade ISS Slave
 - drop the current ISS implementation
   - full backwards compatibility of all commandline flags is left as an implementation detail, it's probably achievable
   - dropping the old ISS implementation could save about ~9k LOC, as measured by `cloc backend/server/importlib/*Import* backend/satellite_exporter/ backend/satellite_tools/exporter/ backend/satellite_tools/*sat*ync*`
+- integrate with the Content Lifecycle Management functionality: automatically sync after content is finalized
 - support syncing repo metadata, in order for them not to necessarily be regenerated on the ISS Slave
 
 # Unresolved questions
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -122,7 +122,7 @@ At this point it is believed/assumed that:
 
 #### Second core idea: SQL script is generated via introspection
 
-The SQL script would be produced by a program that looks at the ISS Master's schema and, given a small set of starting tables (eg. `rhnchannel`) and starting row ids (eg. `channel_id = 103`), figures out:
+The SQL script would be produced by a program that looks at the ISS Master's schema and, given a small set of starting tables (eg. `rhnchannel`) and starting rows given by a criterion (eg. `channel_id = 103` or `modified_date > '2019-01-01'`), figures out:
 
 - the graph of dependent/dependency tables (from the starting set)
 - for each table:
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -172,7 +172,7 @@ The first implementation of this RFC will:
   - the ISS Slave tool will mainly copy files to correct locations and execute the SQL script
   - triggering of the ISS Slave tool might happen via Salt, this is left as an implementation detail
   - transfer proper could be done via `rsync` or `rclone`, which provide transparent compression and delta encoding
-- all new code will be typed Python 3, connecting to Postgres via psycopg2 and no dependencies to the old Python stack
+- all new code will be either typed Python 3, connecting to Postgres via psycopg2 and no dependencies to the old Python stack, or Go. Choice is deferred at implementation time
 - inspection of the schema can be implemented via `SELECT` queries on Postgres's internal tables
   - current and new implementations would thus be completely independent, so could coexist for an initial period of time, as the new implementation matures
 
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -184,6 +184,8 @@ When that happens, there is a chance of regressions for ISS users, which represe
 # Drawbacks
 [drawbacks]: #drawbacks
 
+This will only work with PostgreSQL, but that is not expected to be a problem given it's the only supported database backend at this point.
+
 Syncing content between different versions of Uyuni on the ISS Master and on the ISS Slave may not work, if the database schema changes in non-compatible ways.
 
 This risk is fundamentally unavoidable at a conceptual level, although different mitigations can be put in place and the current implementation actually allows for a broader set of mitigations than the proposal in this RFC can allow.
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -227,8 +227,9 @@ General recommendation would be to upgrade an ISS Master, then upgrade ISS Slave
 - drop the current ISS implementation
   - full backwards compatibility of all commandline flags is left as an implementation detail, it's probably achievable
   - dropping the old ISS implementation could save about ~9k LOC, as measured by `cloc backend/server/importlib/*Import* backend/satellite_exporter/ backend/satellite_tools/exporter/ backend/satellite_tools/*sat*ync*`
-- integrate with the Content Lifecycle Management functionality: automatically sync after content is finalized
 - support syncing repo metadata, in order for them not to necessarily be regenerated on the ISS Slave
+- allow downloading of files from locations other than the ISS Master - SCC, a closer Proxy, third-party repos...
+- integrate with the Content Lifecycle Management functionality: automatically sync after content is finalized
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -240,6 +240,7 @@ General recommendation would be to upgrade an ISS Master, then upgrade ISS Slave
   - dropping the old ISS implementation could save about ~9k LOC, as measured by `cloc backend/server/importlib/*Import* backend/satellite_exporter/ backend/satellite_tools/exporter/ backend/satellite_tools/*sat*ync*`
 - support syncing repo metadata, in order for them not to necessarily be regenerated on the ISS Slave
 - allow downloading of files from locations other than the ISS Master - SCC, a closer Proxy, third-party repos...
+- offer a UI for ISS in general. Current implementation does not have one
 - integrate with the Content Lifecycle Management functionality: automatically sync after content is finalized
 
 # Unresolved questions
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -1,0 +1,236 @@
+- Feature Name: hub_iss
+- Start Date: 2019-10-17
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Uyuni Server has a functionality called Inter-Server Synchronization (abbreviated to ISS from now on) that allows users to transfer channel data (packages and metadata) from one Server (called an ISS Master) to another Server (called an ISS Slave).
+
+Current implementation [is documented in Uyuni manuals](https://www.uyuni-project.org/uyuni-docs/uyuni/administration/iss.html).
+
+With the Hub architecture, there will be multiple Servers that receive content managed in the Hub, hence ISS will be a required mechanism.
+
+This RFC describes an overhaul of the ISS implementation in order to cover more use cases (in particular, synchronization of content different than channel data) and with better performance to attain higher scalability.
+
+See the [Hub general RFC](accepted/00060-hub-general.md) for an introduction to the Hub project.
+
+# Motivation
+[motivation]: #motivation
+
+ISS is currently limited to software channels, and it is not sure whether it can scale up above tens of Servers.
+
+From the requirements:
+
+> * Managed content can be selectively synced from Hub to Servers
+>   * In order of importance:
+>    - Configuration Management data (states, pillars...)
+>    - OS Images
+>    - Container Images
+>   * Synchronization has to be atomic
+>   * Synchronization has to be doable via a storage media instead of a network for certain network-constrained environments (defense, China)
+>   * Nice-to-have: avoid the download of content from the Hub if it is faster to retrieve it from SCC
+>   * Nice-to-have: on-demand downloading from the Hub should be considered (Proxy/Squid model)
+> [...]
+> * Performance: support high number of clients and Servers
+> [...]
+>  * maximum expected number of Servers: 9k
+
+# Detailed design
+[design]: #detailed-design
+
+## Main concept
+
+Any kind of managed content on a Uyuni Server can be thought as:
+  - a set of database records describing that content and, optionally,
+  - a set of files on a filesystem referred to by those records
+
+One way to look at the ISS problem is to provide a way of exporting from an ISS Master, transferring and importing into an ISS Slave those two pieces of information.
+
+### File export/import
+
+The ISS Master, given a specified piece of content to sync, should be able to identify the set of relevant files and "export" it. Then it should be transfered to the ISS Slave and "imported" (see "Transport" below).
+
+The current implementation has a code on the ISS Master to "export" files and code on the ISS Slave to "import" them - this concept is not supposed to change.
+
+### Database data import/export
+
+The ISS Master, given a specified piece of content to sync, should be able to export relevant data from the ISS Master database, to be transported and imported on the ISS Slave side.
+
+The current implementation serializes ISS Master database data to XML and deserializes it on the ISS Slave. Relatively complicated logic is in place on both communication ends to convert to and from this format.
+
+**First core idea** of this RFC is to **export database data from the ISS Master in form of a SQL script to be executed as-is on the ISS Slave** (eg. via `psql`).
+
+**Second core idea is to generate that SQL script from the ISS Master data mainly by inspecting the schema programmatically** (eg. figuring out by schema inspection most of the involved tables, key relationships, ordering, etc.).
+
+A proof-of-concept Python script that implements partially those ideas is available at: https://github.com/SUSE/spacewalk/blob/Manager-4.0-iss2/iss2/iss2.py
+
+#### First core idea: export as SQL script
+
+The SQL script exported from the ISS Master will have to be idempotent. For example:
+
+```sql
+INSERT INTO "rhnchecksumtype" (
+    "id",
+    "label",
+    "description"
+)
+VALUES
+    (1, 'md5', 'MD5sum'),
+    (2, 'sha1', 'SHA1sum'),
+    (3, 'sha256', 'SHA256sum'),
+    (4, 'sha384', 'SHA384sum'),
+    (5, 'sha512', 'SHA512sum')
+ON CONFLICT ("label") DO UPDATE
+    SET "description" = excluded."description";
+```
+
+The `ON CONFLICT DO UPDATE` clause is called an "upsert" effectively turning the `INSERT` into an `UPDATE` if a row with a conflict on the specified column (`label` in this case) is found.
+
+Note that additional mechanisms are needed to add correct values in the `id` column, if they come from a sequence. For example:
+
+```sql
+INSERT INTO "web_customer" (
+    "id",
+    "name",
+    "modified"
+)
+VALUES
+    (nextval('web_customer_id_seq'), 'SUS', '2019-10-17T10:09:11.750508+02:00'::timestamptz)
+ON CONFLICT ("name") DO UPDATE
+    SET "modified" = excluded."modified";
+```
+
+For tables with foreign keys, `SELECT` statements on parent tables will have to be used to link rows correctly, for example:
+
+```sql
+INSERT INTO "rhnchannelarch" (
+    "id",
+    "label",
+    "arch_type_id",
+    "name"
+)
+VALUES
+    (500, 'channel-ia32', (SELECT id FROM rhnarchtype WHERE label =  'rpm'), 'IA-32' - ),
+    (501, 'channel-ia32-deb', (SELECT id FROM rhnarchtype WHERE label =  'deb'), 'IA-32 Debian' - )
+ON CONFLICT ...
+```
+
+At this point it is believed/assumed that:
+  - ISS is by and large an additive process, so cleanups should not be needed for most tables. For those where it is necessary (eg. linking tables in an m:n relationship), `DELETE` statements either before or after `INSERT`s suffice
+  - most tables can be treated in substantially the same way, not much table-specific logic has to be implemented (in case it is, it is believed a solution can be found via pl/SQL)
+
+#### Second core idea: SQL script is generated via introspection
+
+The SQL script would be produced by a program that looks at the ISS Master's schema and, given a small set of starting tables (eg. `rhnchannel`) and starting row ids (eg. `channel_id = 103`), figures out:
+
+- the graph of dependent/dependency tables (from the starting set)
+- for each table:
+  - all columns
+  - the primary key (if any), and columns involved
+  - the sequences feeding the primary key if numeric (heuristics with exceptions, or database schema naming rules, might be needed)
+  - the foreign keys (if any), and the columns involved
+  - the rows directly or indirectly connected to the starting row id
+
+With the above information, it should be possible to write the SQL script in a completely or almost-completely automated way, in particular:
+  - determining the right order of `INSERT` and `DELETE` statements given table dependencies
+  - determining which column values have to be `SELECT`ed, because they are foreign keys, and which are constant
+  - determining the conflict column set for `ON CONFLICT` clauses
+  - the actual row data to insert
+
+Expected benefits:
+  - the ISS programs need (potentially) no updating when the database structure changes
+  - supporting a new managed content type is as easy as specifying a new set of starting tables
+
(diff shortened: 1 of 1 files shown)


accepted/00063-hub-iss.md
```diff
@@ -0,0 +1,236 @@
...
+3. Finally, yet other changes could really introduce a backwards compatibility problem and, since the protocol is versioned, [the import program will send an alert and abort](https://github.com/SUSE/spacewalk/blob/Manager-3.2/backend/satellite_tools/xmlSource.py#L271-L298).
+
+With the proposal in this RFC, it is possible to implement protection from case 3. via a similar versioning mechanism, partially against 2. ignoring missing tables but only offer limited resilience against case 1., as we would not have a real database-agnostic representation of data.
+
+pl/SQL could still be used, as a last resort, to apply different commands to different ISS Slave versions, but that would be hardly maintainable, so it would only be suggested to handle critical situations temporarily.
+
+General recommendation would be to upgrade an ISS Master, then upgrade ISS Slaves before attempting any new Inter-Server Synchronization.
+
```
admd: I would expect Hub to be helpful here and provide a way to upgrade the ISS slaves in regard to database changes as well.

moio: Yes, that is part of an upcoming Hub RFC, adding note.

moio:
Hey Joe, thanks for your answer, I think it's actually a good idea to dig deeper into the topics you mentioned. Let me ask a few more questions before I adapt the text.

> On second thought, this may be a relatively low-maintenance, pragmatic approach to speed up the current XML-RPC-based approach and make it easier to maintain/extend. I guess that's the main motivation, right?

Yes - maintainability (adding content types, not to have to change the code frequently when the DB schema changes, moving on from a complex codebase that's hard to reason about) is the first motivation, directly followed by performance. The bet about the latter is that by removing translation layers we can implement a faster solution.

> I'd still like to see some basic discussion on why alternatives like the pulp stack used in Katello haven't been considered.

My understanding of Pulp is that it implements a solution to the more general problem of content management, with content distribution being kind of a "side effect" given one needs to cache content anyway in order to implement management.

I would be open to evaluate replacing the whole content management functionality in Uyuni with Pulp, but that's a different problem/discussion than just transferring content between two Servers.

Do I read correctly that you are suggesting to evaluate Pulp just for content distribution here, while keeping content management features in Uyuni as they are?

> We have a slightly different use case where customers would like to see a full export of all the configuration and artifacts required for installing a system from, let's say, a USB thumb drive.
> 
> For that use case we'd need a different output format (e.g. "rendered" YUM repos plus Salt states) rather than SQL scripts. While I don't think we can solve that use case with the exact same export mechanism anyway, we should keep an eye on it and at least try to prepare for synergies.

Do you mean installing a system from a thumb drive without an Uyuni Server at all?

> A question about transaction-safety and frequency of exports/imports:
> 
> In traditional ISS environments, the syncs happen as part of a multi-hour or even multi-day patch management cycle.
> 
> This new approach will likely also be used much more frequently, as an almost real-time way of keeping distributed "proxies" in sync with the master.

Not necessarily. My thinking here is that an "export" from the ISS Master should be triggered as soon as a content management cycle is finished, so that a certain state is "frozen" in an exported directory. Then "imports" should be triggered to ISS Slaves (eg. via a Salt state), so that when the process is concluded, they will all be aligned.

Using a SQL file makes it very easy to make the "import" really atomic - care should be taken for the plain file part to avoid leaving ISS Slaves in a halfway state. But I do not expect major difficulties there - in the end I expect the whole "import" process to be atomic.

> Have we thought about how we can ensure that actions aren't triggered _before_ the necessary data has been properly synced? This was one of the reasons why an earlier Salt Syndic with git concept was dropped (gitfs not allowing to ensure data is present when it's needed by Salt in realtime).

Not explicitly. But it will be possible to use the fact the import is a Salt state application to make sure Actions are not started before application is complete. I can also add a bullet point in the Next Steps section to think about some more explicit locking mechanism.

> On the master server, could introducing such a transaction system negatively affect performance?

I do not think so, the lock will have to be on ISS Slaves to prevent Actions from being executed.

> Related to that, what's the plan if, let's say, over a thousand "slave" servers are to be synced regularly? Can the exports be parallelized?

The idea is that only one export will happen, and then data from the export will be transferred as plain files (SQL + RPM/deb/images etc) as many times as needed. In a high-scale situation, I am basically expecting API-based transports to work worse than storage-based ones.

If different content is needed on different ISS Slaves, multiple exports (I hope, a handful or so) can be created even in parallel.

> How much load do we expect on the server? And will this in turn match our scalability requirements? 

Hopefully less than what we have now, hopefully meeting requirements yes. In any case there is no way around get data out of the ISS Master DB + network transfer + getting data in the ISS Slave DB. The design idea behind this RFC is to trim any other operation to the absolute minimum.

That said we won't know until we have at least a complete PoC.

> As the proposal looks today, it's a good replacement for a full ISS. But is it the right approach for the distributed (retail-style) use case, where we may explicitly _not_ want to sync whole channel trees, but only what's absolutely necessary at that location?

This proposal will allow to select any managed content "entity" (eg. channels) by any criteria (eg. by ID, or their content by date range). Basically this will selectively export entities. This is something I could make clearer in the text.

> For example, we could improve on the current proposal by syncing all the DB data completely, but lazy-loading packages on demand (which would then of course in turn require a mechanism for starting the sync of what's going to be required for the next update run, so that the update run is offline-capable).

I forgot that in the Next Steps section and will add it.

> Another general/almost philosophical question is whether we should rather integrate with ideas like GitOps on the node level (which would mean that a lot of the data would be distributed directly via git checkouts etc.) or only interface with those tools on the central instances and then view the rest of the infrastructure as black boxes (i.e. we can use "hacks" like the SQL-based syncing without problems)?

Sorry but I could not really understand your writing here. What is a node in this context?

> Same for syncing images (in particular container images), which could mean using existing distributed artifact repos (docker registries etc.) or an Uyuni-specific (but highly efficient) implementation.

Well, images are already distributed via registries, so this RFC wouldn't change that. What we would sync is the rest of the metadata living in the Uyuni database, something else would be needed to sync registries assuming there is one close to the ISS Master and one close to the ISS Slave.

> As for the API-based transport: I'd mainly look into network topologies and ports. It would be great if we could get away with just the Salt ports (i.e. use the Salt filesystem or command/even bus as a transport) and potentially plain https://, so that things just work behind a NAT. If we can't (e.g. because the Salt filesystem doesn't scale/perform well enough), we should make sure we don't introduce any additional complications (think making all that work through a reverse ssh setup!) like extra open ports and one-way routeable network connections.

Will keep this in mind.

> The whole proposal assumes that we keep the SQL-centric backend for Uyuni. Particularly for channel management, of course there are alternatives. This (IMHO) doesn't directly affect the RfC as those changes would, if we decide to implement them, be much more long-term than the scope of this RfC.

Right. Or at least, some SQL metadata will always be needed, but could be much smaller in scope should we change Uyuni substantially in that direction.

joesusecom:
> > I'd still like to see some basic discussion on why alternatives like the pulp stack used in Katello haven't been considered.
 
> My understanding of Pulp is that it implements a solution to the more general problem of content management, with content distribution being kind of a "side effect" given one needs to cache content anyway in order to implement management.

Yes and no. Katello, from day one, has had a concept of capsules, where you may run multiple instances of, let's say, the repo server, for scalability and to match your topology. Not sure how well that works, but I'd assume that you could run a central full Katello with local "headless" capsules very simiar to how what we are planning.

> I would be open to evaluate replacing the whole content management functionality in Uyuni with Pulp, but that's a different problem/discussion than just transferring content between two Servers.

I don't think that would be a short-term option. But I _do_ think that we should know why it isn't.

> Do I read correctly that you are suggesting to evaluate Pulp just for content distribution here, while keeping content management features in Uyuni as they are?

Kind of. Certainly not to replace Manager with the whole Katello. And remember: Pulp is just one example. Maybe Artifactory is closer to our use case.


> Do you mean installing a system from a thumb drive without an Uyuni Server at all?

Correct.

The one where you still have a server also exists. But it's really just a special case of PXE boot where you boot your image from a media.

> > This new approach will likely also be used much more frequently, as an almost real-time way of keeping distributed "proxies" in sync with the master.

> Not necessarily. My thinking here is that an "export" from the ISS Master should be triggered as soon as a content management cycle is finished, so that a certain state is "frozen" in an exported directory. Then "imports" should be triggered to ISS Slaves (eg. via a Salt state), so that when the process is concluded, they will all be aligned.

I see. In that case, wouldn't it make most sense if that was deeply integrated with the concept of lifecycle management environments/stages? I.e., the point where we "promote" to the production environment is when we trigger the exports?

BTW, the design we have now (and the reason why we're talking about promoting environments) was inspired by how Katello does it, which in turn, I think, is due to the fact that to make things work there you actually need to do an explicit step to sync out all your changes (as opposed to the more "real-time" experience in Spacewalk, which isn't 100% real-time either because things like re-generating repo data on disk may take a while in the background).

> Using a SQL file makes it very easy to make the "import" really atomic - care should be taken for the plain file part to avoid leaving ISS Slaves in a halfway state. But I do not expect major difficulties there - in the end I expect the whole "import" process to be atomic.

Ok. I was thinking about scenarios where the whole central server would potentially have to be locked for a while (and not able to do other important stuff) to ensure that the snapshot is atomic.

On the other hand I realized later that maybe such a central server _is_ meant to spend most of the time syncing with the slaves, not doing "action work". Because it won't see any of the individual node's incoming Salt traffic. Reporting is of course another story, but can hit the central server on very different code paths.
 
> > Have we thought about how we can ensure that actions aren't triggered _before_ the necessary data has been properly synced? This was one of the reasons why an earlier Salt Syndic with git concept was dropped (gitfs not allowing to ensure data is present when it's needed by Salt in realtime).

> Not explicitly. But it will be possible to use the fact the import is a Salt state application to make sure Actions are not started before application is complete. I can also add a bullet point in the Next Steps section to think about some more explicit locking mechanism.

See above. Coupling the sync explicitly to a "stage transition" would make this much easier. We'd still need to define though whether a client that gets a patch update run triggered while we're still syncing the new environment version should

* see the old environment instead or
* be blocked until the environment is fully updated

> > On the master server, could introducing such a transaction system negatively affect performance?

> I do not think so, the lock will have to be on ISS Slaves to prevent Actions from being executed.

Ok.

> > Related to that, what's the plan if, let's say, over a thousand "slave" servers are to be synced regularly? Can the exports be parallelized?

> The idea is that only one export will happen, and then data from the export will be transferred as plain files (SQL + RPM/deb/images etc) as many times as needed. In a high-scale situation, I am basically expecting API-based transports to work worse than storage-based ones.

Ok, so a true 1:many scenario. We'll obviously have to validate that with reality. I _expect_ this to be the most common case, but we may also run into, let's say, retailers that run several different generations of stores and would have to sync a mildly (hopefully not wildly) different set of resources/data to each and every store.

> If different content is needed on different ISS Slaves, multiple exports (I hope, a handful or so) can be created even in parallel.

Ok, let's assume we don't have more than a couple (three to a dozen maybe). But with the current proposal, we wouldn't be able to use synergies between such export runs, right?

> > How much load do we expect on the server? And will this in turn match our scalability requirements?

> Hopefully less than what we have now, hopefully meeting requirements yes. In any case there is no way around get data out of the ISS Master DB + network transfer + getting data in the ISS Slave DB. The design idea behind this RFC is to trim any other operation to the absolute minimum.

Right. The actual sync via the network could be the bottleneck. May require a broadcast approach like we did locally with tftp-multicast in SLEPOS (when most stores still had non-switched 10MBit networks that would let you wait for hours when you want to install a dozen systems at the same time).

> That said we won't know until we have at least a complete PoC.

Understood.

> > As the proposal looks today, it's a good replacement for a full ISS. But is it the right approach for the distributed (retail-style) use case, where we may explicitly _not_ want to sync whole channel trees, but only what's absolutely necessary at that location?

> This proposal will allow to select any managed content "entity" (eg. channels) by any criteria (eg. by ID, or their content by date range). Basically this will selectively export entities. This is something I could make clearer in the text.

Ok. Makes sense.

> > Another general/almost philosophical question is whether we should rather integrate with ideas like GitOps on the node level (which would mean that a lot of the data would be distributed directly via git checkouts etc.) or only interface with those tools on the central instances and then view the rest of the infrastructure as black boxes (i.e. we can use "hacks" like the SQL-based syncing without problems)?

> Sorry but I could not really understand your writing here. What is a node in this context?

Basically I'm comparing two different topologies:

1) You have a GitOps workflow that _ends_ with your central Uyuni server. The Uyuni server behaves like a "Cloud API" of sorts that can manage all the moving parts (mostly Linux servers of course). Any communication between your central server and the nodes is not visible to your GitOps layer.

2) You have a GitOps workflow where distributed systems all get their share of truth from (parts of) the Git data (plus artifacts repos). Yes, you centrally coordinate actions from the headquarters, but configuration data, parameters etc. flow more or less directly from the repos to your clients (or at least the local "proxies".

We're currently focusing on 1), while 2) may (or may not) be another valid option, especially if it's hard to say how much centralization is actually desired (e.g. one customer may only want to share "golden update channels" with the local teams, the other wants fully centralized control).

> > Same for syncing images (in particular container images), which could mean using existing distributed artifact repos (docker registries etc.) or an Uyuni-specific (but highly efficient) implementation.

> Well, images are already distributed via registries, so this RFC wouldn't change that. What we would sync is the rest of the metadata living in the Uyuni database, something else would be needed to sync registries assuming there is one close to the ISS Master and one close to the ISS Slave.

I see.

> > The whole proposal assumes that we keep the SQL-centric backend for Uyuni. Particularly for channel management, of course there are alternatives. This (IMHO) doesn't directly affect the RfC as those changes would, if we decide to implement them, be much more long-term than the scope of this RfC.

> Right. Or at least, some SQL metadata will always be needed, but could be much smaller in scope should we change Uyuni substantially in that direction.

Exactly. I was thinking of things like Pillar data as JSON in PostgreSQL, where you'd potentially just replicate the whole "file" to the downstream server.

moio:
> > Do I read correctly that you are suggesting to evaluate Pulp just for content distribution here, while keeping content management features in Uyuni as they are?
> 
> Kind of. [...] Pulp is just one example. Maybe Artifactory is closer to our use case.

Added as alternatives.

> > Do you mean installing a system from a thumb drive without an Uyuni Server at all?
> 
> Correct.

I see that use case, but I think it does not belong to this RFC, and arguably not even in the Hub project in general. We can think about that, but it's definitely a separate effort.

> I.e., the point where we "promote" to the production environment is when we trigger the exports?

Added as next step.

> let's assume we don't have more than a couple (three to a dozen maybe). But with the current proposal, we wouldn't be able to use synergies between such export runs, right?

Right.


Can we get your approval now? :wink:

accepted/00063-hub-iss.md
```diff
@@ -0,0 +1,246 @@
...
+# Motivation
+[motivation]: #motivation
+
+ISS is currently limited to software channels, and it is not sure whether it can scale up above tens of Servers.
+
+From the requirements:
+
+> * Managed content can be selectively synced from Hub to Servers
+>   * In order of importance:
+>    - Configuration Management data (states, pillars...)
```
mcalmer: Should we list also software channels, product and subscription data here? This exists in the current ISS implementation but if you want to draw a full picture I think this is needed.

joesusecom:
> > > Do you mean installing a system from a thumb drive without an Uyuni Server at all?

> > Correct.

> I see that use case, but I think it does not belong to this RFC, and arguably not even in the Hub project in general. We can think about that, but it's definitely a separate effort.

Agreed. See the original comment. It was just about making sure we have that other use case in mind, not making it part of the scope.

> Can we get your approval now? wink

Yes.

accepted/00063-hub-iss.md
```diff
@@ -0,0 +1,246 @@
...
+3. Finally, yet other changes could really introduce a backwards compatibility problem and, since the protocol is versioned, [the import program will send an alert and abort](https://github.com/SUSE/spacewalk/blob/Manager-3.2/backend/satellite_tools/xmlSource.py#L271-L298).
+
+With the proposal in this RFC, it is possible to implement protection from case 3. via a similar versioning mechanism, partially against 2. ignoring missing tables but only offer limited resilience against case 1., as we would not have a real database-agnostic representation of data.
+
+pl/SQL could still be used, as a last resort, to apply different commands to different ISS Slave versions, but that would be hardly maintainable, so it would only be suggested to handle critical situations temporarily.
+
+General recommendation would be to upgrade an ISS Master, then upgrade ISS Slaves before attempting any new Inter-Server Synchronization. Another RFC will detail how Hub could make such upgrades easier.
+
```
mcalmer: With SQL export/import we limit us to stay on postgresql . Another DB might not be possible because of different SQL syntax and features.

We do not plan to support something different. So this is not a big problem.

accepted/00063-hub-iss.md
```diff
@@ -0,0 +1,246 @@
...
+    "name"
+)
+VALUES
+    (500, 'channel-ia32', (SELECT id FROM rhnarchtype WHERE label =  'rpm'), 'IA-32' - ),
+    (501, 'channel-ia32-deb', (SELECT id FROM rhnarchtype WHERE label =  'deb'), 'IA-32 Debian' - )
+ON CONFLICT ...
+```
+
+At this point it is believed/assumed that:
+  - ISS is by and large an additive process, so cleanups should not be needed for most tables. For those where it is necessary (eg. linking tables in an m:n relationship), `DELETE` statements either before or after `INSERT`s suffice
```
mcalmer: When customers do cleanups, they often remove channels, and software packages and Errata. If we start to sync also images and containers I think they will be changed and removed much more often.

I think we should be sure that "Cleanup" is also working. I think this is important and will happen more often then we think.

accepted/00063-hub-iss.md
```diff
@@ -0,0 +1,246 @@
...
+  - the ISS programs need (potentially) no updating when the database structure changes
+  - supporting a new managed content type is as easy as specifying a new set of starting tables
+
+### Transport
+
+Both file and database data need to be transferred somehow from the ISS Master to the ISS Slave.
+
+Current ISS implementation offers:
+  - storage-based transfer (a directory with well-defined structure is created to host files and data in XML form) and
+  - API-based transfer: an XMLRPC request is made from ISS Slave to Server, and the same deta set is sent over XMLRPC (http), streamed
```
mcalmer:
```suggestion
  - API-based transfer: an XMLRPC request is made from ISS Slave to Server, and the same data set is sent over XMLRPC (http), streamed
```

accepted/00063-hub-iss.md
```diff
@@ -0,0 +1,246 @@
...
+[alternatives]: #alternatives
+
+- re-use of the current ISS implementation, extending it to cover other content types
+  - con: current implementation proved difficult to maintain. Complicating factors are:
+    - data is exchanged in XML form which needs explicit marshaling and unmarshaling from the database (types and structure)
+    - marshaling implementation [requires extensive mapping onto the schema](https://github.com/uyuni-project/uyuni/blob/1638692858c3d0a9c8342524c34307842d57136d/backend/server/importlib/backendOracle.py#L32-L651) to be manually kept in sync
+    - supports two backends (Postgres and Oracle), but the second has been dropped
+    - format is versioned and code is meant to cope with version differences
+    - format contains some data summaries for performance reasons (eg. packages and packages_short)
+    - XML files are compressed, as they tend to get large
```
mcalmer: Just to be fair: some of these points will also apply to the SQL variant described in this RFC.

accepted/00063-hub-iss.md
```diff
@@ -0,0 +1,246 @@
...
+
+1. Some changes in the database structure could be "internal only", not really changing the nature of the information that's being exchanged. In those cases, the current implementation could make use of the fact the the XML representation does not change, thus the ISS Slave continues to operate normally
+2. Other changes would be reflected in the XML structure, but still not create problems because, for example, the current implementation ignores unknown tags. This is the case for completely new elements that the ISS Slave would just not import (which is most likely conceptually correct, as tables would also be missing)
+3. Finally, yet other changes could really introduce a backwards compatibility problem and, since the protocol is versioned, [the import program will send an alert and abort](https://github.com/SUSE/spacewalk/blob/Manager-3.2/backend/satellite_tools/xmlSource.py#L271-L298).
+
+With the proposal in this RFC, it is possible to implement protection from case 3. via a similar versioning mechanism, partially against 2. ignoring missing tables but only offer limited resilience against case 1., as we would not have a real database-agnostic representation of data.
```
mcalmer: I think this is the key point and we need to think carefully about it. Is it acceptable that we force a customer to update there whole Hub/Server infrastructure "together" at the same day? In worst case we talk about 1 Hub with 9k Servers.

Otherwise it he is not able to sync content from hub to not migrated servers.

Another point is the "internal only" problem. Are all developers aware that they potentially break ISS when they restruture DB tables and data? It should not be too easy to break it.

accepted/00063-hub-iss.md
moio: Hmm, maybe I am overlooking something, but which ones do you think apply? I am not seeing any.

mcalmer: - Also sql scripts will be big and might need compression
- creating the SQL scripts might be similar effort to create XML snippets

But this was just a note

accepted/00063-hub-iss.md
moio: Copypaste error, this was in the requirements.

https://github.com/uyuni-project/uyuni-rfc/blame/master/accepted/00060-hub-general.md#L34

Fixed.

accepted/00063-hub-iss.md
moio: I should have made this evident in the Drawbacks section, added.

accepted/00063-hub-iss.md
moio: As per our Mumble chat: here I am referring to low-level, database row cleanups and not high-level, "entity" cleanups like removing an entire channel or an image.

The latter are not addressed by the current ISS but are always possible through regular UI/APIs, I expect that not to change with the proposal in this RFC.

I am now marking this resolved

accepted/00063-hub-iss.md
moio: My best idea so far about compression is to use `rsync`, which does that transparently in addition to delta-encoding, so that we do not have to implement that - but that's something for a follow-up RFC when API-based transport is discussed.

As for the creation of the SQL script, I hope it will be faster than the current XML mashaling layer, but of course that can't be determined without a PoC. Maybe we will be able to prepare something in that direction.

I am marking this as resolved for now, feel free to reopen if you think any change in the text is appropriate

accepted/00063-hub-iss.md
```diff
@@ -0,0 +1,250 @@
...
+  - be limited to software channels
+  - be limited to storage-based transfers
+
+## Implementation details
+
+- two commandline tools will be added, one to be used on an ISS Master and one on an ISS Slave
+  - output of the ISS Master tool will be a directory with files and a SQL script
+  - the ISS Slave tool will mainly copy files to correct locations and execute the SQL script
+  - triggering of the ISS Slave tool might happen via Salt, this is left as an implementation detail
+  - transfer proper could be done via `rsync`, which provides transparent compression and delta encoding
```
renner: If an external tool should be used maybe [`rclone`](https://rclone.org/) could be an alternative here: from my experience it has a much more user-friendly CLI than `rsync` (there seems to be a [python wrapper](https://pypi.org/project/python-rclone/) too), additional features (e.g. sync to and from network), support for all kinds of storage types, so API based transport could maybe even be implemented as an Uyuni plugin.

moio:
@mcalmer, I tried solve the update dilemma by the numbers.

I wrote [a script](https://gist.github.com/moio/55459d92bfade2119e6b228a214b7271) that looks into our schema upgrade directories and looks for `ALTER TABLE` commands against a set of known "hot tables". Currently I defined "hot tables" as the tables handled by our current ISS implementation - packaged software substantially.

It found that, in SUSE Manager major versions 3.0 to 4.0:
 - less than 2% of the individual SQL files alter hot tables (8/433)
 - about 11% of the migrations alter hot tables (6/52)

Looking at those 6 breaking migrations 1 is duplicate and 2 are expected work anyway (eg. `DROP` of a column which had a `DEFAULT`), so less than 4% (2/52) would really require an upgrade. Details:

| Upgrade                                                | File                                                        | Breaking?     |
|--------------------------------------------------------|-------------------------------------------------------------|---------------|
| 3.0.20 -> 3.0.21 |                                                             | Yes           |
|                                                        | 001-add-channel-gpgcheck-column.sql.postgresql              |               |
| 3.1.0 -> 3.1.1   |                                                             | Yes           |
|                                                        | 1001-drop-rhnContentSourceSsl-move-data.sql                 |               |
|                                                        | 1004-rhnChannelPackage-autovacuum.sql.postgresql            |               |
|                                                        | 2005-rhnContentSourceSsl-move-data.sql                      |               |
| 3.1.8 -> 3.1.9   |                                                             | No, duplicate |
|                                                        | 001-add-channel-gpgcheck-column.sql.postgresql              |               |
| 3.2.15 -> 3.2.16 |                                                             | No, works     |
|                                                        | 005-drop-productUrl.sql.postgresql                          |               |
| 4.0.3 -> 4.0.4   |                                                             | No, works     |
|                                                        | 005-drop-productUrl.sql.postgresql                          |               |
| 4.1.0 -> 4.1.1   |                                                             | No, works     |
|                                                        | 000-use-blob-instead-of-varchar-rhnpackagechangelogdata.sql |               |

Of course we can't easily predict the change rate, especially on tables representing newer content types such as images, but at least we can guess that in the vast majority of cases (90% or so) ISS will continue to work even on slightly different versions of Masters and Slaves at least as far as packages and repos are concerned.

On those cases where compatibility is broken we could either recommend to align versions or develop a specific patch for that particular case.

How does that sound?

paususe:
A few questions come to my mind:

- Why does the current ISS use XML? For compatibility with several databases? For some other reason?
- "_transfer proper could be done via rsync or rclone, which provide transparent compression and delta encoding_": would it make sense to use git + git lfs + gitfs as a content backing store to put RPMs, DEBs, containers, Salt states/pillars/etc in git and having the metadata in the database? Sync would be as simple as git pull/push with steroids ("steroids" = database stuff, be it plain SQL or serialized XML), export would be git clone --shallow with steroids, etc. Plus git is atomic: either it transfers and applies everything, or it does not. In addition to that, we'd be prepared for more advanced GitOps features.
- With the SQL approach, what happens with e. g. stored procedures? I guess those only happen when a new version of Uyuni is released :-?
- As MC said: can we ask people to update all Servers at the same time? Given that the Hub and the Server would have different versions of the database, ISS would not work, thus making it impossible to transfer the Server update packages via ISS. Is this a catch-22 or am I missing something?

moio:
> * Why does the current ISS use XML? For compatibility with several databases? For some other reason?

First reason is historical - the ISS protocol is the same RHN used (to push content from the Red Hat Network to Red Hat Satellites). Basically the current ISS implementation was a cheap adaptation of that protocol that was basically already there.

Second reason, as you guessed, is that multiple database backends were supported in Spacewalk.

> * "_transfer proper could be done via rsync or rclone, which provide transparent compression and delta encoding_": would it make sense to use git + git lfs + gitfs as a content backing store to put RPMs, DEBs, containers, Salt states/pillars/etc in git and having the metadata in the database? Sync would be as simple as git pull/push with steroids ("steroids" = database stuff, be it plain SQL or serialized XML), export would be git clone --shallow with steroids, etc. Plus git is atomic: either it transfers and applies everything, or it does not. In addition to that, we'd be prepared for more advanced GitOps features.

That could be an option, but we will have to pay attention to bandwidth usage. With `rsync`, it's easy to only download the "delta" in compressed form, now with git this is also possible, but usually one downloads the full history. I guess that will depend on what the use will be on the other end.

> * With the SQL approach, what happens with e. g. stored procedures? I guess those only happen when a new version of Uyuni is released :-?

Stored procedures would not be exported/imported, this is only about table data. That means SPs could change as part of (maintenance) updates just like any other part of the codebase.

> * As MC said: can we ask people to update all Servers at the same time? Given that the Hub and the Server would have different versions of the database, ISS would not work, thus making it impossible to transfer the Server update packages via ISS. Is this a catch-22 or am I missing something?

Fortunately this is not a catch-22 situation. Servers would be minions to the Hub's master, so they will consume channels directly from the Hub. That means they get their updates through regular http repos, not via ISS.

Practically this would mean that:
 - in the [expected majority of cases](https://github.com/uyuni-project/uyuni-rfc/pull/21#issuecomment-547446373) ISS will just work while Hub and Servers are on different versions
 - when a particular update strictly requires Servers to be updated to the Hub version (eg. major releases):
  - users will update the Hub first. At that point ISS will return an error if attempted and not proceed with synchronization further
  - then update the Servers to the same version, via usual content management/patch management from the Hub
  - then ISS will resume working normally

moio:
@paususe, this is now ready to be merged as soon as you give your OK, unless you have further questions of course.

paususe:
> > * Why does the current ISS use XML? For compatibility with several databases? For some other reason?
> 
> First reason is historical - the ISS protocol is the same RHN used (to push content from the Red Hat Network to Red Hat Satellites). Basically the current ISS implementation was a cheap adaptation of that protocol that was basically already there.

Just one more question: do you know why did RHN use this XML-based protocol for RHN-RHS communication?

My point is, given that XML is much more inefficient than SQL, there must be a reason why they used XML back in the day. I wouldn't want to overlook it and then find out the hard way.

moio:
> do you know why did RHN use this XML-based protocol for RHN-RHS communication?

I am not aware of reasons other than multi-database support and the need for maximum compatibility with older RHS versions. That would have represented a Catch-22 situation like the one you described AFAIU.

joesusecom:
> My point is, given that XML is much more inefficient than SQL, there must be a reason why they used XML back in the day. I wouldn't want to overlook it and then find out the hard way.

I don't think there was any particular reason. Back in the day, _everyone_ was using XML for data exchange.

mcalmer:
> > My point is, given that XML is much more inefficient than SQL, there must be a reason why they used XML back in the day. I wouldn't want to overlook it and then find out the hard way.
> 
> I don't think there was any particular reason. Back in the day, _everyone_ was using XML for data exchange.

Right. And it is not just some XML protocol, it is XMLRPC. Maybe at that point in time the most used protocol for exchanging data.

Today everybody is doing REST with JSON, at that point everybody where doing XMLRPC or some even SOAP.

PR #22, opened 2019-10-24
RFC for 'Modular repositories with CLM'

cbbayburt:
Fixes: https://github.com/SUSE/spacewalk/issues/9732


accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,180 @@
...
+The API utilizes the Python port of [libmodulemd](https://github.com/fedora-modularity/libmodulemd), which is used for structural parsing of `modules.yaml` files.
+
+*TODO: Add the file*
+
+An example of such implementation is provided in `attachments/00000-modular_repos_api.py`.
+
+### UI
+The UI consists of a new type of CLM filter called "module". The filter provides inputs for a module name and a stream name.
+Each filter selects a stream for a single module. The stream input can be left empty to select the default stream.
+Since a filter is independent of the source repositories in a CLM project, it is not possible to display a list of modules/streams on the filter form and the inputs can only be simple text inputs.
```
paususe: What would be needed in order to have a list of modules/streams?

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,180 @@
...
+## Components
+
+This section summarizes the different components involved in the design.
+
+### Backend
+The Java backend is responsible for filtering out unselected module streams and building the repository with the desired packages through CLM.
+
+### API
+A Python API parses and interprets the module metadata, resolves modular dependencies and translates specified modules and streams to an actual package list.
+The API utilizes the Python port of [libmodulemd](https://github.com/fedora-modularity/libmodulemd), which is used for structural parsing of `modules.yaml` files.
```
paususe: Libmodulemd is only available in PackageHub and only for Python 3. 

I assume we will distribute in the SUSE Manager channels for SLE15 SP1. What about SP2? Still in our channels?

Is it OK to have it only for Python 3?

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,180 @@
...
+The current version of Uyuni supports modular repositories in the most basic way. Modular repositories are recognized, mirrored correctly and can be assigned systems. However, doing package operations on these systems through the Uyuni UI is not supported. These systems must be managed from their own package managers.
+
+The next step to supporting modular repositories would be to implement modularity in the core with dedicated logic for package operations and a new set of UI. The most obvious benefit of this would be to be able to do package operations through the UI while still providing modular repositories to the clients. However, this implementation would require much more effort and currently beyond the scope.
+
+# Further enhancements
+[enhancements]: #enhancements
+
+### Interactive inputs for filters
+To provide better user experience, the filter inputs can be turned into responsive dropdowns showing a list of module/stream names with realtime feedback.
+However, since the module list can only be retrieved from the project sources and the current implementation of CLM filters are independent of the sources, this is not currently achievable.
```
paususe: What would we need to change in our current implementation of CLM filters? Would it also enhance the experience for other OSs?

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Selective version filtering
+If required, it is possible to selectively filter a specific version of any module by introducing additional types of module filters.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+ - How to communicate between the Python API and the Java backend
+ - What should be the behavior for the unselected modules? Should the modules be disabled unless there is a filter targeting them? Should they be enabled with default streams in the best effort?
```
paususe: What does Fedora/RHEL/CentOS do? We should do the same

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,180 @@
...
+ - What should be the behavior if no module filter is added to a project? Should the resulting repository be "flattened" or should it be modular?
```
paususe: What does Fedora/RHEL/CentOS do? We should do the same

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,180 @@
...
+To provide better user experience, the filter inputs can be turned into responsive dropdowns showing a list of module/stream names with realtime feedback.
+However, since the module list can only be retrieved from the project sources and the current implementation of CLM filters are independent of the sources, this is not currently achievable.
+
+### Selective version filtering
+If required, it is possible to selectively filter a specific version of any module by introducing additional types of module filters.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+ - How to communicate between the Python API and the Java backend
```
paususe: Start a process?
SWIG-generated bindings to the C library? (risk to crash the JVM)
Jep? https://github.com/ninia/jep
VOC? (probably too risky) https://github.com/beeware/voc

accepted/00000-modular-repos-with-clm.md
cbbayburt: By design, CLM filters are not related to any channels at all. Because of that, it's not possible to list the modules initially.
In "Filters" page, there is no channel information that we can use.
However, in the project page, we can conditionally add some kind of autocomplete once some sources are added to the project. But this would be just a supplementary enhancement.

accepted/00000-modular-repos-with-clm.md
cbbayburt: It is only required for SUSE Manager server, so Python 3 will be sufficient.

accepted/00000-modular-repos-with-clm.md
cbbayburt: We will add a new type of filter and the additional logic to process this filter. The current experience won't be affected in any way.

accepted/00000-modular-repos-with-clm.md
cbbayburt: dnf automatically figures out and enables the default stream for the module when a package install is requested. Since SUMA does not recognize modularity at the package operations level, this is not possible for us.

The question is about UX. The point is to set all the modules either enabled or disabled. Let's say I have 10 modules available and I added 4 filters to select streams for 4 modules. What happens to the remaining 6?
 - Should we disable them?
 - Should we enable them with default streams? What if this combination is not possible?

accepted/00000-modular-repos-with-clm.md
cbbayburt: Module filters and flattening a repository into a regular one are the concepts this RFC introduces, inherited from CLM features. `dnf` does not "flatten" a repository. We must pick our own way in this case.

The point to consider is: what does a user want to achieve by adding a modular channel to a CLM project other than selecting some modules?

Also, keep in mind that a modular repository usually contains some regular RPMs as well. They don't exclusively contain modules. What if the user only wants to work on those regular RPMs in CLM?

accepted/00000-modular-repos-with-clm.md
cbbayburt: IMO, implementing another layer in Python on top of `libmodule` makes this communication a lot easier. I've already written a PoC for this purpose (I'll attach it shortly). With that extra layer, we'd only need a couple of method calls from Python. And to keep things as simple as possible, I'd vote for "starting a process".

cbbayburt:
@paususe thanks for the quick review! I'll also try to clarify further in the document the things you pointed out.

accepted/00000-modular-repos-with-clm.md
cbbayburt: We would need to relate filters to projects so that a filter cannot exist outside of a project's scope. But I believe that's a radical change in the current CLM implementation.

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,179 @@
...
+
+This section describes the detailed design for the different phases of a CLM project.
+
+### Project setup
+At the project setup, the backend checks if any of the sources are modular repositories.
+This check is done by doing a `getChannel().getModule()` call on the source object.
+For every modular source, a proper indicator is displayed as described in the UI component section.
+
+Currently, every CLM filter displays a field to choose one of two hard-coded rules: `allow` or `deny`. For module filters, this field must be hidden.
+Each added module filter stores the value of a chosen module/stream in `name` or `name:stream` format.
```
mcalmer: Not sure if this would work, as a customer can add conflicting filters.
Additionally a created filter does not mean that that filter gets assigned to a project.

I think we need to find a way to have only 1 filter per project and you set everything in this one filter.

accepted/00000-modular-repos-with-clm.md
mcalmer: Well, but I think it makes sense to think about it. Here we have the first real use case where it makes sense to create a filter for a specific project.
But on the other hand this is not really true. It is more a filter for specific source selection.

The idea could be to add a validator to every filter. Such a filter can only be added to a project if the validation succeed. Otherwise we could not display it in the list of filters in a project.

accepted/00000-modular-repos-with-clm.md
mcalmer: I would suggest to do it similar to virtual-host-gatherer, subscription-matcher or pam auth.
Just write a program, define API via STDIN and STDOUT (STDERR) - example json in/out and call the program from java

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,179 @@
...
+### API
+A Python API parses and interprets the module metadata, resolves modular dependencies and translates specified modules and streams to an actual package list.
+The API utilizes the Python port of [libmodulemd](https://github.com/fedora-modularity/libmodulemd), which is used for structural parsing of `modules.yaml` files.
+
+An example of such implementation is provided in [attachments/00000-modular-repos-api.py](attachments/00000-modular-repos-api.py).
+
+### UI
+The UI consists of a new type of CLM filter called "module". The filter provides inputs for a module name and a stream name.
+Each filter selects a stream for a single module. The stream input can be left empty to select the default stream.
+Since a CLM filter is not related to any channels, it is not possible to display a list of modules/streams on the filter form and the inputs can only be simple text inputs.
```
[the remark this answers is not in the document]
cbbayburt: In the current implementation, filters and projects have an `n-to-n` relation. Furthermore, the "Filters" page has no relationship with projects at all. I think it would be only possible with some fundamental changes and/or a lot of exceptional code.

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,179 @@
...
+The module information for a modular repository is displayed under an additional "Modules" tab on the "Channel Details" page.
+A modular repository is indicated with an icon next to the name of the repository in the CLM sources list. A click on the icon navigates to the "Modules" tab of the channel details.
```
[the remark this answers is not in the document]
cbbayburt: Assuming we won't have any feedback in the filter form (simple text inputs) this would be the only way to provide information about the modules. So I'd say it's mandatory.

accepted/00000-modular-repos-with-clm.md
mcalmer: My suggestion: disabled by default.

I think it is always possible to disable everything, but not to enable everything.

accepted/00000-modular-repos-with-clm.md
mcalmer: My suggestion: without filter it stay as "modular" repository.

accepted/00000-modular-repos-with-clm.md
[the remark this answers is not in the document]
cbbayburt: Also, this would impose a lot of complexity. For example, what happens when you remove one of the sources after you added filters to the project? We'll have to do a lot of checks to keep the project in a good state.

accepted/00000-modular-repos-with-clm.md
cbbayburt: This is all about the UX. Choosing either way is just a matter of designing the UI accordingly, and serializing the filter data to a string. I guess both are achievable. But since the filter forms are modal dialogs, I wonder if it would be convenient to list all modules in one.

> Not sure if this would work, as a customer can add conflicting filters.

On the other hand, a part of the implementation will be to give user early feedback with red notification errors as soon as they input an invalid combination, I don't think it's a problem.
We can do a sanity check whenever an individual filter gets updated in real-time to check if the update plays well with the existing filters and sources and notify the user.

accepted/00000-modular-repos-with-clm.md
cbbayburt: > The idea could be to add a validator to every filter. Such a filter can only be added to a project if the validation succeed. Otherwise we could not display it in the list of filters in a project.

This is a more realistic approach IMO. We can validate the project setup at every modification and notify the user in real-time.

accepted/00000-modular-repos-with-clm.md
cbbayburt: > We can do a sanity check whenever an individual filter gets updated in real-time to check if the update plays well with the existing filters and sources and notify the user.

I'll add a part to explain this in detail

accepted/00000-modular-repos-with-clm.md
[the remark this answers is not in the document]
mcalmer: Agreed. This highly depends on how we implement the filter.

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,181 @@
...
+```
+
+The goal is to handle modular repositories in a way that the clients can be still managed through the Uyuni UI.
+
+# Detailed design
+[design]: #detailed-design
+
+## Overview
+
+Instead of implementing the modularity logic inside the core of Uyuni, this document describes a design to simulate the modularity behavior by building regular repositories out of modular repositories by filtering out packages from unselected streams.
```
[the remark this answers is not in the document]
cbbayburt: Not much. in the general sense, the concept of modularity is directly related to a repository so that's why I decided to go with it.

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,181 @@
...
+
+## Components
+
+This section summarizes the different components involved in the design.
+
+### Backend
+The Java backend is responsible for filtering out unselected module streams and building the repository with the desired packages through CLM.
+
+### API
+A Python API parses and interprets the module metadata, resolves modular dependencies and translates specified modules and streams to an actual package list.
```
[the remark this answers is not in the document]
cbbayburt: It's a standalone Python executable, and the API is via standard in/out (a similar concept to the salt runners).

cbbayburt:
@moio thanks for the review. I'll clarify your points further in the document.

accepted/00000-modular-repos-with-clm.md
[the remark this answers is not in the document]
moio: This sounds better and I'd suggest to make it explicit in the text here.

accepted/00000-modular-repos-with-clm.md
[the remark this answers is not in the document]
moio: Hmm, that's true as well. I asked because I was confused - at first read it might seem the RFC would result in a feature that has repos as an output (and then users have to create channels out of them), while actually knowing the CLM feature I am sure channels is what comes out at the end.

Feel free to either ignore or clarify if you find a way to. In the end this is nothing substantial really, so I am marking this as resolved.

Thanks!

accepted/00000-modular-repos-with-clm.md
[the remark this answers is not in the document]
ncounter: So, do I understand correctly that you should populate the filter field based on the selected sources? If so, yes at the moment it is completely impossible.
The only option I can see at the moment, without changing the entire architecture, is to populate such a filter field for modules listing everything that is available at the SUSE Manager level. Still detached from the project sources selection set, but a user-friendly way to select modules instead of having a completely free-text input field the user has to populate typing manually. Just my 2c

accepted/00000-modular-repos-with-clm.md
ncounter: I think we can add the warning for limiting to one filter only of this specific type when attaching filters to a project. Then, doing all the module filtering for a modular repository in one filter only, it can be added on top of my previous comment about a select box.
Let's split the filter in two fields: a list of all possible modular repositories, and a list of modules for the selected modular repository. Pretty the same we do in ActivationKeys selecting the parent channel and changing dynamically the list of options for child channels. Would this work?

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,194 @@
...
+
+### Project setup
+At the project setup, the backend checks if any of the sources are modular repositories.
+This check is done by doing a `getChannel().getModule()` call on the source object.
+For every modular source, a proper indicator is displayed as described in the UI component section.
+
+Currently, every CLM filter displays a field to choose one of two hard-coded rules: `allow` or `deny`. Since the module filtering logic works in an implicit "deny all" fashion (See build/promote section), this field must be hidden for module filters.
+Each added module filter stores the value of a chosen module/stream in `name` or `name:stream` format.
+
+Whenever a change is made in any of the filters, a call to the same API endpoint is made to check the validity of the current selections to provide real-time feedback when the current combination is not a valid one. The API's error reporting mechanism is used to determine the failure and provide meaningful messages to the user in the UI.
```
ncounter: At the moment we clearly state that changing a filter will not be reflected in projects where it is in use until a Build is triggered again. So the filtering will not be applied before you Build again the project: adding a warning there could be sufficient, but of course if we can early do in realt-time it would be perfect.

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,194 @@
...
+
+The functionality will be implemented through a new interface called `ProjectEntity`, which will declare a method to list any feedback messages that the entity currently has. An entity can inspect the `ContentProject` object and determine if it has any feedback messages and return them with this method. The method will be called for every entity during the rendering of the project response.
+
+In contrast to using flash messages, the feedback messages generated by the project entities will persist in the UI as long as the project's state is unchanged, making sure the messages get the user's attention and encouraging the user to modify the project to make any warnings/errors disappear.
```
ncounter: :+1: I think the right place to present these messages is right before the Build button and after the Filter box. In case of any `[ERROR]` is present, we can even decide to disable the Build as well.

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,194 @@
...
+
+### Error reporting & feedback interface
+To provide a convenient user experience, we must provide feedback for user's actions as early as possible. For this purpose, the CLM project page will have a new panel to display feedback in the form of a list of info, warning, alert items. Each item will provide feedback for a different aspect of the current project setup. That means, each project entity (a filter, a source, etc.) will report its own feedback in case it has one. This will be implemented in a generic way which can be extended to the existing features, by adding a separate data item to the `ProjectResponse` class. Doing so will make it possible for each entity to report its own messages whenever a project response is rendered.
+
+The functionality will be implemented through a new interface called `ProjectEntity`, which will declare a method to list any feedback messages that the entity currently has. An entity can inspect the `ContentProject` object and determine if it has any feedback messages and return them with this method. The method will be called for every entity during the rendering of the project response.
```
ncounter: While I see this as an harder way to go for it, I really like the approach of independency: the project is not the active actor who checks all the entities status but they will report to it whenever they know about themselves. At the moment the only entity type that can actively do this is the filter, but this could be useful whenever the feature will grow and more content types will be enabled and so on. :+1:

accepted/00000-modular-repos-with-clm.md
ncounter: As long as we can keep the filter detached from the project context, I'd keep it. We started thinking about it and early feedback were "no please, generic filters to be reused are the most important one". Then we got the request for filtering at an Environment level. As always there is at least a use case to have a different implementation, but I'd keep the feature *as simple as possible*, and it is already getting complexity every day.
So to me the solution is to keep the filter context separated from the project: as for the ActivationKey workflow, we can populate the first dropdown box with all possible modular repositories. Based on this first selection, we present a second filter step, where the user can choose the modules he wants. Here at this point we can do the validity check, and we can even disable saving the filter if a bad selection is attempted. Or we flag the filter as *in error* and we reflect this information in projects where the filter is attached, with the status messages mentioned in the `### Error reporting & feedback interface` section of this RFC.

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,194 @@
...
+Each added module filter stores the value of a chosen module/stream in `name` or `name:stream` format.
+
+Whenever a change is made in any of the filters, a call to the same API endpoint is made to check the validity of the current selections to provide real-time feedback when the current combination is not a valid one. The API's error reporting mechanism is used to determine the failure and provide meaningful messages to the user in the UI.
+
+### Build/promote
+At build time, all the module filter values are read in the form of `name:stream` tuples or only `name` in case the default stream is chosen.
+The remaining modules where a selection has not been made are set to disabled (except the case that no module filters are attached at all, see below).
+With a list of name/stream tuples, the API is called to resolve the modular dependencies and the appropriate contexts (see next section).
+If successful, the API returns a list of full RPM names belonging to all the enabled streams.
+Additionally, the backend retrieves a full list of all the modular RPMs and subtracts the enabled ones from this list. The resulting list is translated into corresponding `Package` objects.
```
mcalmer: We need to check what happens to `epoch` in the full package names. In SUSE Manager the epoch is not part of the filename which create some trouble. We should check the artifact list if they contain the epoch or not.
If they do we are fine. If not, we can expect some extra work.

accepted/00000-modular-repos-with-clm.md
```diff
@@ -0,0 +1,194 @@
...
+The Java backend is responsible for filtering out unselected module streams and building the repository with the desired packages through CLM.
+
+### API
+A standalone Python executable that parses and interprets the module metadata, resolves modular dependencies and translates specified modules and streams to an actual package list. The API access is via STDIN (CLI args) and STDOUT in JSON string.
+
+It utilizes the Python port of [libmodulemd](https://github.com/fedora-modularity/libmodulemd), which is used for structural parsing of `modules.yaml` files.
+An example of such implementation is provided in [attachments/00000-modular-repos-api.py](attachments/00000-modular-repos-api.py).
+
+### UI
+The UI consists of a new type of CLM filter called "module". The filter provides inputs for a module name and a stream name.
```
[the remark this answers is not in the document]
ncounter: What about `repo-stream`? Anyway you're right, it makes sense to name it clearly with no overlapping meanings. :+1:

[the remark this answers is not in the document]
cbbayburt: "Module stream" and "Application stream" are some alternatives used throughout the RHEL documentation. So to prevent confusion while keeping it familiar, I'd also suggest "appstream" or some variation of it.

cbbayburt:
@paususe @ncounter @mcalmer final draft ready for reviews

cbbayburt:
> Overall a good RFC. One point is missing we discussed in our meeting. We may need to filter "Errata" in the module repos and filter out updateinfo with packages which do not exist in the repo.

In CLM, this would affect package and module filters alike and there shouldn't be any special handling required in the context of appstreams. I've created https://github.com/SUSE/spacewalk/issues/10131 to test if it's working as expected.

PR #23, opened 2019-10-27
Initial RFC draft about LDAP sync tool

isbm:
RFC about LDAP synchronisation tool, which should completely replace existing Python script `sw-ldap-user-sync`.

[Read it here](https://github.com/uyuni-project/uyuni-rfc/blob/isbm-ldapsync/0000-ldap-sync-tool.md).

### DIFF
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -12,7 +12,7 @@ from a directory service.
 [motivation]: #motivation
 
 Uyuni server can authenticate users from LDAP via PAM. For this, users
-needs to be added to Uyuni with the flag "PAM authentication" and thus
+need to be added to Uyuni with the flag "PAM authentication" and thus
 their credentials will be verified via LDAP directory. Synchronisation
 tool should transparently synchronise users from certain groups and/or
 roles in LDAP and reflect that in Uyuni server, overriding local
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -189,9 +189,24 @@ N/A
 # Alternatives
 [alternatives]: #alternatives
 
+## Native Implementation
+
 Alternative design would be to extend user management in the very
 Uyuni server, making it LDAP-aware natively. However, the amount of
 changes to the existing code might be very large.
+
+## Salt-based Implementation
+
+It could be possible to reuse existing Spacewalk runner in Salt,
+wrapping it into a state module and implement all that as a Salt
+Formula. While such approach would give more flexibility to the user,
+it will also have a number of drawbacks, such as additional formula
+debugging difficulties overhead; user-introduced bugs that might sync
+LDAP wrong with severe consequences; fragility of the very solution:
+it is still an editable Formula.
+
+Salt, however, may call `uyuni-ldap-sync` and/or setup configuration
+for it (PAM, the `uyuni-ldap-sync` configuration etc).
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -5,8 +5,8 @@
 # Summary
 [summary]: #summary
 
-The `mgr-ldapsync` is a tool that performs synchronisation of the
-users between Uyuni server and LDAP.
+The `mgr-ldapsync` is a tool that configures and manages users in Uyuni
+from a directory service.
 
 # Motivation
 [motivation]: #motivation
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -164,6 +164,23 @@ UI should mark external users either with special icon or color etc
 and the page of such user should have disabled "Update" button and an
 explanation for such user is managed elsewhere externally.
 
+# Reference Implementation
+
+This is already implemented in the prototype which is available at 
+[uyuni-ldap-sync repository](https://github.com/isbm/uyuni-ldap-sync).
+
+Current prototype has two releases:
+1. [v0.1](https://github.com/isbm/uyuni-ldap-sync/releases/tag/v0.1) just reimplements
+[existing script in Uyuni](https://github.com/uyuni-project/uyuni/blob/master/utils/sw-ldap-user-sync).
+2. [v0.2](https://github.com/isbm/uyuni-ldap-sync/releases/tag/v0.2) implements the current RFC and adds Active Directory support
+
+Moving from prototype status to a production release, the following
+would be still required:
+
+- Automatic (transparent) PAM authentication setup
+- Add unit tests
+- Possible refactorings
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -177,9 +177,15 @@ Current prototype has two releases:
 Moving from prototype status to a production release, the following
 would be still required:
 
-- Automatic (transparent) PAM authentication setup
 - Add unit tests
 - Possible refactorings
+
+## Additional Features
+
+Features that might come with the next release:
+
+- Automatic (transparent) PAM authentication setup/un-setup
+- Daemon mode
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -121,6 +121,20 @@ will be shown.
 The output should be stored either to a configured log file or to the
 STDOUT, if requested.
 
+# Uyuni UI Enhancements
+
+Considering that this tool a new way of creating/modifying users
+beside the UI and API in an automated way, there should be an API
+endpoint that allows to set a flag in User Management that gives
+admins the clear indication that the certain user is managed externally.
+
+Main reason is the role mapping that is included. An admin currently
+easily can change the user settings in the UI and will be wondering
+why his changes get reset to the original settings. Therefore, Uyuni
+UI should mark external users either with special icon or color etc
+and the page of such user should have disabled "Update" button and an
+explanation for such user is managed elsewhere externally.
+
 # Drawbacks
 [drawbacks]: #drawbacks
 
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -18,9 +18,8 @@ tool should transparently synchronise users from certain groups and/or
 roles in LDAP and reflect that in Uyuni server, overriding local
 changes.
 
-Often LDAP configuration at infrastructures is unpredictable. This
-tool should bring certain boundaries and over up most common LDAP
-setups for users and their roles management.
+Due to the nature of LDAP flexibility, the synchronisation tool should
+be flexible enough to reflect a range of setups.
 
 # Detailed design
 [design]: #detailed-design
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -100,9 +100,16 @@ treated as "Uyuni local users" and be ignored therefore.
 
 ## Clashed UIDs
 
+Any UID that exists in the Uyuni server but has different
+authentication method than PAM or is mentioned as "frozen users"
+should be marked as conflicting and therefore ignored.
+
 Conflicting UIDs should be not processed, but reported to the log file
 as erroneous. However this should not prevent to process the rest
 of the existing users.
+
+However, any UID that has PAM authentication mode configured or is not
+mentioned as "frozen user" should be managed by LDAP.
 
 ## Usage
 
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -206,13 +206,26 @@ changes to the existing code might be very large.
 It could be possible to reuse existing Spacewalk runner in Salt,
 wrapping it into a state module and implement all that as a Salt
 Formula. While such approach would give more flexibility to the user,
-it will also have a number of drawbacks, such as additional formula
-debugging difficulties overhead; user-introduced bugs that might sync
-LDAP wrong with severe consequences; fragility of the very solution:
-it is still an editable Formula.
+it will also have a number of drawbacks, such as:
 
-Salt, however, may call `uyuni-ldap-sync` and/or setup configuration
-for it (PAM, the `uyuni-ldap-sync` configuration etc).
+- Additional formula debugging difficulties overhead
+
+- User-introduced bugs that might sync LDAP wrong with severe
+  consequences 
+
+- Fragility of the very solution, since it is still an editable
+  Formula
+
+- No daemon mode. Every time LDAP database update, changes supposed to
+  be just propagated transparently. For that, the job in Salt should
+  be scheduled, and thus calling Salt locally would be too much of overhead
+
+Salt, however, could periodically call `uyuni-ldap-sync` and/or setup
+configuration for it (PAM, the `uyuni-ldap-sync` configuration etc).
+
+With daemon mode implemented, Salt supposed to only setup the
+`uyuni-ldap-sync` once and delegate the rest to the
+`uyuni-ldap-sync`.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/0000-ldap-sync-tool.md
+++ b/0000-ldap-sync-tool.md
@@ -1,0 +1,132 @@
+- Feature Name: `mgr-ldapsync`: LDAP sync tool
+- Start Date: 2019, Oct 28
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The `mgr-ldapsync` is a tool that performs synchronisation of the
+users between Uyuni server and LDAP.
+
+# Motivation
+[motivation]: #motivation
+
+Uyuni server can authenticate users from LDAP via PAM. For this, users
+needs to be added to Uyuni with the flag "PAM authentication" and thus
+their credentials will be verified via LDAP directory. Synchronisation
+tool should transparently synchronise users from certain groups and/or
+roles in LDAP and reflect that in Uyuni server, overriding local
+changes.
+
+Often LDAP configuration at infrastructures is unpredictable. This
+tool should bring certain boundaries and over up most common LDAP
+setups for users and their roles management.
+
+# Detailed design
+[design]: #detailed-design
+
+The tool should work as an aggregator, collecting users from LDAP and
+Uyuni server synchronising them.
+
+## Functionality Aspects
+
+The following functions should be implemented:
+
+- Distinguish between local users in Uyuni and LDAP users.
+- Prevent UID clashes between the local users and LDAP users.
+- Prevent accidental modification or removal of the local user.
+- Flexible support various LDAP schemas, such as eDirectory, POSIX,
+  Active Directory etc.
+- Role management.
+- Dry run mode (preview only).
+
+## Configuration
+
+Configuration should be default in a separate file
+`/etc/rhn/ldapsync.conf`. Alternatively, user should be able to
+specify own path, in case the default one is not created or used.
+
+The main principle of the tool is to map specific roles and or groups
+in LDAP to the roles in the Uyuni server via the configuration. Adding
+or removing LDAP users from these groups or roles should reflect in
+Uyuni without additional manual changes in the Uyuni server.
+
+Configuration should be just a YAML file and the mapping should
+implement a relation between LDAP group to a set of roles from
+Uyuni. So each user in LDAP that is a member of such group would
+inherit these roles. For example:
+
+```yaml
+directory:
+  groups:
+    cn=sysop,ou=groups,dc=example,dc=com:
+      - activation_key_admin
+      - channel_admin
+      - config_admin
+```
+
+The configuration above would allow creating `groupOfNames` class in
+LDAP and add users via `member` attribute. As long as DN of that
+member is valid LDAP object and is not clashing with the local user in
+the Uyuni server, it should be synchronised. Similarly roles LDAP
+objects are configured as same as `groups` but via `roles`.
+
+## Frozen Users
+
+Frozen user is a local user in Uyuni, which has verified `org_admin`
+privileges (all) and is mentioned as "frozen" in the
+configuration. Such user is usually used as an emergency user, in case
+LDAP fails to respond and/or authenticate anyone. LDAP synchronisation
+should not proceed unless there is at least one "frozen" user.
+
+Frozen users configuration option can be also used to explicitly lock
+certain UIDs, even if they are listed in LDAP. In this case "frozen"
+UID should be entirely ignored by the `mgr-ldapsync` tool.
+
+Configuration:
+
+```yaml
+directory:
+  frozen:
+    - administrator  # This is the UID
+```
+
+## LDAP vs Local Users
+
+The `mgr-ldapsync` should distinguish between LDAP users and Uyuni
+users, and operate adding, modifying or removing them only in LDAP
+context. As long as the user is not specified in LDAP, it should be
+treated as "Uyuni local users" and be ignored therefore.
+
+## Clashed UIDs
+
+Conflicting UIDs should be not processed, but reported to the log file
+as erroneous. However this should not prevent to process the rest
+of the existing users.
+
+## Usage
+
+Typical use of the `mgr-ldapsync` should be straightforward by
+periodically run in cron. In console mode it should also provide a
+possibility of "dry-run" where only planning of actions to be taken
+will be shown.
+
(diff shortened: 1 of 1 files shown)


0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,139 @@
...
+[summary]: #summary
+
+The `mgr-ldapsync` is a tool that performs synchronisation of the
+users between Uyuni server and LDAP.
+
+# Motivation
+[motivation]: #motivation
+
+Uyuni server can authenticate users from LDAP via PAM. For this, users
+needs to be added to Uyuni with the flag "PAM authentication" and thus
```
paususe: Typo: "need"

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,139 @@
+- Feature Name: `mgr-ldapsync`: LDAP sync tool
+- Start Date: 2019, Oct 28
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The `mgr-ldapsync` is a tool that performs synchronisation of the
```
paususe: uni-directional or bi-directional synchronization?

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,139 @@
...
+[motivation]: #motivation
+
+Uyuni server can authenticate users from LDAP via PAM. For this, users
+needs to be added to Uyuni with the flag "PAM authentication" and thus
+their credentials will be verified via LDAP directory. Synchronisation
+tool should transparently synchronise users from certain groups and/or
+roles in LDAP and reflect that in Uyuni server, overriding local
+changes.
+
+Often LDAP configuration at infrastructures is unpredictable. This
```
paususe: What kind of "unpredictability" are we talking about?

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,139 @@
...
+- Prevent accidental modification or removal of the local user.
+- Flexible support various LDAP schemas, such as eDirectory, POSIX,
+  Active Directory etc.
+- Role management.
+- Dry run mode (preview only).
+
+## Configuration
+
+Configuration should be default in a separate file
+`/etc/rhn/ldapsync.conf`. Alternatively, user should be able to
```
paususe: Why, oh why, we keep using /etc/rhn and not something like /etc/uyuni ? :-(

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,139 @@
...
+
+N/A
+
+# Alternatives
+[alternatives]: #alternatives
+
+Alternative design would be to extend user management in the very
+Uyuni server, making it LDAP-aware natively. However, the amount of
+changes to the existing code might be very large.
+
```
paususe: Another alternative would be to extend the current sw-ldap-user-sync tool

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,139 @@
...
+the Uyuni server, it should be synchronised. Similarly roles LDAP
+objects are configured as same as `groups` but via `roles`.
+
+## Frozen Users
+
+Frozen user is a local user in Uyuni, which has verified `org_admin`
+privileges (all) and is mentioned as "frozen" in the
+configuration. Such user is usually used as an emergency user, in case
+LDAP fails to respond and/or authenticate anyone. LDAP synchronisation
+should not proceed unless there is at least one "frozen" user.
```
paususe: Does this mean if Uyuni Server loses connection to e. g. the Active Directory Domain Controller, then all the LDAP users (but the frozen ones) will be removed ? In the first try? Not even after some retries, timeouts or honoring AD credential cache time?

Looks dangerous to me.

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,139 @@
...
+      - activation_key_admin
+      - channel_admin
+      - config_admin
+```
+
+The configuration above would allow creating `groupOfNames` class in
+LDAP and add users via `member` attribute. As long as DN of that
+member is valid LDAP object and is not clashing with the local user in
+the Uyuni server, it should be synchronised. Similarly roles LDAP
+objects are configured as same as `groups` but via `roles`.
```
paususe: It's not clear to me how this tool works. Does it create local Uyuni Server users for the LDAP imported users? Does it create organizations, groups, etc too? Or it just synchronizes user data for existing Uyuni Server users?

0000-ldap-sync-tool.md
mcalmer: Well, everybody knows this directory. I would not change it as long as we change it for all the config files. This is of cause a bigger task which could break a lot of things.

We would need to dedicate some time to such a change and we need to do it on a new major version. It would also need a migration old-path to new path.

0000-ldap-sync-tool.md
mcalmer: Loud thinking:

... configures and manages users in Uyuni from a directory service.

0000-ldap-sync-tool.md
paususe: > Well, everybody knows this directory. I would not change it as long as we change it for all the config files. This is of cause a bigger task which could break a lot of things.
> 
> We would need to dedicate some time to such a change and we need to do it on a new major version. It would also need a migration old-path to new path.

I know, I know. It was a rhetorical question :-(

0000-ldap-sync-tool.md
mcalmer: There exists multiple schemas to configure users and groups in directory services.
Every Directory Service use there own standards.
And very often admins create additional schemas on top or use there own definitions.

Not sure if "unpredictable" is the right word for it.

0000-ldap-sync-tool.md
isbm: @paususe because:
1. It has historical reasons
2. Everyone knows it and very likely hard-coded it in their own scripts (we can symlink it though for backward compatibilities)
3. It is already there, so why I would add another directory? If we migrate it one day, then we migrate it all together

0000-ldap-sync-tool.md
isbm: This is exactly the opposite: in case one adds to the LDAP a DN with the corresponding `cn` as same user as in Uyuni, say `cn=admin` and then syncs it and then removes that member from the DN group keeping the DN in the users class, then the Uyuni user will be removed permanently locking Uyuni. :wink:

This is exactly to make sure this never happens.

0000-ldap-sync-tool.md
isbm: It creates users in Uyuni, based on LDAP layout and assignes roles accordingly. All you change in LDAP reflects in Uyuni. Local Uyuni users (those aren't in LDAP and has no PAM auth marked) are ignored and should be managed locally using Uyuni UI.

0000-ldap-sync-tool.md
isbm: @mcalmer but those are still DN paths to those groups and users and those are still POSIX or Active Directory (lately almost compatible with POSIX) schemas.

isbm: What word you would use instead?

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Frozen Users
```
mcalmer: I think we do not need to define this at all. A local user has nothing in common with a ldap managed user.
We can document this as best practice but it should be obvious that you cannot login when your directory service is down and you have PAM auth configured.
On the other hand: when your directoy service is down you really have other problems as you may not login to your workstation.

0000-ldap-sync-tool.md
mcalmer: I think we can detect a "connection refused" or "no route to host" error and in such cases we change nothing.
Only when we can connect to the LDAP server and a user configuration has changed or a user was removed we change things in Uyuni

0000-ldap-sync-tool.md
isbm: And by "extend" that would mean "rewrite from scratch", because existing tool is written in Windows `.bat` way without a single function inside.

0000-ldap-sync-tool.md
isbm: Local user still might have UID which is the same as in LDAP. Also you may want to freeze LDAP user as well for some reasons. Think of it as a guard to protect clashing UIDs or someone mistakenly adds a user in a wrong directory.

0000-ldap-sync-tool.md
isbm: @mcalmer copypaste that to summary section, thanks! :wink:

0000-ldap-sync-tool.md
isbm: pushed

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,139 @@
...
+inherit these roles. For example:
+
+```yaml
+directory:
+  groups:
+    cn=sysop,ou=groups,dc=example,dc=com:
+      - activation_key_admin
+      - channel_admin
+      - config_admin
+```
```
mcalmer: I think this example is incomplete. You say below, that the class is `groupOfNames` and there should be a `member` attribute which specify the user.

Other schemas uses other classes and other attributes. I think the config need to reflect this and the tool needs to work with these different schemas.

Maybe complete examples of the 2 or 3 most used schemas would be a good idea to have better overview what we need to implement. 

- AD
- POSIX
- SUSE?

isbm: > Other schemas uses other classes and other attributes. 

Yes, AD (which is not POSIX) is using `group` instead. That is simply included in the search filter making literally no difference by default, but I can still mention that if you want.

I don't know anything about SUSE schema (it should be still POSIX), since I never found a documentation of it, but only followed standards of Active Directory and POSIX. If you could provide examples of it, then it would be very helpful.

0000-ldap-sync-tool.md
paususe: > On the other hand: when your directoy service is down you really have other problems as you may not login to your workstation.

No credential caché on the workstation side? That's a given in Active Directory and Windows workstations.

0000-ldap-sync-tool.md
thallma: "As there are several ways and approaches to configure users and groups in directory services, this tool aims to work with most of them with a simple configuration setup." ?

thallma:
Considering that you have with this tool a new way of creating/modifying users beside the UI and API in an automated way, I think we should consider creating a flag in User Management that gives admins the clear indication that the user they are looking at has been touched by this tool. Main reason is the role mapping that is included. Now an admin easily can change the user in the UI and wonders why his changes get wiped away later by the tool. Having a distinction between manual and automated touching of users could prevent additional support tickets, by giving a hint, that there has been something automated and the admin should resolv this first.

rjmateus:
This tool will connect directly to the database or use the XMLRPC-API? If it connects directly to the database why was that the chosen design?

isbm:
> This tool will connect directly to the database or use the XMLRPC-API? If it connects directly to the database why was that the chosen design?

It connects over API.

isbm:
@thallma That's a very valid point and this certainly needs to be enhanced in Uyuni UX, adding something like a marker attribute for the Uyuni user. But I doubt this is directly related to this RFC, since there no such API. Sure, as long as appropriate API call appears, we can always enhance it. And such enhancement shouldn't really take much effort. It is basically a boolean field in the DB and a disabled "Update" button with the note "This user is managed externally".

0000-ldap-sync-tool.md
isbm: Well, while it is a valid point, I still believe this is beyond of the scope of this RFC and the tool itself. The tool is supposed to do only one job: preview the status, telling who is new and who is old and who is removed and then sync it, making sure no kittens were harmed. Done. How Uyuni is dealing _afterwards_ with the faulty/flaky LDAP through the PAM — is something else. Can we agree here?

0000-ldap-sync-tool.md
isbm: @thallma this is a poor explanation. :wink: What I wanted to say here is that there are few standard ways of managing groups/roles in LDAP. So this tool supposed to handle them all, having good 80% covered. 20% of those extraordinary setups I leave behind, without saying "No, never". But in those cases we would need a clear examples why current tool cannot work in their setup to find out what we can further help them as well.

Does it make more sense now?

thallma:
> @thallma That's a very valid point and this certainly needs to be enhanced in Uyuni UX, adding something like a marker attribute for the Uyuni user. But I doubt this is directly related to this RFC, since there no such API. Sure, as long as appropriate API call appears, we can always update this tool in no time. And such Uyuni enhancement shouldn't really take much effort. It is basically a boolean field in the DB and a disabled "Update" button with the note "This user is managed externally".

How to proceed then, to not get this improvement idea lost?

isbm:
> How to proceed then, to not get this improvement idea lost?

OK, you've got me. I've [added that note](https://github.com/uyuni-project/uyuni-rfc/pull/23/commits/8af4e82419be0504dc25d16208110728027cf170) to the RFC. Hopefully, it clarifies everything well.

0000-ldap-sync-tool.md
isbm: @paususe I've updated this.

0000-ldap-sync-tool.md
isbm: @mcalmer I've added [Attribute Remapping section](https://github.com/uyuni-project/uyuni-rfc/pull/23/commits/bf10f44b5ab179ffb51846d5a28c62ea8c30aa1f) specifically for this. That would allow support custom schemas.

isbm:
I've discussed with @mcalmer custom "crazy" schemas possibility and added another section to handle them as well. So @paususe @thallma please take another look at this RFC.

We are close to finalise it! :+1:

0000-ldap-sync-tool.md
isbm: @mcalmer  [New commit ID](https://github.com/uyuni-project/uyuni-rfc/pull/23/commits/37490a2cda3f564169b81b236d3ce7bce4e9d0e4) for the [Attribute Remapping](https://github.com/uyuni-project/uyuni-rfc/pull/23/commits/37490a2cda3f564169b81b236d3ce7bce4e9d0e4) section (force-pushed)...

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,182 @@
+- Feature Name: `mgr-ldapsync`: LDAP sync tool
+- Start Date: 2019, Oct 28
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The `mgr-ldapsync` is a tool that configures and manages users in Uyuni
```
moio: Can you please add a link to the current PoC code somewhere?

moio: Please also reference the software this RFC is intended to replace, which is:

https://github.com/uyuni-project/uyuni/blob/master/utils/sw-ldap-user-sync

isbm: Added.

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,233 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+## Native Implementation
+
+Alternative design would be to extend user management in the very
+Uyuni server, making it LDAP-aware natively. However, the amount of
+changes to the existing code might be very large.
+
+## Salt-based Implementation
```
moio: Does this alternative have any overlap with https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00061-hub-users.md ?

isbm: Frankly, I was considering about doing it with Salt at the beginning. I was also considering having this LDAP sync as an Ansible style module one can just call via Salt Ansible Gate (btw, still possible to add that interface).  The https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00061-hub-users.md is that alternative. 

The difference between current RFC is that it is a daemon performing on its own, transparently reflecting changes in LDAP, without extra intervention and manually spawn Salt every time one wants to update LDAP data. Putting such Salt job to a cron would be an overkill and would introduce other issues (two minions issue, jobs racing etc). Also, Salt can just call this tool or we can have "Ansible module mode", so it will return output in Ansible-compatible JSON protocol and Salt can leverage that via earlier mentioned Ansible Gate subsystem.

I personally think that "Salt calls a tool Foo" is a much better aproach than "Salt implements Foo in Formula on its own" from reliability, testing and debug-ability perspectives.

moio: > Putting such Salt job to a cron looks like an overkill and might even introduce additional issues we could just avoid by design.

What issues can you anticipate?

> firing Salt [...with...] periodical crontab [...] is probably not the best idea [...]. This makes little to none sense to me [...]

I hear you when you mention that, from a software design perspective, a standalone daemon could look better than `cron` + `salt`.

Do you see additional reasons for this idea to be superior, on top of the simpler design?

> Salt [...] can just call this tool from a runner [...] I personally think that "Salt calls a tool Foo" is a much better aproach than "Salt implements Foo in Formula on its own"

OK, but that approach is not really proposed in this RFC, so while I could agree on the general idea, it is not really being discussed here.

> Last and not least, Salt [`pillar_ldap`](https://docs.saltstack.com/en/latest/ref/pillar/all/salt.pillar.pillar_ldap.html#configuration) does not support attribute and objects remapping. 

So, IIUC, with RFC 61 users are supposed to specify the mapping from LDAP attributes to Uyuni configuration (Users/Groups/permissions and so on) via Salt templates (Jinja, Python, etc.). With this RFC, the tool would have to include additional configuration options (worst case: a complete "mini-language") to specify that mapping - complexity would depend on how complicated the mapping needs to be. Would that be accurate?

Thanks

isbm: > Do you see additional reasons for this idea to be superior, on top of the simpler design?

Thank you for the compliment that my design is simpler!

Yes, sure:

- Easier to debug than formulas. App  has full-blown logging facility built-in through the code. Salt Formulas, however, one would need to [modify Formula adding print-style logging](https://docs.saltstack.com/en/latest/topics/jinja/index.html#debugging) in order to narrow down the possible problem. Partially compiled state might only add to the  confusion and could increase debugging efforts.
- Possibility to clearly outline what is supported and what is not, keeping users from randomly modifying things. Feature request for missing parts.
- Keeps it _configurable_ at configuration files vs  _implementable_ in Jinja formulas. Salt is for configuration, daemon is for delivery. My approach forces you to keep things simple and segregated: let Salt deploy the configuration and run the daemon. It would make more sense from Salt perspective to do this:

```yaml
acme:
  ldapsync.running
    uyuni:
      user: pillar ......
    directory:
      cn: .......
```

> So, IIUC, with RFC 61 users are supposed to specify the mapping from LDAP attributes to Uyuni configuration (Users/Groups/permissions and so on) via Salt templates (Jinja, Python, etc.). With this RFC, the tool would have to include additional configuration options (worst case: a complete "mini-language") to specify that mapping - complexity would depend on how complicated the mapping needs to be. Would that be accurate?

Yes (reimplement sync in pure Jinja)  and no (add mini language for remap):

1. I could hardly call that "configuration". I would rather call it _full-blown implementation_. Implementation, one is going to debug. Debug at the Formula level. Worse than that, due to deviation from LDAP standard schema (or existing of POSIX and AD at first place), there consequently will be at least few if not more of those sets of Jinja templates with some slight subtle changes (read: more reported bugs). Bugs, that are only fixable to _that particular custom case_ and not applicable elsewhere. So this brings also whole new layer of versioning difficulty overhead. Therefore ideally one would like to rather implement LDAP sync itself in a separate Salt module, operating only data structures from Salt Formulas. This would allow keeping Jinja Formulas separated, taking care only of remappings/etc (again, this additional overhead only due to `pillar_ldap` limitations) and thus the Formula would format some commonly accepted structure that the module we are talking about would receive it then sync with the Uyuni. But such setup I find fragile and complicated.

2. No mini language is needed for remapping. Remapping is [already implemented as per RFC](https://github.com/isbm/uyuni-ldap-sync/commit/4123b5de2783ed9016544089bb35f47d534cda2b). You need to just write CN, proper LDAP filters and manually map custom objects to POSIX or AD in the config. This seems also _often_ needed, since admins tend to often deviate from the standard LDAP schemas writing their own stuff. Mini language will be actually required in Formulas: having additional couple of loops to figure out what to remap with before syncing.

3. My approach is basically alongside how [`slapd.conf`](https://linux.die.net/man/5/slapd.conf) does and the feature-set I am trying to reach how [Novell Filr](https://www.novell.com/documentation/novell-filr-1-2/filr-1-2_admin/data/ldap.html) does.

4. User `present`/`absent`/`updated` actions are still **insufficient** to properly sync users. Still somehow Salt Formula needs to know which user is local _prior to the sync_ and thus "untouchable". Same what Novell Filr does, BTW. So if there is a user with the ID `john` in LDAP with the role "Read Only" and the same _local user UID_ is in Uyuni server as `org_admin`, then you are asking for a trouble. You also cannot wipe out everyone from Uyuni, because you need to login to XML-RPC using local user that has `org_admin` perms, so at least one user needs to be reserved, which I foresee a hard-coded UID in the Formulas.

Hopefully this helps.

moio:
Personally I still see substantial overlap with RFC 61, which is already approved, and that makes me wonder if it is worth it to pursue both approaches, especially simultaneously, from the perspectives of risk, development resources and maintenance resources.

I lack the authority to answering those questions, and I will leave them to SUSE Manager Team management to answer, unless community members volunteer to bring this forward.

isbm:
@moio this is done project, passing R&D. PoC prototyping and initial implementation phases. Left refactoring and testing, while your Hub, on the other hand, is on hold right now. So I think anything what is already working and raised interest from other departments is worth finishing.

Having it in community-supported tools (like `spacecmd`) won't do any impact on resources.

0000-ldap-sync-tool.md
```diff
@@ -0,0 +1,233 @@
...
+configuration for it (PAM, the `uyuni-ldap-sync` configuration etc).
+
+With daemon mode implemented, Salt supposed to only setup the
+`uyuni-ldap-sync` once and delegate the rest to the
+`uyuni-ldap-sync`.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+N/A
```
mbologna: Context: SSO via SAML.
Can we use the `uyuni-ldap-sync` tool create an on-the-fly user that is authenticated in the SSO but not existing in the Uyuni user database and let him/her login?

isbm: Uhm, but how would you imagine that? The SAML part is a completely different approach.

The current sync does exactly how it is designed for and how it is called: _synchronising_ LDAP data. Same applies to moio's Salt-based idea, except it is just slower and much more fragile. Basically, syncing works by just duplicating everything into local DB of the Uyuni from the LDAP, so the *standard* Uyuni's *built-in* auth routines will apply on request and issue a session.

Now, if you want single sign on via SAML (or oAuth), you still want to have an identity provider somewhere that will sign the responses for you, which you will process directly in Tomcat. That said, the user is authenticated elsewhere (at the provider) and provider sends your app (the Uyuni on Tomcat) a signed message for "yes, let him in" or "no, kick him away", as there is no longer any user database available at the Uyuni side per se (hence why it is called SSO, not redistributed DB of users).

Hope that clarifies.

mbologna: > Now, if you want single sign on via SAML (or oAuth), you still want to have an identity provider somewhere that will sign the responses for you, which you will process directly in Tomcat. That said, the user is authenticated elsewhere (at the provider) and provider sends your app (the Uyuni on Tomcat) a signed message for "yes, let him in" or "no, kick him away", as there is no longer any user database available at the Uyuni side per se (hence why it is called SSO, not redistributed DB of users).

Yes, this is what I was referring to: the SAML gets back with "authentication for $username successful". 
The question is: do you think we can reuse part of the solution you devised to create an on-the-fly user in the Uyuni/SUSE Manager database (that we mandatory need - think of user roles) upon successful login message from the SAML entity (IdP)?

isbm: As per conversation in chat, seems I was misled by the original question formulation, so we lost common ground:

> Can we use the uyuni-ldap-sync tool create an on-the-fly user that is authenticated in the SSO **but not existing in the Uyuni user database** and let him/her login?

The correct question would be:

> Can we use `uyuni-ldap-sync` tool to pre-sync users in the local database so they can be authorised after being authenticated with SAML?

In that case, yes.

isbm:
The tool resides in https://github.com/isbm/uyuni-ldap-sync as community-supported.

paususe:
Thanks!

PR #2, opened 2019-02-21
Add 'kubernetes-integration'

cbbayburt:
Add the 'Kubernetes integration' RFC files.

PR #24, opened 2019-11-04
Hub XMLRPC API RFC - alternatives and clarifications

moio:
This PR is to address two comments on the first text that was merged:

 - describing alternatives to writing from scratch
 - clarify how the retry mechanism can be configured
 - defer the programming language choice

### DIFF
--- a/accepted/00062-hub-xmlrpc-api.md
+++ b/accepted/00062-hub-xmlrpc-api.md
@@ -129,6 +129,8 @@ Alternatives that were considered come from:
 | apicast              | 56k    | Lua              |                                                   |
 | expressgateway       | 33k    | Javascript       |                                                   |
 
+Note: Lua projects are actually Nginx plugins.
+
 ## Features
 * Authorization and authentication
   * Alternative idea 1: a configuration file on the Hub contains a list of credentials and `serverIds`. Hub would automatically maintain a pool of Server `sessionKey`s from those credentials
--- a/accepted/00062-hub-xmlrpc-api.md
+++ b/accepted/00062-hub-xmlrpc-api.md
@@ -33,7 +33,9 @@ Content management APIs will work as of today. Inter-Server Synchronization (ISS
 
 ### New service and endpoint
 A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
-  * Implementation will be in Python based on top of Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library
+  * Implementation will either be (choice delegated to the implementation phase):
+    * in Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library, in order for the resulting program to use exclusively async I/O
+    * in Go, using goroutines to attain the same result
   * all I/O should be handled asynchronously via Tornado. Calls to several Servers should happen in parallel, with a maximum timeout value for unreachable/not responding Servers
   * there will be no backing database. All new methods will delegate to existing XMLRPC APIs (either the Hub's or individual Servers')
     - some kind of storage might be needed for performance/caching purposes, that is left as an implementation detail
--- a/accepted/00062-hub-xmlrpc-api.md
+++ b/accepted/00062-hub-xmlrpc-api.md
@@ -33,11 +33,11 @@ Content management APIs will work as of today. Inter-Server Synchronization (ISS
 
 ### New service and endpoint
 A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
-  * Implementation will be based on asynchronous I/O. The exact technology choice is delegated to the implementation phase and could be:
+  * Implementation will be based on asynchronous I/O. Calls to several Servers should happen concurrently  
+  * The exact technology choice is delegated to the implementation phase and could be:
     * typed Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/). Rationale: learning Tornado, which is also used in core Salt
     * typed Python, based on the native [asyncio library](https://docs.python.org/3.6/library/asyncio.html) (new to 3.4, similar in intent to Tornado). Rationale: learning latest Python technologies
     * in Go, using goroutines. Rationale: learning Go
-  * all I/O should be handled asynchronously via Tornado. Calls to several Servers should happen in parallel, with a maximum timeout value for unreachable/not responding Servers
   * there will be no backing database. All new methods will delegate to existing XMLRPC APIs (either the Hub's or individual Servers')
     - some kind of storage might be needed for performance/caching purposes, that is left as an implementation detail
 
--- a/accepted/00062-hub-xmlrpc-api.md
+++ b/accepted/00062-hub-xmlrpc-api.md
@@ -33,9 +33,10 @@ Content management APIs will work as of today. Inter-Server Synchronization (ISS
 
 ### New service and endpoint
 A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
-  * Implementation will either be (choice delegated to the implementation phase):
-    * in Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library, in order for the resulting program to use exclusively async I/O
-    * in Go, using goroutines to attain the same result
+  * Implementation will be based on asynchronous I/O. The exact technology choice is delegated to the implementation phase and could be:
+    * typed Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/). Rationale: learning Tornado, which is also used in core Salt
+    * typed Python, based on the native [asyncio library](https://docs.python.org/3.6/library/asyncio.html) (new to 3.4, similar in intent to Tornado). Rationale: learning latest Python technologies
+    * in Go, using goroutines. Rationale: learning Go
   * all I/O should be handled asynchronously via Tornado. Calls to several Servers should happen in parallel, with a maximum timeout value for unreachable/not responding Servers
   * there will be no backing database. All new methods will delegate to existing XMLRPC APIs (either the Hub's or individual Servers')
     - some kind of storage might be needed for performance/caching purposes, that is left as an implementation detail
--- a/accepted/00062-hub-xmlrpc-api.md
+++ b/accepted/00062-hub-xmlrpc-api.md
@@ -35,7 +35,7 @@ Content management APIs will work as of today. Inter-Server Synchronization (ISS
 A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
   * Implementation will be based on asynchronous I/O. Calls to several Servers should happen concurrently  
   * The exact technology choice is delegated to the implementation phase and could be:
-    * typed Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/). Rationale: learning Tornado, which is also currently used in core Salt (rumored to be potentially replaced in future)
+    * typed Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/). Rationale: learning Tornado, which is also currently used in core Salt (but rumored to be potentially replaced with [trio](https://github.com/python-trio/trio) in future)
     * typed Python, based on the native [asyncio library](https://docs.python.org/3.6/library/asyncio.html) (new to 3.4, similar in intent to Tornado). Rationale: learning latest Python technologies
     * in Go, using goroutines. Rationale: learning Go
   * there will be no backing database. All new methods will delegate to existing XMLRPC APIs (either the Hub's or individual Servers')
--- a/accepted/00062-hub-xmlrpc-api.md
+++ b/accepted/00062-hub-xmlrpc-api.md
@@ -96,6 +96,37 @@ All code would be new in a new component, so no change to existing components is
 
 # Alternatives
 [alternatives]: #alternatives
+## Build on an existing API gateway
+
+Several open source project exist that offer API gateway functionality on any kind of network API (often in practice limited to http protocols). Those projects typically offer a range of functionalities such as configurable retry patterns, additional security checks, logging, monitoring, rate limiting, circuit breaking, caching and in some cases API aggregations and manipulation of requests and responses. In several cases, such engines provide plugin mechanisms or anyway lend themselves to implementation of ad-hoc modules that could fit in the requirements from this RFC. Summarizing:
+ - pros: potentially many features would be provided by such tools. Only automatic retrying is required at this point, but others might also become relevant in future
+ - cons: code bases are typically much bigger than the expected size of this project (1K - 2K LOC), and the type of request modifications required will still need they have to be implemented as plugins/extensions, which might lead to an approximately-similar coding effort
+
+Alternatives that were considered come from:
+ - GitHub projects with at least 1000 stars
+ - [the CNCF category](https://landscape.cncf.io/category=api-gateway&format=card-mode&grouping=category)
+ - a list provided as comment to the original version of this RFC
+ 
+| Name                 | LOC    | Main languages   | Notes                                             |
+|----------------------|--------|------------------|---------------------------------------------------|
+| gravitee             | 500k   | Java, TypeScript | Supports MongoDB, Redis, ElasticSearch...         |
+| tyk                  | 100k   | Go               | No request modification support, security only    |
+| wso2                 | 5.636M | Java             | 28 main git repos                                 |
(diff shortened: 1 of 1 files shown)


accepted/00062-hub-xmlrpc-api.md
```diff
@@ -33,7 +33,9 @@ Content management APIs will work as of today. Inter-Server Synchronization (ISS
 
 ### New service and endpoint
 A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
-  * Implementation will be in Python based on top of Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library
+  * Implementation will either be (choice delegated to the implementation phase):
+    * in Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library, in order for the resulting program to use exclusively async I/O
+    * in Go, using goroutines to attain the same result
   * all I/O should be handled asynchronously via Tornado. Calls to several Servers should happen in parallel, with a maximum timeout value for unreachable/not responding Servers
```
isbm: I guess, this needs to be adjusted because of the above. But nor Tornado neither `asyncio` are about _parallelism_. They are about _concurrency_. Hello, GIL.

paususe:
> This PR is to address two comments on the first text that was merged:
> 
>     * describing alternatives to writing from scratch
> 
>     * clarify how the retry mechanism can be configured
> 
>     * defer the programming language choice

Is there anything actionable? I can see added text but no decisions or changes from what was already merged.

Here's one proposal about the API gateway: collect a list of features offered by Gravitee (the richest one IIRC), Tyk, Goku, etc and ask our stakeholders which ones of those features will be important for the hub XMLRPC API (now, in 6 months, in 1 year, in 2 years). That should help making a decision for or against an API gateway.

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -33,7 +33,9 @@ Content management APIs will work as of today. Inter-Server Synchronization (ISS
 
 ### New service and endpoint
 A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
```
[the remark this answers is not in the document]
paususe: > Similarly, we don't need to _change_ the API signatures as per documentation and use multitude of auth keys from _each_ SUMA server or an avalanche of parameters in APIs, assuming there might be 500 SUMAs connected. We should have only one. And in case we would need to use a subset of SUMA servers, we could have something lile this:
> 
> ```python
> proxy.filter("ignore-me.acme.com")
> 
> ...
> 
> proxy.filter("ignore-me-too.acme.com", "and-me.acme.com")
> 
> ...
> 
> proxy.clearFilter()
> ```
> 
> All works as above, except `ignore-me.acme.com` would be... ignored completely (not even popping out in the return structure).
> 
> Does it make sense to you?

That looks to me like a different use case.

Case 1: hub administrator creates some keys for user A so that user A can act on servers 1, 3 and 5. Rationale for the apikeys: the end user should not have access to all of the servers. This is much like GitHub keys: key user and key creator are different entities, key creator does not trust key user to have full access to everything.

Case 2 (what you describe here) : hub administrator wants to perform some action on a subset of servers and limits himself to that subset. Key user may have full access but decides to drop (filter out) some of his privileges.

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -33,7 +33,9 @@ Content management APIs will work as of today. Inter-Server Synchronization (ISS
...
-  * Implementation will be in Python based on top of Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library
+  * Implementation will either be (choice delegated to the implementation phase):
+    * in Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library, in order for the resulting program to use exclusively async I/O
```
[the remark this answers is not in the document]
moio: Good point. The reasoning behind Tornado was to grow familiarity with it, as it's also in core Salt, so a good investment to learn. Adapting text.

accepted/00062-hub-xmlrpc-api.md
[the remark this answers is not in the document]
moio: Meta-note: aou are making a lot of different points here, not all related to this position in the text and some not related to this PR at all. Please divide better next time.

> essentially it is a multicast microservice

Yes.

> I would suggest that the caller method simply always returns a map string/object (host/result) 

The advantage of using ordered arrays as input parameters and ordered arrays as outputs, which is proposed here, is composability gets a bit easier. Often an API call will need input coming from output of other API calls, like this:

```python
def ids_from(struct_array):
  [e.id for e in struct_array]

systems_per_server = hub.system.listActiveSystems(key)
errata_per_server = hub.errata.findByCve(key, [cve] * n)
hub.system.scheduleApplyErrata(key, ids_from(systems_per_server), ids_from(errata_ids))
```

Using Server-indexed maps will make it more difficult to pass data out of a function and into another.

> proxy = xmlrpc.client.ServerProxy("https://hub.acme.com:12345")
> auth_key = proxy.login(USER, PASSWORD)
> result = proxy.call("system.listSystems", auth_key)

Point of this RFC is to make it easier for script developers to script the Hub, so I think the proposed syntax above is much clearer than a separate `proxy` object. Also not changing the text in this respect for the time being.

> That way anything can be called as-is on all servers simultaneously.

That is not different from what it is being proposed:

https://github.com/uyuni-project/uyuni-rfc/pull/24/files#diff-299b7fe4661063d97fab6412711ed861R36

> Exception here would be only login part, where it obviously wouldn't return all tokens to all servers! But instead it would return only one token from the Hub, and so then Hub would map that token to all the connected server tokens internally and pass per a target (keeping all that story about auto-reconnect/token-caching etc etc)

That does not sound different from what it is being proposed:

https://github.com/uyuni-project/uyuni-rfc/pull/24/files#diff-299b7fe4661063d97fab6412711ed861L54-L67


> proxy.filter("ignore-me.acme.com")
> proxy.filter("ignore-me-too.acme.com", "and-me.acme.com")
> proxy.clearFilter()
> All works as above, except ignore-me.acme.com would be... ignored completely (not even popping out in the return structure).
> Does it make sense to you?

Problem with this approach is that any call will get the same parameter values to all Servers, which might happen but it's not the most frequent case. In general, one wants to pass different parameter values to different Servers. Consider the example above: one wants to schedule an errata update on all clients - but errata ids are not guaranteed to be the same across Servers. One needs to use the right errata ids on the right Servers.


So far I do not see reasons to change the current text, I am resolving this conversation.

moio:
> Is there anything actionable? I can see added text but no decisions or changes from what was already merged.

Well, that's what this PR is about - my proposal is indeed not to change what was proposed, as I believe it is not worth it to build upon a third-party tool at this time.

This is the right place to have a discussion before starting with the next implementation step, so I'll kick it off directly.

moio:
@joesusecom, @mbrookhuis: please take a look at those changes, particularly:

https://github.com/uyuni-project/uyuni-rfc/pull/24/files#diff-299b7fe4661063d97fab6412711ed861R103

In [another discussion](https://github.com/uyuni-project/uyuni-rfc/pull/19#discussion_r340376251) @paususe suggested we should evaluate existing API gateways to build upon instead of starting from scratch. Here I am proposing to keep those alternatives in mind for the future, but not as the first implementation step (note: implementation could start as early as next week).

Pros/cons balance could change if requirements change, in particular, if any of the following functionality is to be implemented as part of the first step:
 - caching
 - rate limiting
 - extra security layers
 - usage analytics
 - monitoring
 - filtering
 - ...

You can get an idea of what can be done from the various API gateway websites. My understanding is that none of this is really required at this time, so I'd go with the leanest choice for now.

Please speak up if you think the text should be changed, and feel free to pull in others in this discussion as needed!

accepted/00062-hub-xmlrpc-api.md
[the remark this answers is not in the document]
isbm: > Please divide better next time.

Yes, I just reused the current PR to see if I would need to go forward. Guilty as charged.

> errata ids are not guaranteed to be the same across Servers

Right. So then I see my proposal as an optional add-on for convenience in bulk calls, beside the main text. Let's keep your text for now as is.

isbm:
> Gravitee (the richest one IIRC), Tyk, Goku, etc 

Hmm... I tend to downvote this. For me it is still hard to see why would we need it in context of Hub. The API gateway is a provider of a single point of entry for a defined group of services. So is Hub. So far, so same. But Hub will unlikely be "drinking from the firehose" (to what API Gateways are used for) but have very little load. Second, Hub is sort of "controller/multicaster", generally speaking "a client to many servers", where it accepts a specific task and just multicasts that equally to the amount of Uyuni servers. Alas, it provides its own APIs.

But do we expect couple of thoudands of requests per a second to the Hub's APIs? Do we need analzysis of API usage, extra security layers, throttling, caching etc etc to just satisfy ONE (max few) user interfaces? That so far sounds to me quite orthodogonal to what Hub is, and also too big so it will consequently bring much more complicated maintenance.

Please correct me if I am wrong.

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -33,8 +33,11 @@ Content management APIs will work as of today. Inter-Server Synchronization (ISS
 
 ### New service and endpoint
 A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
-  * Implementation will be in Python based on top of Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library
-  * all I/O should be handled asynchronously via Tornado. Calls to several Servers should happen in parallel, with a maximum timeout value for unreachable/not responding Servers
+  * Implementation will be based on asynchronous I/O. Calls to several Servers should happen concurrently  
+  * The exact technology choice is delegated to the implementation phase and could be:
+    * typed Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/). Rationale: learning Tornado, which is also used in core Salt
```
isbm: Actually, there are quite strong rumors in Salt community that asyncore is much more appealing option than Tornado (and it technically is better anyway). It is just not *yet* possible due to the necessity of Py2.x support and no official decision has been made yet. But as long as the minimal Python version stays on 3.6, very much chances are that Tornado actually might be dropped entirely. Especially taking to the account all the current austere issues with it in the Salt code. According to that,  https://github.com/python-trio/trio (which I myself as well like much more) is one of the candidates to replace Tornado. Ideally, if asyncore finally gets rid of event loop concept and just uses groups, similar to Go.

Thus I doubt if the rationale is strong enough here.

moio: Will update the text soon, thanks for this bit of information

isbm:
Guys, I gave another round of thoughts on my previous comment. Things I've missed to mention:

- API Management (e.g. "allow to user X call only specific APIs")
- Reporting of API usage
- Compliances

I quite doubt API segregation is a special topic here (look at `spacecmd`: it wants admin mode and it is how everyone is using it in scripts). Seems that for the SUMA it was never been a problem to give all API usage to a privileged user and script it up. Because there is no point to have less usage access anyway: all it mostly does via APIs that it installs some patches or upgrades. So why would be that a problem to the same `spacecmd`, for example, call Hub now, getting the same uer IDs or list the same channels or apply errata, just on the global scale? But still, in case we need such access segregation, user roles should be sufficient (users getting to be sync'ed across all the instances).

As of API reporting, compliances and their management — this needs to be more clarified with the stakeholders is that at all required. Still, to my knowledge, I see no mentions of requiring even *in a quite distant future* something similar to CORS (Cross Origin Resource Sharing) or URL rewriting or offline operation communication. APIs won't do any payment-related operations either. :wink:  Thus I still see a very little to none need of something like Red Hat 3scale to be _as a Hub_ or be one of the functions of the Hub. So far,  for the next years, I don't see Hub more than just (roughly speaking) little "synchronizer" microservice that sits in front of a bunch of SUSE Manager instances.

So my proposal is to:
1. Pre-agree to drop API Gateway idea.
2. Listen to @joesusecom @mbrookhuis thoughts and requirements on Hub
3. Revert #1 if #2 reveals missing requirements we didn't heard of yet.

We need a very clear consensus here, as I see it.

joesusecom:
I tend to agree with @isbm and @moio. The API aggregator is an implementation detail of scaling Uyuni. Let's keep it simple and focus on the one goal of making a whole cluster of SUSE Manager servers feel like one from an API point of view.

Adding a whole API gateway implementation just to reach that goal seems to be overkill to me.

Customers can always add additional stacks on top of this if they have more complex needs in terms of security filtering, combining tools, etc.

That said, @paususe has raised a few very valid concerns about making the stack robust enough in case certain nodes don't come back in time. Just remember the early days of our Java Salt API integration, where a single non-responding minion could let the SUSE Manager UI freeze. :-)

Those cases will probably be rare, and we could just implement a "graceful failure" mode. But let's make sure things are easy to debug and well documented (e.g. in log messages) when anything goes wrong.

paususe:
> I tend to agree with @isbm and @moio. The API aggregator is an implementation detail of scaling Uyuni. Let's keep it simple and focus on the one goal of making a whole cluster of SUSE Manager servers feel like one from an API point of view.
> 
> Adding a whole API gateway implementation just to reach that goal seems to be overkill to me.
> 
> Customers can always add additional stacks on top of this if they have more complex needs in terms of security filtering, combining tools, etc.
> 
> That said, @paususe has raised a few very valid concerns about making the stack robust enough in case certain nodes don't come back in time. Just remember the early days of our Java Salt API integration, where a single non-responding minion could let the SUSE Manager UI freeze. :-)
> 
> Those cases will probably be rare, and we could just implement a "graceful failure" mode. But let's make sure things are easy to debug and well documented (e.g. in log messages) when anything goes wrong.

My main concern with the non-API gateway solution is I've seen this many times.

We start by saying "it makes no sense to use an API gateway because we will only need a very small number of features". And then we do it ourselves.

And it starts to grow.

And at some point, we will start not doing some feature requests because they will be too difficult with our non-API gateway solution (which in fact will be a tailor-made, small API gateway).

And then we need to throw it away and start from scratch with an actual market API gateway, and as it's new code, new bugs and problems appear.

:-(

mbrookhuis:
I have to agree with @isbm and @moio.

Adding new external software makes us again depended of what they are doing. We see it now already with salt and python. SALT will force us to go to a python version we don't have in our products available on the moment. And taking support in consideration,  how will this work for an added product? 

For me the current API works, it is quick and logs all relevant data I need to know if there is a problem. The number of problems is limited. @isbm is correct with saying that we need more control in the API who could do what. But this is an old discussion and is a problem in everything we do in SUSE Manager. 

Adding another product also rises the question of how long it will take before we have added it to the product? Guess we could hit many problems in the beginning. And there I see another problem. As consultant I see an increasing number of customers with 10,000+ systems. Currently I am involved at 2 customers of which one expect to grow to 30,000+ in 3 years. We all know that the current SUSE Manager will have problems with such installations and we need to find a solution. Them waiting for a solution and then running into "beginners" bugs, will not make them happy and could force them to look for other solutions.

accepted/00062-hub-xmlrpc-api.md
```diff
@@ -33,8 +33,11 @@ Content management APIs will work as of today. Inter-Server Synchronization (ISS
 
 ### New service and endpoint
 A new XMLRPC API endpoint will be created, implemented by a new service called the "XMLRPC Gateway API" (simply called "Gateway" from now on in this document). Technology-wise:
-  * Implementation will be in Python based on top of Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/) library
-  * all I/O should be handled asynchronously via Tornado. Calls to several Servers should happen in parallel, with a maximum timeout value for unreachable/not responding Servers
+  * Implementation will be based on asynchronous I/O. Calls to several Servers should happen concurrently  
+  * The exact technology choice is delegated to the implementation phase and could be:
+    * typed Python, based on Tornado and [tornado-xmlrpc](https://pypi.org/project/tornado-xmlrpc/). Rationale: learning Tornado, which is also currently used in core Salt (rumored to be potentially replaced in future)
```
isbm: I would suggest to put more full version of the rumour:
`(rumored to be potentially replaced with asyncio-based [trio](https://github.com/python-trio/trio) in future)`

isbm:
Just a note: a proposed YAML won't work and I suggest one improvement over it:

```yaml
acme_org:
  uyuni.user_present:
    - name: john
    - password: foobar
    - org: ACME corp
    - roles:
      - org_admin
```

The namespace won't work: we need to modify `state.py` for this. I suggest `user.present` as `user_present`.

Also, instead of hardcoding _all possible roles_ as parameters to the functions (state functions don't like `**kw` as a single parameter), the `roles` should be just a list and whatever is there, should be set.

accepted/00062-hub-xmlrpc-api.md
moio: Addressed

moio:
> Just a note: a proposed YAML won't work [...]

I agree with the substance but that refers to a different RFC/PR, I guess it was posted by mistake.


Given I addressed your remaining comment as you suggested I am proceeding to merge this PR.

Thanks everybody for the cooperation

PR #25, opened 2019-11-07
Integrate translate for react RFC with latest outcome

ncounter:
Integrate translate for react RFC with latest outcome.

https://github.com/SUSE/spacewalk/issues/9003

### DIFF
--- a/accepted/00016-translate-4-react.md
+++ b/accepted/00016-translate-4-react.md
@@ -68,6 +68,22 @@ Once that this file is downloaded by the browser, it is mapped in *javascript* a
 
 A complete demo is available in the folder */text/attachments/translate-4-react-demo/*.
 
+### Adding a translated language into the current structure
+
+In case of adding a new language, we would need to implement the following steps:
+
+**Old pages**:
+
+* Translate the file `StringResource_en_US.xml` for the desired language.
+
+**New React Pages**:
+In the new pages developed by React all the texts are already wrapped by the function `t()`.
+
+* Generate a JSON file for the wanted language `{"key": "translation"}`, where the key is the string in English. 90% of the keys could be automatically generated but in some places, they are being injected by the server, so these ones have to be added manually
+* Review wrong usages of the function `t()`: for instance t(`Items ${itemsVar}`) should be refactored to t("Items {0}", itemsVar) , there are some occurences of this misuse format
+* Review if there are places where wrapping texts with the function `t()` was forgotten. This could be achieved changing the function `t()` to: `function t() { return "translated" }` and review all pages.
+* Add warnings or javascript lint rules to avoid missing translations in the future
+
 
 # Drawbacks
 


paususe:
What's the implementation of t()?

Should we use react-gettext instead of Polyglot.js ? It generates standard POT files, which will be easier for translators to work with.

Do we need RFCs for translation of the docs and CLI tools?

mcalmer:
I think for them we do not need a rfc. Commandline tools have already translation process in place.
For the docs i hope the team has ideas how to translate them

ncounter:
> What's the implementation of t()?

It is described [few lines above](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00016-translate-4-react.md#current-proposal-bypass) and it is the [current implementation](https://github.com/uyuni-project/uyuni/blob/master/web/html/javascript/susemanager-translate.js#L6) alive in the codebase and it is used everywhere [lilke this](https://github.com/uyuni-project/uyuni/blob/master/web/html/src/manager/systems/bootstrap/bootstrap-minions.js#L159)

paususe:
> > What's the implementation of t()?
> 
> It is described [few lines above](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00016-translate-4-react.md#current-proposal-bypass) in the existing RFC and it is the [current implementation](https://github.com/uyuni-project/uyuni/blob/master/web/html/javascript/susemanager-translate.js#L6) alive in the codebase and it is used everywhere [lilke this](https://github.com/uyuni-project/uyuni/blob/master/web/html/src/manager/systems/bootstrap/bootstrap-minions.js#L159)

That implementation does not currenty translate anything, does it? 

What would be the implementation of a t() that actually translates strings, takes into account plurals, numbers, dates, etc? The JSON key:value format described [here](https://github.com/uyuni-project/uyuni-rfc/blob/00b232dbffc9e2278fb29dad33c33407324401ce/accepted/00016-translate-4-react.md#the-test) doesn't seem to be prepared for that.

How would we extract the strings POT, XLIFF or some other common format translators already know and have tools for? 

Should we consider an already-established format and external library instead to overcome the [known drawbacks](https://github.com/uyuni-project/uyuni-rfc/blob/00b232dbffc9e2278fb29dad33c33407324401ce/accepted/00016-translate-4-react.md#drawbacks) of the previous RFC? E. g. gettext via [gettext.js](https://github.com/guillaumepotier/gettext.js/) ?

moio:
> That implementation does not currenty translate anything, does it?

Correct. `t` is just a passthrough now, ready to be re-implemented when the need for translation comes.

> What would be the implementation of a t() that actually translates strings, takes into account plurals, numbers, dates, etc? The JSON key:value format described [here](https://github.com/uyuni-project/uyuni-rfc/blob/00b232dbffc9e2278fb29dad33c33407324401ce/accepted/00016-translate-4-react.md#the-test) doesn't seem to be prepared for that.
> 
> How would we extract the strings POT, XLIFF or some other common format translators already know and have tools for?
> 
> Should we consider an already-established format and external library instead to overcome the [known drawbacks](https://github.com/uyuni-project/uyuni-rfc/blob/00b232dbffc9e2278fb29dad33c33407324401ce/accepted/00016-translate-4-react.md#drawbacks) of the previous RFC? E. g. gettext via [gettext.js](https://github.com/guillaumepotier/gettext.js/) ?

These are all valid open questions and I would suggest to discuss them in the context of a new RFC to actually implement the translations.

My understanding is the intent of this PR was just to include research results in the original RFC text, to serve as a reference when that time comes.

ncounter:
> These are all valid open questions and I would suggest to discuss them in the context of a new RFC to actually implement the translations.

I totally agree. Once we will have to take the path to reach translations, we will start with a new small RFC about the proper implementation. This RFC was only about how to implement a bypass-true at the time we started with React writing new pages in the new stack but still keeping the possibility to add a translation mechanism in the future.

> My understanding is the intent of this PR was just to include research results in the original RFC text, to serve as a reference when that time comes.

Exactly, this RFC integration was just to update the current status with latest outcome. Nothing here is taking decision on how `t()` should be refactored with what in order to have the translation in place and working.


Because of the above discussion, I will merge this integration PR.
Thank you all for your comments.

PR #26, opened 2019-11-08
CaaSP nodes operation locking

mbologna:
This RFC describes how we are going to operation lock the CaaSP nodes registered to Uyuni/SUSE Manager.

- Formatted text: https://github.com/uyuni-project/uyuni-rfc/blob/886bfbd4a006e3ddfdfcbf71204442295b77a009/accepted/0000-caasp-nodes.md
- Text with inline comments: https://github.com/uyuni-project/uyuni-rfc/pull/26/files

EPIC reference: https://github.com/SUSE/spacewalk/issues/8307#issuecomment-550400131

TODO:
- [ ] (@mbologna & @Loquacity): document forbidden operations in doc-susemanager
- [x] (@mbologna) Change filename before merging
- [x] (@mbologna) Add PR link in the RFC text


accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,274 @@
...
+- What is the impact of not doing this? -->
+
+## Future work
+
+In the future, we can team up with cluster providers and ask to `skuba` and other [cluster provider managers](https://trello.com/c/2X01ypO4/10-cluster-awarenesss-workshop-suse-manager-caasp-ha-cap-ses#comment-5dc1bc4567c6031ae57bcb04) to implement all these forbidden actions that require special handling. In that case, Uyuni/SUSE Manager can remove all the specific features described in this RFC and offload the task to the cluster provider manager.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* Deciding which action locking mechanism to take: Salt minion blackout or Uyuni/SUSE Manager action locking
```
mateiw: I think both are needed. If only the actions are locked in the UI/API the user still has the option to use the Salt command line or the Remote Command UI to execute any of the forbidden actions.

mbologna: If we go with Salt minion blackout then we do not need Uyuni/SUSE Manager action locking (unless we want to forbid the forbidden actions when blackout is disabled).
With minion blackout, every action scheduled by Uyuni/SUSE Manager will fail - at Salt level the action will not be executed (even `cmd.run` if not whitelisted).

mateiw: > With minion blackout, every action scheduled by Uyuni/SUSE Manager will fail - at Salt level the action will not be executed (even `cmd.run` if not whitelisted).

That would do the trick for the first iteration but I think it's not very user friendly. IMHO users should see clearly that a minion is locked for certain operations.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,274 @@
...
+
+NOTE: It is outside of this RFC to discuss automatic registration methods of the CaaS Platform nodes.
+
+## Technical description of the problem
+
+### The forbidden actions
+
+[`skuba-update`](https://documentation.suse.com/suse-caasp/4/single-html/caasp-admin/#_base_os_updates) systemd timer will take care of [patching](https://github.com/SUSE/skuba/blob/master/skuba-update/skuba_update/skuba_update.py#L301-L305) all installed packages locally to every node of the cluster and optionally reboot the node if `zypper` signals that a reboot is required.
+
+Uyuni/SUSE Manager does not offer the possibility to reboot a registered CaaS Platform node.
```
moio: Is this because Power Management is still not available for minions, or am I missing something?

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,274 @@
...
+
+  * obscure corner cases
+  * will it impact performance?
+  * what other parts of the product will be affected?
+  * will the solution be hard to maintain in the future? -->
+
+The user can craft a Salt state that executes one of the forbidden operations and breaks the cluster. This possibility is outside the scope of the RFC.
+
+# Alternatives
+[alternatives]: #alternatives
```
moio: Could enhancing the minion blackout feature in Salt, so that individual states can be whitelisted, be another option?

accepted/0000-caasp-nodes.md
mbologna: It is a typo: "does not" -> "should not". Fixed

accepted/0000-caasp-nodes.md
mbologna: Yes, it might work indeed. Adding it to the alternatives.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+
+Any node of a CaaS Platform cluster can be registered to Uyuni/SUSE Manager.
+Registering a CaaS Platform node to Uyuni/SUSE Manager brings the following notable advantages:
+
+- The user can see the patch level status of the CaaS Platform nodes (at the packages level, not at the container level - the latter is called "[container staging](https://confluence.suse.com/display/~dmacvicar/Container+Staging+Breakout)" and it is outside of the scope of this RFC)
+- The user can perform configuration management operations using Salt from Uyuni/SUSE Manager
+- The user can assign a different set of channels (with channel staging and/or CLM filters) to different clusters and manually control from Uyuni/SUSE Manager which version of CaaS Platform gets automatically installed in its CaaSP clusters<sup>1</sup>.
+
+However, if the user does some specific actions on any node of the cluster via Uyuni/SUSE Manager, the cluster may result temporarily or permanently unusable or broken.
```
moio: I am confused a bit now: is this RFC exclusively about actions (in the Uyuni concept sense of the word) or are we thinking about both actions and Salt commands?

IOW, are we happy to block some buttons in the UI for now, or do we want to prevent also users calling Salt directly, or writing their own states, from breaking the cluster?

Note that the answer can very well be "we want to do both, but this RFC is just the first step so it only concentrates on the former for now"

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+
+In this RFC, we will use:
+
+* CaaS Platform node: any node that comprises the cluster, either a control plane node or a worker node
+* Bootstrap/registration in the Uyuni/SUSE Manager context: registering a node to Uyuni/SUSE Manager via [the documented methods](https://opensource.suse.com/doc-susemanager/suse-manager/client-configuration/registration-overview.html).
+At the time of writing, registering a CaaS Platform node to Uyuni/SUSE Manager has to be manually performed by the user with:
+  * Bootstrap script
+  * UI: Systems > Bootstrapping
+  * `bootstrap` XMLRPC call
+  * Not usual but still possible and hackier: mass bootstrap via salt-ssh and a user-generated roster file
```
moio: Will plain key acceptance never work? Why?

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+
+## Assumptions
+
+We assume that:
+* The user is registering the CaaS Platform nodes as Salt clients
+* The registration of every CaaS Platform node to Uyuni/SUSE Manager is already completed by the user using the aforementioned methods
+* The user has already created an activation key and associated it with the onboarding. The CaaS Platform nodes are registered correctly and with the proper CaaS Platform channel(s) assigned.
```
moio: Why is it required to have an activation key?

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+
+### The forbidden actions
+
+[`skuba-update`](https://documentation.suse.com/suse-caasp/4/single-html/caasp-admin/#_base_os_updates) systemd timer will take care of [patching](https://github.com/SUSE/skuba/blob/master/skuba-update/skuba_update/skuba_update.py#L301-L305) all installed packages locally to every node of the cluster and optionally reboot the node if `zypper` signals that a reboot is required.
+
+Uyuni/SUSE Manager should not offer the possibility to reboot a registered CaaS Platform node.
+
+In general, patch and package upgrade/removal are safe operations for a CaaS Platform cluster: on the next trigger of `skuba-update`, if there are packages that required a reboot, `skuba-update` will orchestrate the reboot.
+
+Package upgrade/removal and patch apply is not a problem if issued from Uyuni/SUSE Manager, unless the involved packages of the action are one of the following:
```
moio:
```suggestion
As an exception to the above rule, the following packages should never be upgraded or removed via Uyuni/SUSE Manager:
```

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+
+- `kubernetes-kubeadm`
+- `kubernetes-kubelet`
+- `kubernetes-client`
+- `cri-o`
+- `cni-plugins`
```
moio: What about dependencies of these packages, do they have any?

What if I remove one such dependency, will it remove the blacklist package as well?

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+### Actions to forbid
+
+Decision criteria:
+
+* Either Uyuni/SUSE Manager allows patching and package upgrade/removal or it does not allow those operations.
+* Additionally, the list of the "forbidden" packages above may vary with time and with CaaS Platform upgrades. Currently, there is no programmatic way to gather a list of those packages from a CaaS Platform cluster. What if Uyuni/SUSE Manager hardcodes that list of packages and the list changes? An update of Uyuni/SUSE Manager would be required.
+
+For the reasons above, Uyuni/SUSE Manager should not interfere with these actions to avoid breaking the cluster. The following actions should not be executed from Uyuni/SUSE Manager:
+
+- System reboot
```
moio: For completeness' sake I'd add power management actions here, and SP migration

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+- `cri-o`
+- `cni-plugins`
+
+NOTE: [installing NEW (not already installed) packages is allowed](https://documentation.suse.com/suse-caasp/4/single-html/caasp-admin/#_existing_cluster).
+
+At the time of the writing, Uyuni/SUSE Manager or Salt cannot forbid the package operations based on the target packages (there is no such kind of granularity).
+
+### Actions to forbid
+
+Decision criteria:
```
moio: Personal taste: put the proposal first, and the explanation about why it was chosen that way afterwards.

In this case, it seems to me the real motivation is "implementation complexity" - we could go for blocking package upgrades/removals completely as a first iteration, and then making it more granular (adding package name-level filtering) as a next step, if needed and when needed.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+
+The functionality would be used also for troubleshooting and installing troubleshooting tools and issuing commands.
+
+The risk of breaking the cluster by issuing any forbidden operation is still present while the blackout is disabled.
+
+Value of this solution:
+* Easy and quick to implement
+* The user can still break its cluster by issuing any forbidden action while the cluster is not in the blackout
+* The Workflow for the user is different and mechanical, can lead to errors
+
```
admd: [disable_modules/whitelist-modules](https://docs.saltstack.com/en/latest/ref/configuration/minion.html#std:conf_minion-disable_modules) could be another alternative to blackout.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+- Patch apply:
+  - If the targeted minion is a `caasp_node` system type, the action for applying a patch:
+    - Must be hidden from the UI and the corresponding action must fail: `/systems/details/ErrataList.do`, `/systems/ssm/ListErrata`, `ErrataManager#applyErrata`
+    - The corresponding action must fail if scheduled via XMLRPC: [`SystemHandler#scheduleApplyErrata`]
+
+Value of this solution:
+* This workflow does not constitute a radical change for a user
+* Change does not involve Salt, the user can still break its cluster by issuing Salt commands
+* Requires changes in Uyuni/SUSE Manager core code
+
```
admd: I am not sure if both **Uyuni/SUSE Manager action locking** & **Minion blackout** methods are mutually exclusive in its current form. We will probably need both in one way or the other.

IIUC, if we decide for **Minion blackout**, we will still need to hide the pages from UI & adjust XMLRPC API points to throw an error in such cases. It wouldn't be efficient to actually reach at salt level and then return the error, we should not even allow to create action/serveraction instance in DB in such cases.

 Similarly, if we only go for "**Uyuni/SUSE Manager action locking**" proposal, then we left susemanager exposed as you mentioned above and customer can do whatever she wants using salt commands directly.

accepted/0000-caasp-nodes.md
mbologna: The first step is to limit the damage that a user can trigger via Uyuni/SUSE Manager.
Secondly, limit the damage via Salt.
Does it make sense?

accepted/0000-caasp-nodes.md
mbologna: That could work as well, but it not published in the documentation as an "official" bootstrap method.
Do you think it would make sense to add it here?

accepted/0000-caasp-nodes.md
mbologna: The activation key is required to automatically assign CaaSP software channels upon registration and have `skuba-update` do the automatic update described [here](https://github.com/uyuni-project/uyuni-rfc/pull/26/files#diff-014d415bc2076bf85e73014398351426R262)

accepted/0000-caasp-nodes.md
moio: It does, and it may make sense to treat the latter as a "Next Step" or even separate RFC.

accepted/0000-caasp-nodes.md
mbologna: Good catch. It is right now unclear (on the CaaSP side) whether to consider forbidden the dependencies of the forbidden packages as well.
It is also difficult to construct the whole dependency tree of the dependencies with `zypper` (it will be available on 1.14.33) but right now it is a matter of iterating over `zypper info` and `zypper search`.

accepted/0000-caasp-nodes.md
mbologna: Right, I will add it

accepted/0000-caasp-nodes.md
mbologna: What you are suggesting is the following: whitelist the `state` module and disable the other execution module. In other words: SUSE Manager will be able to issue package/patch operations but users from the Salt command line will not.

Do I get it right?

accepted/0000-caasp-nodes.md
mbologna: Yes, very much agreed. I will push a correction here.

accepted/0000-caasp-nodes.md
mbologna: From a UI and from a [perf POV](https://github.com/uyuni-project/uyuni-rfc/pull/26#discussion_r345249600) I agree. Pushing a correction in this.

accepted/0000-caasp-nodes.md
moio: Yes.

But IIUC that would mean the list would comprise all possible registration methods we have, so maybe you could even consider getting rid of the list altogether and just say "with one of the available registration methods".

accepted/0000-caasp-nodes.md
moio: Hmm and what would prevent users to still install packages via a state in that case?

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,326 @@
...
+  * UI: Systems > Bootstrapping
+  * `bootstrap` XMLRPC call
+  * Not usual but still possible and hackier: mass bootstrap via salt-ssh and a user-generated roster file
+
+## Assumptions
+
+We assume that:
+* The user is registering the CaaS Platform nodes as Salt clients
+* The registration of every CaaS Platform node to Uyuni/SUSE Manager is already completed by the user using the aforementioned methods
+* The user has already created an activation key and associated it with the onboarding. The CaaS Platform nodes are registered correctly and with the proper CaaS Platform channel(s) assigned. The activation key is needed to have `skuba-update` automatically patch the cluster with the latest patches available<sup>1</sup>.
```
moio: Just for extra clarity, I would reorder the list items here so that's clear the AK must be created prior onboarding

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,326 @@
...
+- `supportutils-plugin-suse-caasp`
+
+and their dependencies<sup>3</sup>. At the time of writing, it is not easy to [build the dependency tree using plain `zypper`](https://stackoverflow.com/questions/12183757/reverse-dependency-generation-with-zypper) (it will be available with `zypper` 1.14.33) and it is not clear (from the CaaS Platform side) whether a forbidden package dependency should be considered forbidden as well. For example, `iptables` is a dependency of `cri-o`, but `cri-o` does not specify which versions of `iptables` are supported.
+
+To be on the safe side, Uyuni/SUSE Manager must not offer the possibility to issue the following actions on a registered CaaS Platform node:
+
+* Upgrade a package
+* Remove a package
+
+NOTE: [installing NEW (not already installed and not conflicting with already installed) packages is allowed](https://documentation.suse.com/suse-caasp/4/single-html/caasp-admin/#_existing_cluster).
```
moio: In general, it is difficult for the Server to determine if a package conflicts with another, as in principle that requires dependency solving.

So there is still a tiny chance installing a new package might break a CaaSP cluster - but I wonder if preventing any package installation is the right thing to do here.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,326 @@
...
+
+### Implementation details
```
moio:
```suggestion
### Description of current implementation
```

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,326 @@
...
+At the time of writing, Uyuni/SUSE Manager or Salt cannot forbid the package operations based on the target packages (there is no such kind of granularity).
+As a future implementation, Uyuni/SUSE Manager _might_ consider adding another level of granularity: filter actions depending on the argument passed (e.g. forbid install of a package when the package is `kubernetes-kubeadm`).
+
+NOTE: It is also implied that the [forbidden operations on a CaaS Platform node are also described in the Uyuni/SUSE Manager documentation](https://github.com/SUSE/doc-susemanager/pull/882/files).
+
+## Design
+
+### Identifying a CaaS Platform node
+
+To identify a CaaS Platform node it is enough to check if the `patterns-caasp-Node` pattern is installed. In the future, we might think to add a check for all CaaS Platform workloads to be installed as well.
```
moio: Does CaaSP provide a product package?

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,326 @@
...
+
+Internally to Uyuni/SUSE Manager, storing that a node is a CaaS Platform node can be fulfilled by adding a new system entitlement (e.g. `caasp_node`) to the node.
```
moio: Does this mean the entitlement will have to be added manually/via AK?

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,326 @@
...
+
+#### Perform an SP migration
+
+An SP migration is scheduled using Salt `state.apply` with the [distupgrade state](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/distupgrade/init.sls).
+
+#### Package upgrade
+
+A package upgrade is scheduled using Salt `state.apply` with the [pkginstall state](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/packages/pkginstall.sls).
+
+As a special case when issuing a package upgrade, a package installation from API is scheduled using Salt `state.apply` with the [pkginstall state](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/packages/pkginstall.sls).
```
moio: Do I get correctly that a package upgrade results in a different state being applied depending on whether it is scheduled via API or UI?

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,326 @@
...
+* Reboot a system
+* Apply a patch
+* Mark a system to automatically install patches
+* Issue any power management action via Cobbler
+* Perform an SP migration
+* Upgrade a package
+* Remove a package
+* Install a package (via API only)
+
+The first step is to block these actions at the Salt level, and later at the Uyuni/SUSE Manager level.
```
moio: This seems in contradiction with what you wrote in https://github.com/uyuni-project/uyuni-rfc/pull/26#discussion_r345679341

Is this RFC about blocking actions, or Salt commands, or am I missing something?

accepted/0000-caasp-nodes.md
admd: Ok didn't think it through. Now the problem that I see with [disable_modules/whitelist-modules ](https://docs.saltstack.com/en/latest/ref/configuration/minion.html#std:conf_minion-disable_modules) is that they are part of minion configuration file which means minion will need to be restarted every time we change something(update modules list) there.

While the `blackout` is more flexible in the sense that it can be switch on/off with just pillar update.


> Hmm and what would prevent users to still install packages via a state in that case?

For sles/opensuse, if  `zypperpkg` module is defined under `disable_modules` then user cannot install or perform any package related operation using `state` or directly calling the `pkg` module. 
Errors are not pretty but at least it works.I quickly execute a state on a minion to install a package where `zypperpkg` was in the list of `disabled_modules` and I got the following error.

```
  ID: test
    Function: pkg.installed
      Result: False
     Comment: State 'pkg.installed' was not found in SLS 'teststate'
              Reason: 'pkg' __virtual__ returned False
     Changes:   
```

accepted/0000-caasp-nodes.md
mbologna: OK

accepted/0000-caasp-nodes.md
mbologna: No, the entitlement will be automatically added upon registration, provided that the initial package list of the node return the pattern `patterns-caasp-` installed

accepted/0000-caasp-nodes.md
mbologna: In addition to the difference that @admd has just explained, the difference between blackout and whitelist is that blacklist allows to granularly whitelist functions, while whitelist of modules enables all the functions in a module.

accepted/0000-caasp-nodes.md
mbologna: Although I do not recall any package that might be in conflict with the forbidden packages right now, there might be a possibility. Let's keep open the possibility to install packages right now - we can disallow it in the future.

accepted/0000-caasp-nodes.md
mbologna: The only one that seems to be a product package would be `caasp-config`. Why?

accepted/0000-caasp-nodes.md
mbologna: No. A package upgrade is always issued with a `state.apply packages.pkginstall`.
The difference is that it is not possible to install a non-installed package on the target system from the UI (you can only select a non-installed package from the list).
But it is possible to upgrade an installed package using `system_installpackage` from the API (see https://github.com/uyuni-project/uyuni-rfc/pull/26/files#diff-014d415bc2076bf85e73014398351426R128-R138 )

accepted/0000-caasp-nodes.md
mbologna: Yes, my bad, after a round of thoughts I imagined to block the problem at the base. First Salt, then Uyuni/SUSE Manager.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,370 @@
...
+
+* CaaS Platform node: any node that comprises the cluster, either a control plane node or a worker node
+* Bootstrap/registration in the Uyuni/SUSE Manager context: registering a node to Uyuni/SUSE Manager via [the documented methods<sup>4</sup>](https://opensource.suse.com/doc-susemanager/suse-manager/client-configuration/registration-overview.html).
+At the time of writing, registering a CaaS Platform node to Uyuni/SUSE Manager has to be manually performed by the user.
+
+## Assumptions
+
+We assume that:
+* The user has already created an activation key and associated it with the onboarding. The CaaS Platform nodes are registered correctly and with the proper CaaS Platform channel(s) assigned. The
+activation key is needed to have `skuba-update` automatically patch the cluster with the latest patches available<sup>1</sup>.
```
flavio: I would say it's need to allow CaaSP nodes to consume repositories managed by SUMA. As a nice "side-effect" `skuba-update` will also apply the patches published by SUMA.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,370 @@
...
+NOTE: It is outside of this RFC to discuss automatic registration methods of the CaaS Platform nodes.
+
+## Technical description of the problem
+
+### The forbidden operations
+
+#### Patch and power management operations
+
+[`skuba-update`](https://documentation.suse.com/suse-caasp/4/single-html/caasp-admin/#_base_os_updates) systemd timer will take care of [patching](https://github.com/SUSE/skuba/blob/master/skuba-update/skuba_update/skuba_update.py#L301-L305) all installed packages locally to every node of the cluster and optionally reboot the node if `zypper` signals that a reboot is required.
+In the case that a patch that requires reboot has been installed by Uyuni/SUSE Manager before `skuba-update` applies it, then the node will not be tagged as pending a reboot.<sup>2</sup>.
```
flavio: SUMA can tag a node to be rebooted by simply `touch`-ing the `/var/run/reboot-required` file. The rebooting is done by the [kured](https://github.com/weaveworks/kured) service, which is deployed by k8s on all the nodes of the cluster.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,370 @@
...
+
+For the reasons above, Uyuni/SUSE Manager must not offer the possibility to issue the following operations on a registered CaaS Platform node:
+
+* Reboot a system
```
flavio: This could be translated with a `touch` of the reboot sentinel file.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,370 @@
...
+* Apply a patch
```
flavio: It's fine to apply patches, as long as these are *not* interactive ones. Also remember to create the reboot sentinel file.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,370 @@
...
+- `cri-o`
+- `kubernetes-client`
+- `kubernetes-kubeadm`
+- `kubernetes-kubelet`
+- `patterns-base-basesystem`
+- `patterns-caasp-Node-x.y`<sup>5</sup>
+- `skuba-update`
+- `supportutils-plugin-suse-caasp`
+
+and their dependencies<sup>3</sup>. At the time of writing, it is not easy to [build the dependency tree using plain `zypper`](https://stackoverflow.com/questions/12183757/reverse-dependency-generation-with-zypper) (it will be available with `zypper` 1.14.33) and it is not clear (from the CaaS Platform side) whether a forbidden package dependency should be considered forbidden as well. For example, `iptables` is a dependency of `cri-o`, but `cri-o` does not specify which versions of `iptables` are supported.
```
flavio: I think the only package you should care about is the `patterns-caasp-Node-x.y`. This one has all the other packages as runtime dependencies. The version of these files is kept in tight control by this pattern package. You can find more information about that inside of [this RFC](https://github.com/SUSE/caasp-rfc/blob/master/accepted/0040-platform-upgrade.md#packaging-approach).

As for the "second level" dependencies (eg: iptables), I think you should just ignore them. Keep your life easy!
The requirement of a specific version of iptables should be formulated at the `.spec` level by our packages, this shouldn't be your problem. (BTW: CRI-O has no relation with iptables, this is something used by kube-proxy right now - k8s is a complex beast, let us take care of that).

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,370 @@
...
+
+To be on the safe side, Uyuni/SUSE Manager must not offer the possibility to issue the following operations on a registered CaaS Platform node:
+
+* Upgrade a package
+* Remove a package
```
flavio: This is fine as long as:

  * No interactive package is installed/upgraded
  * The dependencies of `patterns-caasp-Node-x.y` are not broken
  * The pacakge `patterns-caasp-Node-x.y` is not touched, the upgrade of this package should be done only via skuba

accepted/0000-caasp-nodes.md
flavio: Again, just focus on keeping `patterns-caasp-Node-x.y` happy

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,370 @@
...
+Given all the considerations above, the following operations must not be scheduled by Uyuni/SUSE Manager for a CaaS Platform node:
+
+* Reboot a system
+* Apply a patch
+* Mark a system to automatically install patches
+* Issue any power management operation via Cobbler
+* Perform an SP migration
+* Upgrade a package
+* Remove a package
+* Install a package (via API only)
```
flavio: I think this can be revisited based on what I wrote earlier.

flavio:
Ouch, github made a bit of a mess... I started to review the RFC some days ago, but didn't have the time to commit it. Today I resumed the work but github seemed to have lost all my previous comments. Apparently that wasn't the case!

Just ignore the commends made against outdated text and forgive me about the accidental noise

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,277 @@
...
+NOTE: [installing NEW (not already installed) packages is allowed](https://documentation.suse.com/suse-caasp/4/single-html/caasp-admin/#_existing_cluster).
+
+At the time of the writing, Uyuni/SUSE Manager or Salt cannot forbid the package operations based on the target packages (there is no such kind of granularity).
+
+### Actions to forbid
+
+Decision criteria:
+
+* Either Uyuni/SUSE Manager allows patching and package upgrade/removal or it does not allow those operations.
+* Additionally, the list of the "forbidden" packages above may vary with time and with CaaS Platform upgrades. Currently, there is no programmatic way to gather a list of those packages from a CaaS Platform cluster. What if Uyuni/SUSE Manager hardcodes that list of packages and the list changes? An update of Uyuni/SUSE Manager would be required.
```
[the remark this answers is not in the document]
mbologna: > If you try to upgrade any of the above packages zypper will complain. This is described in depth here.

I manually created a situation where an updated version of `cri-o` is available:

```
ip-172-27-1-209:~ # zypper in --oldpackage cri-o-1.15.0-1.4
Loading repository data...
Reading installed packages...
Resolving package dependencies...

The following package is going to be downgraded:
  cri-o

1 package to downgrade.
Overall download size: 11.4 MiB. Already cached: 0 B. No additional space will be used or freed after the operation.
Continue? [y/n/v/...? shows all options] (y): y
Retrieving package cri-o-1.15.0-1.4.x86_64                                                                                                                                                                (1/1),  11.4 MiB ( 50.9 MiB unpacked)
Retrieving: cri-o-1.15.0-1.4.x86_64.rpm .................................................................................................................................................................................................[done]

Checking for file conflicts: ............................................................................................................................................................................................................[done]
(1/1) Installing: cri-o-1.15.0-1.4.x86_64 ...............................................................................................................................................................................................[done]
There are running programs which still use files and libraries deleted or updated by recent upgrades. They should be restarted to benefit from the latest updates. Run 'zypper ps -s' to list these programs.
 ```

and I upgraded it: `zypper` does not complain - is this because I manually created the setup?

```
ip-172-27-1-209:~ # rpm -qa | grep -iw cri-o
cri-o-1.15.0-1.4.x86_64
cri-o-kubeadm-criconfig-1.15.0-3.3.2.x86_64
ip-172-27-1-209:~ # rpm -qa | grep -i patterns-caasp
patterns-caasp-Node-1.15-1.2-1.1.x86_64
ip-172-27-1-209:~ # zypper up cri-o
Loading repository data...
Reading installed packages...
Resolving package dependencies...

The following package is going to be upgraded:
  cri-o

1 package to upgrade.
Overall download size: 11.3 MiB. Already cached: 0 B. No additional space will be used or freed after the operation.
Continue? [y/n/v/...? shows all options] (y): y
Retrieving package cri-o-1.15.0-3.3.2.x86_64                                                                                                                                                              (1/1),  11.3 MiB ( 50.9 MiB unpacked)
Retrieving: cri-o-1.15.0-3.3.2.x86_64.rpm ...................................................................................................................................................................................[done (3.3 MiB/s)]

Checking for file conflicts: ............................................................................................................................................................................................................[done]
(1/1) Installing: cri-o-1.15.0-3.3.2.x86_64 .............................................................................................................................................................................................[done]
There are running programs which still use files and libraries deleted or updated by recent upgrades. They should be restarted to benefit from the latest updates. Run 'zypper ps -s' to list these programs.
 ```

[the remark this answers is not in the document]
mbologna: > @ereslibre: can you imagine other packages SUMA should not touch?

I think `kernel` is one of those: if Uyuni/SUSE Manager install an upgrade or patch the kernel before `skuba-update`, the node does not have the reboot sentinel set. Have I understood it correctly?

accepted/0000-caasp-nodes.md
flavio: This is a package we would like to get rid of. `patterns-caasp-Node` is the one to look for.

accepted/0000-caasp-nodes.md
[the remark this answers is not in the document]
flavio: To be more specific: the dependency mechanism implemented by the `patterns-caasp-Node-<major>.<minor>` package:
  * **allows** patch release updates (like the one you similated before)
  * **prevents** major version updates
  * **minor** major version updates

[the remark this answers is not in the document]
flavio: > I think `kernel` is one of those: if Uyuni/SUSE Manager install an upgrade or patch the kernel before `skuba-update`, the node does not have the reboot sentinel set. Have I understood it correctly?

Yes, it wouldn't have the reboot sentinel. That why I suggested [above](https://github.com/uyuni-project/uyuni-rfc/pull/26#discussion_r345285894) that you should `touch /var/run/reboot-required` when needed

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,436 @@
...
+### Step 1: operations blocking at Salt level
+
+By default, every node that is a CaaS Platform node is not allowed to perform Salt commands.
+This is achieved by issuing a [minion blackout](https://docs.saltstack.com/en/latest/topics/blackout/) upon registration to Uyuni/SUSE Manager.
+
+Blackout works by blocking every command targeting the minion, except for a whitelist of functions allowed during the blackout. To enable or disable a blackout, a `saltutil.refresh_pillar` has to be issued on the minion.
+
+The allowed list of Salt commands can only contain:
+
+- `test.ping`
```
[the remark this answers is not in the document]
moio: Looking at that file, other calls seem sensible to have: config.get, status.uptime, saltutil.sync_modules, saltutil.sync_grains... I'd suggest either doing a thorough review or leaving this detail for the implementation phase.

Just `test.ping` is not going to work.

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,480 @@
...
+
+<!-- - What other designs/options have been considered?
+- What is the impact of not doing this? -->
+
+### Disallow forbidden actions at both levels
+
+As an alternative to "Package locking at `zypper` level", we can consider adding another level of granularity: filter operations depending on the packages involved (e.g. forbid install of a package when the package is `kubernetes-kubeadm`).
+The idea is that:
+
+* At the Salt level: when the user tries to issue a command (e.g. `pkg.upgrade`) involving a forbidden package, the command will fail.
```
mateiw: I think this would require some more changes to Salt. We could avoid this by using package locking on the forbidden packages + minion `blackout` with `blackout_whitelist`.

mbologna: Isn't that covered in https://github.com/uyuni-project/uyuni-rfc/pull/26/files#diff-014d415bc2076bf85e73014398351426R289 ?

accepted/0000-caasp-nodes.md
```diff
@@ -0,0 +1,372 @@
...
+#### Patch and power management operations
+
+[`skuba-update`](https://documentation.suse.com/suse-caasp/4/single-html/caasp-admin/#_base_os_updates) systemd timer will take care of [patching](https://github.com/SUSE/skuba/blob/master/skuba-update/skuba_update/skuba_update.py#L301-L305) all installed packages locally to every node of the cluster and optionally reboot the node if `zypper` signals that a reboot is required.
+In the case that a patch that requires reboot has been installed by Uyuni/SUSE Manager before `skuba-update` applies it, then the node will not be tagged as pending a reboot.<sup>2</sup>.
+
+For the reasons above, Uyuni/SUSE Manager must not offer the possibility to issue the following operations on a registered CaaS Platform node:
+
+* Apply a patch (if the patch is marked as interactive)
+* Mark a system to automatically install patches
+* Issue any power management operation via Cobbler
```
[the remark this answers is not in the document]
moio: Why was reboot removed from this list? (maybe I am missing something)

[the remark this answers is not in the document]
mbologna: Reboot has now its own section after the patching operation/action - are you on the latest version?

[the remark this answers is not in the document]
moio: I was not, sorry.

moio:
I think this can be merged now, especially considered that implementation is already well underway.

accepted/0000-caasp-nodes.md
mateiw: Yes. I'm just suggesting that disallowing actions might not be robust enough and it might be turn out to be too complex, as you said.
The two options could be combined:
1. At the SUMA level prevent actions on forbidden packages
2. Forbid some Salt states and modules using the `blackout` mechanism (e.g. reboot) but allow states that upgrade/install packages. No need for more fine grained filtering at this level (e.g. no need to forbid certain packages)
3. Lock packages to prevent accidental changes from the Salt CLI or from the system CLI.

mbologna:
RFC merged. Implementation underway, thanks to everyone who contributed!

PR #27, opened 2019-11-22
Automatic cluster node registration

mbologna:
This RFC describes how we are going to automatically register a cluster node to Uyuni/SUSE Manager upon `kubeconfig` import.

- Formatted text: https://github.com/uyuni-project/uyuni-rfc/blob/c7bc8e6360a34ae8e51c88266957ed195ab3b91d/accepted/0000-caasp-registration.md
- Text with inline comments: https://github.com/uyuni-project/uyuni-rfc/pull/27/files

EPIC reference: https://github.com/SUSE/spacewalk/issues/8307#issuecomment-550400131

TODO:
- [x] (@mbologna) Change filename before merging
- [x] (@mbologna) Add PR link in the RFC text

### DIFF
--- a/accepted/0000-caasp-registration.md
+++ b/accepted/0000-caasp-registration.md
@@ -1,0 +1,130 @@
+- Feature Name: automatic registration of Kubernetes nodes
+- Start Date: 2019-11-20
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Automatically register to Uyuni/SUSE Manager nodes all the nodes reachable via a `kubeconfig` import.
+
+# Motivation
+[motivation]: #motivation
+
+<!-- - Why are we doing this?
+- What use cases does it support?
+- What is the expected outcome?
+
+Describe the problem you are trying to solve, and its constraints, without coupling them too closely to the solution you have in mind. If this RFC is not accepted, the motivation can be used to develop alternative solutions. -->
+
+When dealing with a Kubernetes or a CaaS Platform cluster, the user is required to manually register all cluster nodes to Uyuni/SUSE Manager.
+In this RFC we will use the term "cluster node" to refer to a generic Kubernetes cluster node.
+
+We are going to describe a way to automatically register all the cluster nodes upon `kubeconfig` import.
+
+# Detailed design
+[design]: #detailed-design
+
+Under "Virtual Host Managers > Kubernetes" cluster, there is already the possibility of importing a `kubeconfig` file (obtained from a Kubernetes cluster).
+After the import, the hostname of each node of the cluster is displayed, along with basic information: operating systems, CPU(s) information, architecture, and memory.
+
+The idea is to introduce a new checkbox on this page called "Register the cluster nodes to Uyuni/SUSE Manager".
+If the user does not click on the checkbox, the workflow is unchanged.
+We are going to describe it in the next chapter what happens when the user clicks on the checkbox.
+
+## The new workflow
+
+When the checkbox is selected, a new UI form will appear containing asking for the following information:
+
+- SSH port
+- SSH username with `sudo` rights
+- SSH password
+- Activation Key
+- Proxy
+
+Most experienced readers will recognize that these details are the same asked in the "Boostrap minions" page except for the host (the hosts will be gathered from the `kubeconfig` import call).
+
+**Assumptions**: it is assumed that all the details provided by the user are valid for all nodes of the cluster, which means that all the cluster nodes have the same:
+
+- SSH port
+- `root` SSH access or a user with `sudo` rights
+- password for the user above
+
+After the form has been filled out and the `kubeconfig` import has begun, the usual workflow is followed.
+The additional result for the final user would be that each system listed reached after the `kubeconfig` has been imported will be bootstrapped to Uyuni/SUSE Manager.
+
+## How each cluster node is registered
+
+The SSH details supplied by the user in the workflow will be used to bootstrap each reachable node of the cluster. We want to reuse as much code as possible: a call to `XmlRpcSystemHelper#bootstrap` will be issued for each node of the cluster. For the sake of simplicity and feature-wise, each node will be registered as a Salt minion and not as a salt-ssh minion.
+Scalability-wise:
+
+* From the field, it has been reported that a Kubernetes cluster should not be comprised of more than 100 clusters: managing 100 registrations at the same time should not constitute a performance problem for a healthy Uyuni/SUSE Manager.
+* Theoretically, a cluster [must not have more than 5000 nodes](https://cloud.google.com/solutions/scope-and-size-kubernetes-engine-clusters): the same considerations apply as per the above.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+<!-- Why should we **not** do this?
+
+  * obscure corner cases
+  * will it impact performance?
+  * what other parts of the product will be affected?
+  * will the solution be hard to maintain in the future? -->
+
+Some users already have their automation stack (e.g. puppet or cloud-init) to automatically use the Uyuni/SUSE Manager bootstrap script upon cluster node provisioning.
+In that case, the feature introduced with this RFC is not appealing to this segment of users.
+
+# Alternatives
+[alternatives]: #alternatives
+
+<!-- - What other designs/options have been considered?
+- What is the impact of not doing this? -->
+
+## Bootstrapping by pasting private SSH key
+
+This is an alternative to using SSH username and passwords to automatic register the cluster nodes.
+
+**Assumptions**: it is assumed that all the cluster nodes have the same:
+
+- `root` SSH access or a user with `sudo` rights
+- An SSH key already deployed for the user above in the `authorized_keys` file
+- SSH logins via SSH keys enabled
+
+The form for `kubeconfig` import will be changed to two inputs:
+
+- User
+- Private SSH key for that user<sup>1</sup>
+
+The private key will be used to register the nodes to Uyuni/SUSE Manager. At the time of writing, Uyuni/SUSE Manager does not have a way to bootstrap a system using SSH keys and this feature must be implemented if we choose this alternative.
+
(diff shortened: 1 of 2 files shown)


mbologna:
@moio : feel free to review just the scalability paragraph, thanks!

moio:
@mbologna would it be possible/would it make any sense to extend the kubeconfig format so that SSH details/AK could be directly specified in it?

mbologna:
> @mbologna would it be possible/would it make any sense to extend the kubeconfig format so that SSH details/AK could be directly specified in it?

Technically, I think so. But the question would be if the k8s cluster distributions (in our case CaaSP but not limited to that) would want to implement this change.

moio:
I was thinking about adding custom data in the kubeconfig in a non-obtrusive way - let's say some `uyuni` specific structure. As long as we are reasonably sure that k8s will ignore those (and I wonder if there might be already a way to specify custom metadata) we should be safe even without a patch upstream.

mbologna:
@flavio : do you have any inputs here?

accepted/0000-caasp-registration.md
```diff
@@ -0,0 +1,130 @@
...
+Some users already have their automation stack (e.g. puppet or cloud-init) to automatically use the Uyuni/SUSE Manager bootstrap script upon cluster node provisioning.
+In that case, the feature introduced with this RFC is not appealing to this segment of users.
+
+# Alternatives
+[alternatives]: #alternatives
+
+<!-- - What other designs/options have been considered?
+- What is the impact of not doing this? -->
+
+## Bootstrapping by pasting private SSH key
```
mateiw: Yet another alternative would be to use SSH key forwarding:
1. the user forwards to the SUMA server a SSH key that is authorized on all cluster nodes
2. in the bootstrap UI instead of providing a username/password the user can select to a forwarded SSH key
3. once bootstrapping is done, the key forwarding is not needed anymore. The use can terminate the SSH connection that forwarded the key.

flavio:
> @flavio : do you have any inputs here?

AFAIK there's no way to extend the kubeconfig file with custom metadata. I also don't know how strict the different parsers are about that. Maybe they will ignore the unknown values, maybe they will just choke on them...

What problem would you like to solve with that? I didn't get it from @moio's message.

moio:
@flavio: problem is that SUSE Manager needs more data than what is provided in kubeconfig in order to implement automatic node cluster registration. Of course we could add some UI or an extra "sumakubeconfig" file, but I guess it would just be a lot more convenient to just sneak in the needed bits in the existing file - assuming of course parsers wouldn't complain.

flavio:
I see, sorry... I lost tracks of the SUMA RFCs. This is the one I don't understand the purpose yet.

This is the motivation written in the RFC:
>When dealing with a Kubernetes or a CaaS Platform cluster, the user is required to manually register all cluster nodes to Uyuni/SUSE Manager. In this RFC we will use the term "cluster node" to refer to a generic Kubernetes cluster node.
>
> We are going to describe a way to automatically register all the cluster nodes upon kubeconfig import.

I don't know how often a customer is going to: deploy a CaaSP cluster using SCC and then register it against SUMA.

The majority of customers I'm aware of want to register the nodes since their creation against SUMA. They want to do that because of different reasons like: air-gapped installations, have full control of the software that is used inside of the cluster itself.

Challenging question: does it really make sense to invest time developing this RFC instead given it would IMHO cover a small use case?

mbologna:
> I don't know how often a customer is going to: deploy a CaaSP cluster using SCC and then register it against SUMA.
> 
> The majority of customers I'm aware of want to register the nodes since their creation against SUMA. They want to do that because of different reasons like: air-gapped installations, have full control of the software that is used inside of the cluster itself,...
> 
> Challenging question: does it really make sense to invest time developing this RFC instead given it would IMHO cover a small use case?

IMHO the effort required to develop the RFC as described is not prohibitive and I consider it rather straightforward; with this effort, we can cover also customers that do not have `cloud-init` (or similar) ways to automatically register CaaSP nodes.

@joesusecom @paususe : what is the voice from the field?

flavio:
The real question is: how often do you expect customers to deploy CaaSP (that includes also kubernetes) and **then** register the nodes against SUMA? IMHO I don't think this is a common workflow.

I can definitely imagine a customer booting some machines and then, before proceeding with CaaSP installation, register them against SUMA (for example to get access to the mirrored channels of CaaSP available on SUMA). In that case there would be no kubeconfig file to handle.

joesusecom:
I agree with Flavio that the "onboard and existing cluster" case isn't that relevant in real life. So I think it's perfectly fine if there's a manual step needed (either for the whole cluster or for each individual node) to provide valid SSH credentials "out of band". From a security point of view, I'm not even sure that one would _want_ that information to be part of the kubeconfig "contract".

In general, I think that the workflow described in the RfC for providing the SSH info/keys makes perfect sense.

Just two additional comments:

1) There's an old open feature request for allowing mass-onboarding of _any_ servers to SUSE Manager that we haven't implemented yet. So _if_ we implement the SSH key option and/or the "shared root password" option for CaaSP, we should do this in a generic way that lets a user also use the same UI experience, but provide, let's say, a CSV file of server names/IPs to target.
2) We've had the concept of "bootstrap nodes" (systems that have been created in the DB, but aren't onboarded yet) for quite a while. Those systems can either be created by using the bootstrap image via Cobbler or by creating the systems via the API. Wouldn't it make sense to allow creating a system group for the Kubernetes cluster and "bootstrap nodes" in it for all those systems that were discovered via the Kubernetes API in SUSE Manager, regardless of whether the SSH-based onboarding has already happened or not? Those nodes would be marked with a different icon and only provide data that's available from the API endpoint, but an admin could already start assigning configuration to them etc.

accepted/0000-caasp-registration.md
```diff
@@ -0,0 +1,130 @@
...
+
+We are going to describe a way to automatically register all the cluster nodes upon `kubeconfig` import.
+
+# Detailed design
+[design]: #detailed-design
+
+Under "Virtual Host Managers > Kubernetes" cluster, there is already the possibility of importing a `kubeconfig` file (obtained from a Kubernetes cluster).
+After the import, the hostname of each node of the cluster is displayed, along with basic information: operating systems, CPU(s) information, architecture, and memory.
+
+The idea is to introduce a new checkbox on this page called "Register the cluster nodes to Uyuni/SUSE Manager".
```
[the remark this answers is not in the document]
mbologna: If we don't implement the refresh, how the workflow would be? Just re-import the `kubeconfig` to take into account newly added nodes to the cluster?
(removed nodes will not be cleaned up by Uyuni/SUSE Manager)

accepted/0000-caasp-registration.md
mbologna: IIUC, from the point of view of the user, it should be something like:

1. `ssh -A root@susemanager`
2. In Uyuni/SUSE Manager UI Java backend, gather the forward keys for a particular user (in this case `root`) and use them to bootstrap the nodes

I am not sure if 2. is feasible from within Java and an `ssh` user.
Am I missing anything?

accepted/0000-caasp-registration.md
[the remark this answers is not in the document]
mateiw: I think in the end it's the same effort even if we just re-import the `kubeconfig` file. In both cases some diff will be needed between the nodes reported by the cluster and the nodes registered to Suma.
I wonder if this is something that can happen frequently. If nodes are not added/removed very often then manual registration might be enough.

accepted/0000-caasp-registration.md
mateiw: I haven't tested but maybe forwarded keys can be retrieved using a script with special `sudo` privileges. We already have a few such scripts (e.g. for restarting Suma from the UI).

accepted/0000-caasp-registration.md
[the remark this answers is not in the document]
mbologna: Let's stick with just `kubeconfig` import and then go from there.

accepted/0000-caasp-registration.md
mbologna: I see that this option might have a lot of additional moving parts compared to the one described in the RFC; for this reason, I would consider the described option in the RFC preferable.

mbologna:
> I can definitely imagine a customer booting some machines and then, before proceeding with CaaSP installation, register them against SUMA (for example to get access to the mirrored channels of CaaSP available on SUMA). In that case there would be no kubeconfig file to handle.

I think this can cover the scenario where the user does not have a CaaSP cluster.
But what happens in the scenario where the user already has one (or more) CaaSP clusters and a SUSE Manager and wants to register those machines to SUSE Manager to cherry-pick CaaSP patches in the associated channels?

flavio:
> I think this can cover the scenario where the user does not have a CaaSP cluster.
> But what happens in the scenario where the user already has one (or more) CaaSP clusters and a SUSE Manager and wants to register those machines to SUSE Manager to cherry-pick CaaSP patches in the associated channels?

I might sound a bit brutal, but in this case the user would have to script the execution of the bootstrap script in some way (he could do that with a simple for loop in bash for example).

In an ideal world we should have a beautiful UX for this scenario, but in the real world we have to prioritize. I think there are way more cooler integration stories for SUMA and k8s that would be worth to be explored before taking this scenario into account. Just my 2 humble cents :pray:

mbologna:
I got in direct contact with the field: my understanding is that this feature is interesting but it is not something that customers desperately need. This is coherent with what @flavio expressed.

From a mere technical perspective, I did not receive any complaint or understood any roadblock in this RFC. Are we all on the same page that, technically, this can work?

If so, here is my proposal: we are going to consolidate the research we did for this RFC by merging this PR. The implementation timing will be decided later in a follow-up.

flavio:
> From a mere technical perspective, I did not receive any complaint or understood any roadblock in this RFC. Are we all on the same page that, technically, this can work?

Yes, this could work. There are some details to sort out when it comes to obtain additional information from the kubeconfig file, but we could talk about that when the implementation phase is closer.

moio:
I agree with merging the text as it is for the time being

PR #28, opened 2019-11-26
Initial draft of Salt Minion updates

isbm:
Salt Minion update on different architectures and platform versions in order to keep Minions in sync to the latest Uyuni-supported version, aligned across all the managed system versions.

[Read it here](https://github.com/uyuni-project/uyuni-rfc/blob/isbm-salt-minion-update-cycle/accepted/00065-salt-minions-update-cycle.md)


accepted/00065-salt-minions-update-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+2. General statically compiled Python vanilla environment is provided as an RPM package from Uyuni repository and is allowed to be installed **(2)** on a Managed System, e.g. in `/opt/python3.8` or whatever version is needed. That way it is also allowed to have many Python versions as well as its modifications, if ever needed.
+3. Once virtual environment (not shown on the _Figure 1_) is created for the the separate Python interpreter, PIP is bootstrapping the entire Salt Minion against Uyuni-served PyPi server with all the required Salt Minion dependencies.
+
+This should result into running fully featured Salt Minion of latest version in an isolated virtual environment on an OS-isolated Python interpreter, which is aligned with the current Uyuni Server. Once Uyuni Server is updated to another version, so is the PyPi container image is also upgraded. Thus, Uyuni Server should issue mass-upgrade to all Salt Minions that should PIP-upgrade themselves, bumping up to the latest available version.
+
```
aplanas: How is this different from when a normal SLE product is updated? First update the SMT repositories, zypper up the main node, restart the services in the main node, go one by one (or via an salt / chef) to update the workers.

IIUC you are only replacing zypper up / apt-get update, with something more brittle: `pip install --upgrade <package-list>`. The order of this package list can break your venv (or site-packages)

accepted/00065-salt-minions-update-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+Consequently, this mechanism will allow to keep always aligned Salt Minions across the entire infrastructure consistent and identical.
+
+## Fail-proof Updates
+
+Having running Salt Minion inside a virtual environment, allows to deliver fail-proof updates. The basic mechanism is to clone Python Virtual Environment from the current active Salt Minion and update the Salt Minion there.
```
aplanas: Actually I was wondering about this.

With venvjail we created a tgz for each architecture. All of them lives in the master node and are scp to the target. From there are untarged and the systemd service was restarted (pointing into the new venv).

The goal was exactly this: rollback. But you need to implement all the mechanism of venv distribution, fail detection and rollback. We used ansible, so not agent was required, but I wonder how this can be implemented in salt, when we are updating the same agent that will do the replacement, the fail detection and the rollback.

accepted/00065-salt-minions-update-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+
+If update is finished successfully, Salt Minion is restarted into a new environment, which is set as default. Previos environment is kept, and is purged over one update cycle (one after current update). This way administrator always has an ability to roll-back to the previous known working version.
+
+If update failed, Salt Minion keeps running without changes, report is created and new Virtual Environment is purged immeditately.
+
+As shown on the _Figure 2_, the difference between RPM-based updates is that the software components are overwritten, instead of installed nearby. Therefore once Salt Minion is broken on upgrade for whatever reasons, it might be no longer possible to revert this with the faulty Salt Minion itself.
```
aplanas: Using a venv as a poor-man container is dangerous, and you need to implement of the logic behind the orchestration, upgrade, replacement and rollback.

accepted/00065-salt-minions-update-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+
+![Layout](images/00065-ve.png)
+
+_Figure 2_
+
+According to the _Figure 2_, if updated Salt Minion is successfully updated and is trusted enough, then the statuses of each virtual environment are shifted, and "Previous Salt Minion" is purged.
+
+Naturally, the same way it is possible to update the entire Python version on production and roll-back to the previous Python version, in case things go wrong.
```
aplanas: Yes. Something nice of this idea is that rm -fr /opt will not break the system. You can bootstrap it again.

This makes me wonder, in your idea, why to detach the python binary from the salt-minion code? Deploy all together as salt-bin is doing, and you avoid the risk of failing during the creation of the venv on the site.

accepted/00065-salt-minions-update-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+
+This RPM package remains being vendor supported.
+
+### Step 2: Virtual Environment
+
+Once Python interpreter is delivered and installed on a managed system, Virtual Environment is created. It should contain Package Installer for Python (PIP) and Salt Minion required packages list.
+
+Salt Minion is no longer installed from the RPM package, but bootstrapped by PIP environment. However, the distributed package version that is on the PyPi server remains vendor supported.
+
+> Idea for audit: checksums of all the files in the Virtual Environment can be calculated and stored in the database that might run on PyPi container. This way it is possible to detect if the environment was tampered.
```
aplanas: You need more than that: https://github.com/SUSE-Cloud/venvjail/blob/master/venvjail.py#L426

accepted/00065-salt-minions-update-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Step 3: Running Salt Minion
+
+Ready to go Salt Minion on the Virtual Environment is linked for the `systemd` as default and is started.
+
```
aplanas: https://github.com/SUSE-Cloud/venvjail/blob/master/venvjail.py#L279

accepted/00065-salt-minions-update-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+Automatically restore and reconnect previous Salt Minion, once new Salt Minion upgrade went wrong and new Salt Minion is either fails to start at all or fails to communicate/respond or has any other major fatal issues.
+
+**Solution**
+One of the ideas is to switch Minions with a simple watchdog function, which has only one purpose: to watch status of some cookie file for a specific amount of time and restore previous minion, in case timeout is over. This function could be a part of the mechanism that takes care of managing Python Virtual Environments.
+
+E.g. if new Salt Minion restarted successfully and reported back, Salt Master can issue another command to "finalise upgrade" and so Salt Minion updates cookie file. In this case watchdog function will get a signal that the previous Virtual Environment restoration is no longer needed. Otherwise, previous Salt Minion is restored back.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
aplanas: I still do not see the advantage of this idea.

* In an RPM / DEB based solution you need to create the packages, here too, and you need to convert them is something that can be installed via pip.

* In an RPM / DEB based solution, updating a single node is a matter of zypper up / apt-get update: this will take care of replacing the code, adjusting the migrations required for the new service, and restarting the service. In this idea you need to build a full infrastructure to create the venv, replace them, reconfiguring the services to point to the new venv, some watchdog to monitor the current status, etc, etc.

Willing to maintain a full Python for each arch / OS, I wonder how different is using a full RPM / DEB system where we provide our Python in /opt, and the minion and all the new libraries are living in /opt/pythonx.y/lib/site-packages/

aplanas:
BTW. If we decide to move in the direction of creating wheels, or anything installable via PIP, we have also this effort, that try to enable in OBS a new file format that can be generated: https://confluence.suse.com/pages/viewpage.action?pageId=352583689

accepted/00065-salt-minions-update-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+
+The principle is to run own minimal [PyPi](https://pypi.org/project/pypiserver/) local server on Uyuni cluster. In order to install a Salt Minion, a managed system should use PIP againstlocal PyPi server, as shown on Figure 1 below:
+
+![Layout](images/00065-layout.png)
+
+_Figure 1_
+
+Process update or installation flow as follows:
+1. One RPM general package of Salt Minion **(1)** is provided for all possible platforms and architectures and is installed on PyPi server. However, Salt Minion is not _installed_ per se in the sense as for end-user, but as a PyPi package into the PyPi repository location. This way patches are still tracked at RPM `.spec` level.
+2. General statically compiled Python vanilla environment is provided as an RPM package from Uyuni repository and is allowed to be installed **(2)** on a Managed System, e.g. in `/opt/python3.8` or whatever version is needed. That way it is also allowed to have many Python versions as well as its modifications, if ever needed.
```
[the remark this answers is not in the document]
isbm: Naturally, as long as we want any kind of software, other then comes with the current distribution — that is always an extra task. Question is how much we could reduce it?

As of Python3.x I was thinking reusing the same RPM, except building it statically on a specific platforms and install Python itself in `/opt` instead.

accepted/00065-salt-minions-update-cycle.md
```diff
@@ -0,0 +1,127 @@
...
+3. Once virtual environment (not shown on the _Figure 1_) is created for the the separate Python interpreter, PIP is bootstrapping the entire Salt Minion against Uyuni-served PyPi server with all the required Salt Minion dependencies.
```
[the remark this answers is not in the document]
isbm: I was not thinking about about Salt SSH. If you had read carefully, I was already mentioned watchdog function in the section of Fail-proof installation, as a part of the virtual env manager. The Virtual Env manager is not much of a rocket science, but just a simple program that would switch links between shell scripts that starts environments. The same program (implemented in whatever way) would also happily create virtual environments.

The OBS venvjail is an interesting topic. However, it is operating in different paradigm? I was not thinking applying PTF on each _client Minion_ separately, but once on the _Uyuni Server's PyPi server_, so then Salt Minion will just update itself from it, applying the patch.

accepted/00065-salt-minions-update-cycle.md
isbm: It is different that you don't have to generate another tons of RPM/deb packages that installing things separately in static way. You have to take care only of installation sources that are deployed on PyPi server. But is it brittle or not robust — isn't it just a technical details? As PoC of pip-boostrapped Salt can be found [here](https://github.com/saltstack/salt-bootstrap/blob/develop/bootstrap-salt.sh), although it seems used in production.

accepted/00065-salt-minions-update-cycle.md
isbm: Use Ansible to install Salt? 😄 Well, as I mentioned earlier, a venv manager would be needed to implement. Something that would a) prepare static Python and b) create/manage venvs and relink them for systemd. Actually, Ansible won't "fly" either, if there is no correct Python yet at all installed.

accepted/00065-salt-minions-update-cycle.md
isbm: No, venv is not any dangerous and is not a "poor-man container" either. It is just different python path to load modules, and nothing else. Set shell's variables different and you have different modules to load for the Python interpreter. The same identical way works PyInstaller, which you are advocating for.

accepted/00065-salt-minions-update-cycle.md
isbm: I am not sure, are you actually fully understand how PyInstaller works? It does not installs any Python. It is a C-written bootstrap program-carrier, which sets up a temporary virtual environment, reads itself and, depending on the form of the distribution (directory or one file) loads the modules into Python interpreter shipped as a library.

You also keep missing the fact that PyInstaller does not bundle `libc`, which is not any kind of static distribution. To make it actually static, you would need to use something like [StaticX](https://github.com/JonathonReinhart/staticx) including libc.

So we basically have a choice:

a) Generate RPMs for each arch with all the static libraries inside, deps and everything and support all that.

Or:

b) Let every machine pip-install to build a Minion in gentoo-like fashion, completely cutting all the packaging overhead from a).

accepted/00065-salt-minions-update-cycle.md
isbm: Not for this paradigm.

accepted/00065-salt-minions-update-cycle.md
isbm: Advantage is not to create lots of dependencies on each arch, but all that delegate to PyPi. You not supposed to create lots of packageS, but one packagE: a completely static vanilla Python, which supposed to be enough to run venvs and pip, so the rest can come from PyPi local server. Unfortunately, in practice, however, it likely will need more than one static package for crypto, SSL... 😞

isbm: I also thought it is quite obvious: if you want Python 3.8 on SLES11, running on S390, then probably it needs to be extra compiled there _anyway_, isn't it? 😉 It is straight-forward to [compile Python static](https://wiki.python.org/moin/BuildStatically) into one single RPM package. Basically the goal here is to have zero `Require`, something like `snap` image does.

accepted/00065-salt-minions-update-cycle.md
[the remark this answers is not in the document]
isbm: The whole idea is to detach Salt Minion from the system completely, so direct removal of it by `rm -rf /opt/salt` won't hurt anything of the operating system itself. I don't mind shipping static Python as just a tar archive, served somewhere on Uyuni server over HTTP.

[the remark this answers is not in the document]
aplanas: > I was not thinking about about Salt SSH. If you had read carefully, I was already mentioned watchdog function in the section of Fail-proof installation,

I did. My question remains. Who will create the venv?

You say this watchdog (even watchgod are usually only for monitoring and triggering). Ok, the watchdog will create the venv.

So the client will install the watchdog in the node that want to register. I guess that it found the there is no venv, somehow will find the pip registry in the DNS, somehow will know where the static Python is living, will also know the order of package installation (or where is living the requirements.txt for the first venv), etc, etc.

I think that you need to dig into those details, as will help how big / small are the real work.

Also managing the live-cycle of a venv is not trivial. I know that because in Ardana we had exactly this approach that you are talking about, but using Ansible (that resolve the bootstrap question). Until we moved to OBS + venvjail, the machinery required to bootstrap the initial venv for each service was insane. Full of details and exceptions, put in place based on real-world experience.

> as a part of the virtual env manager — something that I don't find yet a time to talk about.

So, please, dig here into the rabbit hole.

> The OBS venvjail is an interesting topic. However, it is operating in different paradigm? I was not thinking applying PTF on each _client Minion_ separately, but once on the _Uyuni Server's PyPi server_, so then Salt Minion will just update itself from it, applying the patch.

Sure, but when a single minion is failing, the first thing that you need to know is what is inside the venv. The exact version that is running, the changelogs, etc.

Do you expect that all the venvs will be flawless updated, atomically, ever? Let me put some examples:

* The fact that the pypi local registry have PyYAML version 2.4.0-45, that have this important patch released by Pablo two weeks ago, do not imply that is available in the minion that is failing.

* L3 needs to send a PTF, but is a test one, that will be target to one single node. How can be done without replacing the wheel in the local registry? How can be targeted to a single node?

* How can you trace the wheel to the exact version in IBS / OBS of the package, to check if a CVE was already fixed or not?

> Generally, speaking, it should be possible to just flush the entire Salt Minion away and re-bootstrap it, if it is needed (preserving keys, config and machine ID parts).

I agree. But this is not how the daily support works. You need to make it easy to find the root cause, and track the problem and the fix. All this can be fixed and provided, but is not a small amount of work.

accepted/00065-salt-minions-update-cycle.md
aplanas: > It is different that you don't have to generate another tons of RPM/deb packages that installing things separately in static way.

In that case I did misunderstand the proposal. As I did understand, you are proposing to create the package in OBS, and transform it later to something that can be registered in a local PyPi, so no patch get lost.

If this is correct, yes, you indeed need to generate another tons of RPM/deb packages, together with a way of making them installable with PIP.

This PoC is not answering how the live-cycle of those wheels is managed, that was my question.

Later I attached a proposal of creating in OBS wheels that can be installed in PIP. It is interesting, but is using this approach: you first create the RPM / DEB and OBS takes care of generating the PIP package.

But there is no escape of repackaging all for the patches and the support.

So my question remains: as you already have a bunch of packages, and a distribution channel (SMT), how this proposal is different?

Please, unresolve this comment.

accepted/00065-salt-minions-update-cycle.md
aplanas: : D I do not know!

Can you attach into the RPF this workflow? And from where the metadata required to build the venv comes from? For example, how this agent living in the node (the venv manager) will discover all the data that requires to bootstrap the first venv.

accepted/00065-salt-minions-update-cycle.md
aplanas: venv is for development, no for production. Is brittle, full of escapes that can break the system, the scripts and services needs to be patched in place, etc. Is a really nightmare when you use them outside development.

> The same identical way works PyInstaller, which you are advocating for.

Did I?

accepted/00065-salt-minions-update-cycle.md
aplanas: I indeed do not dig into PyInstaller, and I did not asked about it. I am not convinced that is not silver-bulled, if this counts.

What I asked is that, if you are willing to do the job and provide an static Python, why do you need now a venv? You can install this static Python in /opt/salt-x.y and all the dependencies (python and non-python .so, .a, whatever) in this namespace (/opt-salt-x.y/lib, /opt-salt-x.y/lib/site-packages, /opt-salt-x.y/usr, /opt-salt-x.y/bin, etc, etc)

> You also keep missing the fact that PyInstaller does not bundle libc

I am not sure, as I did not discussed here anything about PyInstaller. But I think that this is not a problem, no?

Go generate static binaries, and rust binaries link to `libc`, `libthread` and so on, and what?

How many bugs can you expect from this? Is the proposal generate a single binary per arch, and forget about the version of the OS? So for all and each binary `ldd` will be empty?

> a) Generate RPMs for each arch with all the static libraries inside, deps and everything and support all that.

> b) Let every machine pip-install to build a Minion in gentoo-like fashion, completely cutting all the packaging overhead from a).

I am clearly misinterpreting something, as this b) option will require from the client to have GCC, python-devel, openssh-devel and many other packages before the venv can be build a la gentoo.

So I guess that you prefer the option a) here, but in that case no venv/pip is required.

accepted/00065-salt-minions-update-cycle.md
aplanas: Ouch. So no L3 support.

accepted/00065-salt-minions-update-cycle.md
aplanas: >  Unfortunately, in practice, however, it likely will need more than one static package for crypto, SSL... disappointed

Exactly.

accepted/00065-salt-minions-update-cycle.md
isbm: Not the way you imagine.

aplanas: That is fair. Maybe you can elaborate it in the doc?

accepted/00065-salt-minions-update-cycle.md
isbm: > What I asked is that, if you are willing to do the job and provide an static Python, why do you need now a venv? You can install this static Python in /opt/salt-x.y and all the dependencies (python and non-python .so, .a, whatever) in this namespace (/opt-salt-x.y/lib, /opt-salt-x.y/lib/site-packages, /opt-salt-x.y/usr, /opt-salt-x.y/bin, etc, etc)

https://github.com/uyuni-project/uyuni-rfc/blob/isbm-salt-minion-update-cycle/accepted/00065-salt-minions-update-cycle.md#fail-proof-updates

accepted/00065-salt-minions-update-cycle.md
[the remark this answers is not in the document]
isbm: > when a single minion is failing, the first thing that you need to know is what is inside the venv. The exact version that is running, the changelogs, etc.

This is no different how SaltSSH is failing.

accepted/00065-salt-minions-update-cycle.md
joesusecom: > a) Generate RPMs for each arch with all the static libraries inside, deps and everything and support all that.

> b) Let every machine pip-install to build a Minion in gentoo-like fashion, completely cutting all the packaging overhead from a).

You're conveniently leaving out the path that `salt-bin` is actually taking if I understand things correctly: Using the oldest libc that you need to support when you build the PyInstaller "package". This isn't going to 100% guarantee that things will work, but chances are high that a single binary would be pretty portable from RHEL 6 up to SLES 15 and RHEL 8 (as long as you are on the same HW architecture).

accepted/00065-salt-minions-update-cycle.md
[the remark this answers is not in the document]
isbm: The whole approach is boiling down to make Salt Minion volatile in the same fashion as SaltSSH is already doing anyway. So the whole story of the patching applies the same path as it is applies to patch SaltSSH `thin.tgz`.  However, it seems unclear from the RFC at the moment.  I will update the RFC and re-request review.

accepted/00065-salt-minions-update-cycle.md
isbm: The `salt-bin` is just a [wrapper script](https://github.com/saltstack/salt-bin/blob/master/build.py#L126) around [PyInstaller](http://www.pyinstaller.org) providing [pre-defined spec](https://github.com/saltstack/salt-bin/blob/master/build.py#L14-L47), so let's just focus on what PyInstaller does. And the [bootstrap](https://github.com/pyinstaller/pyinstaller/tree/develop/bootloader) loader from PyInstaller does not provide `libc`, consequently making the resulting binary dynamic.

That said, to ensure Python environment works 100% on RHEL6, SLES11 etc, the `libc` should be also brought together. Granted, using [StaticX](https://github.com/JonathonReinhart/staticx) together with PyInstaller would include `libc`, but that is even worse hack.

accepted/00065-salt-minions-update-cycle.md
isbm: Yes, I will update the RFC, since it clearly missing many parts. Thanks for help!

moio:
As a different solution to this problem is currently being implemented (design is internally available [here](https://confluence.suse.com/display/SUSEMANAGER/Tiamat+vs+venvjail) and will be published to the community soon) and the author of this PR has left the Team, I am for the time being archiving this PR.

Please note this is only in the interest of readability of the repo and GitHub project - we are open to revisit different ideas whenever there is a concrete need.

In any case thanks for this contribution @isbm

PR #29, opened 2020-01-02
modularisation rfc

lucidd:
RFC about cleaning up the Uyuni codebase by modulerizing it for better long term maintainablilty and reduced developer load. [rendered](https://github.com/uyuni-project/uyuni-rfc/blob/9c645c3ab8b60aec7b3ded8b3fe5e2706488326e/accepted/00065-modularisation.md)

### DIFF
--- a/accepted/00065-modularisation.md
+++ b/accepted/00065-modularisation.md
@@ -72,7 +72,12 @@ In tomcat the situation is maybe a little bit more complicated since our busines
 
 
 
+## Minimal example
 
+This ia a minimal example extracting most of the interactions with salt into an interface. Its not complete
+as it lacks more refinement i.e getting rid of JsonElement in result or splitting it up into more topical interface i.e having libvirt interaction in its own interface. But it should give a general idea of how this rfc looks applied to our code.
+
+[salt-branch](https://github.com/uyuni-project/uyuni/compare/salt-interface?expand=1)
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/accepted/00065-modularisation.md
+++ b/accepted/00065-modularisation.md
@@ -1,0 +1,85 @@
+- Feature Name: modular_codebase
+- Start Date: 2.1.2020
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This proposal is about improving the code quality, long term maintainablilty and testabilty of Uyuni by
+creating a set of guidelines based on best practices and retroactively applying them to the existing codebase.
+
+
+# Motivation
+[motivation]: #motivation
+
+The codebase of Uyuni has become increasingly hard to understand and maintain. The majority of time is spent on fixing bugs, debugging, fiddling with complex test setups and simply trying to understand what existing code even does (this includes new code). Even the most knowlegable people on the team don't always know how Uyuni would behave in a certain scenario or how its supposed to behave. Those problems are symptoms of the same underlying issue which is only adding new features without any big picture or architecture in mind and the complexity and lack of quality resulting from that. This creates a feedback loop where too much busywork leaves not enough time for proper reseach and development leading to rushed features leading to more bugs, complexity and busywork. This proposal is an attempt at breaking this loop by investing into a maintainable codebase that requires less busywork and allows for more time on the things that matter.
+
+# Detailed design
+[design]: #detailed-design
+
+The core of this proposal will be modulerizing our codebase driven by the following aspects:
+
+- testabilty (make it easy to test invidual components of Uyuni in isolation without a full multivm setup running.)
+- understandability (make the APIs and Components clear so it is possible to understand how things are supposed to work without the need to look at its implementation and all its interactions.)
+- reusablity (decouple components as much as possible to the point where they can be easily reused as a library or run standalone.)
+
+The goal of those aspects is to cut down on busywork and make working with the code more productive.
+
+The proposal is structured in a way that each area can be tackled indiviually and the refractoring can be done gradually in multiple repeatable steps.
+After each step the code should be in a compilable state with all tests still passing.
+
+
+# Preparation 
+
+- pick one of the areas like: salt, filesystem, database, task scheduling, different topics of buisness logic like cve audit, etc.
+- The first step is to gather all the requirements Uyuni has towards this area. This means finding all operations Uyuni currently needs from i.e salt. From this set of operations we create an interface.
+- Next step is to implement this interface with the existing code we already have for those operations. This can be done by moving code or delegating in case moving requires a lot of extra refactorings.
+Note: only put methods in the interface that are actually used outside the implementation of the interface. Everything else will be private and will be considered an implementation detail.
+- Then we need to point exising use sites of those operations to use this interface.
+- At this point we should have a pretty good overview of what set of operations we have and we have centeralized them in an interface that is now used. We should be in a state where everything works as before.
+
+
+## Refinement 
+
+- review the interface by consolidation of all the operations, dropping possible duplicates or deprecated functionality, merge similar methods etc.
+- Refine the interface with more precise types, consistent method names, signatures and error handling. (in case of salt there are some wildcard methods that let you do anything which leads to people putting fragile parsing logic inside business logic)
+- Make batch operations the default and single target a special case. This is so implementations can be done with scalability in mind.
+- Merge related existing implementations across the codebase to implement the interface
+- Make sure the interface does not imply a specific implementation (i.e don't reference salt if all you do is gathering information from a system.)
+- Refactor other components to take an implementation of this interface as an argument instead of reaching into global space.
+- write new and reuse existing tests for the interface
+- move interface with related datatypes and implementations into package with fitting name for the component.
+
+### Interface guidelines?
+
+- It should be focused on a particular area and not mix different topics.
+- It should be minimal meaning only the minimal set of methods needed by our business logic.
+- It should give immediately useful information and not burden the user to do much additional work to get to the useful information. (i.e don't return plain json and let business logic parse it).
+- it shoud be precise in its types and minimize invalid usage.
+- Don't be scared to define multiple interface to separate areas properly and let them build ontop of each other.
+
+
+## Bootstrapping
+
+Bootstrapping is about creating our instances of our different "services" and assembling them to then pass on into our business logic.
+The assembly should happen at the very start and then everything below that will only get things passed down but not instanciate any services themselves.
+The Start in this case is kind of multiple places tomcat and taskomatic are differnt processes and have there own entry points so both need to have some bootstrapping code.
+In tomcat the situation is maybe a little bit more complicated since our business logic lives below filters which are instanciated by tomcat logic based on xml configuration.
+
+- One way to go could be looking into programatic initialisation and registration of filters.
+- Use some limited shared global state just to pass the initial instances between entry points.
+- in the worst case duplicating instances per filter so spark and xmlrpc would not share the same instances.
+
+
+
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Refactorings in areas that are not well understood by anyone may be hard and lead to short term breakage.
+- measures will have to be found to ensure established interfaces will continue to be respected.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions


moio:
While I may not necessarily agree 100% on the motivations as they are laid down in this proposal I do recognize the good intent and can agree on the general spirit where this is headed.

Nevertheless, I do have concerns, the main one being that this proposal reads more like an "umbrella" RFC than an actual RFC - being so high-level, I fear it does not convey sufficient concrete detail to really be agreed-upon from a technical standpoint.

(Aside: the purpose of RFCs is not really to produce "statements of political intent", rather, they are a way to submit one's practical idea to solve a given problem and solicit feedback from others before proceeding. Approving/merging means that others do not see significant obstacles/downsides to the proposed text or, if they did, those were considered, documented and accepted with respect to visible alternatives at the time of writing. Whether that particular solution is then actually implemented, and when, are important but separate questions)

I'll provide you some example questions that came to me while reading - not necessarily for you to respond or respond immediately, but to hopefully provide inspiration about how this text could be changed or augmented in order to make it something that can actually be discussed and agreed-upon from a technical standpoint (FMPOV):

 - is this only or mainly about the Java portion of the codebase, or does it also cover Salt internals, Cobbler, the rest of the Python codebase, the Javascript codebase,...?
 - how does this relate to existing interfaces we already have? Eg. [salt-netapi-client](https://github.com/SUSE/salt-netapi-client) for Salt, `CobblerCommand` classes that interface with Cobbler, the frontend API endpoints in the `Router` class consumed by React components, the `TaskomaticApi` class to access Taskomatic from Tomcat...
 - how would such modules exchange data? Are you thinking about an intra-process interface (eg. Java interface/classes), or an inter-process interface (eg. ZeroMQ over sockets, similar to what `salt-master` and `salt-api` use) or a network-ready interface (eg. http: REST, RPC...)
 - what would be the first such component to tackle, why, and how exactly?


Thanks for your efforts

isbm:
@lucidd I would suggest to bring this to the upcoming event live in some sort of presentation/talk instead, and thus discuss it with everyone explicitly.

joesusecom:
Adding to what Silvio has already commented: I think the intention is really good. But in its current form this RFC isn't actionable.
Given the limited resources our project has, I'd also suggest that _if_ we refactor, we always refactor along the lines of a new feature or performance/scalability optimization path that we have to work on anyway.
As Bo has pointed out, this could be a good topic for our next workshop. But let's also keep the community involved by sharing inputs and results on the GitHub project. Not everyone is going to be at the workshop.
While we're at it, let me share some thoughts that go into your direction, but with a bit of a different angle:
In the last couple of months I've thought a lot about how we could _fundamentally_ improve Uyuni, and one of the things that hit me is that as features were added over time, we often mixed three domains that we should have kept more separate, leading to reliability and scalability challenges:

### Realtime monitoring/reporting and task execution ###

Especially since we started using Salt, we added a lot of code that is pretty "transactional" and expects things to be done in realtime. There are lots of examples that prove how that led to bad system design because core Java code would wait for a synchronous Salt call to come back.

In the meantime, we've fixed lots of them, but it would have been better to start with a clear understanding of whether a certain piece of functionality is actually supposed to be executed in realtime, with the requirement of providing realtime feedback to the UI or CLI/API, or is material for a scheduled (batch-style) operation.

### Batch processes with the goal of eventual consistency ###

The batch processing concept is really how most of the old Spacewalk code was designed. It wasn't bad scalability-wise, but of course it had its weirdnesses.

### Regular Long-running tasks like consolidated reporting, content syncing ###

This is all the magic that's happening behind the scenes and often isn't very obvious, thus leading to seemingly strange behavior like CVE reports not being up to date.

## Bottom line ##

I'd suggest that before we even start looking into improving modularity per se, we should look into categorizing features and tasks into those three categories (or similar ones, this is just a starting point).

Just to give you and idea how that may matter:

Really "real-time" monitoring data may end up taking very different code paths than stuff that needs to scale, but doesn't have to be particularly "fast". E.g., Prometheus could be used instead of the main Java/Salt stack.

If we become more aware of the background tasks and how they affect foreground tasks, we could make the user experience much more solid. For example, a channel clone job that's run as part of the content lifecycle management would be blocked from execution (or scheduled for later) if there is an ongoing channel sync with the customer center, or a CVE search would return a warning that informs the user of potentially outdated data if the last update run was hours ago.

Over time, while refactoring the codebase, we would then re-categorize some of those features and, for example, make functionality more asynchronous that caused scalability issues, but doesn't have to be realtime, or refactor batched code into something more synchronous if the expectation for the feature is that really current data matters.

paususe:
What @moio said, and I would suggest a different approach: let's identify a few features that are mostly independent from the rest of Uyuni and try to make them fully pluggable.

First examples that come to my mind: 

- Visualization. Clearly a plugin so that only the interested users install it.
- Systems > Software Crashes. It is only supported on RHEL/CentOS/SLES-ES, so clearly a plugin that should not be installed by default.
- Audit. And I could even go 1 level deeper and ask for the different features to be different plugins:
  - CVE audit
  - Subscription matching
  - OpenSCAP
- Images. Say I want to have an Uyuni Server that I will only use to define and build images. Or that I do not want to build images at all. This plugin should be optional.
- Content Lifecycle Management. The whole thing could be a plugin.
- Configuration management. I may want Uyuni to do the content management and leave configuration management to some other tool. Or the other way around. Clearly a plugin (or even several). But maybe currerently we depend on having the System
- A more radical one: Salt vs traditional stack. Say I want to have an Uyuni Server that will only be able to manage Salt minions. Why should I install all the traditional stack stuff? (This will help in the future when we finally deprecate the traditional stack)
- Virtualization. What if I already have VMware and I'm not interested in Uyuni's virtualization features? Why should I install them. Plugin.

Researching and carving out any of those is an actionable item and will benefit all our users. Some of them (e. g. visualization and software crashes) seem rather easy.

isbm:
@lucidd given that nobody understood what is RFC is all about, I would suggest you to update its narrative so nobody thinks you are talking about plugins... :disappointed:

renner:
This proposal is more for those people who are actually working on the codebase, I guess. As someone who worked with it a lot, who maybe even created some "hard to understand" parts of it, my feeling is that @lucidd has a point in that it might be time for a clean-up. And he is looking to do some really simple things like creating a clean interface that contains all our calls to the Salt API or to the filesystem, simply in order to avoid duplicates or to remove dead code. To ease testing and maintainability. It should have been written like that from beginning on, but of course it grew over time and no one gets to spend time for fixing the structure.

I agree with other commenters that the current text is lacking details about what exactly should be done first and why / how exactly. As I understand the suggested changes would affect only the Java code, *modules* are basically service interfaces and would be initialized on application startup (mentioned as *bootstrap*). Some example code might help here, or a link to a prototypical branch.

At this point I would not subscribe to *modularizing* the whole codebase, but cleaning up some of the most important interfaces, and e.g. transforming the Java packages into a maven style structure, those are things I would have liked to see since years. Without being a huge investment such refactorings could be really beneficial for developers working on the code every day, and in the longer run even to the product quality. From my perspective this effort is more about the low level interfaces, and every attempt to create e.g. pluggable features or frameworks for sync/async execution of tasks would benefit from it as well.

paususe:
> At this point I would not subscribe to _modularizing_ the whole codebase, but cleaning up some of the most important interfaces, and e.g. transforming the Java packages into a maven style structure, those are things I would have liked to see since years. Without being a huge investment such refactorings could be really beneficial for developers working on the code every day, and in the longer run even to the product quality. From my perspective this effort is more about the low level interfaces, and every attempt to create e.g. pluggable features or frameworks for sync/async execution of tasks would benefit from it as well.

I partly agree.

My disagreement comes in the approach: instead of taking a "theoretical" approach in modularizing, I think we will find out what kind of modularization we need / must have / can have with reasonable effort by trying: let's try to convert one feature in a pluggable module and find out what interfaces we discover we need, what rough edges we find, etc. The first one will be easy: everything can be tailor-made for that pluggable module. 

The second one will be hard: we will discover the decisions taken in the first modularization were not that good (no matter what decisions will be made, it always happens :-) ), and some rewriting and redefinitions of those interfaces will be required / discovered.

After a few features are converted to pluggable modules,  we will have the actual guidelines for developers. The ones our code tolerates. I am pretty sure we will need a clean Salt API, a clean filesystem API, etc (in fact, probably several implementations of those, one in each language) and doing the real thing will help us to be pragmatic.

lucidd:
@moio 

> is this only or mainly about the Java portion of the codebase, or does it also cover Salt internals, Cobbler, the rest of the Python codebase, the Javascript codebase,...?

This proposal is generally applicable to any code but my focus is on the java codebase because splitting server functionality between python and java does more harm then good. 

> how does this relate to existing interfaces we already have? Eg. salt-netapi-client for Salt, CobblerCommand classes that interface with Cobbler, the frontend API endpoints in the Router class consumed by React components, the TaskomaticApi class to access Taskomatic from Tomcat...

TaskomaticApi is probably closest to what i'm describing in this RFC with a small exceptions i.e pulling configuration out of global state or using raw types. Given that its already that close it would be an easy target to convert. The bigger problem i remember with our interaction with Taskomatic over this API is that we don't enforce a proper order of operation were we tell taskomatic about an action before we even saved it to the db.

Without having looked much into our usage of cobbler at first glance i can't get an overview of what exactly we are doing with it on a high level. There is the `org.cobbler.*` classes which looks more like a API to me and there are the the `CobblerCommand` classes in `com.redhat.rhn.manager.kickstart.cobbler.*` which uses `org.cobbler.*` and then both seem to be mixed togehter in business logic like `PowerManagementAction`. In short there is a lot of noise making it hard to get a highlevel overview of what functionality uyuni needs and what/how its implemented with cobbler. In comparison with what this RFC suggests we would have high level interfaces describing directly what operations we need i.e `PowerManagement`. Cobbler itself is not accessible to the rest of the application only this high level interface and all cobbler implementation details of providing this functionality are hidden away in the implementation of the interface. So you would have business logic -> high level interface -> low level implementation. The high level interface is where you get your overview, its where you can easily provide "mock" implementations acting a certain way to test different scenarios and its also where you can hook in if you want to provide a different backend implementation.

salt-netapi-client falls into the same category as cobbler. I added a link to an example branch to the RFC which covers salt so you can maybe get some idea there.

Router is an API consumed from our javascript so yes you could to the same here in javascript. Like i said i focus only on the java part for now since you already lose benefits going cross language because you can't share code/types/interfaces and has to do everything from scratch and keep it in sync over time.

> how would such modules exchange data? Are you thinking about an intra-process interface (eg. Java interface/classes), or an inter-process interface (eg. ZeroMQ over sockets, similar to what salt-master and salt-api use) or a network-ready interface (eg. http: REST, RPC...)

Its just java interface/classes and simple dependency management via constructors (no access to global state) that's pretty much the core of this RFC.

> what would be the first such component to tackle, why, and how exactly?

Like i said there is an example branch for doing this with salt usage which is a pretty big mess and can really use some simplifications. In addition to that it helps towards making uyuni more independent of salt. But as you pointed out TaskomaticApi it seems like an easy candidate to look at.

lucidd:
@joesusecom 
> Given the limited resources our project has ...

This RFC is pretty much about freeing up resources that are currently wasted on simple task that drag on way to long because of the complexity of our codebase.

> I'd also suggest that if we refactor, we always refactor along the lines of a new feature or performance/scalability optimization path that we have to work on anyway.

That has been the directive so far and what lead to this mess in the first place.

joesusecom:
@paususe, you're proposing a pretty radical approach. :-)

Why? Because as of today we rely so much on the one database with its one schema that ripping out many of the items you've proposed will be pretty challenging. But I think it's ultimately the right thing to do. Once we even start considering ripping those tightly integrated features out, we'll start thinking differently on how we use the database, apply schema upgrades, etc. etc.

>     * Visualization. Clearly a plugin so that only the interested users install it.

Actually, this one should be easy because it already relies on a separate (undocumented) API endpoint that we only use for the visualization feature.

That said, the _original_ design proposal from PM for this wasn't to hide it away under its own menu item, but make it an alternative view/tab in the system overview just like the lists views we use. So you'd be able to apply filters, store them, and go back and forth between looking at graphical views and list views as needed.

So in a way, doing this right would require it to be a different _kind_ of plugin that introduces its own dimension of pluggability: A view that can take the data that comes from a filterable API that provides node information and represents it in different ways.

Such views should also allow for additional "data plugins" to get them enriched (e.g. with monitoring status data coming from Prometheus).

There's some prior art for that kind of tech (e.g. our own Stratos project and https://github.com/weaveworks/scope).

>     * Systems > Software Crashes. It is only supported on RHEL/CentOS/SLES-ES, so clearly a plugin that should not be installed by default.

+1

>     * Audit. And I could even go 1 level deeper and ask for the different features to be different plugins:
    
>       * CVE audit

A bit of a stretch goal. :-)

>       * Subscription matching

That's another good candidate as it's pretty modular already.

Actually, _within_ subscription matching there are at least two more dimensions of "pluggability":

* node data providers (AKA Virtual Host Managers); those would overlap with what we'd require for the visualization feature
* pluggable rules (e.g. if customers or 3rd party software providers want to extend SUSE Manager to track other software licenses, or if we want to ship customer-specific rules for customers with special deal contracts)

>       * OpenSCAP

Again a good candidate.

>     * Images. Say I want to have an Uyuni Server that I will only use to define and build images. Or that I do not want to build images at all. This plugin should be optional.

Another example of a feature comes with its own dimensions of pluggability:

* Different build implementations (Kiwi vs. Dockerfile)
* Different build schedulers (e.g. use physical host, auto-spawned "build containers", VMs, or a full-blown OBS instance)
* Different template management backends (Git!)
* Different image stores (container registries, plain directories on disk, OpenStack Glance-style image libraries, Artifactory, ...)

>     * Content Lifecycle Management. The whole thing could be a plugin.

Not sure whether _that's_ a good candidate from a UI POV. I'd consider Content Lifecycle Management _the_ core feature of SUSE Manager.

But of course we can still build it in a very pluggable way (again with several dimensions like making the filter rules easy to extend or adding additional aspects like config file staging).

>     * Configuration management. I may want Uyuni to do the content management and leave configuration management to some other tool.

Yes and no.

On the one hand, we _do_ have demand from customers to make it very easy to use SUSE Manager together with an existing Salt (or other configuration management) setup, where Manager would mainly (or only) be used as the repository manager.

On the other hand, there are things we need to do _with_ content management, like setting up the repos.

By making Salt our configuration and client communication engine/stack (as opposed to having the  traditional stack that didn't really get into the way of configuration management, unless you explicitly _enabled_ and _used_ the config management features), we changed the picture a lot.

This is actually one of the points where we may not have done the best job with the original design:

* We didn't ensure from day one that customers (or other product teams) can use _one_ Salt Master for both what they want to do and what SUSE Manager needs to do via Salt.
* In hindsight, maybe we should have kept the "traditional stack" around on Salt minions until we've really reached feature parity.

One of the reasons why this didn't work out perfectly is that we didn't have a clear understanding on whether we wanted Salt to be an implementation detail (in which case we probably should have assumed that Manager runs its own dedicated Salt minion that doesn't interfere with other people's Salt minions) or make SUSE Manager a generic Salt Master with benefits.

Now that lots of customers are starting to use more of Salt than just the built-in features in SUSE Manager, we really need to fix this.

> Or the other way around. Clearly a plugin (or even several). But maybe currently we depend on having the System

>     * A more radical one: Salt vs traditional stack. Say I want to have an Uyuni Server that will only be able to manage Salt minions. Why should I install all the traditional stack stuff? (This will help in the future when we finally deprecate the traditional stack)

This makes sense.

Let me add (in the context of what I've written above) that we may even want to go one step further and try to build a SUSE Manager that can work with a completely separate Salt Master that's much more loosely coupled than it is today.

>     * Virtualization. What if I already have VMware and I'm not interested in Uyuni's virtualization features? Why should I install them. Plugin.

That's even a paid-for feature. _And_ it's being developed by a different team. So yes, definitely plugin material.
 
> Researching and carving out any of those is an actionable item and will benefit all our users. Some of them (e. g. visualization and software crashes) seem rather easy.

Very good points. Now, the challenge is to prioritize the areas that matter most.

joesusecom:
> Without having looked much into our usage of cobbler at first glance i can't get an overview of what exactly we are doing with it on a high level. There is the `org.cobbler.*` classes which looks more like a API to me and there are the the `CobblerCommand` classes in `com.redhat.rhn.manager.kickstart.cobbler.*` which uses `org.cobbler.*` and then both seem to be mixed together in business logic like `PowerManagementAction`. In short there is a lot of noise making it hard to get a highlevel overview of what functionality Uyuni needs and what/how its implemented with cobbler.

Cobbler is a good example. What I've _heard_ (without looking at the code myself) is that it would be really hard to replace or just uninstall Cobbler because of all the dependencies it has into many core parts of the codebase. If that's true, it would certainly show a lot of design technical debt, because as a feature, bare metal deployment _shouldn't_ obviously have any deep dependencies into the rest of the system, and the actual bare metal deployer (Cobbler, MaaS, plain Salt like RBS is doing it) should be pluggable.

> In comparison with what this RFC suggests we would have high level interfaces describing directly what operations we need i.e `PowerManagement`. Cobbler itself is not accessible to the rest of the application only this high level interface and all cobbler implementation details of providing this functionality are hidden away in the implementation of the interface. So you would have business logic -> high level interface -> low level implementation. The high level interface is where you get your overview, its where you can easily provide "mock" implementations acting a certain way to test different scenarios and its also where you can hook in if you want to provide a different backend implementation.

Exactly. BTW, the PowerManagement example is a good one, because what we're using here is a part of Cobbler that actually is a wrapper around the (already highly abstracted) fencing agent stack. There's little value added by using those _through_ Cobbler.

> > how would such modules exchange data? Are you thinking about an intra-process interface (eg. Java interface/classes), or an inter-process interface (eg. ZeroMQ over sockets, similar to what salt-master and salt-api use) or a network-ready interface (eg. http: REST, RPC...)
> 
> Its just java interface/classes and simple dependency management via constructors (no access to global state) that's pretty much the core of this RFC.

That part really didn't get too obvious from reading the original draft. :-)

And I still think that, while those Java-internal refactorings are valuable on their own, we should tackle the big ones first (e.g. making sure you can use Salt and the Java/Tomcat stack in different nodes/containers with a clearl API/contract). Where the Java code already touches such external systems like Salt or Cobbler, the code refactoring will come as part of the "subsystem refactoring".

> > what would be the first such component to tackle, why, and how exactly?
> 
> Like i said there is an example branch for doing this with salt usage which is a pretty big mess and can really use some simplifications. In addition to that it helps towards making Uyuni more independent of salt. But as you pointed out TaskomaticApi it seems like an easy candidate to look at.

Both look like good candidates. The Salt one is probably more important though.

joesusecom:
> > Given the limited resources our project has ...
> 
> This RFC is pretty much about freeing up resources that are currently wasted on simple task that drag on way too long because of the complexity of our codebase.

But you first have to put _more_ time into the refactoring, which will mean less features short-term, which we can't really afford.

Now what if the code you've just refactored is dropped later because we don't need it any more?

Starting any refactoring where you're already touching code avoids that.

> > I'd also suggest that if we refactor, we always refactor along the lines of a new feature or performance/scalability optimization path that we have to work on anyway.
> 
> That has been the directive so far and what lead to this mess in the first place.

I don't agree on that one. Refactoring along the lines of the code we are touching anyway doesn't mean _not refactoring_. It just means that you start with touching the parts that you need to fix most urgently.

Where we've added spaghetti code in the past was when there was _no_ refactoring, just gluing new stuff to the old code. I also fear that the code review wasn't always rigid enough. Code reviews should not just check whether the code works as expected, but also whether it introduces extra dependencies, risks, or side effects that aren't obvious at a first glance.

Or to put it differently: It doesn't matter much if code that you don't have to touch because it works is hard to maintain. Refactoring such code is just a waste of time. You can consider it a black box until you either throw it away or _have_ to touch it anyway (because you need to extend it or because someone hit a bug).

_If_ you need to touch code anyway, with the goal of not just adding stuff, but also refactoring things properly, and then one thing leads to another, and you end up having to clean up a whole subsystem, so be it. In that case you just hit one of the codepaths that need refactoring most urgently.

isbm:
While @paususe and @joesusecom having nice conversation, I still can't see the wood for the trees. But @renner nailed it down par excellence, hence I would suggest @lucidd stick to it and prepare a short talk/presentation for the summit, thus let everyone of our team express their concerns.

lucidd:
@paususe @joesusecom 

Just to respond to the ongoing sub discussion about plugins. This RFC is not about this its not about plugins nor about microservices. Yes this clean separation of components will make things like plugins or microservices easier to do in the future but that's not an immediate goal of this RFC.

> But you first have to put more time into the refactoring, which will mean less features short-term, which we can't really afford.

Yes it means less features short term but more more resources and faster development long term. 

> Now what if the code you've just refactored is dropped later because we don't need it any more?

- We rarely drop any code let alone entire subsystems. We still tag along code for products that are long out of support.
- Good. Dropped code is efficient code.
- During those refractorings you will encounter common problems and patterns that are transferable to other areas so even if the code you refactor does not survive, the lessons learned will in other components.
- Even during the small incomplete refactoring i did for the example branch i found dozens of IDE warnings, unused methods and test with mocking that's not unused.


I think @renner already put it best with saying that his is an RFC for people who work on the code because only then you will feel those pains.

dmacvicar:
> @lucidd given that nobody understood what your RFC is all about, I would suggest you to update its narrative so nobody keep thinking you are talking about plugins... disappointed
> 
> @joesusecom this is a very good topic for the winter summit in few weeks, technical part. This is not about features or plugins, but about the way we code and test... @lucidd should just be more verbose and precise, I believe.

I was probably the original culprit of bringing the topic of plugins into the discussion, but I explained in length the reason already, and what it has to do with how we code and test.

dmacvicar:
> > Starting any refactoring where you're already touching code avoids that.
> 
> In theory yes but that not what happens. If you work on feature X then that is the priority and usually has to be shipped at a certain time. Everything else is sidelined and second class. There won't be much effort put into refactorings and they are usually kept minimal in this mode. No major improvements or redesigns are gonna come out of this. Its more likely that some hacky workaround is chosen then a refractoring that would take time.

But that is the moment to suggest a concrete refactoring and action, and schedule it appropriately.

Which is exactly the main problem we always have moving this topic forward. It seems to either go into hopelessness to suggest we assign more people to dig the rabbit hole and start improving parts without a clear definition of improving and which parts.

I still think the RFC goes into the right direction.

> pick one of the areas like: salt, filesystem, database, task scheduling, different topics of buisness logic like cve audit, etc.

The first step I would take is to document the current architecture. I call this "emerging it" because it is sink in the bottom of the sea. Document the different parts that we can identify, and document the main motivations for certain decisions as we understand them from our own experience. The benefits the current architecture provides, as we have benefited from some decisions as well.

That architecture and technical documentation needs to then be evolved as we change decisions. The areas you mention in the RFC are at the end services consumed by different parts of the applications, and with an initial documented architecture we could identify those missing or misplaced services. Example: filesystem, which does not bring you much value to think about it at the Java API level, as we may need to consume files from different services.

>The first step is to gather all the requirements Uyuni has towards this area. This means finding all operations Uyuni currently needs from i.e salt. From this set of operations we create an interface.
>Next step is to implement this interface with the existing code we already have for those operations. >This can be done by moving code or delegating in case moving requires a lot of extra refactorings. >Note: only put methods in the interface that are actually used outside the implementation of the interface. Everything else will be private and will be considered an implementation detail.

How is that different than today? There is a list of Salt functionality that Uyuni consumes, and methods exposed to it. Why can't we extract today an interface from what it is used?.

>Then we need to point exising use sites of those operations to use this interface.
>At this point we should have a pretty good overview of what set of operations we have and we have centeralized them in an interface that is now used. We should be in a state where everything works as before.

That sounds totally valid. I hope we can bring the language of the discussion one level up from the source code, as lot of the interfaces in the architecture conceptually should be similar if they were Java APIs, remote APIs, etc.

I support your approach, but in summary I would add:

* Creating a living document for the current architecture and technical details/design
* Identify areas and services that are used between components.
* Chose one area, as you described, where we need to do some immediate work
* Create an RFC to modularize that specific area, refactoring the necessary services while at it

And that is probably the main point I want to make. I don't think you can abstract Salt in a single shot. You can abstract Salt when you modularize one area that depends on Salt, but that abstraction will not be perfect, and will need a revisit when modularizing the next area that depends on Salt.

dmacvicar:
what would be a project out there that shares some similarities with our project on some dimension (target audience, technology, size), which you would consider a role model in terms of their architecture, technical design process, technical documentation and continuous refactoring?

isbm:
> I was probably the original culprit of bringing the topic of plugins into the discussion, but I explained in length the reason already, and what it has to do with how we code and test.

Actually, what you were saying about plugins (Confluence) absolutely makes perfect sense. However I personally find it quite separate topic, as well as being ahead of its time (sadly). So **this RFC** is exactly to bring more features in a near future but dramatically reduce the sheer load of bugs we are getting. And @lucidd has a solution how to do that.

Since we want plugins-based architecture — sure! But before we even start getting there, @lucidd 's RFC should by no means go first and prepare basic ground "autobahns" for plugin-based logistical ideas. Otherwise we have a very high impact risk of ending up with a very bad ad-hoc code.

Besides, I see the communication problem here. The issue is actually three-dimensional:

- Java codebase in terms of classes/ interfaces/ abstractions.
- Python codebase in terms of package components and plugin interfaces.
- All SW components of Uyuni in a whole as tools and how they are interact between.

Again, to boost our team delivery performance, I would strongly advice @lucidd to bring this topic well prepared to a technical part of our winter summit and discuss it there with everyone focusing on how to get it practically done, who would participate and how to actually organise that. I would not mind to join, refreshing my Java skills. :wink:

lucidd:
> How is that different than today? There is a list of Salt functionality that Uyuni consumes, and methods exposed to it. Why can't we extract today an interface from what it is used?.

That's essentially the step i started with in the example branch. Basically finding the usage of raw callSync and callAsync usages throughout the codebase and moving them behind a properly named method in an interface. This is one of the first and easiest steps since consolidation and refinement are put in a separate step. Its still not trivial in the case of salt since our interaction with salt spans quite some areas. We have the netapi interaction with salt and salt-ssh which both have evolved differently due to subtle differences in behavior. We interact with salt though the filesystem by writing pillar, state and roster files. And we consume salt events from our database.

accepted/00065-modularisation.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+The codebase of Uyuni has become increasingly hard to understand and maintain. The majority of time is spent on fixing bugs, debugging, fiddling with complex test setups and simply trying to understand what existing code even does (this includes new code). Even the most knowlegable people on the team don't always know how Uyuni would behave in a certain scenario or how its supposed to behave. Those problems are symptoms of the same underlying issue which is only adding new features without any big picture or architecture in mind and the complexity and lack of quality resulting from that. This creates a feedback loop where too much busywork leaves not enough time for proper reseach and development leading to rushed features leading to more bugs, complexity and busywork. This proposal is an attempt at breaking this loop by investing into a maintainable codebase that requires less busywork and allows for more time on the things that matter.
+
+# Detailed design
+[design]: #detailed-design
+
+The core of this proposal will be modulerizing our codebase driven by the following aspects:
```
mbologna: Let's talk about design: do we refer to the module and their interactions as in the mediator pattern? How would the modules talk to each other?

accepted/00065-modularisation.md
```diff
@@ -0,0 +1,90 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Refactorings in areas that are not well understood by anyone may be hard and lead to short term breakage.
+- measures will have to be found to ensure established interfaces will continue to be respected.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
```
mbologna: How would the traditional (legacy) stack fit into this picture? Would it be still there in your imagined scenario?

isbm: @mbologna I guess, something like that:

```java
interface InterfaceName {
    
    // declare constant fields
    // declare methods that abstract 
    // by default.
}
```

😉 

Imagine you want to un-install that traditional thing away. So you would `zypper rm uyuni-traditional-connector` or something like that. But before that, you probably would like to crawl across the whole codebase, rip off all traditional-related bare methods and random classes, then place them behind some common interface, so they became an implementation detail.

Which is the same as the rest of Kevin's proposal. So I don't see it as an "unresolved question". But let's re-ask this tomorrow at his session maybe?

accepted/00065-modularisation.md
isbm: No, it is barely at that level yet. For now the idea is to stop creating a spaghetti and switch to lasagna complexity.

accepted/00065-modularisation.md
mbologna: > Imagine you want to un-install that traditional thing away. So you would `zypper rm uyuni-traditional-connector` or something like that. But before that, you probably would like to crawl across the whole codebase, rip off all traditional-related bare methods and random classes, then place them behind some common interface, so they became an implementation detail.
> 
> Which is the same as the rest of Kevin's proposal. So I don't see it as an "unresolved question". But let's re-ask this tomorrow at his session maybe?

Is it still unresolved: while doing the refactoring, should we keep the traditional stack or should we drop it? It sounds like, from your proposal, that you want to keep it. But I haven't got Kevin's POV from the RFC text.

accepted/00065-modularisation.md
mbologna: What would be the interface between modules then? Are the modules going to be talking to each other using a well-defined protocol or using a pattern?
I think it's not a low-level detail but rather a design decision that has to be taken better earlier than later and specified in the RFC.

accepted/00065-modularisation.md
isbm: > should we keep the traditional stack or should we drop it?

I am not sure if this is relevant to the current RFC.

chiaradiamarcelo:
My 2 cents:
There are 2 main concepts on the broad picture of OOP (actually on any programming paradigm that uses some kind of module structures) which are cohesion and coupling.
Cohesion is a measure of how related are the components inside a module, coupling is a measure of how tight is the relation between modules. When modules are tightly coupled one to each other, making a small change for fixing a bug/adding a new feature usually implies editing many parts of the code, in different modules, which is more error prone than just touching one class and call it a day, which is the case if the classes wouldn't be so dependent one from each other as it is right now.

IMHO, currently our code base is in a bad shape, many classes present no cohesion at all, meaning they are in charge of doing multiple things, meaning a fix on one of those things might break the flow of the others, and many classes are highly coupled one to the other, meaning a change in one of them requires touching many components. This on top of the wasted time that takes to understand what some parts of the code actually do, and the quite long curve for someone new to get familiar with it.

Again IMHO, what Kevin is proposing is nothing fancy, it is just common sense, and it's about re-arranging the code in order to put together those components that should work together inside a module (cohesion) and abstract the functionality of classes for "de-coupling" the modules, will help to better code, less error prone and less time consumed to understand what the code actually does.

chiaradiamarcelo:
What is the current status of this? Were there any agreements regarding this RFC during the winter summit that I missed?

renner:
I am closing this PR in agreement with the author (@lucidd). His currently ongoing effort to create an internal interface for interactions with Salt (https://github.com/uyuni-project/uyuni/pull/1954) will continue as sufficient communication has happened and there was agreement between the developers working on the codebase. The current text has been [transcribed and moved into the wiki](https://github.com/uyuni-project/uyuni/wiki/Modularization).

Any future refactoring efforts with a bigger impact on the codebase will as well require consensus between the community and codebase owners. Publishing an RFC and finding agreement on the PR might still be appropriate in many cases.

PR #30, opened 2020-01-27
RFC: automated Ubuntu onboarding without manual preparation

mantel:
https://github.com/uyuni-project/uyuni-rfc/blob/master-ubuntu-sudo/accepted/0000-ubuntu-sudo.md


accepted/0000-ubuntu-sudo.md
```diff
@@ -0,0 +1,63 @@
...
+Onboarding Ubuntu minions
+
+- What is the expected outcome?
+
+Ubuntu clients should be as easy to onboard as any other client. This means it should be sufficient to just enter the root password in the SUSE Manager UI without also manually configuring the to-be-onboarded Ubuntu minion.
+
+# Detailed design
+[design]: #detailed-design
+
+When onboarding a minion via ssh, SUSE Manager asks for a user and password with sufficient privileges. On non-Ubuntu systems this typically is the root user. On Ubuntu systems, there typically is just an ordinary user, but this user can do almost anything via sudo. This user is referred to as "ubuntuser" in this document. Unfortunately it is necessary to supply the user's password to sudo. So currently the SUSE Manager documentation asks the admin to enable some special commands in sudo to be executed even without providing the password. The goal is to do this modification directly from SUSE Manager, so there is no need to manually prepare the system for onboarding.
```
moio: Do I get it correctly that you are addressing the onboarding from the Web UI only in this RFC?

(ie. onboarding via manual execution of a bootstrap script and onboarding via key acceptance do not really change)

accepted/0000-ubuntu-sudo.md
```diff
@@ -0,0 +1,63 @@
...
+
+So the actual workflow would be like this:
+
+SUSE Manager UI asks for user and password like with any other client. The password needs to get stored in some local temporary file, so it can be transferred to the minion without being visible in the process list:
+
+1. umask 0377 && echo $PASSWORD > /run/usr/$UID/pass.txt
+2. scp -p /run/usr/$UID/pass.txt ubuntuser@ubuntu-minion.target.system:/run/usr/$UID/pass.txt
```
moio: I understand the idea behind those commands, but I wonder how we would actually implement something like that. At the moment Java code does not shell out to ssh, rather we call the Salt API to drive Salt SSH (to apply some states).

How would you imagine sneaking in those commands (or something else to a similar effect) via Salt SSH, and how would Java code drive that?

accepted/0000-ubuntu-sudo.md
mantel: Yes, this is only about bootstrapping via the UI. If the user is logged into the minion and manually executes the bootstrap script, there is not much to win anyway as he can also give herself the needed privileges.

accepted/0000-ubuntu-sudo.md
mantel: Oh, I was not aware of the actual current implementation. But isn't there some mechanism we can use? How is the onboarding performed in cases where no salt agent is installed? So far this onboarding mechanism was a black box to me, maybe Matei can comment as he implemented it initially.

paususe: > I understand the idea behind those commands, but I wonder how we would actually implement something like that. At the moment Java code does not shell out to ssh, rather we call the Salt API to drive Salt SSH (to apply some states).
> 
> How would you imagine sneaking in those commands (or something else to a similar effect) via Salt SSH, and how would Java code drive that?

I'm not sure I understand what the problem is.

There are a number of Java libraries providing an SSH implementation. If we are not using any of them, we should pick one and add it to our dependencies. E. g. here is an example of what we'd need, implemented with JSch (pure Java library, BSD-licensed)
https://medium.com/@ldclakmal/scp-with-java-b7b7dbcdbc85
http://www.jcraft.com/jsch/

moio: What we do from the UI is to use salt-ssh (aka masterless Salt) to apply a state which does the same things a regular bootstrap script does.

So when it comes to minions, Java code knows nothing about the SSH layer, all it does is to delegate to Salt via the usual Salt APIs. I am not sure we want to introduce anything jsch based on the Java side to go straight to the minion before handing control over Salt SSH - it sounds like a layering violation to me.

@mateiw would it be possible to still use Salt SSH to apply a first state that configures sudo, followed by our existing states to bootstrap everything else?

@paususe the class you mentioned implements part of the SSH Push client protocol, aka one of the possible modes of the traditional stack. Specifically traditional cliends have a (pull-based) agent service by default, and on top of that we developed the so-called "SSH Push" mode to deal with those cases when an active agent is not acceptable, but plain SSH is an option.

paususe: > What we do from the UI is to use [salt-ssh](https://docs.saltstack.com/en/latest/topics/ssh/) (aka masterless Salt) to apply a state which does the same things a regular bootstrap script does.
> 
> So when it comes to minions, Java code knows nothing about the SSH layer, all it does is to delegate to Salt via the usual Salt APIs. I am not sure we want to introduce anything jsch based on the Java side to go straight to the minion before handing control over Salt SSH - it sounds like a layering violation to me.
> 
> @mateiw would it be possible to still use Salt SSH to apply a first state that configures sudo, followed by our existing states to bootstrap everything else?

What would be the advantage of using Salt SSH or JSch? In the end, we need no Salt feature here.
 
> @paususe the class you mentioned implements part of the SSH Push client protocol, aka one of the possible modes of the traditional stack. Specifically traditional cliends have a (pull-based) agent service by default, and on top of that we developed the so-called "SSH Push" mode to deal with those cases when an active agent is not acceptable, but plain SSH is an option.

Thanks for the information but I only mentioned that class because it already uses JSch, so it's not like we are introducing a new dependency here.

moio: > What would be the advantage of using Salt SSH or JSch? In the end, we need no Salt feature here.

Avoiding to add yet another new completely separate code path to achieve the same goal: running an arbitrary command via SSH.

paususe: > > What would be the advantage of using Salt SSH or JSch? In the end, we need no Salt feature here.
> 
> Avoiding to add yet another new completely separate code path to achieve the same goal: running an arbitrary command via SSH.

What about efficiency?

JSCh => salt-ssh (involves copying Salt) => run 1-liner shell script that requires nothing provided by Salt

vs

JSch => run 1-liner shell script that requires nothing provided by Salt

I don't understand using salt-ssh, which we do not need at all, for such a simple operation.

accepted/0000-ubuntu-sudo.md
```diff
@@ -0,0 +1,63 @@
...
+
+Now that the password exists on the minion, the needed modifications to the sudo configuration can be done from the SUSE Manager server by running the following ssh command on the minion:
+
+3. sudo -S sh -c "umask 0377 && echo 'ubuntuser ALL=NOPASSWD: /usr/bin/python, /usr/bin/python2, /usr/bin/python3' > /etc/sudoers.d/00001-susemanager" < /run/usr/$UID/pass.txt
+
+Now the system can be boostrapped like any other client because ubuntuser can do all the needed steps via sudo without password. Finally both copies of the file containing the password are deleted:
+
+4. rm -f /run/usr/$UID/pass.txt && ssh ubuntuser@ubuntu-minion.target.system rm -f /run/usr/$UID/pass.txt
```
paususe: Actually, it will probably be more efficient like this (it will save the second ssh connection to the client)

3. sudo -S sh -c "umask 0377 && echo 'ubuntuser ALL=NOPASSWD: /usr/bin/python, /usr/bin/python2, /usr/bin/python3' > /etc/sudoers.d/00001-susemanager" < /run/usr/$UID/pass.txt && sudo rm -f /run/usr/$UID/pass.txt

4. rm -f /run/usr/$UID/pass.txt

accepted/0000-ubuntu-sudo.md
moio: We are using salt-ssh anyway for the rest of the onboarding process, so performance wise copying and all is a sunk cost FMPOV.

paususe: According to the salt-ssh docs, NOPASSWD in sudoers is required for salt-ssh to work:
https://docs.saltstack.com/en/latest/topics/ssh/

That 1-liner is what adds NOPASSWD to the sudoers file. 

Will we have a chicken and egg problem here if we try to run it with salt-ssh? 

Or can we use "salt-ssh - r" , set "sudo: False" in the roster file and then make sudo part of the command (ie what the RFC contains)? 

Also, I can't remember at the moment if logging out if required to enable the changes in the sudoers file (and I cannot test atm)

accepted/0000-ubuntu-sudo.md
mantel: > Actually, it will probably be more efficient like this (it will save the second ssh connection to the client)

I do not think efficiency is relevant at all in this particular case. Onboarding is a one-time thing and there are many operations that will consume much more time than one single additional ssh command.

accepted/0000-ubuntu-sudo.md
moio: That was what I had in mind to research about, thanks for starting it.

I plan to give the topic an in depth look from this angle as soon as time allows.

paususe: Why you? Is this scalability-related?

moio: It's not in my area, but I do try to take a look at all incoming RFCs in general and help where I can :-)

paususe: @meaksh Could you please shed some light?

paususe:
@mantel 
After discussion today, it seems this can be written as a Salt state and transfer the password in a pillar. Could you please modify the RFC accordingly?

moio:
@mateiw would it be possible to have a small state with the equivalent of the following two lines:

https://github.com/uyuni-project/uyuni-rfc/pull/30/files#diff-6a014a07f2bdea99da8fd9824fa3a869R37-R38

and apply it, possibly only for Ubuntu minions, as part of the bootstrap mods, ie.

https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/controllers/utils/RegularMinionBootstrapper.java#L70-L74

so that we could remove the manual instructions here?

https://opensource.suse.com/doc-susemanager/suse-manager/client-configuration/clients-ubuntu.html#_root_access

mateiw:
> @mateiw would it be possible to have a small state with the equivalent of the following two lines:
> 
> https://github.com/uyuni-project/uyuni-rfc/pull/30/files#diff-6a014a07f2bdea99da8fd9824fa3a869R37-R38
> 

Yes, I did some quick and dirty testing and it seems to be feasible. Both creating the passfile on the client and creating the sudoers file could be done using a salt state.
It's even easier because `salt-ssh` can take care under the hood of supplying the password to `ssh`. If using only plain `ssh` the pass needs to be supplied using tricks like `expect` or `sshpass`.

paususe:
> > @mateiw would it be possible to have a small state with the equivalent of the following two lines:
> > [uyuni-project/uyuni-rfc/pull/30/files#diff-6a014a07f2bdea99da8fd9824fa3a869R37-R38](https://github.com/uyuni-project/uyuni-rfc/pull/30/files#diff-6a014a07f2bdea99da8fd9824fa3a869R37-R38)
> 
> Yes, I did some quick and dirty testing and it seems to be feasible. Both creating the passfile on the client and creating the sudoers file could be done using a salt state.
> It's even easier because `salt-ssh` can take care under the hood of supplying the password to `ssh`. If using only plain `ssh` the pass needs to be supplied using tricks like `expect` or `sshpass`.

@mantel can you please work with @mateiw to update this RFC and maybe even include Matei's quick hack for reference?

paususe:
I wonder if this can be implemented more easily using Salt SSH pre-flight as implemented in Salt 3001:
https://docs.saltstack.com/en/master/topics/releases/3001.html#ssh-pre-flight

@brejoc ?

moio:
Dear contributors, I am closing this PR in the interest of keeping our repo clean due to inactivity.

The discussion is still completely valid though and it can be reopened anytime there's an interest.

Thanks for the collaboration so far!

PR #31, opened 2020-01-30
Selective blocking of minion actions

mateiw:
The goal is to selectively block certain action types from being executed on a Salt minion.

[RFC text](https://github.com/uyuni-project/uyuni-rfc/pull/31/files?short_path=8fcaa98#diff-8fcaa9826254d01ebdc04d65ec1ac38e)


accepted/00000-minion-selective-action-blocking.md
```diff
@@ -0,0 +1,139 @@
...
+   Column    |           Type           
+ ------------+--------------------------
+  server_id  | numeric                  
+  locker_id  | numeric                  
+  reason     | character varying(4000)  
+  created    | timestamp with time zone 
++ profile_id | numeric nullable references rhnLockProfile(id)
+
+```
+The new column `rhnserverlock.profile_id` will reference the locking profile to be used. If it's `NULL` the global configuration from `rhnActionType.unlocked_only` will be used.
```
moio: What advantages or disadvantages you see with the choice of having such profiles defined in the database vs. in code?

accepted/00000-minion-selective-action-blocking.md
```diff
@@ -0,0 +1,139 @@
...
+2. Feature parity with traditional clients - use global locking configuration that's currently defined in table `rhnActionTypes`
+3. Enhance db schema to add locking profiles
+4. Disable UI elements and API calls depending depending on action type when a minion is locked
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+In some cases it's not fined grained enough. Some actions might be forbidden only in certain cases. E.g. in the case of CaaSP, package operations are allowed as long as they don't impact the CaaSP packages.
+
+Such cases need a different solution that could be complementary to system locking, e.g. package locking.
```
moio: IIUC the only real solution to that is zypper-level package locking on clients, correct?

accepted/00000-minion-selective-action-blocking.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Salt level selective action blocking
+
+The Salt blackout mechanism works by defining a special pillar called `minion_blackout` to tell the minion that it must reject all remote executions. Exceptions can be configured using the `minion_blackout_whitelist` pillar. This accepts a list of Salt functions that are allowed during blackout, e.g. `test.ping`, `state.apply`, etc.
+
+However, in Uyuni there are multiple actions that use `state.appply` under the hood. Some actions can change the state of the system while others are read-only but they all use `state.apply`. If `minion_blackout_whitelist` is used to allow `state.apply` then all actions that use this function are allowed.
+
+The blackout mechanism must be enhanced to allow a more fine grained filtering. One approach would be to check the metadata that can be attached to a Salt call. 
+
+A metadata whitelist would be defined. If the Salt call contains metadata that matches the whitelisted metadata then the call would be allowed.
```
moio: @brejoc, is this big enough to warrant a Salt RFC from your perspective, or we just go ahead with a PR?

accepted/00000-minion-selective-action-blocking.md
mateiw: I think both approaches are valid. For the initial implementation we could maintain the profile in code. In the future, with a plugin system, the profile could be defined in code by the plugin. 
OTOH having the profile in the DB is consistent with the current implementation, it makes the profile easier to change and we could even create some UI/API to allow an admin user to define what actions are allowed on certain machines for certain users.

accepted/00000-minion-selective-action-blocking.md
mateiw: Right, so far I don't see any other solution.

accepted/00000-minion-selective-action-blocking.md
moio: My personal opinion is that we have far too many tables which are actually constants, and constants that the code needs to know about anyway, so in those cases they should not really be in the DB at all, following the principle that DBs are for variable data, not programming enums.

That said, I can see the case of having a table here if it is actually a good choice: fed via a product-specific plugin or maybe SCC (I would not particularly trust users for this use case). Until we have such variability, I would still recommend keeping this in code.

I agree with you this is inconsistent with some of the mechanisms we have elsewhere, but IMO we should rather refactor those and get rid of some tables (eg. `rhnActionStatus`, `rhnActionType`, `rhnChecksumType`...) - or at least avoid adding more. Having the same constant in the DB and code is just more maintenance burden without a clear advantage FMPOV.

mateiw: That's fine with me. Let's keep this in code for now. We can always add tables later if there's a need for it.

accepted/00000-minion-selective-action-blocking.md
```diff
@@ -0,0 +1,93 @@
+- Feature Name: minion_selective_action_blocking
+- Start Date: 2020-01-21
+- RFC PR: (leave this empty)
```
mbologna: nit: link the RFC URL

accepted/00000-minion-selective-action-blocking.md
```diff
@@ -0,0 +1,93 @@
...
+* Feature parity with traditional clients. System locking is available for traditional clients but no such functionality exists for Salt minions.
+
+Usability:
+* Salt has a blackout mechanism. This is used for the first version of CaaSP support but it's not flexible enough for Uyuni's use case. Uyuni makes heavy use of `state.apply` to execute actions. Salt blackout allows whitelisting which modules are allowed during blackout. In the case of Uyuni, whitelisting `state.apply` means allowing basically all actions. Additionally the user experience of using Salt blackout is not great, the user can schedule any actions from the UI/API but then gets an error message at execution time that the action is not allowed by Salt.
+
+
+# Detailed design
+[design]: #detailed-design
+
+Traditional clients can be locked by the user. When locked, only certain action types are allowed. Actions that change the state of a system are blocked (package install, upgrade, apply highstate, etc) while read-only actions are allowed (hardware refresh, package profile refresh, etc).
```
mbologna: There is one notable example that works for traditional systems even when locked: remote commands.
Do you think we should enable it for minion system locked as well? In sake of feature parity, I would say we should mimic the same behavior.

accepted/00000-minion-selective-action-blocking.md
```diff
@@ -0,0 +1,93 @@
...
+
+The same should be possible for Salt minions. However, in the case of Salt minions, additional flexibility is needed to allow blocking some actions depending on the the products installed on the minion. This is needed to accommodate CaaSP and maybe other clustering products in the future.
+
+To prevent the user from accidentally executing forbidden actions, the selective blocking must work both at the level of the Uyuni UI/XML-RPC API and at the level of the Salt command line.
+
+## Salt level selective action blocking
+
+The Salt blackout mechanism works by defining a special pillar called `minion_blackout` to tell the minion that it must reject all remote executions. Exceptions can be configured using the `minion_blackout_whitelist` pillar. This accepts a list of Salt functions that are allowed during blackout, e.g. `test.ping`, `state.apply`, etc.
```
mbologna: "all remote executions" = all Uyuni/SUSE Manager actions _and_ Salt commands, right?

accepted/00000-minion-selective-action-blocking.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Uyuni level selective action blocking
+
+For traditional clients the action types that are allowed when a system is locked are configured in the db. The configuration is global and applies to all clients indiscriminately.
+
+In the case of Salt minions, the same approach will be used. The action types that are allowed when a system is locked will be stored in the db.
+
+Additionally, in order to make the approach more flexible and to accommodate CaaSP a new object called a "locking profile" will be introduced.
+
+The locking profile will contain a list of permitted actions. A profile can be reused across multiple minions. A minion will have only one locking profile assigned when locked.
```
mbologna: Would it be possible a scenario where we would need two or more locking profiles associated with a minion?
In that case, what would the resulting operation for the locking profiles, union?

accepted/00000-minion-selective-action-blocking.md
```diff
@@ -0,0 +1,93 @@
...
+
+For traditional clients the action types that are allowed when a system is locked are configured in the db. The configuration is global and applies to all clients indiscriminately.
+
+In the case of Salt minions, the same approach will be used. The action types that are allowed when a system is locked will be stored in the db.
+
+Additionally, in order to make the approach more flexible and to accommodate CaaSP a new object called a "locking profile" will be introduced.
+
+The locking profile will contain a list of permitted actions. A profile can be reused across multiple minions. A minion will have only one locking profile assigned when locked. 
+
+The locking profile will be configured in code. Based on the installed products or any other arbitrary criteria defined in code Uyuni will apply a locking profile. The profile will be applied automatically when the software profile is refreshed (either as a result of registration or when triggered at a later time).
```
mbologna: The locking profile will be deployed as a Salt formula?

accepted/00000-minion-selective-action-blocking.md
mbologna: While I agree with keeping with the principle that databases should contain variable data, aren't we creating a potential pitfall for managing the locking action in the future? Because if allowed actions for traditional systems will be in the database, whilst the allowed actions for minions will be in code.

accepted/00000-minion-selective-action-blocking.md
mateiw: To keep things simple for now I would say to have only one locking profile per minion. This can be easily changed if such a need appears in the future.

accepted/00000-minion-selective-action-blocking.md
mateiw: It will be hardcoded in java, ideally in a plugin.

accepted/00000-minion-selective-action-blocking.md
mateiw: Right.

accepted/00000-minion-selective-action-blocking.md
mbologna: Ok I think it's worth adding it to the RFC text.

accepted/00000-minion-selective-action-blocking.md
mateiw: Yes, I think for now we should be consistent with the traditional clients.

mbologna: Please add this bit of information as it is a notable exception IMHO

PR #33, opened 2020-02-20
Autogenerate Bootstrap Repositories RFC

mcalmer:


mcalmer:
CC @ncounter @juliogonzalez - just in case you are interested

accepted/00000-auto-generate-bootstrap-repositories.md
```diff
@@ -0,0 +1,71 @@
...
+
+A bootstrap repository provide the software for the supported operating systems with its dependencies. The content for the
+bootstrap repositories come from the normal channels which must be synchronized and updated on regular base.
+
+Users very often do not know this, or forget to create them when they synchronize a new product. Also the bootstrap repository
+is only updated when the user execute the generator script again.
+
+This RFC should describe how to change the current tools to automatically 
+execute the generation of bootstrap repositories after
+a new product was added and after a channel changed when a maintenance update was synced to Uyuni Server.
```
paususe: Do we also need to create/update the bootstrap repository when the product itself is synchronized, e. g. new or updated packages in the openSUSE Leap Updates repository potentially every day?

accepted/00000-auto-generate-bootstrap-repositories.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+We maybe 15 minutes late with the bootstrap repository or maybe longer when a channel failed to sync or does not sync at all.
+
+# Alternatives
+[alternatives]: #alternatives
+
+1. Execute the `mgr-create-bootstrap-repo` one time per day in the morning via taskomatic expecting the nightly reposync is finished.
```
mbologna: Can we steer away from Taskomatic and use a systemd timer to automate the execution of `mgr-create-bootstrap-repo`?

accepted/00000-auto-generate-bootstrap-repositories.md
```diff
@@ -0,0 +1,71 @@
...
+the number of runs.
+
+Step 2: add a log file to `mgr-create-bootstrap-repo`. Currently it has no log file. But when this tool is running in the background
+we need the possibility to debug problems. All messages should go to a log file.
+
+Step 3: let `mgr-create-bootstrap-repo` use a configuration file. All commandline options should be possible to set via
+configuration file. This allows to influence the behavior even in automatic mode when it is not possible to change the commandline
+options.
+
+To trigger the re-generation we implement the following algorithm:
```
paususe: Instead of comparing the last_modified and modified_date, would it make sense for reposync to add a flag (in the DB or even something as simple as touching a file on disk) when the reposync brings new/updated packages? If there are no new/updated packages, then no need to compare anything.

Or might that fail in the case of custom repositories?

accepted/00000-auto-generate-bootstrap-repositories.md
```diff
@@ -0,0 +1,71 @@
...
+- if any `last_modified` timestamps are newer then the timestamp of the main metadata file and the `modified date` of the main metadata
+  file is older than 26 hours, re-generate the bootstrap repository. This is for the case a channel is not synced. It can happen on
+  custom repositories needed for bootstrap repositories.
+
+Provide a `--force` option to enforce recreation of all bootstrap repositories without checking if needed. Mostly for manual usage.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+We maybe 15 minutes late with the bootstrap repository or maybe longer when a channel failed to sync or does not sync at all.
```
paususe: What happens if the bootstrap repository creation takes more than 15 min? Will another run of the algorithm and `mgr-create-bootstrap-repo be started? How do we protect against that?

Can we do this based on actual packages being new/updated, and then run the check for the flag every 1 min, so that if there is no new/updated packages nothing will be done (so check finishes early) and if there are new/updated packages the bootstrap repository is immediately generated?

accepted/00000-auto-generate-bootstrap-repositories.md
paususe: It will cause less trouble than now but still cause trouble

accepted/00000-auto-generate-bootstrap-repositories.md
```diff
@@ -0,0 +1,71 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+1. Execute the `mgr-create-bootstrap-repo` one time per day in the morning via taskomatic expecting the nightly reposync is finished.
+
+2. Every finished repo sync trigger a `mgr-create-bootstrap-repo --for_updated_channel <label>`. It check if this channel is used
```
paususe: It sounds like this is the same proposal I mentioned above. What are the pros/cons of this one?

accepted/00000-auto-generate-bootstrap-repositories.md
juliogonzalez: Not just products or Maintenance Updates.

Think for example about Ubuntu, as it requires the user to add the Ubuntu main repositories, and some packages for the bootstrap repositories are taken from there.

I think it should be more: when a repository is added or when gets updates.

accepted/00000-auto-generate-bootstrap-repositories.md
```diff
@@ -0,0 +1,71 @@
...
+
+This RFC should describe how to change the current tools to automatically 
+execute the generation of bootstrap repositories after
+a new product was added and after a channel changed when a maintenance update was synced to Uyuni Server.
+
+# Detailed design
+[design]: #detailed-design
+
+The currently exiting tool `mgr-create-bootstrap-repo` list all available bootstrap repositories it could generate.
+This list depends on the products the user has synchronized.
```
juliogonzalez: Not sure if I'd use the word `products` here. Sounds too much like `SUSE Products`.

The list will also include `Debian` soon, or it includes `Ubuntu for Uyuni`.

accepted/00000-auto-generate-bootstrap-repositories.md
juliogonzalez: Let's think that the user is enabling a SUSE product from WebUI, or adding something external manually or with `spacewalk-common-channels`.

How would the user know when it's possible to start bootstrapping stuff?

paususe: > Let's think that the user is enabling a SUSE product from WebUI, or adding something external manually or with `spacewalk-common-channels`.
> 
> How would the user know when it's possible to start bootstrapping stuff?

You are right: we should provide feedback to the user that the bootstrap repository is ready. 

Do we provide any feedback now?

Would it make sense to consider a repository/channel is not fully synced until the bootstrap repository creation is not finished?

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: yes, exactly. I expect they will be re-generated daily.

mcalmer: This is what I mean when I write `a new product was added and after a channel changed`. On initial creation and on update.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: Tell me a better word. OSes?

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: AFAIK last_modified is only updated when really something changed. So I think this is the flag we look for.
I do not like to add 20 different ways to achieve the same thing.

Of cause I will check if last_modified really only change when a new package was added.

paususe: Ah OK, perfect then

accepted/00000-auto-generate-bootstrap-repositories.md
juliogonzalez: That's a good question :-\

We can keep `products` as long we  don't think about `SUSE Products from SCC` during the implementation :-)

Maybe a not to explain what `product` at the start of the RFC would help:

> ** NOTE: When talking about "products" at this RFC, we are talking about either products from SUSE SCC, or other OS added manually by the user or with `spacewalk-common-channels` such as Ubuntu, CentOS, Debian, etc.**

accepted/00000-auto-generate-bootstrap-repositories.md
juliogonzalez: But it says `when a maintenance update was synced to Uyuni Server.` And a channel doesn't change only because of Maintenance Updates, as it depends on the kind of channel.

Only those from SCC change because of MUs.

I suggest the following:

`a new product was added and after a channel changed when underlying repositories synced new packages`.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: Running multiple `mgr-create-bootstrap-repo` at the same time can be prevented with taskomatic or locking.
spacewalk-repo-sync does the same. We can copy that mechanism if we want. I will add the locking to the RFC.

About the second question: It is very hard as we have the data only inside of the `mgr-create-bootstrap-repo` tool.
This is the reason why I want to call the tool, and the tool check if there is something to do or not.
In case there is nothing to do, it should exit as soon as possible.
We can talk about the interval when we know how long it would take to find out that there is nothing to do.
But doing things every minute will add load on the server. And when I add a new product we have to wait anyway for all required channels to finish syncing. Depending on the product it takes between 30 Minutes and multiple hours.
So I think 15 minutes is not that much.

mcalmer: This is not a new problem. It already exist. What I can add to the RFC is, that we can send a notification to users.
But I see this as optional and a nice to have enhancement.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: Well, currently everything we schedule in SUSE Manager is done by Taskomatic. Schedules for taskomatic jobs can be changed using the Web UI. So my first option is taskomatic for this reason.

If taskomatic should be used as less as possible, we can do this also via systemd timer. As we do not need to trigger this from the java code, we can easily do it.

I add this to the open questions for now. It is more a policy question.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: The difference is, that this is triggered by reposync itself and it was not planned to check for changes.
But you are right, they are pretty similar.

We could remove sheduling via taskomatic or systemd and call it directly from reposync. This would
remove the need of an extra trigger job.
What do you think?

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: ok, will change

accepted/00000-auto-generate-bootstrap-repositories.md
juliogonzalez: FMPOV, the problem is that until now they had to run it manually, but they knew that if the script didn't show errors, they could already bootstrap.

While now they won't need this, but they don't really know when they can start to bootstrap.

So yeah, maybe having a notification is optional and a "nice to have", 

I am OK with but if we do it this way, we need to clearly document at the docs that they should track the logs.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: I used products/channels now.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: I added a section about Notifications to the text

paususe: _Running multiple mgr-create-bootstrap-repo at the same time can be prevented with taskomatic or locking.
spacewalk-repo-sync does the same. We can copy that mechanism if we want. I will add the locking to the RFC._

I don't know the internals of mgr-create-bootstrap-repo but problem should not be running multiple mgr-create-bootstrap-repo at the same time but multiple mgr-create-bootstrap-repo at the same time *for the same product*, which is what may clash.

accepted/00000-auto-generate-bootstrap-repositories.md
paususe: I think I like it. In the end, bootstrap repository generation should be a sequential step after reposync, so if we can avoid the trigger job and pose no disadvantages, I'm all for it.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: current idea is to trigger `mgr-create-bootstrap-repo` from `spacewalk-repo-sync`. At least no extra taskomatic job anymore.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: Right.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: Agreed. I change the RFC

mcalmer:
@paususe @juliogonzalez @mbologna - last call. If this is now ok, I would like to merge. Or do you think a final meeting is needed?

accepted/00000-auto-generate-bootstrap-repositories.md
```diff
@@ -0,0 +1,91 @@
...
+
+To trigger the re-generation we implement the following algorithm:
+
+- call `mgr-create-bootstrap-repo` from a successful finished reposync
+- It calculate all channels which are needed to create the bootstrap repository for every available OS
+- It compare the `last_modified` timestamps of the channels (DB) with the `modified date` of the main metadata file
+  (repomd.xml or Release)
+- if all `last_modified` timestamps are newer then the timestamp of the main metadata file, re-generate the bootstrap repository
+- if any `last_modified` timestamps are newer then the timestamp of the main metadata file and the `modified date` of the main metadata
+  file is older than 26 hours, re-generate the bootstrap repository. This is for the case a channel is not synced. It can happen on
```
paususe: Why this complex logic and why the 26 hours? Given that `mgr-create-bootstrap-repo` will be called from reposync, why not just "if any of the `last_modified` timestamps for one of the repos of that channel is newer that the last modified bootstrap repository?" I don't think we need to make a difference between custom repos and vendor repos.

accepted/00000-auto-generate-bootstrap-repositories.md
```diff
@@ -0,0 +1,91 @@
...
+As this now run in background we should add notifications for the users. The following are useful:
+
+1. On Error: when regeneration of a bootstrap repository failed.
+2. On Success: when a regeneration successfully finished.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- We maybe 15 minutes late with the bootstrap repository or maybe longer when a channel failed to sync or does not sync at all.
```
paususe: Is this still true? I think this drawback has gone away, since we are now running the bootstrap repository creation/update after reposync, not with a 15 min timer.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: Only when all repos are synced you can get a successful run of the tool.

Example - first time sync after the product was added:

- tools update channel finished sync => error with not found packages
- tools pool channel finished sync => error with not found packages
- sles pool channel finished sync => success, but you may have outdated packages
- sles update channel finished sync => success, now you are really up to date.

With notifications it would produce at least 2 errors and user start L3s.

Also in case update just syncing updates there is the chance that new required packages cause errors.

The 26 hours come from the initial idea to run it at least 1 time per day. I think with the current approach we can go down to some hours (4) and check that all repos were synced at least 1 time before.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: correct. Removed

accepted/00000-auto-generate-bootstrap-repositories.md
```diff
@@ -0,0 +1,90 @@
...
+Also the bootstrap repository is only updated when the user execute the generator script again.
+
+This RFC should describe how to change the current tools to automatically 
+execute the generation of bootstrap repositories after a new product or channel was added,
+and after a channel changed because the underlying repositories synced new or updated packages.
+
+# Detailed design
+[design]: #detailed-design
+
+The currently exiting tool `mgr-create-bootstrap-repo` list all available bootstrap repositories it could generate.
```
juliogonzalez: Typo `exiting` -> `existing`.

accepted/00000-auto-generate-bootstrap-repositories.md
paususe: Sorry but I do not understand it yet.

I get the "all repos were synced at least 1 time before", I'm in 100%.

But why the 4 hours? Where does that magic number come from?

You cannot count on a repository being synced only once a day. User may trigger reposync manually and if new/updated packages come, bootstrap repository needs to be created/updated.

mcalmer: > But why the 4 hours? Where does that magic number come from?

When we talk about custom repositories the user is free to disable syncing. When he knows that the content does not change he could decide to configure it that way.
When this happens, we never reach all repos re-synced. So we need a number as a grace period we wait (maybe it is just syncing later) before we just re-generate it even if not all repos were re-synced.

If you think it is ok that we generate them even when not all repos are re-synced. We can do this, but it would cause 4 times re-generation of the repo for every normal SLE product.

With this approach I wanted to limit it and try to generate it only when we are really "complete" and have a consistant state.
Getting an inconsistant state will happen only 1-2 times per year. But we know it will happen.

accepted/00000-auto-generate-bootstrap-repositories.md
mcalmer: fixed

accepted/00000-auto-generate-bootstrap-repositories.md
paususe: I am all for whatever avoids inconsistent states :-)

paususe:
@mcalmer 
I just realized: are there any implications in terms of GPG key acceptance?

When `mgr-create-bootstrap-repo` is run manually, the user can manually accept the GPG key on the console. Now that it will be run automatically, how will that happen?

mcalmer:
> @mcalmer
> I just realized: are there any implications in terms of GPG key acceptance?
> 
> When `mgr-create-bootstrap-repo` is run manually, the user can manually accept the GPG key on the console. Now that it will be run automatically, how will that happen?

No, GPG key acceptance is done in `spacewalk-repo-sync`. This is a different tool.

paususe:
> > @mcalmer
> > I just realized: are there any implications in terms of GPG key acceptance?
> > When `mgr-create-bootstrap-repo` is run manually, the user can manually accept the GPG key on the console. Now that it will be run automatically, how will that happen?
> 
> No, GPG key acceptance is done in `spacewalk-repo-sync`. This is a different tool.

Ah right, sorry I don't know why I messed those two.

PR #34, opened 2020-02-24
Clustering architecture for Uyuni Server

isbm:
Short overview for clustering architecture for Uyuni Server. [Read it here](https://github.com/uyuni-project/uyuni-rfc/blob/6ecfe093e15800addc5405e3fa79264a9756ead7/accepted/00099-uyuni-cluster.md)


moio:
My personal take: lots of ideas are proposed here, and I am sure some of them can be made to work to attain scalability. With one inescapable condition that is implied even from the PR title, which is: a cluster is available.

One of the fundamental characteristic of a cluster is that all nodes are (substantially) **co-located and connected by a high speed network - that is the only physically possible way to attain low latency between nodes, which is a working hypothesis for most clustering software design**.

Per [Wikipedia](https://en.wikipedia.org/wiki/Computer_cluster):
> The components of a cluster are usually connected to each other through fast local area networks

Without that condition such designs will simply not work well, and that reflects in implementations and their requirements. Examples cited in this PR include:

 - Kafka ([according to Cloudera](https://docs.cloudera.com/documentation/enterprise/6/release-notes/topics/rg_hardware_requirements.html#concept_qkj_ymn_3bb)) requires "Gigabit Ethernet or 10 Gigabit Ethernet. Avoid clusters that span multiple data centers.". According to [Confluent](https://docs.confluent.io/current/kafka/deployment.html#network): "A fast and reliable network is an essential performance component in a distributed system. Low latency ensures that nodes can communicate easily, while high bandwidth helps shard movement and recovery. Modern data-center networking (1 GbE, 10 GbE) is sufficient for the vast majority of clusters"
 - etcd, needed for Kubernetes, [has similar recommendations](https://etcd.io/docs/v3.4.0/op-guide/hardware/): "Deploy etcd members within a single data center when possible to avoid latency overheads and lessen the possibility of partitioning events"
 - [Ceph](https://docs.ceph.com/docs/jewel/start/hardware-recommendations/#networks) even recommends two LAN cards: "We recommend that each host have at least two 1Gbps network interface controllers (NICs)"
 - [TiKV](https://tikv.org/docs/4.0/tasks/deploy/introduction/#production-specifications) specifies "10 Gigabit ethernet (2x preferred)"


So any adoption of these ideas has to take into account that low-latency is a requirement, and this limits the applicability of the proposal here.

PR #35, opened 2020-03-04
align RFC to implementation

mcalmer:
During implementation of this RFC some decissions were made which differ from the RFC text.
This PR should align RFC and code.

- the spacewalk configfile was already used and required for DB access. Another config file is not so useful. Additional values where specified in that config but others could not be done in simple `key/value` style. So let's see if somebody need them before introducing a 3rd config file.
- use DB column `last_synced` instead of `last_modified` as it is the better value to test with
- the algorithm was described wrong when a sync should be triggered when not all repos are synced yet.
- a force parameter is not needed as the same can achieved by calling the tool in interactive mode.


mcalmer:
Two approvers are enough. Merging.

PR #37, opened 2020-03-14
Akka

moio:
## [Formatted version here](https://github.com/uyuni-project/uyuni-rfc/pull/37/files?short_path=d5dbe9a#diff-d5dbe9a50b195fa403d616f70c756bb9)

@rjmateus: please take a close look particularly at the "Hackweek shortcuts" list, TIA


accepted/00070-akka.md
```diff
@@ -0,0 +1,134 @@
...
+1. in Akka dependencies can be expressed by actors firing messages to other actors when done
+2. any other part of the code base can ask an Akka actor to do something by sending it a message
+3. [Akka has a concept of priority messages to actors](https://doc.akka.io/docs/akka/2.5/mailboxes.html)
+4. Akka seems to be regarded as high performance, has thorough and well-documented tuning knobs
+5. [Akka has a lot of utilities for monitoring](https://developer.lightbend.com/docs/telemetry/current/home.html)
+6. Akka is generally known and well documented, has books, videos and other material available
+
+## Detail design ideas
+
+ - The Akka-based asynchronous facility will be made available via a facade (`ActorManager` in the above PR)
```
rjmateus:
```suggestion
 - The Akka-based asynchronous facility will be made available via a facade (`ActorManager` in the bellow PR)
```
Link for PR is two paragraphs bellow. Should it be added to the beginning of the test?

accepted/00070-akka.md
```diff
@@ -0,0 +1,134 @@
...
+ - new methods/classes do not have new accompanying unit tests
+ - logging is incomplete
+ - code style of Akka-related parts follows the Akka coding conventions, which are different from the CheckStyle rules currently applied. We will either need to adapt new code to the existing conventions or relax them/create different rules
+ - error handling was not tested thoroughly, might need rework
+ - the Cucumber integration testsuite was not run on the resulting code
+ - performance tests will have to be conducted to make sure there is no regression (and evaluate possible speedups)
+ - the reflections-based mechanism to find out workers will have to be considered and it might be removed in exchange for an explicit approach
+ - most but not all Actor Messages were made `Serializable`, more refactoring might be needed to ensure this is always the case
+ - Akka tuning knobs were not all exposed, or exposed by reusing existing configuration variables. New variables might need to be introduced and properly documented
+ - the `DeferredActor` implementation is at the moment not completely written in functional style
```
rjmateus: The name is DeferringActor, and I think it is written in a functional style (but I may be missing something): java/code/src/com/suse/manager/tasks/actors/DeferringActor.java

accepted/00070-akka.md
```diff
@@ -0,0 +1,134 @@
...
+ - pro: with one DB connection pool and one Akka thread pool, it should result in an easier-to-tune Server
+ - pro: currently many support issues boil down to undetected Taskomatic failures. With one only web app, it should be easier to surface errors in a more user-visible way
+ - con: unexpected Taskomatic errors may bring Tomcat down as well. At the moment this does not happen, but as we rely on Taskomatic much more than in original Spacewalk, in many cases we already end up with a non-functioning Server anyway
+
+### Leverage Akka Clustering to offload computation onto other machines
+
+Akka has an optional module called Clustering that allows to transparently move computation across machines, specifically by instantiating Actors on remote nodes and letting Akka route messages to them.
```
rjmateus: makes sense to refer it allows across different machines or different processes on the same machine? For example, we can keep just one machine as we have now, move taskmatic event firing to inside tomcat, but the execution could be done in a worker in the same machine.

accepted/00070-akka.md
```diff
@@ -0,0 +1,134 @@
...
+
+ - unit tests have been arbitrarily disabled or left broken
+ - new methods/classes do not have new accompanying unit tests
+ - logging is incomplete
+ - code style of Akka-related parts follows the Akka coding conventions, which are different from the CheckStyle rules currently applied. We will either need to adapt new code to the existing conventions or relax them/create different rules
+ - error handling was not tested thoroughly, might need rework
+ - the Cucumber integration testsuite was not run on the resulting code
+ - performance tests will have to be conducted to make sure there is no regression (and evaluate possible speedups)
+ - the reflections-based mechanism to find out workers will have to be considered and it might be removed in exchange for an explicit approach
+ - most but not all Actor Messages were made `Serializable`, more refactoring might be needed to ensure this is always the case
```
rjmateus: also no exhaustive testing in the refactorings we have done for messages when transforming messages to be serializable. I remember some classes we move some Id's to the message but the old code was doing some processing with it, and we replace it by computing the data again. I Don't know the impact of this behavior change.

accepted/00070-akka.md
moio: Right, I meant stateless but wrote functional. The only detail I would change is the mutable `accumulator` Map in it, but it's definitely a minor point.

rjmateus: I think it cannot be stateless since we need to keep tracking of the deferrer events we should be able to fire in the future.
One of the actores component is the state (https://doc.akka.io/docs/akka/current/general/actors.html#state) which is propagated in the  accumulator variable.
How do you think it should be changed?

moio: One way is to construct a new Behavior and return it. In the following example, a `Counter` Actor keeps a counter (duh), and the "natural" way to implement it coming from the Java world is to keep a mutable variable inside the Actor itself (this is called "object-oriented style" and it is the second example).

The alternative way, more coming from immutable/FP point of view, is to construct and return an entirely new Behavior upon reception of the "increment" Message (the first example):

https://doc.akka.io/docs/akka/current/typed/style-guide.html#functional-versus-object-oriented-style

I think it should be possible to change `DeferringActor` in such a way it works without mutable variables.

rjmateus: But In the current implementation (i may seeing it wrong) we are recreating a new behavior every time we receive a message, and don't have any field in the actor class, using this way the FP approach.Sorry to ask again, but what I'm missing.
Link to implementation:
https://github.com/uyuni-project/uyuni/pull/1903/files#diff-1ccb348d144c3a941b3dcb1a6a6378d5

moio: The only small thing I see missing is that we keep and modify an existing `HashSet` as accumulator, instead of creating one new Set with just one extra element (and discarding the old one).

https://github.com/uyuni-project/uyuni/pull/1903/files#diff-1ccb348d144c3a941b3dcb1a6a6378d5R65-R67

I share that this is very minor and at this point it amounts to nitpicking. Removing from text.

rjmateus: AH!! Ok got it :)
Thanks for the explanation :+1:

moio:
@mcalmer I would like to have an opinion from you as well - just in case you see anything big that we might have missed.

accepted/00070-akka.md
```diff
@@ -0,0 +1,137 @@
...
+
+After all asynchronous computation is moved to Akka, it might make sense to consider moving all Taskomatic code out of the currently separated daemon snd into the main Tomcat web app.
+
+Architecturally, such a change would have pros and cons:
+ - pro: we could remove the internal Tomcat XMLRPC API, eliminating substantial complexity and the synchronization/availability problems we currently face (eg. [the race condition in `MinionActionExecutor`](https://github.com/uyuni-project/uyuni/blob/4.1-alpha1/java/code/src/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java#L82-L102), as per [downstream issue 10657](https://github.com/SUSE/spacewalk/issues/10657))
+ - pro: with one DB connection pool and one Akka thread pool, it should result in an easier-to-tune Server
+ - pro: currently many support issues boil down to undetected Taskomatic failures. With one only web app, it should be easier to surface errors in a more user-visible way
+ - con: unexpected Taskomatic errors may bring Tomcat down as well. At the moment this does not happen, but as we rely on Taskomatic much more than in original Spacewalk, in many cases we already end up with a non-functioning Server anyway
```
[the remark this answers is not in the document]
moio: > One major point of having taskomatic is, that it run as root and can execute actions which must have root privileges. Is this part of Akka?

No, that's a problem I was aware of but didn't connect the dots. Will add it as a con.

> What do you mean with internal Tomcat XMLRPC API?

That's a typo, I meant "internal Taskomatic XMLRPC API". Will correct, thanks.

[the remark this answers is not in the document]
moio: Clarified in 58fd7d0a78bb05b1e9c6b23de541309061611008, thanks

accepted/00070-akka.md
```diff
@@ -0,0 +1,139 @@
...
+## Following Steps
+
+Following points need further research when points 1.-6. are completed, and possibly need follow-up RFCs, still are reported here for completeness.
+
+### Move Taskomatic code in the Tomcat web app
+
+After all asynchronous computation is moved to Akka, it might make sense to consider moving all Taskomatic code out of the currently separated daemon snd into the main Tomcat web app.
+
+Architecturally, such a change would have pros and cons:
+ - pro: we could remove the internal Taskomatic XMLRPC API, eliminating substantial complexity and the synchronization/availability problems we currently face (eg. [the race condition in `MinionActionExecutor`](https://github.com/uyuni-project/uyuni/blob/4.1-alpha1/java/code/src/com/redhat/rhn/taskomatic/task/MinionActionExecutor.java#L82-L102), as per [downstream issue 10657](https://github.com/SUSE/spacewalk/issues/10657))
```
mcalmer: The `feature` of unauthenticated requests to the taskomatic API was really used by python scripts.
We would need to find a new way of connecting to it.

moio: Will add a note, however, keeping the endpoint is not a problem (we just need to make sure it ends up using the same TCP port).

moio: Fixed

mcalmer: I was not sure if this is working with what you have in mind.

accepted/00070-akka.md
```diff
@@ -0,0 +1,141 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+The only considered alternative at this point is to roll our own thread pool model and infrastructure and use it to replace all current implementations. This is expected to be much more difficult and error-prone.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Just how far do we feel like going in the steps listed above?
```
renner: I would feel confident about a first implementation round covering steps 1-4 mentioned above, but converting Taskomatic to be completely based on Akka seems like a potentially independent effort to better be tackled once the first and easier steps are done and better experience is gathered? It is good to mention the whole plan here, but I could imagine that some technical details are still unknown by now that might require an additional RFC or an extension to this one later. As I understood the prototype does not cover anything related to Taskomatic yet.

accepted/00070-akka.md
```diff
@@ -0,0 +1,141 @@
...
+6. learnability (by new developers and existing developers unfamiliar with the interested code area, eg., for bugfixing purposes)
+
+At this point the main goal is to **improve internal quality by rationalizing the internal rearchitecture**.
+
+# Detailed design
+[design]: #detailed-design
+
+## Core idea
+
+Remove homegrown solutions and replace them with [akka.io](https://akka.io/), a well-known open source tookit for concurrent/reactive/distributed applications.
```
renner:
```suggestion
Remove homegrown solutions and replace them with [akka.io](https://akka.io/), a well-known open source toolkit for concurrent/reactive/distributed applications.
```

accepted/00070-akka.md
moio: Step 5 should not be very difficult, considering that some Taskomatic tasks to use the `MessageQueue`, and that means we will have an instance of the `ActorManager` in the Taskomatic process already by step 4.

Step 6 is more uncertain I can agree. Certainly it won't be started until 1-5 are completed, and at that point some research and even a second RFC might be needed. I can make that explicit in the text, moving step 6. to the "Next steps", if you feel it's wiser to do so.

mbologna: Yes, I agree, please clearly separate what you are aiming for in this RFC and what's left for the future.

accepted/00070-akka.md
```diff
@@ -0,0 +1,141 @@
...
+ - new methods/classes do not have new accompanying unit tests
+ - logging is incomplete
+ - code style of Akka-related parts follows the Akka coding conventions, which are different from the CheckStyle rules currently applied. We will either need to adapt new code to the existing conventions or relax them/create different rules
+ - error handling was not tested thoroughly, might need rework
+ - the Cucumber integration testsuite was not run on the resulting code
+ - performance tests will have to be conducted to make sure there is no regression (and evaluate possible speedups)
+ - the reflections-based mechanism to find out workers will have to be considered and it might be removed in exchange for an explicit approach
+ - most but not all Messages were made `Serializable`, more refactoring might be needed to ensure this is always the case
+ - Akka tuning knobs were not all exposed, or exposed by reusing existing configuration variables. New variables might need to be introduced and properly documented
+ - ideas from [the normalisation PR](https://github.com/uyuni-project/uyuni/pull/1954) should probably be integrated
```
mbologna: It is not clear to me what are the action point to be taken here, can you please elaborate?
For example: it the normalization PR in the scope of the RFC?

accepted/00070-akka.md
```diff
@@ -0,0 +1,141 @@
...
+   - if some really need to be retained then it is a possibility to keep the endpoint on its current (Taskomatic-specific) TCP port or to map the methods to the Tomcat XMLRPC TCP port and adapt scripts using them 
+ - pro: with one DB connection pool and one Akka thread pool, it should result in an easier-to-tune Server
+ - pro: currently many support issues boil down to undetected Taskomatic failures. With one only web app, it should be easier to surface errors in a more user-visible way
+ - con: unexpected Taskomatic errors may bring Tomcat down as well. At the moment this does not happen, but as we rely on Taskomatic much more than in original Spacewalk, in many cases we already end up with a non-functioning Server anyway
+ - pro: currently Taskomatic runs as `root`, but not all tasks really require `root` privileges. Attack surface can thus be reduced by running code with Tomcat privileges by default
+ - con: Taskomatic tasks that do require root privileges will need some different permissions to run. One way is to use Akka Clustering (described below) with two processes (one `root` and one non-`root`) on the same machine
```
mbologna: Why is root required? I suppose for opening a well-know port (< 1024). Is my assumption correct?

accepted/00070-akka.md
```diff
@@ -0,0 +1,141 @@
...
+ - is the extra administration overhead worth it? In which cases? Maintaining a cluster is much more work-intensive than maintaining a single VM
+ - how can we deal with upgrades? Akka Clustering has primitives for versioning, but making them work with existing packaging and delivery processes requires some study
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+ - as with all refactorings, we risk introducing regressions at least initially
+ - Akka itself is a complicated artifact. Although external quality seemed very good during HackWeek (no bugs or strange/undocumented behavior was noticed, great documentation) there is always a chance some internal bug hits us, and might be difficult to solve
+ - we get to maintain a packaged version of Akka
+ - performance has to be measured - we expect it to improve but that is not guaranteed
```
mbologna: Can we already define some metrics to measure the performance before and after the Akka introduction?

accepted/00070-akka.md
moio: Done

accepted/00070-akka.md
moio: Hmm, I think you are right, it could be phrased better.

The normalization PR has nothing to do with this RFC, other than the fact it happened in parallel. The moment it is merged, it will be encouraged to use interfaces and implementation with explicit dependencies between components expressed via constructor parameters rather than the `static`-based approach otherwise common in our codebase (see `*Manager` classes) and also used in the HackWeek PR.

So the gist is: if (as expected) the normalisation PR is merged before work on Akka is merged, the latter should be done in a way to reflect the former's design - which is not the case in the HackWeek patches above.

moio: Addressed.

accepted/00070-akka.md
moio: Not really, opening ports is not the main concern.

Problem here is that some tasks require `root` to be executed successfully, for example in order to invoke `spacewalk-repo-sync`.

accepted/00070-akka.md
moio: One possible idea is to measure onboarding time and upgrade time of a given number of minions without and with the patch, as those operations exercise the Salt event code path and MessageQueue heavily.

I'd go incrementally though, creating specific benchmark for each migrated subsystem (before moving on to the next!)

Do you feel like adding some text in that direction is necessary, or maybe do you have any other suggestion?

mbologna: In the Motivation section, you point out that you are aiming at - among the other aspects - to performance.
So, IMHO it makes sense to define a rough idea of what would be the figure of merit for every aspect you want to improve (or at least the one with the heaviest impact, like performance). No need to be super detailed, onboarding time and upgrade time are already OK in my opinion.

moio: Right, added, thanks

accepted/00070-akka.md
```diff
@@ -0,0 +1,144 @@
...
+
+All of the above was implemented (as proof-of-concept) as part of [this HackWeek PR](https://github.com/uyuni-project/uyuni/pull/1903)  (excluding commits after February 15, which deal with Akka Clustering) which can serve as a reference for implementation details.
+
+Some shortcuts have been taken during HackWeek and will not be acceptable for proper implementation, particularly:
+
+ - unit tests have been arbitrarily disabled or left broken
+   - in particular, when Message formats were significantly altered from the original code
+ - new methods/classes do not have new accompanying unit tests
+ - logging is incomplete
+ - code style of Akka-related parts follows the Akka coding conventions, which are different from the CheckStyle rules currently applied. We will either need to adapt new code to the existing conventions or relax them/create different rules
```
paususe: Does adapting to the CheckStyle rules imply any risk, e. g. performance?

Does it make sense to write Akka-related code using Akka code conventions, and non-Akka code using CheckStyle conventions? E. g. would that duality make it easier to debug, discuss, find code snippets in the web, etc, or is it irrelevant?

accepted/00070-akka.md
paususe: > Not really, opening ports is not the main concern.
> 
> Problem here is that some tasks require `root` to be executed successfully, for example in order to invoke `spacewalk-repo-sync`.

Why does `spacewalk-repo-sync` require root in the first place?

After task 1-5 are complete, some research would be needed to identify which actions Taskomatic currently runs as root, why those need root, and whether we can make them not need root.

accepted/00070-akka.md
```diff
@@ -0,0 +1,144 @@
...
+ - we get to maintain a packaged version of Akka
+ - performance has to be measured - we expect it to improve but that is not guaranteed. At minimum onboarding and patching time for a definite set of minions should be measured before/after implementation of this RFC
+
+# Alternatives
+[alternatives]: #alternatives
+
+The only considered alternative at this point is to roll our own thread pool model and infrastructure and use it to replace all current implementations. This is expected to be much more difficult and error-prone.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
```
paususe: Another unresolved question: would it be possible (i. e. sensible effort) to implement Akka for tasks 1-5 as an opt-in feature, while still keeping the existing codepaths?

This way we can develop and test and at some point, declare stable, or kill Akka.

accepted/00070-akka.md
moio: > Does adapting to the CheckStyle rules imply any risk, e. g. performance?

No, CheckStyle rules only impact code conventions (eg. naming of variables, whitespace...) and will not really impact semantics.

> Does it make sense to write Akka-related code using Akka code conventions, and non-Akka code using CheckStyle conventions? E. g. would that duality make it easier to debug, discuss, find code snippets in the web, etc, or is it irrelevant?

My personal opinion is that yes, it will make sense to keep Akka code under Akka conventions, as it will be easier to pick up looking at Akka usage in the outside world. Also, some of the CheckStyle rules we currently apply only really made sense when the upstream was Spacewalk and possibly we might want to just relax them.

In all, I think all we need is some sort of poll, and either we will adapt code or CheckStyle rules, but that should not be of a major impact in any case.

accepted/00070-akka.md
moio: Anything is possible but such a solution would be much more complex and hard to debug, because we would have two mechanisms and have to pay a lot of attention not to mix them. I cannot presently imagine an easy way to make it pluggable. Given this is impacting very critical areas of the codebase, I would not recommend that.

Or maybe, @rjmateus, do you have some ideas in this direction?

rjmateus: IIRC, when we introduce AKKA, we replaced just a few lines in each caller, to invoke akka instead of message queue.
Could be a possibility to have flag to decide which code paths should be used, and turn on/off the execution of akka code path. Imagine an IF that says if we use message queue or akka, based on a parametrized boolean flag.
This needs to be analyzed in more detail for all specific implementation cases.
Cons of this approach:
- Two different code path to do the same thing could lead to divergent code. Imagine someone solving a bug in the old code path and forget to port to the new one.
- Debug possible problems reported by client would be more difficult.

moio: Call sites are the lesser evil - we could have a Manager in front of ActorManager. Problem is the backend

rjmateus: In this case, backend should be duplicated, with two execution paths, one is the current and another implemented using akka. With the flag we decide on the fly which code path to follow.
This could never be a long term solution since it duplicates the maintenance cost.

paususe: > In this case, backend should be duplicated, with two execution paths, one is the current and another implemented using akka. With the flag we decide on the fly which code path to follow.
> This could never be a long term solution since it duplicates the maintenance cost.

Right, this is of course not a long term solution. 

My proposal would be, if feasible, to implement Akka like we did for the SPA: opt-in (e. g. via rhn.conf), then declare stable and remove the old code.

moio: I understand the intent but I do not think it makes sense to duplicate thousands of lines (not less than 4K given the HackWeek PR) and then ask every developer to make sure they land any modification to both versions.

Rather, I would suggest a gradual migration, a few well tested classes at a time, to be merged first in Uyuni and then backported even to released product versions. That will:

1. spread the risk instead of concentrating it all on one big release,
2. keep the code distance between Uyuni and released products small. Any developer needing to modify any of those 4K lines (eg. because of a bug fix) will have to do so once, as patches will apply cleanly regardless of the branch. Thus avoiding time-costly and error-prone porting
3. provide the Akka infrastructure very early so that new features/code can start making use of it at their first version, instead of targeting old frameworks first and then needing a port later

Of course the big catch is that we get an obligation to work with a higher-than-usual attention to quality, particularly regression testing. After some reflections I do not see this as a big obstacle: very many use cases are already covered by our testsuites and the remaining should be covered anyway, so we better spend some time to adding missing tests. About internal quality, the best approach we found so far is pairing, which I would definitely recommend in this case.

rjmateus: Two different execution path for the same logic is normally bad, and that was why try to make a point that it would duplicate maintenance cost.
I think the approach you suggested is the best one in this case.

The pairing is a really good option here. As you said, we need to be extra careful and in complex implementations like this one, I think it's the most productive and secure approach.

moio: I am not adding configurability to the text at this time.

As always course might be reversed later if contrary evidence surfaces.

PR #38, opened 2020-03-25
RFC about providing and trusting GPG keys from other Vendors in Uyuni

mcalmer:


accepted/00000-trust-vendor-gpg-keys.md
```diff
@@ -0,0 +1,75 @@
...
+
+After we got the permissions from the admin we can import the key in the keyring and also deploy the key
+to clients when they use the channels.
+
+# Detailed design
+[design]: #detailed-design
+
+## How to provide the keys?
+
+We should build a package which contains all the keys and get installed as files at `/usr/share/susemanager/gpg-keys/<keyid>.key` ?
```
paususe: Is this it? `distribution-gpg-keys`

https://software.opensuse.org/package/distribution-gpg-keys

mcalmer: Well, this package has a lot of keys, but all in different formats and sometimes multiple in 1 file. So it seems to be better to create a new package with content we can really use.

msmeissn:
looks good fwiw :)

accepted/00071-trust-vendor-gpg-keys.md
```diff
@@ -0,0 +1,99 @@
...
+## How to trust the keys?
+
+When the admin go into the products page and click on synchronizing products, we collect a unique list of keyids
+which needs to be accepted. Keys which were already accepted can be skipped.
+
+This should be done by an Ajax call to the backend. We send the channel identifiers and the backend lookup the
+keyids, reduce them by keyids already imported and send back detailed information about the ids which needs to be accepted.
+
+The User Interface show a popup and present all keys with User ID and fingerprint and ask the admin if he trust the keys
+and if they should be imported. To make it easier we allow only all presented keys or none.
```
jsegitz, line 62: This is for us a crucial point. The UI needs to be good to ensure that the user understands the implications of this decision. As it's an all-or-nothing decision it needs to be shown which key corresponds to which product so that the user can cancel, remove a product and redo this if he doesn't trust a key. 
There should be an explanation of the implications of trusting a key and and a link to documention that explains how they can verify keys, not just a "do you trust this key" question

accepted/00071-trust-vendor-gpg-keys.md
```diff
@@ -0,0 +1,99 @@
...
+
+In case of `mgr-sync` we need to do the same from the command line. The information about missing GPG keys should be
+provided via a XMLRPC API.
+
+
+## How to update keys?
+
+When keys expire and get extended we need to update them. To do this we should update the key in the package.
+The package should get a `post` script which lookup which keys are expired in the keyring. Only for these keys
+it try to update them from the new files in the package.
```
jsegitz, line 79: This needs to be specified more clearly. How do you know that the new keys is trustworthy?
- Is the expectation that the new key is signed by the old key? 
- Or do you only do this for keys that have a new expiration date but are otherwise the same?

accepted/00071-trust-vendor-gpg-keys.md
```diff
@@ -0,0 +1,99 @@
...
+
+
+## How to trust keys when using spacewalk-common-channels?
+
+In spacewalk-common-channels we can configure GPG key URLs. We can download the file and present it to the admin
```
juliogonzalez, line 84: @mcalmer problem about this: Some distributions do not provide the GPG on an URL, as you already discovered.

What are the plans in this case?

accepted/00071-trust-vendor-gpg-keys.md
mcalmer: The idea was just update when the key has a new "expiration date".

accepted/00071-trust-vendor-gpg-keys.md
mcalmer: We wanted to collect all the keys and install them locally with an RPM.
We could point to these certificates using a file url.

accepted/00000-trust-vendor-gpg-keys.md
Conan-Kudo: I know this is many months later, but if you have a problem with how it's structured, talk to @xsuchy (upstream developer) and me (openSUSE maintainer) about it. Please don't make another random package for this because it just makes life miserable for everyone.

xsuchy: *nod* I have no problem modifying the structure of distribution-gpg-keys. Just state your problem and suggested solution - ideally on https://github.com/xsuchy/distribution-gpg-keys/issues - and I will be happy to work with you on the solution.

PR #39, opened 2020-03-31
Maintenance Windows RFC

mcalmer:
How to implement Maintenance Windows in SUSE Manager.

References:
- docs about ITSM tools - https://docs.servicenow.com/search?q=Create+a+maintenance+plan
- https://checkmk.com/cms_basics_downtimes.html

### DIFF
--- a/accepted/00000-maintenance-windows.md
+++ b/accepted/00000-maintenance-windows.md
@@ -103,6 +103,12 @@ Every `VEVENT` describe a maintenance window.
 Every `VEVENT` must have a `SUMMARY`. All events with the same "summary" belong to the same schedule.
 The summary must match the schedule name in the Uyuni Database.
 
+#### Timezones
+
+The ICalendar entry defines the timezone. Uyuni must of cause respect it. When showing the Maintenance Windows
+in the UI, they should be converted into the timezone set in the user preferences.
+
+
 ### Adding a new Maintenance Schedule
 
 In the left sidebar a new page should be added below `Schedule`.
--- a/accepted/00000-maintenance-windows.md
+++ b/accepted/00000-maintenance-windows.md
@@ -259,7 +259,8 @@ For setting a schedule should be possible for multiple systems using an array of
 
 The current status of the system should be reflected in the User Interface and API.
 
-If a system is in Maintenance Mode, the User Interface could show a special icon [ || ] (pause symbol of an audio player).
+If a system is in Maintenance Mode, the User Interface could show a special icon.
+Suggestion: https://www.iconfinder.com/icons/2639855/maintenance_icon
 
 In the XMLRPC API a flag could be added to `system.getDetails()`.
 A special function `system.listSystemsInMaintenanceMode()` should be added as well.
@@ -387,8 +388,6 @@ It should be possible to add a URL to an iCalendar file and maybe implement CalD
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-
-- Pause icon could conflict with virtualization management. Is it a good idea?
 
 # Info
 
--- a/accepted/00000-maintenance-windows.md
+++ b/accepted/00000-maintenance-windows.md
@@ -336,10 +336,22 @@ More details will follow when we know more about the cluster implementation.
 
 ## Second Iteration
 
+### Visualize Maintenance Schedules, Maintenance Windows and Scheduled Actions
+
+When dealing with multiple Maintenance Schedules which have multiple Maintenance Windows it would be good
+to get an overview by showing the calendar with all these items:
+
+- Maintenance Schedules
+- Maintenance Windows (colors belong to schedules)
+- Scheduled Actions (nice to have)
+
+
 ### Implement a Maintenance Mode override
 
 Put a system manually into "Maintenance Mode". This is useful in emergency situations where a system is down
 and needs to be fixed quickly.
+Another use case is to better cooperate with 3rd party ITSM tools which could use the API to enable the Maintenance Mode.
+
 An API call should be implemented to do it. The required permissions needs to be discussed.
 Either a normal admin of the system could get the permissions to do this, or it still required organization admin permissions.
 This requires further discussion.
@@ -352,7 +364,6 @@ could be implemented in the second iteration.
 
 All this would require a trigger mechanism which execute changes on the system to reflect the maintenance mode.
 
-
 #### Emergency Maintenance Mode
 
 When a system is put into Maintenance Mode manually, it mostly happen because of an emergency case.
@@ -361,6 +372,16 @@ back online.
 
 When a system is currently in Maintenance Mode, it should be possible to schedule actions "now" and for the next minute.
 
+
+#### 3rd party ITSM tool without calendar
+
+Without a calendar schedule actions upfront is not possible.
+Such a 3rd Party Tool could enable the maintenance mode, but nothing is prepared at that point in time.
+This makes sense only when the admin sitting in front of Uyuni Server and execute the actions "now".
+
+Another option would be to change the recurring action feature.
+An action could be prepared as recurring action which run automatically when the maintenance window is turned "on".
+The change of the maintenance mode would be the trigger to start the recurring action.
 
 #### Enforcing the Maintenance Mode on the client side
 
@@ -393,6 +414,10 @@ It should be possible to add a URL to an iCalendar file and maybe implement CalD
 ### Assign a Maintenance Schedule to a System
 
 - adding a system to a Maintenance Schedule via activation key at onboarding time. This requires changes on the activation key.
+
+### Schedule Editor
+
+Implement an integrated calendar editor to write own schedules without 3rd party tools.
 
 
 # Drawbacks
--- a/accepted/00000-maintenance-windows.md
+++ b/accepted/00000-maintenance-windows.md
@@ -103,10 +103,10 @@ Every `VEVENT` describe a maintenance window.
 Every `VEVENT` must have a `SUMMARY`. All events with the same "summary" belong to the same schedule.
 The summary must match the schedule name in the Uyuni Database.
 
-#### Timezones
+#### Timezone
 
 The ICalendar entry defines the timezone. Uyuni must of cause respect it. When showing the Maintenance Windows
-in the UI, they should be converted into the timezone set in the user preferences.
+in the User Interface, they should be converted into the timezone set in the user preferences.
 
 
 ### Adding a new Maintenance Schedule
@@ -322,10 +322,21 @@ This Maintenance Window overflow should be show with a notification.
 Ideally we should have only one notification per maintenance window, not per client which cause the overflow.
 It is not an `error`, but it show the admin that the window might needs to get increased for the next runs.
 
+## Cluster
+
+Cluster defined in Uyuni needs to be handled in a special way. All nodes of the cluster are readonly and changes
+are applied only via a dedicated Cluster Manager.
+
+For Maintenance Windows we can define a `None Schedule` which defines no Maintenance Window and assign this schedule to
+all nodes of a cluster. This prevent to schedule any action on the cluster nodes.
+
+The Cluster Provider Manager may not be a system object and therefor need a special assignment of a maintenance schedule.
+More details will follow when we know more about the cluster implementation.
+
 
 ## Second Iteration
 
-### Implement a Maintenance Mode
+### Implement a Maintenance Mode override
 
 Put a system manually into "Maintenance Mode". This is useful in emergency situations where a system is down
 and needs to be fixed quickly.
--- a/accepted/00000-maintenance-windows.md
+++ b/accepted/00000-maintenance-windows.md
@@ -196,20 +196,20 @@ We define a second flag which say, if that action can be executed on systems not
   32 | rhnsd.configure                                 | SUSE Manager Network Daemon Configuration                                                          | N
   33 | packages.verify                                 | Verify deployed packages                                                                           | N
   34 | rhn_applet.use_satellite                        | Allows for rhn-applet use with an Spacewalk                                                        | N
-  35 | kickstart_guest.initiate                        | Initiate an auto installation for a virtual guest.                                                 | Y
-  36 | virt.shutdown                                   | Shuts down a virtual domain.                                                                       | Y
-  37 | virt.start                                      | Starts up a virtual domain.                                                                        | Y
-  38 | virt.suspend                                    | Suspends a virtual domain.                                                                         | Y
-  39 | virt.resume                                     | Resumes a virtual domain.                                                                          | Y
-  40 | virt.reboot                                     | Reboots a virtual domain.                                                                          | Y
-  41 | virt.destroy                                    | Destroys a virtual domain.                                                                         | Y
-  42 | virt.setMemory                                  | Sets the maximum memory usage for a virtual domain.                                                | Y
-  43 | virt.schedulePoller                             | Sets when the poller should run.                                                                   | Y
+  35 | kickstart_guest.initiate                        | Initiate an auto installation for a virtual guest.                                                 | N
+  36 | virt.shutdown                                   | Shuts down a virtual domain.                                                                       | N
+  37 | virt.start                                      | Starts up a virtual domain.                                                                        | N
+  38 | virt.suspend                                    | Suspends a virtual domain.                                                                         | N
+  39 | virt.resume                                     | Resumes a virtual domain.                                                                          | N
+  40 | virt.reboot                                     | Reboots a virtual domain.                                                                          | N
+  41 | virt.destroy                                    | Destroys a virtual domain.                                                                         | N
+  42 | virt.setMemory                                  | Sets the maximum memory usage for a virtual domain.                                                | N
+  43 | virt.schedulePoller                             | Sets when the poller should run.                                                                   | N
   44 | kickstart_host.schedule_virt_host_pkg_install   | Schedule a package install of host specific functionality.                                         | N
   45 | kickstart_guest.schedule_virt_guest_pkg_install | Schedule a package install of guest specific functionality.                                        | N
   46 | kickstart_host.add_tools_channel                | Subscribes a server to the Spacewalk Tools channel associated with its base channel.               | N
   47 | kickstart_guest.add_tools_channel               | Subscribes a virtualization guest to the Spacewalk Tools channel associated with its base channel. | N
-  48 | virt.setVCPUs                                   | Sets the Vcpu usage for a virtual domain.                                                          | Y
+  48 | virt.setVCPUs                                   | Sets the Vcpu usage for a virtual domain.                                                          | N
   49 | proxy.deactivate                                | Deactivate Proxy                                                                                   | N
   50 | scap.xccdf_eval                                 | OpenSCAP xccdf scanning                                                                            | N
   51 | clientcert.update_client_cert                   | Update Client Certificate                                                                          | Y
--- a/accepted/00000-maintenance-windows.md
+++ b/accepted/00000-maintenance-windows.md
@@ -248,7 +248,10 @@ There could be multiple ways to assign systems to maintenance schedules.
 - It should be possible to change the maintenance schedule for multiple systems using SSM
 - In the details of the Maintenance Schedule we could have a systems page where add/remove is possible.
 
-If a Maintenance Schedule is assigned and which one, should be visible in the systems details page.
+When a system is assigned to a new Maintenance Schedule, all future actions for this system should be canceled.
+The list of canceled actions should be shown.
+
+The information if a Maintenance Schedule is assigned to a system and which Schedule, should be visible in the systems details page.
 Implementation of XMLRPC functions for getting and setting the current maintenance schedule must be implemented.
 For setting a schedule should be possible for multiple systems using an array of system ids.
 
@@ -384,8 +387,6 @@ It should be possible to add a URL to an iCalendar file and maybe implement CalD
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-
-- Adding a Maintenance Schedule to a system: what should happen with existing scheduled action? They may be scheduled outside of the defined maintenance windows. It could be actions scheduled for multiple systems. Re-scheduling might not be the right option here. Maybe we do not allow it?
 
 - Pause icon could conflict with virtualization management. Is it a good idea?
 
--- a/accepted/00000-maintenance-windows.md
+++ b/accepted/00000-maintenance-windows.md
@@ -205,7 +205,7 @@ We define a second flag which say, if that action can be executed on systems not
   47 | kickstart_guest.add_tools_channel               | Subscribes a virtualization guest to the Spacewalk Tools channel associated with its base channel. | N
   48 | virt.setVCPUs                                   | Sets the Vcpu usage for a virtual domain.                                                          | Y
   49 | proxy.deactivate                                | Deactivate Proxy                                                                                   | N
-  50 | scap.xccdf_eval                                 | OpenSCAP xccdf scanning                                                                            | Y ?
+  50 | scap.xccdf_eval                                 | OpenSCAP xccdf scanning                                                                            | N
   51 | clientcert.update_client_cert                   | Update Client Certificate                                                                          | Y
  500 | image.deploy                                    | Deploy an image to a virtual host.                                                                 | N
  501 | distupgrade.upgrade                             | Service Pack Migration                                                                             | Y
@@ -213,7 +213,7 @@ We define a second flag which say, if that action can be executed on systems not
  503 | states.apply                                    | Apply states                                                                                       | Y
  504 | image.build                                     | Build an Image Profile                                                                             | N
  505 | image.inspect                                   | Inspect an Image                                                                                   | N
- 506 | channels.subscribe                              | Subscribe to channels                                                                              | Y ?
+ 506 | channels.subscribe                              | Subscribe to channels                                                                              | N
 
 (That list is currently the full action list. Not all actions are still in use)
 
@@ -384,14 +384,6 @@ It should be possible to add a URL to an iCalendar file and maybe implement CalD
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-- Subscribe to channels: should also be possible outside of a maintenance window? 
-  * Pro:
-    * Some actions can only be prepared when the system has the correct channel assignment.
-    * Pre download of packages will only succeed when the correct channels are assigned.
-  * Con:
-    * we change the repository file on the system
-
-- oscap scan: possible outside of maintenance windows?
 
 - Adding a Maintenance Schedule to a system: what should happen with existing scheduled action? They may be scheduled outside of the defined maintenance windows. It could be actions scheduled for multiple systems. Re-scheduling might not be the right option here. Maybe we do not allow it?
 
--- a/accepted/00000-maintenance-windows.md
+++ b/accepted/00000-maintenance-windows.md
@@ -1,0 +1,409 @@
+- Feature Name: Maintenance Windows
+- Start Date: 2020-03-31
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+Define Maintenance Windows in Uyuni and assign them to systems
+
+# Motivation
+[motivation]: #motivation
+
+Customers have company policies to perform changes on systems only at specific points in time.
+This is typically called "Maintenance Window".
+
+Uyuni should be able to define them and prevent executing of actions outside of such a window.
+
+# Architecture
+[architecture]: #architecture
+
+A system is in "Maintenance Mode" or not.
+When a system is not in "Maintenance Mode" no action can be executed.
+Setting the Maintenance Mode "on" and "off" requires Organization Admin permissions.
+A "Maintenance Window" does nothing else than turning Maintenance Mode "on" when the window starts
+and "off" when it ends.
+
+Defining Maintenance Windows requires Organization Admin permissions.
+
+Maintenance Windows are defined on system level.
+Every System can get 1 Schedule assigned.
+A Schedule can be used for multiple systems.
+Inside of the Schedule, multiple Maintenance Windows can be defined.
+They can be defined as individual or as recurring events.
+
+To assign a Schedule to multiple systems, SSM should be used.
+An equivalent XMLRPC function should be implemented as well taking a list of system ids.
+
+
+# Detailed design
+[design]: #detailed-design
+
+## First Iteration
+
+In the first iteration we limit the implementation to load and update maintenance schedules
+and scheduling actions inside of maintenance windows.
+
+Enforcing Maintenance Windows on lower level is part of the future iterations and will be defined later
+in updates of this RFC.
+
+
+### Implement a Maintenance Mode
+
+For now the Maintenance Mode is a method called on a system object which lookup the schedule and check
+if we are currently inside of a defined maintenance window or not. There maybe some functionalities
+which require this information.
+
+Putting a system manual into maintenance mode is a feature of the second iteration.
+
+
+### Maintenance Window
+
+A Maintenance Window specify a time when changes on a system can be applied.
+- It has a start date-time and an end date-time.
+- It can be defined as recurring interval (RRULE).
+- Exceptions should be supported (EXDATE)
+- Events for a full day (anniversary) or multiple days must be considered.
+- The Summary may define the schedule name (See also Maintenance Schedule).
+
+For practical reasons the window describes when an action can be `started`.
+As it is unpredictable to know how long an action will take, we cannot calculate,
+if the action can also be `finished` inside of the Maintenance Window.
+
+A started Action should always be finished to not risk a broken system.
+
+#### Action Chains
+
+Action Chains with reboots are split into multiple actions.
+As the goal is, that everything what was started inside of a maintenance window should also be finished,
+we need to finish whole action chains also after the window closed.
+
+If we won't do it, we would end up with broken systems.
+
+
+### Maintenance Schedule
+
+A Maintenance Schedule is a set of Maintenance Windows. A Maintenance Schedule has a name/label.
+The data should be stored as ICalendar data (https://tools.ietf.org/html/rfc5545).
+In the first iteration importing the data created by an external tool is required.
+Uyuni will not provide a User interface for generating or manipulating Maintenance Windows.
+
+This is a task for future iterations.
+
+To parse the windows we need a library for ICalendar files. E.g. ical4j(https://github.com/ical4j/ical4j).
+
+There could be two types of Schedules for the first iteration
+
+1. One ICalendar file represent one schedule
+
+Every `VEVENT` describe a maintenance window.
+
+2. One ICalendar file contains multiple schedules
+
+Every `VEVENT` must have a `SUMMARY`. All events with the same "summary" belong to the same schedule.
+The summary must match the schedule name in the Uyuni Database.
+
+### Adding a new Maintenance Schedule
+
+In the left sidebar a new page should be added below `Schedule`.
+In this page you can see a list of existing schedules. You can edit them, when you select an existing schedule.
+There should be a "Create Maintenance Schedule" link in the upper left corner which open a new page where the following values can be specified:
+
+- Schedule Name
+- Schedule Type (Single, Multi, ....)
+- Data (ICalendar content) with file upload option
+
+A "Create" button will create the new schedule. The ICalendar data are stored in the database.
+
+Adding a Maintenance Schedule via XMLRPC API must be possible as well.
+
+
+### Updating/Removing a Maintenance Schedule
+
+A Maintenance Schedule will change over the time. A Maintenance Window will be moved or removed on short notice,
+or a new Maintenance Window slot will be created.
+
+This happens while updating the ICalendar data of a maintenance schedule.
+
+While adding a new slot is a non-issue, the movement or removal of a window needs to be handled properly.
+
+As it is nearly impossible to identify a change of a special event we need to follow a different strategy.
+When we refresh or upload a changed ICalendar file, we need to re-evaluate all scheduled actions of systems using the
+new schedule.
+We also need to provide a resolve strategy. This can happen while implementing modules to move actions
+to a different time. Possible modules could be:
+
+- Nearest: Use the nearest maintenance window
+- Previous: Use the maintenance window before the scheduled date
+- Next: Use the maintenance window after the scheduled date
+- Cancel: Cancel the Action
+- Fail: do not import the new schedule at all and keep the existing one
+
+The customer can configure the order:
+
+  resolveStrategy=Nearest,Previous,Next,Fail
+
+The resolve strategy is also needed when updating the Maintenance Schedule via XMLRPC API.
+
+Removing a maintenance schedule from a system can happen at any point in time as all scheduled actions remain valid.
+
+
+### Actions and Maintenance Windows
+
+Some Actions can be executed at any time, others should be executed only inside of Maintenance Windows.
+Every Action in Uyuni has an entry in the Database which already defines if it can be executed on "Locked" Systems.
+We define a second flag which say, if that action can be executed on systems not in maintenance mode.
+
+ id  |                      label                      |                                                name                                                | maintenance_mode_only 
+-----|-------------------------------------------------|----------------------------------------------------------------------------------------------------|----------------------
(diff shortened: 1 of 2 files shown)


accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,99 @@
...
+
+For the first iteration we want to block all actions when the system is not in maintenance mode.
+In future iterations we can consider to allow some actions which are mostly read-only (package list refresh or hardware refresh).
+Also actions which prepare things for the maintenance window could be considered (pre-fetching packages for update).
+
+
+## Maintenance Schedule
+
+A Maintenance Schedule is a set of Maintenance Windows. A Maintenance Schedule has a name/label.
+The data should be stored as ICal data (https://tools.ietf.org/html/rfc5545).
```
hustodemon, line 54: I wonder if we want to store the iCal data as a varchar in the db. If we do it this way, we need to be prepared that some queries, we might need, can be costly (in general, it would mean selecting everything to memory and do the filtering on backend).

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,99 @@
...
+Watching details of a schedule could be done in a calendar widget.
+
+
+## Maintenance Window
+
+A Maintenance Window specify a time when changes on a system can be applied.
+It has a start date-time and an end date-time.
+It can be defined as recurring interval.
+
+When a Maintenance Window start, it turn the maintenance mode "on".
```
hustodemon, line 75: We also need to clarify "how" does the switch happen. We need some mechanism, that is "triggered" when the maint. window starts/ends. The first idea that comes to my mind is that when we create the maintenance window, we need to create 2 taskomatic schedules (1 for turning on the maint. mode for the systems, 1 for turning it off). The one problem I could foresee now is that taskomatic only support cron expressions and I fear iCal allows defining more elaborate frequencies (last day of month).

mcalmer: Do we really need a "value" or could it be a function `isInMaintenanceMode()` which lookup the calendar to decide if it is or not?

accepted/00000-maintenance-windows.md
mcalmer: Or we convert ical data as single events into a DB structure. This requires of cause update capability. On the other hand it would make it easier to get these data also in python and other languages when needed.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,99 @@
...
+To assign a Schedule to multiple systems, SSM should be used.
+
+
+# Detailed design
+[design]: #detailed-design
+
+## Implement a Maintenance Mode
+
+Check, how System Locking is implemented. Copy the mechanism for Maintenance Mode.
+Does it make sense to use System Locking as "Maintenance Mode"?
```
hustodemon, line 44: Publishing a comment about my research of system locking. There are 3 types of locking in Uyuni:

### Package locking
(in the web UI: `Software -> Packages -> Lock`)
- this allows locking some package versions on the system (probably uses zypper locking mechanism under the hood?)
- looks like we need something more general

### System locking (traditional systems only)

(in the Web UI, go to system details, click `lock system`)

- systems locked in such way will not allow running _some_ actions:
- the (black)list of forbidden actions is set in the DB: `select * from rhnActionType where unlocked_only = 'Y';`
- the blacklist is not customizable by the user
- all actions (even the blacklisted) can still be scheduled, when the system is locked
- the actions not in the blacklist are run normally
- as for the blacklisted actions: as soon as the system is unlocked, these scheduled actions are run (i.e. on next execution of the `rhn_check`, supposing the system is unlocked at the time of execution)
- implementation: system lock is an object in the database, there is a `1:1` relation with the system (or server) object
- **problems**
  - if we use it for the maintenance mode, we could collide with the existing locks (the admin can set the lock, independent of any maintenance window. Maybe we don't want to unlock the system that was locked explicitly by some administrator). This could be solved by documenting the behavior ("system locks can be overridden by the maintenance windows mechanism").

### Blackout mode (minions only)
- i.e. locking on "salt level"
- Implementation via a formula (see `system-lock-formula` RPM)
- when the formula is assigned to a system (group, org), the salt module calls will not be executed on the system(s) 
- difference to the system lock for trad. systems: when the formula is assigned, any actions run will **fail**
- allows whitelisting of allowed actions (`test.ping`, `grains.items`, `grains.item`)
- **problems**:
  - same as the previous case

### Q&A
- Q: we need to decide, how we should react when an action is executed outside of the maintenance window
  - A1: Postpone
  - A2: Fail
  - ~A3~: Don't allow executing that action at all. That would be very hard to achieve: actions can be scheduled for future. Maintenance window that forbids this action can be created **after** the action is scheduled.

- Q:  do we want to support both traditional clients and salt minions? if yes, we need to unify the fail/postpone semantics above
  - A: We focus primarily on minions. We could use the blackout mechanism for minions and go with the "fail" semantics above.

- Q: Can we actually use the blackout mechanism for minions?
  - pros:
    - already implemented mechanism
    - works on the salt modules level, which should be powerful and flexible enough
  - cons:
    - can collide with existing code that uses blacklisting (we would need a special handling for CaaSP nodes, which should always stay in blackout mode).
    - implementation via formula: it inherits the common formula integration drawbacks, which we might take into account (assignment on filesystems, not in the database: slower querying, no support for migration scripts)

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,142 @@
...
+- when an action was scheduled as "in next MU", should we re-schedule it to the new window when it appears earlier?
+
+
+## Traditional Clients
+
+They may ask for new jobs only every 2 - 6 hours. Then the maintenance window might be over.
+What to do?
+
+- Execute even after MU is over? (this is what may happen without changes)
+- New features are made for salt - just mention it in the docs and suggest to use osad, a short checking frequence or a very long Window?
```
paususe, line 140: This.

accepted/00000-maintenance-windows.md
parlt91: Agreed we could perform this check every time when scheduling an action and would not need this `maintenance mode` flag. The org admin could get a warning in this case (could decide to perform the action anyway), other users would get an error stating that the system isn't in `maintenance mode` at that time.

accepted/00000-maintenance-windows.md
parlt91: If we convert the iCal into single events how would we do the updating? I guess we don't want to override the entire table every time. (Could do a check what changed on frontend. What about xmlrpc?)

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,142 @@
...
+When its end, it turns the maintenance mode "off".
+
+
+## Assign a Maintenance Schedule to a System
+
+There could be multiple ways to assign systems to maintenance schedules.
+
+- In the systems details page we need to show the selected schedule with a view and a change option.
+- It should be possible to change the maintenance schedule for multiple systems using SSM
+- In the details of the Maintenance Schedule we could have a systems page where add/remove is possible.
```
parlt91, line 85: As I understood it changing the schedule would change the schedule for all systems that follow that schedule.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,142 @@
...
+
+## Schedule Actions
+
+How should it look like?
+
+- limit the datetime picker to show only dates and times with a maintenance window?
+- automatically choose next maintenance window to the given date?
+- set the datepicker to the next maintenance window?
+- change the datepicker into a simple combobox where we display the next maintenance windows
+
```
parlt91, line 120: These datepickers make use of the `bootstrap-datepicker` component that does not allow excluding individual dates, so limiting these to just the ones inside of maintenance windows would be difficult.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+- Fail: do not import the new schedule at all and keep the existing one
+
+The customer can configure the order:
+
+  resolveOrder=Nearest,Previous,Next,Fail
+
+
+### Schedule an Action
+
+When a system is assigned to a schedule, only schedules withing a maintenance window should be allowed.
```
parlt91, line 128: Should be: `...schedule, only actions within a ...`

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+times in the time box change.
+
+When using SSM it is possible that systems have different maintenance windows. The combination systems belong
+to one maintenance schedule and systems without maintenance schedule is not critical. The action would be scheduled
+at a time which is inside of a maintenance window of the systems which belong to the schedule.
+
+Systems in different schedules needs special handling:
+
+1. do not allow systems with different schedules in SSM
+2. divide systems by maintenance schedules and display a combo box with the names. Let the user choose for which set he want to schedule it. He need to repeat it for the rest.
```
parlt91, line 140: Only concern here is that it might get messy to display the systems if many systems belong to either schedule

parlt91:
This looks already pretty good to me.
Double checking with my notes I found that we should probably mention something about the "Emergency mode" (allow scheduling actions for certain users under certain circumstances)

accepted/00000-maintenance-windows.md
mbrookhuis: It should be possible to create the schedule via the API

accepted/00000-maintenance-windows.md
joesusecom: From a usage point of view, there are two different interactions I can imagine:

- admin defines a maintenance schedule and expects the system to be in maintenance mode at the given times
- admin (or an API call triggered by a third party tool/scheduler) puts the system into planned or emergency maintenance mode _now_ or switches it off again.

Of course you can implement both in different ways. For example, if there's only a method to check whether the system has a maintenance mode every time an action is triggered, you'll need to create a (temporary) schedule when the system is set to maintenance manually.

But it may be easier to do it the other way around: Manual maintenance mode changes the flag immediately, and the schedules are checked at certain intervals to set/unset maintenance mode as needed.

You may want to look into the potential performance implications of both, for a typical scenario (i.e. is it going to create more load to set/unset the flags for thousands of machines when a window is hit or to calculate the status on the fly based on the schedule). There must be prior art. Calendaring tools are doing this (busy status) all the time. 😄 

Additional considerations:

- Would we want to bring back a system from maintenance mode only at the end of the window or should there be away of switching it off once all maintenance tasks have been finished (which could be much earlier than the end of the window)? I'd think that most admins would _love_ to have the systems come back sooner (which could also be used in uptime statistics to prove that SLAs are met/improving). But the corner case is when an admin figures out later (but still within the maintenance window originally scheduled) that there's more to do.
- What happens if a system is in manual maintenance mode and the scheduled maintenance window kicks in? I guess an admin wouldn't want a system to come back from maintenance automatically at the end of the window if the reason for the manual override hasn't been resolved yet.
- What if a long-running task _exceeds_ the scheduled window and can't be rolled back quickly? In other words, unless the task gets finished, the system will be in an unusable state. Yes, this should never happen. But if it _does_, I'd think the admin would rather see the system flagged as in maintenance until things are finished than get the system back to production while it's still broken.

In the end, this could mean that we need to separate between "actual maintenance mode" and "scheduled maintenance mode" (with the actual one being driven manually or through exit criteria (e.g. full maintenance task action chain has been completed successfully) and the scheduled on just indicating what _should_ be based on the schedule.

mbrookhuis:
can we define also a set of actions that should be executed during the maintenance and this should apply for all re-occurring appointments too. For actions think of:
- highstate
- remote command
- patch
- reboot
- highstate

And report the assigned admins (and when enabled) that the maintenance is finished or having issues.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+
+## What does it mean to turn maintenance mode "on" and "off"
+
+- A flag in DB which needs to be changed?
+- We evaluate the calendar and calculate the state of the system?
+- mix: taskomatic job (every minute) which evaluate the calendar and switch the flag (and trigger any other action which is needed)
+
```
joesusecom, line 171: The part that's missing here is that systems that are using Maintenance Windows should probably be locked (like cluster nodes) while they're _not_ in maintenance mode. This will likely affect the implementation, as the locking needs to be applied and unapplied (so just doing a DB query `is_in_maintenance_mode` or so won't do the job.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+## Traditional Clients
+
+They may ask for new jobs only every 2 - 6 hours. Then the maintenance window might be over.
+What to do?
+
+- Execute even after MU is over? (this is what may happen without changes)
+- New features are made for salt - just mention it in the docs and suggest to use osad, a short checking frequency or a very long Window?
+- Special implementation?
+
```
joesusecom, line 183: My guidance for traditional clients is: Focus on Salt, but if there's no practical difference, I'm super-fine with getting the traditional stack support for free.

_If_ we end up being able traditional clients, my expectation (to be validated by someone in the field) would be that Uyuni strictly enforces the windows. So if a client checks in too late, bad luck and come back in the next window.

That said we'd most likely be able to just point to OSAD.

BTW, one thing we may want to check is whether admins are _currently_ using the run_check to implement their own "maintenance windows" (by only running the run_check when they actually want to do maintenance).

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+[drawbacks]: #drawbacks
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+
```
parlt91, line 165: What to do with **recurring actions**?
I see two possibilities:
1) Don't allow scheduling recurring events / disable existing events if there is a maintenance window defined.
2) Only execute the (recurring) action if the moment of execution is inside a defined maintenance window.

accepted/00000-maintenance-windows.md
mcalmer: Added some notes about XMLRPC API usage to the RFC.

accepted/00000-maintenance-windows.md
mcalmer: To be able to plan, we need to know about maintenance windows upfront.
A 3rd party application which just set the maintenance mode "on" and "off"
is not helpful.

The emergency case is special. It can set the maintenance mode "now".
You cannot predict when an emergency happens, so there also no need to "plan"
actions for it.
When it happen, you set the system in maintenance mode and create new actions
wich should be exectuted "now".

This means when we check the schedule for maintenance windows we need to consider
the current state. If the system is currently in maintenance mode, we allow to schedule
actions "now" - maybe also for the next minute.

I will add a section about this case

accepted/00000-maintenance-windows.md
mcalmer: as we decided to not enforce the maintenance mode on the system itself in the first iteration I will mark this discussion as resolved.
This will come back when we discuss the 2nd iteration.

accepted/00000-maintenance-windows.md
mcalmer: adapting the datepicker is tricky. Already suggest the combobox variant.

accepted/00000-maintenance-windows.md
mcalmer: fixed

accepted/00000-maintenance-windows.md
mcalmer: I will add "next available" to the option.
It is a bit unspecified, but as we defined a resolve strategy for removing a window, it is a doable option.

mcalmer:
> can we define also a set of actions that should be executed during the maintenance and this should apply for all re-occurring appointments too. For actions think of:
> 
> * highstate
> * remote command
> * patch
> * reboot
> * highstate
> 
> And report the assigned admins (and when enabled) that the maintenance is finished or having issues.

You want actions which can be executed independent of the maintenance mode.
Well, this is tricky.

With the currently planned 1st iteration it would be possible to define "actions" which need to keep the maintenance mode into acount and "actions" which can be executed always at any time.

But when we start to enforce the maintenance mode on the client, we cannot differentiate anymore between "highstate" and "package list refresh" as both are a state.apply.

If this is wanted, we need to decide that maintenance mode is a "SUSE Manager" feature and pure salt can still overwrite everything.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+Customers have company policies to perform changes on systems only at specific points in time.
+This is typically called "Maintenance Window".
+
+Uyuni should be able to define them and prevent executing of actions outside of such a window.
+
+# Architecture
+[architecture]: #architecture
+
+A system is in "Maintenance Mode" or not.
+When a system is in "Maintenance Mode" no action can be executed.
```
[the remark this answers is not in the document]
mcalmer: totally right. I will change it

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+Setting the Maintenance Mode "on" and "off" requires Organization Admin permissions.
```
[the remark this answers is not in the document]
mcalmer: Sounds like an interesting idea.
I will add this to the open questions for now to have a discussion around it.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+Defining Maintenance Windows requires Organization Admin permissions.
+
+Maintenance Windows are defined on system level.
+Every System can get 1 Schedule assigned.
+A Schedule can be used for multiple systems.
+Inside of the Schedule, multiple Maintenance Windows can be defined.
+They can be defined as individual or as recurring events.
+
+To assign a Schedule to multiple systems, SSM should be used.
+
```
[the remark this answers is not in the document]
mcalmer: > If I'm not mistaken, this also means that we would easily be able to create API queries or
> UI views that show all systems that have a certain schedule assigned, right?

Yes, should be possible

I also added a "Second Iteration" section to the RFC and added there the activation key story.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+- Exceptions should be supported (EXDATE)
+- Events for a full day (anniversary) or multiple days must be considered.
+- The Summary may define the schedule name (See also Maintenance Schedule).
+
+### Maintenance Schedule
+
+A Maintenance Schedule is a set of Maintenance Windows. A Maintenance Schedule has a name/label.
+The data should be stored as ICalendar data (https://tools.ietf.org/html/rfc5545).
+In the first iteration importing the data created by an external tool is required.
+Uyuni will not provide a User interface for generating or manipulating Maintenance Windows.
```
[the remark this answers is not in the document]
mcalmer: I tried korganizer and office365 outlook. We should definitive test some of them.

mbologna:
An action chain is started in the maintenance window. Before all the actions in the action chains are executed, the maintenance window ends. What happens?

accepted/00000-maintenance-windows.md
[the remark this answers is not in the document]
mcalmer: I add this to the question section.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,191 @@
...
+
+In the left sidebar a new page should be added below `Schedule`.
+In this page you can see a list of existing schedules. You can edit them, when you select an existing schedule.
+There should be a "Create Maintenance Schedule" link in the upper left corner which open a new page where the following values can be specified:
+
+- Schedule Name
+- Schedule Type (Single, Multi, ....)
+- Data (ICalendar content) with file upload option
+
+A "Create" button will create the new schedule. The ICalendar data are stored in the database.
```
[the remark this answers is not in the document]
mcalmer: I added this to "Second Iteration" to not forget about it.

accepted/00000-maintenance-windows.md
mcalmer: Well, we put exactly that to future iterations as this brings a lot of extra work.
When doing this in the first iteration we may not be able to deliver in time.

This is also the reason why it is in the open questions section.

On the other hand I am not 100% sure if this is really needed in the end.
Without it, it is a "SUSE Manager" feature and when you connect directly via salt nobody will hold you back on doing stupid things.
But maybe this is good enough for a lot of people. I think we will find out via feedback on this.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,208 @@
...
+# Detailed design
+[design]: #detailed-design
+
+## First Iteration
+
+In the first iteration we limit the implementation to load and update maintenance schedules
+and scheduling actions inside of maintenance windows.
+
+Enforcing Maintenance Windows on lower level is part of the future iterations and will be defined later
+in updates of this RFC.
```
mbologna, line 47: IIUC, this means that the implementation logic that actually blocks an action from being executed will be defined in future updates? I spoke with @hustodemon about Salt and traditional locking, and their conceptual differences. It might be a good occasion to unify the concept of "locking" in case we want to use it as a way to prevent action execution outside maintenance windows.

accepted/00000-maintenance-windows.md
mcalmer: This is similar to what I think. When customer uses osad, it should just work. We schedule for a time in the maintenance window and osad will trigger at that time.

Only when not using osad possible problems might happen.
But with documentation and long enough maintenance windows it should be doable.

accepted/00000-maintenance-windows.md
moio: So do I get it correctly that this will be enforced for Actions only, and not at the Salt level?

accepted/00000-maintenance-windows.md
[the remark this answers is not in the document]
hustodemon: I was also thinking like that subconsciously: having a system in _that_ mode means the system is "locked". Maybe it's because I'm used to the "lock" and "blackout" mode that this similar semantics.

I _really_ like the maintenance mode/window/schedule terminology and can't think of anything better with the "inverse" semantics (the best I could come up with, was "production" or "protected", but I don't like this, or "locked", which we already use), but maybe somebody else has a better idea. For me, the "maintenance" is still the best option.

accepted/00000-maintenance-windows.md
hustodemon: At least initially, yes. As written above, the "blackouting" (locking on the salt modules level) the minions **could** come up later (2nd iteration).

accepted/00000-maintenance-windows.md
mcalmer: thanks. added

accepted/00000-maintenance-windows.md
mcalmer: Just found out that we might not be able to enforce this on minion level.
It would effect docker and kiwi image building. But these actions should happen when the system is **not** in maintenance mode.
If we really want to enforce it, we need a big refactoring and need to implement an own daemon to build images and exclude it from `salt`.

moio: From a tech POV, I presently cannot imagine a unique solution to enforce this both at the Action level and the Salt level - I guess we will always somehow need two almost-separate code changes, as underlying mechanisms are completely different.

From a product POV, I also recognize this concept clicks much better with the Action model, and could be seen as added value on top of plain Salt.

On the other hand though, I see users having to or preferring to avoid Actions and work directly with Salt, at least in part, in many cases.

So my only recommendation would be: if possible, try to confirm from more-user-oriented colleagues that addressing Actions first is actually more valuable than having an (initial) Salt-only solution (I can easily imagine that being actually the case, just playing devil's advocate here).

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,390 @@
...
+
+Defining Maintenance Windows requires Organization Admin permissions.
+
+Maintenance Windows are defined on system level.
+Every System can get 1 Schedule assigned.
+A Schedule can be used for multiple systems.
+Inside of the Schedule, multiple Maintenance Windows can be defined.
+They can be defined as individual or as recurring events.
+
+To assign a Schedule to multiple systems, SSM should be used.
```
moio, line 35: SSM is a GUI-only concept. Can we expect an equivalent on the API side?

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,390 @@
...
+ 506 | channels.subscribe                              | Subscribe to channels                                                                              | Y ?
+
+(That list is currently the full action list. Not all actions are still in use)
+
+
+### Schedule an Action
+
+When a system is assigned to a maintenance schedule, actions which are flagged as `maintenance_mode_only` should be executed only within a maintenance window.
+In the User Interface we should show in these cases not the datepicker, but a simple Combo box with the Date and Time of the next 20 Maintenance Windows.
+All actions will be scheduled at the begin of the Maintenance Window.
```
moio, line 222: If we constrain all `maintainance_mode_only` Actions to be scheduled at the beginning of a Maintenance Window, does it make sense to talk about Maintenance Windows at all? Is this constraint expected to be permanent, or to be lifted in a successive iteration?

If it is expected to be permanent, wouldn't it be easier to "rename" this whole concept to "Maintenance Start Times" and ignore lengths altogether?

(the only difference would be to prevent out-of-window scheduling via XMLRPC - arguably though if a user already uses XMLRPC he might have a scheduling already)

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,390 @@
...
+- Schedule Name
+- Schedule Type (Single, Multi, ....)
+- Data (ICalendar content) with file upload option
+
+A "Create" button will create the new schedule. The ICalendar data are stored in the database.
+
+Adding a Maintenance Schedule via XMLRPC API must be possible as well.
+
+
+### Updating a Maintenance Schedule
```
moio, line 120:
```suggestion
### Updating/Deleting a Maintenance Schedule
```

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,390 @@
...
+### Assign a Maintenance Schedule to a System
+
+There could be multiple ways to assign systems to maintenance schedules.
+
+- In the systems details page we need to show the selected schedule with a view and a change option.
+- It should be possible to change the maintenance schedule for multiple systems using SSM
+- In the details of the Maintenance Schedule we could have a systems page where add/remove is possible.
+
+If a Maintenance Schedule is assigned and which one, should be visible in the systems details page.
+It should be also part of the `system.getDetails()` XMLRPC API function.
```
moio, line 246: Or possibly a separate endpoint, as `getDetails` contains already quite a lot of information.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,390 @@
...
+
+## System in Maintenance Mode
+
+The current status of the system should be reflected in the User Interface and API.
+
+If a system is in Maintenance Mode, the User Interface could show a special icon [ || ] (pause symbol of an audio player).
```
moio, line 252: Not sure if such an icon is a good idea, it could be ambiguous with virtualization pausing which some hypervisors support. I agree with the general idea to have it reflected in the UI somehow, just maybe we will want a different icon or just a label.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,390 @@
...
+- korganizer
+- Lightning (Thunderbird Extension)
+- evolution
+
+
+## Notifications
+
+When a system is assigned to a maintenance schedule, every finished action should check, if the system is still in maintenance mode.
+If this is not the case it means, the action was started inside of the maintenance window, but was not able to finish in time.
+This Maintenance Window overflow should be show with a notification.
```
moio, line 304: Ideally we will not have one notification per system, or those could easily overflow the UI (imagine a big batch of systems all more-or-less simultaneously missing their window).

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,390 @@
...
+[drawbacks]: #drawbacks
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- performance: what happens when a lot of actions are scheduled at the beginning of a maintenance window? Will this create a performance problem? Is salt batching ready to handle such situations properly?
```
moio, line 379: We should be fine as long as we create one Action per type and timestamp with multiple clients associated. Specifically, we should try to avoid as much as possible the situation in which there is one Action per client.

Batching will limit the number of clients executing an Action (only considering that one Action).

My impression is that this proposal would not make the situation worse than it is now - of course users will still have to pay a bit of attention not to create too many separate Actions or batching won't help them. But by and large, they should be able to create Actions targeting many clients just like before - just "taking one pre-defined start datetime from a list" instead of a specifying one "freely".

We will have to pay attention when, for example, rescheduling Actions due to updates to their Maintenance Schedule - clients will have to stay grouped together as much as possible.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,390 @@
...
+As it is nearly impossible to identify a change of a special event we need to follow a different strategy.
+When we refresh or upload a changed ICalendar file, we need to re-evaluate all scheduled actions of systems using the
+new schedule.
+We also need to provide a resolve strategy. This can happen while implementing modules to move actions
+to a different time. Possible modules could be:
+
+- Nearest: Use the nearest maintenance window
+- Previous: Use the maintenance window before the scheduled date
+- Next: Use the maintenance window after the scheduled date
+- Cancel: Cancel the Action
```
moio, line 138: Would it be possible to only implement a subset of these, for example only Cancel, for a first version?

I might be wrong, but in my eyes this functionality is quite an important complexifier.

accepted/00000-maintenance-windows.md
paususe: > If we constrain all `maintainance_mode_only` Actions to be scheduled at the beginning of a Maintenance Window, does it make sense to talk about Maintenance Windows at all? Is this constraint expected to be permanent, or to be lifted in a successive iteration?

I don' think we should schedule everything at the beginning of the maintenance window, at least if there are multiple actions to be scheduled, i. e. do not overlap actions.

> If it is expected to be permanent, wouldn't it be easier to "rename" this whole concept to "Maintenance Start Times" and ignore lengths altogether?
> 
> (the only difference would be to prevent out-of-window scheduling via XMLRPC - arguably though if a user already uses XMLRPC he might have a scheduling already)

Maintenance windows are a well-established concept in the ITSM world, I don't think there's a need to redefine concepts or try to outsmart sysadmins.

In the real world, it works like this:

* "as my maintenance window is from 2 AM to 6 AM on Saturday and Sunday night, I will plan to start this maintenance procedure at 3 AM on Saturday and I [human] have estimated this will be finished at 4 AM, because I have done my testing in TEST and PRE [which in theory is identical to PRO] environments"
* Why 3 AM instead of 2 AM, which is the time the maintenance window starts? Because it turns out this week there is some other procedure at 2 AM, at the beginning of the maintenance window. It does not even need to be for this server, but given that a human operator is always supervising or executing changes, a procedure on server A at 2 AM may prevent another procedure on server B at 2 AM due to lack of humans
* If for some reason (human or machine) the procedure does not finish on time, there are two possibilities:
  * Rollback is possible and/or desirable: a final "go/no-go" decision is made at e. g. 5 AM (whatever is enough to cancel the procedure and perform the rollback)
  * Rollback is not possible and/or undesirable: let the procedure finish at e. g. 7 AM, and register an out-of-window event (incident) in the ITSM tool (e. g. ServiceNow or BMC), calculate SLIs/SLAs, etc

moio: > I don' think we should schedule everything at the beginning of the maintenance window, at least if there are multiple actions to be scheduled, i. e. do not overlap actions to be executed on the same client.

Note: both the traditional and the Salt stack queue execution of Actions if they were to be concurrently on the same client.

mbrookhuis: Subscribe to (software) channels should always be possible.

accepted/00000-maintenance-windows.md
mcalmer: will add XMLRPC function for this as well taking a list of systemids

accepted/00000-maintenance-windows.md
mcalmer: It seems there is a preference to make it at least possible to schedule an action later in the maintenance window. I think I have a good idea how to make it easy possible.

We would implement a combobox with the next 20 maintenance windows and additionally an entry field with "Offset in Minutes" which has the default `0`. Using the offset, you can schedule an action 60 Minutes after the maintenance window starts. I also think remember an offset is easier than an exact time.

I will add this to the RFC

mcalmer: > Subscribe to (software) channels should always be possible.

Will add this to the unresolved questions and we decide in the afternoon.

accepted/00000-maintenance-windows.md
mcalmer: Removing a whole schedule is easy, as it does not invalidate any scheduled action. No schedule means permanent maintenance mode. You can execute actions all the time.

But I will add this info to this section.

accepted/00000-maintenance-windows.md
mcalmer: Changed that part and name also "setting" the schedule via API and name also SSM.
I have removed the explicit naming of this function. We will decide at implementation time what is the best thing to do.

accepted/00000-maintenance-windows.md
mcalmer: Noted. Will discuss in the afternoon.

accepted/00000-maintenance-windows.md
mcalmer: added to the RFC.

accepted/00000-maintenance-windows.md
mcalmer: I think it is possible to implement only some of them. But I am not sure if this is the issue.
I see the general check for all scheduled actions as the bigger problem.

We will see how far we can get.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,409 @@
...
+
+For Uyuni users Linux tools might be interesting
+- korganizer
+- Lightning (Thunderbird Extension)
+- evolution
+
+
+## Notifications
+
+When a system is assigned to a maintenance schedule, every finished action should check, if the system is still in maintenance mode.
```
hustodemon, line 307: Moreover, we need to check, if the system is still in the same maintenance **window**. What could happen, is that a long running action could overflow to a next maintenance window and everything would look ok.

accepted/00000-maintenance-windows.md
```diff
@@ -0,0 +1,409 @@
...
+If this is not the case it means, the action was started inside of the maintenance window, but was not able to finish in time.
+This Maintenance Window overflow should be show with a notification.
+Ideally we should have only one notification per maintenance window, not per client which cause the overflow.
+It is not an `error`, but it show the admin that the window might needs to get increased for the next runs.
+
+
+## Second Iteration
+
+### Implement a Maintenance Mode
```
hustodemon, line 316: "Maintenance mode" has been mentioned multiple times before, and now there is "Implement a Maintenance Mode" section. I think it'd be better to explicitly mention, that this is the maintenance mode "override" (or "flag").

```suggestion
### Implement a Maintenance Mode override
```

accepted/00000-maintenance-windows.md
moio: Just a reminder: changing software channel subscriptions for minions might imply product package installation!

accepted/00000-maintenance-windows.md
mcalmer: From my point of view not a hard requirement. At least not for the first iteration.
I think a typical customer define for a system only 1 window per week.
If that overflow to the next week without anybody notice it, they have a different problem.

accepted/00000-maintenance-windows.md
mcalmer: But only when a highstate is called. (Traditional clients are different, but they are somehow excluded when thinking about this feature)

moio: Right, I forgot about that. Should not be a worry then.

mcalmer:
> Some final comments:
> 
> * Being able to toggle (switch on/off) the maintenance mode (without having to assign a new schedule) seems to be a very important feature to me, especially when it comes to integrating with tools like ServiceNow or check_mk. Fine with labeling it phase two, but let's address it soon after, before putting more effort into other "refinements".

I will create an issue for it which can be scheduled by Pau

> * The "resolve strategy" for handling scheduled actions that wouldn't fit into a window after the schedule has changed seems overly ambitious/complicated to me. No text changes required, but let's make sure we don't get held up by this. In most cases, admins will know what they're doing, and having to reschedule tasks manually after the maintenance window has changed sounds like a perfectly valid approach.

We will try to keep the time short here. But remember that you cannot "re-schedule" an action from UI or API. You can only cancel it and re-create it. So the idea was for MVP to just offer "Cancel" but provide a list of actions which will be canceled. With this list the admin could hopefully re-create them.

If time permits, we may implement also another strategy. Maybe also as Maintenance Update.

> * One thing that doesn't become very clear in the RfC is that adding/removing a schedule basically also means that a system changes from being unprotected from admins to being protected during the non-scheduled times. We should make that clear both in naming the APIs and in the UI.

Added to doc issue

> * Maintenance mode vs. locking: It could be very confusing to explain what happens to a system that is both locked and uses maintenance windows. I'm fine per se with keeping those two concepts separate, but we really need to document well (both in the status UI and the actual documentation) what one can do to a system at any given point in time and who can change the mode/lock how.

Added to doc issue

> * Can we please put the list of actions that are allowed/blocked into the documentation? This is great stuff, but shouldn't require a database inspection tool to verify. :-)

Added to doc issue

> * the virt.* actions and kickstart_guest.initiate functions need to be excluded from the maintenance-only mode: Running VMs is production use for a VM host!

Done

> * We'll need to plan for a follow-up on handling VMs. For example, as long as we don't have automated live migration of VMs, a VM will automatically be affected by maintenance mode when the host is rebooted or becomes unstable because of ongoing package updates! So, from an SLE perspective, _all_ VMs on a host that's under maintenance may be down! On the other hand, the VMs can't just inherit the schedule because
>   
>   * when the host is down, you _can't_ actually update them
>   * you may want to have maintenance windows for them at times when the host is definitely up!

Added to 2nd iteration epic as topic (to not forget it)

> * For scheduling actions, we really need to have another RfC related to this that handles storing re-usable action chains. Having to schedule tasks one by one with enough "safety margin" is just painful.

We have already an issue for it https://github.com/SUSE/spacewalk/issues/3439 . It was created when we made the first step for action chains. I will just add it to the planning board.

> * Idea for later: We could display all scheduled actions and windows in a web calendar at some point!

Added to 2nd Iteration epic

mcalmer:
> Add-on comments on the 2nd iteration chapter:
> 
> **Maintenance Mode**
> The "Maintenance Mode override" only covers one of the two use cases: Emergencies. The other one is when a third party tool (e.g. the monitoring tool, service management application, or CMDB) actually controls the maintenance modes. In that case, instead of having to export calendars (which those tools may not even support), we'd need an action API anyway that allows one to toggle the maintenance mode for one or more systems.
> 
Well, but without calendar you cannot schedule anything. This is a problem. Such a Tool could enable the maintenance mode, but nothing is prepared at that point in time. So this makes sense only when the admin sitting in front of SUSE Manager and execute the actions "now".

Other option would be to change the recuring action feature. You prepare a recurring action which run automatically when the maintenance window is turned "on".
Here we need to enhance the RFC a bit more.

I just added this to the RFC
> 
> **Salt**
> Whether or not to enforce the maintenance mode on the Salt side is a design question that goes beyond this particular topic. We haven't implemented ACLs for Salt either. So what we need to discuss is whether it's generally fine even in the future that only actions that go _through_ SUSE Manager (UI or API) are protected by our policies (ACLs, maintenance windows, locking settings, cluster awareness, ...) or whether we also need to implement them for _native_ Salt (salt CLI, Salt API) at some point.

Aggreed, needs further discussions and we need to bring more people into it.

> **Native editing**
> While using the iCalendar approach is a great design idea, we may also want to integrate an actual schedule editor at some point (which could be just a generic calendar plugin that translates the UI input into an iCalendar file).

Added to RFC 2nd It.

> **Visualizing the calendars**
> We could have added a sentence or two about the plan to visualize calendars.

Added to RFC 2nd It.

PR #40, opened 2020-07-03
Basic support for Salt Orchestration in Uyuni.

mateiw:
Bits and pieces for making use of Salt orchestration in Uyuni. The initial scope is limited to the use cases needed for clustering support.

[TEXT](https://github.com/uyuni-project/uyuni-rfc/pull/40/files?short_path=b0c7d9f#diff-b0c7d9fbfc123b808005dd45d09f6775)

Prototype: https://github.com/uyuni-project/uyuni/pull/2390

### DIFF
--- a/accepted/0000-basic-orchestration.md
+++ b/accepted/0000-basic-orchestration.md
@@ -35,7 +35,7 @@ Salt supports asynchronous runner calls but it doesn't allow attaching metadata 
 In order to store the `jid` and the executions outcome a new table is needed:
 
 ```sql
-CREATE TABLE IF NOT EXISTS rhnActionSaltRunnerJob (
+CREATE TABLE IF NOT EXISTS rhnOrchestrationAction (
 
     action_id       NUMERIC NOT NULL
                             CONSTRAINT rhn_actsaltrunjob_aid_fk
@@ -115,8 +115,8 @@ Why should we **not** do this?
 # Alternatives
 [alternatives]: #alternatives
 
-- What other designs/options have been considered?
-- What is the impact of not doing this?
+- Use the Thorium Complex Reactor instead of Orchestration ?
+- Create a separate entity hierarchy for orchestration instead of reusing Action ?
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/accepted/0000-basic-orchestration.md
+++ b/accepted/0000-basic-orchestration.md
@@ -10,9 +10,10 @@ Implement basic support for Salt Orchestration in Uyuni.
 # Motivation
 [motivation]: #motivation
 
-The initial use case if for improving clustering support. There are two main use cases requiring orchestration:
+The initial use case if for improving clustering support. Use cases requiring orchestration:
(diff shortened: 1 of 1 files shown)


mateiw:
CC @moio @mcalmer @mbologna @cbosdo

mbologna:
I am generally in favor of this change and of the RFC per se, but I wonder:

- How the orchestration would compare with Action Chains? I see some overlapping for minions (action chains are for traditional systems too). Would the action chains be superseded by this minion-only feature?

- How this RFC would compare with https://github.com/uyuni-project/uyuni-rfc/blob/master/unimplemented/00000-salt-orchestration.md ?

accepted/0000-basic-orchestration.md
```diff
@@ -0,0 +1,125 @@
...
+
+  * obscure corner cases
+  * will it impact performance?
+  * what other parts of the product will be affected?
+  * will the solution be hard to maintain in the future?
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Use the Thorium Complex Reactor instead of Orchestration ?
```
mbologna, line 118: What would be the consequences of using Thorium? 
Additionally, can we use original Salt Reactor and Thorium at the same time?

accepted/0000-basic-orchestration.md
```diff
@@ -0,0 +1,125 @@
...
+
+The `salt-netapi` library already support calling Salt runner. 
+
+The Uyuni actions must be enhanced to make Salt runner calls besides minions calls.
+
+## Action enhancements.
+
+A runner action won't need a collection of `ServerAction` objects because the action is not executed on a set of minions directly.
+
+Salt supports asynchronous runner calls but it doesn't allow attaching metadata like `action-id`. Therefore the `jid` returned by Salt must be stored in the database in order to match the return event to the originating action.
```
moio, line 31: What would be pros and cons of extending Salt so that runner calls can also have arbitrary metadata?

accepted/0000-basic-orchestration.md
```diff
@@ -0,0 +1,125 @@
...
+    </join>
+</subclass>
+
+```
+
+### Action result display
+
+Actions executed on a system are shown in the Events tab of the system details. A similar approach should be used for orchestration results. They should be displayed in the details of the target entity or in a generic orchestration UI.
+
+In the case of clusters, the UI should be enhanced to show a view similar to the system Events tab.
```
moio, line 92: Currently, an event result is nothing more than a status and basically a string with the result itself. Thinking about orchestration, I can imagine we might want a more detailed view of some sort - ie. a single long string might be a bit basic, especially if lots of nodes are involved.

How do you envision such a UI?

moio (commented):
I am interested in how this topic evolves, so left some more questions.

BTW, are there any concrete plans to proceed on this one?

moio:
As cluster support in general is currently being re-evaluated and the author of this PR has left the Team, I am for the time being archiving this PR.

Please note this is only in the interest of readability of the repo and GitHub project - we are open to revisit these ideas whenever there is a concrete need.

In any case thanks for this contribution @mateiw

PR #41, opened 2020-07-30
Proposal of including user stories as part of our Development cycle

srbarrios:
Please read this definition of user stories to understand what I was proposing.
https://en.m.wikipedia.org/wiki/User_story

Here a template:
User_story_template_screenshot

Here a video explaining it:
https://youtu.be/eYZ8vBXL1kk

And here a link explaining the link between user stories to our test cases:
https://blog.testlodge.com/writing-test-cases-from-user-stories-acceptance-criteria/

Hope it helps to understand the missing part that we were rising some time ago. So, what we are missing from an agile development cycle is the **user stories**, to then be able to translate them into code and at same time into test cases, using our BDD framework. So, even if that proposal comes from QA Squad, this is not QA related (nothing like test description, test plan or test case is involved on the issue), instead is just an additional level in the process:

**Business case**, described in the RFC -->  **User Stories**, described in the RFC our outside it ---> **Design** described in the RFC --> **Implementation** --> Ideally **test cases** should go before the implementation to follow BDD/TDD, but currently we are doing after implementation stage (which is good enough)

For more details of the full workflow and the additional step proposed (naming is not important, let's call it user stories or acceptance criteria), read my comment in https://confluence.suse.com/display/qacss/How+to+write+a+Test+Description

### DIFF
--- a/0000-template.md
+++ b/0000-template.md
@@ -57,7 +57,7 @@ Description:
 
 Acceptance criteria:
 - Given `<some context>`
-- When <some action is carried out>
+- When `<some action is carried out>`
 - Then `<a set of observable outcomes should occur>`
  
  What is a user story?
--- a/0000-template.md
+++ b/0000-template.md
@@ -44,3 +44,23 @@ Why should we **not** do this?
 
 - What are the unknowns?
 - What can happen if Murphy's law holds true?
+
+# User stories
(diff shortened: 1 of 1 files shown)


srbarrios:
Kostas will present it on our internal PHP, using an internal card.

PR #42, opened 2020-10-29
Hub ISS RFC: clarifications

moio:
We made a second ISS v2 PoC, and the lessons learned are encoded in the changes to this RFC now.

### DIFF
--- a/accepted/00063-hub-iss.md
+++ b/accepted/00063-hub-iss.md
@@ -187,8 +187,10 @@ The first implementation of this RFC will:
 ## Implementation details
 
 - two commandline tools will be added, "exporter" and "importer"
-  - output of the "exportter" will be a directory with files and a SQL script
+  - output of the "exporter" will be a directory with files, a SQL script, and if needed an extra metadata file
+    - the metadata file will contain any extra piece of information useful at the importing end, such as the export version or the list of the included content identifiers (eg. channel labels)
   - the "importer" program will mainly copy files to correct locations and execute the SQL script
+    - the importer program might as well run some pre-condition check, for example making sure the channel selection is valid (any parent channel exists)
   - triggering of the "importer" tool might happen via Salt, this is left as an implementation detail
 - all new code will be in Go
 - inspection of the schema can be implemented via `SELECT` queries on Postgres's internal schema tables
@@ -196,6 +198,8 @@ The first implementation of this RFC will:
 
 ### Software channel specific implementationd details
 
+- it will not be possible to import child channels without their corresponding parents
+- it will be possible to import cloned channels without their originals, although this might break some features (notably CVE Audit and SP migration)
 Once a channel has been imported, the following needs to happen in order to refresh cache and other ancillary tables:
 - `ANALYZE` should be performed on `rhnChannelPackage` and possibly other heavily-changed tables
 - the `rhn_channel.refresh_newest_package(channel_id, 'inter-server-sync');` stored procedure should be called. This updates the `rhnChannelNewestPackage`/`rhnChannelNewestPackageAudit` tables so that they do not have to be transferred
@@ -216,6 +220,8 @@ INSERT INTO rhnErrataQueue (channel_id, errata_id, next_action)
 ```
 Those instruct Taskomatic to regenerate the repo metadata files and to regenerate entries in `rhnServerNeededCache`.
 
+The decision whether to perform these actions via an XMLRPC API or directly in the exported SQL script is left as an implementation detail.
+
 ## Impact on existing components and users
 All code would be new in a new component, so no change to existing components is expected. No impact on existing users is expected at all until the old implementation is dropped.
 
@@ -223,6 +229,8 @@ When that happens, there is a chance of regressions for ISS users, which represe
 
 # Drawbacks
 [drawbacks]: #drawbacks
+
+## Versioning
 
 Syncing content between different versions of Uyuni may not work, if the database schema changes in non-compatible ways.
 
@@ -239,6 +247,11 @@ With the proposal in this RFC, it is possible to implement protection from case 
 pl/SQL could still be used, as a last resort, to apply different commands to different ISS Slave versions, but that would be hardly maintainable, so it would only be suggested to handle critical situations temporarily.
 
 General recommendation would be to upgrade an ISS Master, then upgrade ISS Slaves before attempting any new Inter-Server Synchronization. Another RFC will detail how Hub could make such upgrades easier.
+
+## Others
+
+- assumption is that a Server will not import content from more than one exporting Server (current implementation allows multiple Masters). This limitation could be lifted at a later stage
+- current assumption is that content will be added to the Organization with the same name at the receiving end. Organization mapping might come as a later step
 
 
 # Alternatives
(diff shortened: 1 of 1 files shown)


accepted/00063-hub-iss.md
```diff
@@ -40,35 +40,48 @@ From the requirements:
...
+**First core idea**: a "database exporter" component runs on a Server and dumps data in form of a special SQL script. On the target Server, data is then imported by executing the SQL script against the local database.
 
-The current implementation serializes ISS Master database data to XML and deserializes it on the ISS Slave. Relatively complicated logic is in place on both communication ends to convert to and from this format.
+Dumped data will have to be limited to a specified piece of content to sync (eg. a software channel label, or a set of such labels).
 
-**First core idea** of this RFC is to **export database data from the ISS Master in form of a SQL script to be executed as-is on the ISS Slave** (eg. via `psql`).
+**Second core idea**: the generated SQL script will be generated without substantial hardcoding of database schema information int the program. Rather, the exporter will inspecting the schema programmatically and, for example, determine most or all of the involved tables, key relationships, ordering, etc.
```
rjmateus, line 72:
```suggestion
**Second core idea**: the generated SQL script will be generated without substantial hardcoding of database schema information in the program. Rather, the exporter will inspecting the schema programmatically and, for example, determine most or all of the involved tables, key relationships, ordering, etc.
```

accepted/00063-hub-iss.md
```diff
@@ -40,35 +40,48 @@ From the requirements:
...
 
-**Second core idea is to generate that SQL script from the ISS Master data mainly by inspecting the schema programmatically** (eg. figuring out by schema inspection most of the involved tables, key relationships, ordering, etc.).
+Notes:
+  - the current implementation serializes ISS Master database data to XML and deserializes it on the ISS Slave. Relatively complicated logic is in place on both communication ends to convert to and from this format
+    - intention of the approach in this RFC is to avoid most of this complication
+  - the current implementation also needs an up-to-date mapping of tables and columns in Python code (ORM), and that has to be kept updated
+    - intention of the approach in this RFC is to avoid most of the manual specification of the mapping, reducing maintenance
 
-A proof-of-concept Python script that implements partially those ideas is available at: https://github.com/SUSE/spacewalk/blob/Manager-4.0-iss2/iss2/iss2.py
+A proof-of-concept Python script that implements partially those ideas is available at: https://github.com/moio/mgr-dump
```
rjmateus, line 80:
```suggestion
A proof-of-concept Golang tool that implements partially those ideas is available at: https://github.com/moio/mgr-dump
```

accepted/00063-hub-iss.md
```diff
@@ -40,35 +40,48 @@ From the requirements:
...
 
 #### First core idea: export as SQL script
 
-The SQL script exported from the ISS Master will have to be idempotent. For example:
+The exported SQL script will have to be idempotent. For example:
```
rjmateus, line 84: Should be pointed that all file should be wrapped inside a single transaction, using a `Begin;` statement, for example?

accepted/00063-hub-iss.md
```diff
@@ -123,7 +136,7 @@ At this point it is believed/assumed that:
```
rjmateus, line 136: At line 134, that statement is not true any more. The linking tables has foreign keys to other tables, so we can replace the ID'd using the same mechanism used in the foreign keys. This approach is implemented and tested in the POC.

accepted/00063-hub-iss.md
```diff
@@ -145,35 +158,40 @@ Expected benefits:
```
rjmateus, line 158: Line 150. we don't need to delete anything.

accepted/00063-hub-iss.md
admd, line 72:
```suggestion
**Second core idea**: the generated SQL script will be generated without substantial hardcoding of database schema information in the program. Rather, the exporter will inspect the schema programmatically and, for example, determine most or all of the involved tables, key relationships, ordering, etc.
```

accepted/00063-hub-iss.md
admd, line 80:
```suggestion
A proof-of-concept Go program that implements partially those ideas is available at: https://github.com/moio/mgr-dump
```

admd (approved):

rjmateus (commented):
LGTM. Just some small notes

accepted/00063-hub-iss.md
moio: Well, in principle we might. What about a channel that used to have a package but now does not have any more? We might want to clean up such a situation.

accepted/00063-hub-iss.md
moio: Just out of excessive care: was it tested in the case one channel used to have, say, 3 packages and now only has 2? Will it be reflected correctly after a new sync round?

accepted/00063-hub-iss.md
moio: To me this is an implementation detail, as the transaction could also be handled by the import tool.

rjmateus: sounds good

accepted/00063-hub-iss.md
rjmateus: If the channel has fewer packages it will not be covered with this implementation. In this case, we need to think about how to deal it.

accepted/00063-hub-iss.md
rjmateus: ok that makes sense

accepted/00063-hub-iss.md
moio: I did reproduce this case with the original ISS and indeed the extra package is removed. I think we should make sure this actually happens the same with the new implementation proposed here.

The idea I had is to issue a `DELETE` statement removing any row which is now unnecesary. I see two alternatives:
1. mass-delete all rows in a table that are affected by this ISS round (eg. `DELETE FROM rhnChannelPackage WHERE channel_label = 'label'`), before any `INSERT` (`ON CONFLICT` clause could in principle be omitted in this case)
2. selectively delete rows with IDs not in the set we are looking at (eg. `DELETE FROM rhnChannelPackage WHERE channel_id, package_id IN (...possibly complicated subselect here...)). Probably slower/more difficult to implement

Some RDBMSs support another (standard) SQL command called `MERGE INTO`, which seems to be doing what we want (INSERT, UPDATE or DELETE depending on conditions). This was scheduled to be merged into PG 11, but [it got reverted](08ea7a2291db21a618d19d612c8060cda68f1892) and indeed we still do not have it in PG 13. Even looking at the [roadmap](https://www.postgresql.org/developer/roadmap/), I see no traces left of it, so I would not consider it as a possibility for the foreseeable future.

At this point I am out of other viable ideas (but would love to hear more, if anybody has).

accepted/00063-hub-iss.md
moio: Same point is being discussed upthread, marking this one resolved

accepted/00063-hub-iss.md
rjmateus: We should also consider that the delete will depend on the data already on the target database, which can be different between the target servers.
Link tables can be deleted with delete all, and re-inserted. Should we also remove entries from `rhnpackage` or `rhnerrata`, or it's enough to remove from the linking tables, like `rhnchannelpackage` and `rhnerratapackage`?

rjmateus (approved):

accepted/00063-hub-iss.md
moio: It depends on the nature of the table's data. I tried to compile a list looking at what the current ISS implementation does and some common sense from my part (hopefully right): tables that need cleanup are marked `// cleanup` here:

https://github.com/uyuni-project/inter-server-sync/blob/360a110b4578fa29dec62788e85f5c3c82f9ca8d/dumper/channelDataDumper.go#L29-L86

As you can see, tables like `rhnErrata` and `rhnPackage` can be left alone and most of linking tables need cleanup, but there is also some others that do not necessarily fit each definition. Ultimately, cleaning never hurts - it's just not necessarily useful.

At the end of the day, implementing a general mechanism will be needed, and we can later judge if we should just mass-apply it to all tables or to a selected inclusion / exclusion list.

From an RFC POV I think the text in line 134 can stay, or do you see a need for any other rewording?

rjmateus: I agree, we need some kind of cleanup mechanism for those linking tables

PR #43, opened 2020-11-04
Retracted updates support draft

hustodemon:
TODO
- [x] Cover ISS

### DIFF
--- a/accepted/00074-retracted-updates-support.md
+++ b/accepted/00074-retracted-updates-support.md
@@ -94,6 +94,8 @@ will implement an `enum`.
 Enhance the `UpdateInfoWriter` class, so that it includes the `status` attribute
 in the output XML. Currently the `status` output is hardcoded to `final`.
 
+## Inter Server Sync
+Make sure that the patch status gets propagated correctly in the ISS slaves.
 
 ## Iteration 2 - Enhance UI and XMLRPC
 
@@ -133,12 +135,21 @@ The following pages must be updated:
   Don't display package versions that are part of a retracted patch.
 
 - New tab *System detail -> Software -> Packages -> Retracted*:
-  This screen should allow user to install packages contained in retracted
-  patches and should contain huge warning about possible consequences.
+  This screen should allow user to install or upgrade packages
+  contained in retracted patches and should contain huge warning about
+  possible consequences. The table should contain packages for both
+  installation and upgrade in the following manner:
 
-  - Q: this wasn't clarified on the mtg: should this allow installation or
-  upgrade of packages contained in a retracted patch? Maybe both, but the UI
-  would get messy.
+  - Installation: display the newest version of the "retracted"
+    package, if it's the newest version in the channel (i.e. there is
+    no non-retracted package that superseedes this one) and the
+    package is not installed on the system.
+
+  - Upgrade: for each package installed on the system, display its
+    retracted upgrade, if there is no superseeding non-retracted
+    upgrade in the channel.
+
+  - We should distinguish the install/upgrade rows graphically
 
 - *System detail -> Software -> Patches*:
   Don't display retracted patches.
--- a/accepted/00074-retracted-updates-support.md
+++ b/accepted/00074-retracted-updates-support.md
@@ -12,6 +12,7 @@ the flag is to signal the fact that the patch had been released, but was then
 taken back (retracted) by its publisher. This can be used, for instance, when
 the publisher issues a patch that is invalid (e.g. can break a system).
 
+
 # Motivation
 [motivation]: #motivation
 
@@ -29,6 +30,7 @@ systems due to broken released patches.
 
 To make Uyuni aware of the `retracted` status of patches and to update
 corresponding pieces of UI/XMLRPC.
+
 
 # Detailed design
 [design]: #detailed-design
@@ -60,7 +62,7 @@ Operations and their outcomes:
   taken into account, since this is the way we upgrade packages in SUMA)
 - installing retracted patch (`zypper in patch:<retracted_patch_id>`):
   yields "patch not needed" and exit code `0`. Consequence: retracted patch
-  can't be applied to a system! Q: somebody double-check this.
+  can't be applied to a system.
 - if a system already has a patch installed, that becomes retracted, zypper has
   no intention to fix this situation (this can be problematic, since downgrading
   RPM is potentially dangerous operation because of rpm scripts)
@@ -79,11 +81,12 @@ of the repository. We need to adjust `spacewalk-repo-sync` (`errataImport.py`)
 so that it parses this attribute during repository synchronization and stores it
 in the database (`rhnErrata.status`).
 
-Q: If the previous question==true: We need to think of consequences of modifying
-   an existing patch (that has been synced in the past with
-   `status=final`). Normally reposync is not designed to modify data. We need to
-   make sure reposync spots this change and maybe we also need to update some
-   caches (system available updates are cached (`rhnServerNeededCache`?)).
+We need to think of consequences of modifying an existing patch (that has been
+synced in the past with `status=final`). We need to make sure reposync spots
+this change and maybe we also need to update some caches (system available
+updates are cached (`rhnServerNeededCache`?)).
+
+Analyze the case, when a stable patchh gets retracted and then stable again in SCC.
 
 ### Java backend
 
@@ -96,6 +99,7 @@ in the output XML. Currently the `status` output is hardcoded to `final`.
 
 ## Inter Server Sync
 Make sure that the patch status gets propagated correctly in the ISS slaves.
+
 
 ## Iteration 2 - Enhance UI and XMLRPC
 
@@ -159,7 +163,6 @@ The following pages must be updated:
 - *Channel detail -> Packages*: an icon for a package contained in a retracted
   patch
 
-
 ### XMLRPC
 The following endpoints must be updated:
 
@@ -169,8 +172,8 @@ The following endpoints must be updated:
 - `ErrataHandler.setDetails` for non-vendor channels the endpoint allows setting
   `status` to either `retracted` or `final`
 
-- Q: `ErrataHandler.create` shall we support creating retracted patches via API?
-  Likely not.
+- `ErrataHandler.create` shall we support creating retracted patches via API?
+   No.
 
 - `PackageHandler.getDetails`, `PackageHandler.findByNvrea` add a boolean flag
   to the return value, that tells if the package is part of a retracted patch
@@ -215,7 +218,6 @@ Early idea:
 
 This must be grouped in a sane way.
 
-Q: does ths ^ make sense?
 
 ## Iteration 4 - deeper UI integration, allow uninstalling a retracted patch
 - Channel detail: a warning icon and a message, that the channel contains a
@@ -230,7 +232,8 @@ Q: does ths ^ make sense?
 - System list: a warning icon in the list, if the system has a retracted patch
   (or its clone)  installed
 
-Q: evaluate query complexity of underlying queries
+As a part of this iteration, evaluate query complexity of underlying queries.
+
 
 # Drawbacks
 [drawbacks]: #drawbacks
@@ -238,6 +241,7 @@ Q: evaluate query complexity of underlying queries
 Why should we **not** do this?
 It adds complexity to the product for relatively sparse use case (SCC is not
 retracting patches _that_ often).
+
 
 # Alternatives
 [alternatives]: #alternatives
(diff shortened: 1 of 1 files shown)


accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,307 @@
+- Feature Name: Retracted updates support
+- Start Date: 2020-11-03
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+The topic of this RFC is implementing the "retracted patches" support in Uyuni.
+
+A retracted patch is a patch with attribut `status=retracted`. The purpose of
```
moio, line 10:
```suggestion
A retracted patch is a patch with attribute `status=retracted`. The purpose of
```

accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,307 @@
...
+the flag is to signalize the fact, that the patch had been released, but was
```
moio, line 11:
```suggestion
the flag is to signal the fact that the patch had been released, but was
```

accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,307 @@
...
+
+Operations and their outcomes:
+- listing (`zypper patches`): lists retracted patches with status `retracted`
+- searching for new versions (`zypper search -v hello`) gets us:
+  `vR | hello | package | 2.10-lp151.4.1` <--- R means retracted
+- upgrading a package (`zypper up hello`):
+  does not upgrade to a version contained in a retracted patch
+- installing specific version (`zypper in hello-2.10-lp151.4.1`):
+  **allows** installing a package contained in a retracted patch (this must be
+  taken into account, since this is the way we upgrade packages in SUMA)
```
moio, line 60: What's the user experience idea in this case? Should they be able to "override" the retracted status like in `zypper in`, or do we plan to be more restrictive in Uyuni?

accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,307 @@
...
+- if a system already has a patch installed, that becomes retracted, zypper has
+  no intention to fix this situation
+
+## Iteration 1 - Minimal Viable Product
+
+### Database
+The `rhnErrata` table must be updated with the `status` column of type
+`VARCHAR(32)`.
+
+Q: Alternatives: 1) `NUMERIC`, 2) normalize the `status` attribute and use
```
moio, line 73: I'd suggest keeping it as a self-descriptive string for ease of debugging's sake

accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,307 @@
...
+
+### UI
+
+The following pages must be updated:
+
+- Patch detail: information about the status
+Q: should this be read-write? so that user can modify the `status` attribute
+using a toggle/drop-down (for non-vendor channels). If we allow this, we need to
+face the consequences of modifying a patch (maybe regenerating some data
+(`rhnServerNeededCache`)).
```
moio, line 111: IIUC patches are fully editable from the UI, so it would make sense that this bit is editable as well.

accepted/00074-retracted-updates-support.md
hustodemon: I've updated the draft now. I'd suggest that we display a big warning in the install/upgrade confirmation screen saying "You're about to install package(s) that belong to a retracted patch". We shouldn't prevent user from installing the packages. Maybe the problem in the retracted patch doesn't apply to them.

hustodemon: Another improvement in this area would be to introduce a toggle on the packages updates list, that would hide the packages that belong to retracted patches.

accepted/00074-retracted-updates-support.md
hustodemon: Totally agree.

cbbayburt (commented):
Looks good so far. I've put some suggestions/answers below.

Do we already have a repository to test retracted patches? If not, how much of an effort is it to create one? I assume we will eventually need it for integration tests. Maybe you can add this task to the RFC?

accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,329 @@
...
+
+We need to enhance the `Errata` class with the `status` attribute, for which we
+will implement an `enum`.
+
+### Repository generating
+Enhance the `UpdateInfoWriter` class, so that it includes the `status` attribute
+in the output XML. Currently the `status` output is hardcoded to `final`.
+
+
+### UI
```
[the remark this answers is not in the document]
hustodemon: Good point. Let's ask @paususe : should the MVP making the SUMA backend aware of retracted patches (i.e. syncing retracted patches is SUMA). Or should the MVP contain UI & XMLRPC as well (as suggested in the current version of the RFC)?

hustodemon:
@cbbayburt 

> Do we already have a repository to test retracted patches? If not, how much of an effort is it to create one? I assume we will eventually need it for integration tests. Maybe you can add this task to the RFC?

I've tested successfuly with this one: http://download.opensuse.org/repositories/openSUSE:/Maintenance:/11749/openSUSE_Leap_15.1_Update/ 
I've used a Leap 15.1 Docker image and the `hello` package (the version of this package installed in the container was lower than the one in the retracted patch).

accepted/00074-retracted-updates-support.md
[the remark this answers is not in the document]
paususe: The former. We can of course quickly follow-up with the UI and XMLRPC after that. But currently, we are losing that status=retracted metadata which, ironically, we have had to use ourselves.

accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,326 @@
...
+The following pages must be updated:
+
+- Patch detail: information about the status. Given the fact we only have 2
+  states now (`retracted`/`final`), this can be a boolean information telling
+  whether the patch is retracted or not.
+  We should implement this as a checkbox, allowing the user to change
+  the attribute (although the change should be disabled for vendor patches).
+  We need to take into account the possible consequences of modifying
+  a patch (maybe regenerating some data (`rhnServerNeededCache`,
+  regenerating repodata of all involved channels)).
```
paususe, line 113: Is this updated per our conversation on Monday?

accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,326 @@
...
+  can be retrieved via existing XMLRPC methods and `ErrataHandler.getDetails`.
+
+
+## Iteration 2 - Content Lifecycle management
+
+Enhance Content Lifecycle Manager and notifications.
+
+### Filters
+
+- add a filter that strips retracted patches from the sources
```
paususe, line 177: Should it be a more generic filter, based on patch/package metadata? Maybe even add one more option to the "Patch (Keyword)" filter, where we have Reboot Required, Package Manager Restart Required, Other keyword <free text>? In fact, would this filter already work if I set "other keyword": "status=retracted" ?

accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,326 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- See the inline "Q:" parts.
+- How to "uninstall" a patch that used to be stable, but is retracted now? Is
+  there some other way than `zypper in --oldpackage pkg-oldver`?
+- What happens if patch=stable -> retracted -> stable? this path should be also
+  supported
+
```
mbrookhuis, line 248: It is also needed to review how this should be done for inter-sync.

mbrookhuis (approved):
Add that we also have to have a look at inter-sync.

accepted/00074-retracted-updates-support.md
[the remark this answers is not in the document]
hustodemon: Makes sense and the recent state of the document reflects that already. Thanks!

accepted/00074-retracted-updates-support.md
hustodemon: Yes. IIRC, there was some feedback against making this modifiable, but in the end there was no agreement not to do it. But we can always start with the simplest thing ("don't allow modification") and change it later, if there is a need for it.

accepted/00074-retracted-updates-support.md
hustodemon: No please. We already have filters for other patch attributes and generalizing those is unrelated to this RFC. I don't know, how long would such a task take, so it should be done separately, if there is a need for it.

accepted/00074-retracted-updates-support.md
```diff
@@ -0,0 +1,363 @@
...
+  - Installation: display the newest version of the "retracted"
+    package, if it's the newest version in the channel (i.e. there is
+    no non-retracted package that superseedes this one) and the
+    package is not installed on the system.
+
+  - Upgrade: for each package installed on the system, display its
+    retracted upgrade, if there is no superseeding non-retracted
+    upgrade in the channel.
+
+  - We should distinguish the install/upgrade rows graphically
```
cbbayburt, line 156: Having the column "Installed Package" as in the "Upgrade" tab, and leaving the cell empty for install rows should do it

cbbayburt (approved):
LGTM :+1: 
Thanks, @hustodemon, for the effort.

I believe we can go on with creating the epic and the individual action items right after we merge this.

cbbayburt:
> Don't allow modification of patch status in the 2nd iteration. We can implement this optionally.

I understand that patch editing is something nobody uses, but the form is still there. If this addition is trivial, maybe we can still do this for the sake of completeness.

accepted/00074-retracted-updates-support.md
hustodemon: Good observation!

hustodemon:
Thanks @cbbayburt for the input. For *now* I'll keep the topic as is (allow modifications), unless @paususe is against it. Anyway, this can be discussed further during detailed design.

I've got one last open question/idea and I'd like to get feedback on it from everybody. It's related to changing the `status=retracted` for an existing patch. When we have a vendor patch in the db, and this will get retracted by SCC, we could propagate the `status=retracted` to all clones of this patch (typically one per organization). This way, all systems subscribed to cloned channels will have the information about the retracted patches (and user will see those in the UI). What do you think?

moio:
> Thanks @cbbayburt for the input. For _now_ I'll keep the topic as is (allow modifications), unless @paususe is against it. Anyway, this can be discussed further during detailed design.
> 
> I've got one last open question/idea and I'd like to get feedback on it from everybody. It's related to changing the `status=retracted` for an existing patch. When we have a vendor patch in the db, and this will get retracted by SCC, we could propagate the `status=retracted` to all clones of this patch (typically one per organization). This way, all systems subscribed to cloned channels will have the information about the retracted patches (and user will see those in the UI). What do you think?

I am not sure that is wise.

On one hand, such a behavior might prevent catastrophic failures (eg. data loss) if SUSE pushed out a completely broken patch by mistake.

On the other hand, such a change breaks the promise of content lifecycle management: as a sysadmin I got a patch from SUSE, I ran it through my QA, it passed, possibly it's fine for my specific use and maybe it never hurts me, possibly it actually fixes a vulnerability **and now you remove it from under my feet without me knowing**. Worst case scenario: my systems are exposed - also not good.

I'd try to seek a statement from SCC colleagues as to what "retracted" really means, as I see two possible interpretations:
1- should we do anything in our power to avoid the patch being installed, at any cost (ie. this is really-really a nuclear option) or
2- should we do anything in our power to raise awareness, maybe lower the probability of silly mistakes from the operators but still leave a possibility open for "corner of corner" cases which can happen

I tend to believe we are really in the business of 2- just because `zypper` allows it too - but I'd suggest touching base with zypp and SCC colleagues, ideally getting some concrete retraction examples, to make sure that is the case.

mcalmer:
> > Thanks @cbbayburt for the input. For _now_ I'll keep the topic as is (allow modifications), unless @paususe is against it. Anyway, this can be discussed further during detailed design.
> > I've got one last open question/idea and I'd like to get feedback on it from everybody. It's related to changing the `status=retracted` for an existing patch. When we have a vendor patch in the db, and this will get retracted by SCC, we could propagate the `status=retracted` to all clones of this patch (typically one per organization). This way, all systems subscribed to cloned channels will have the information about the retracted patches (and user will see those in the UI). What do you think?
> 
> I am not sure that is wise.
> 
> On one hand, such a behavior might prevent catastrophic failures (eg. data loss) if SUSE pushed out a completely broken patch by mistake.
> 
> On the other hand, such a change breaks the promise of content lifecycle management: as a sysadmin I got a patch from SUSE, I ran it through my QA, it passed, possibly it's fine for my specific use and maybe it never hurts me, possibly it actually fixes a vulnerability **and now you remove it from under my feet without me knowing**. Worst case scenario: my systems are exposed - also not good.

Totally agreed. We should never change something which is "owned" by the customer without his agreement.

We should show warnings and we can show a "diff" between vendor channel and clone. But the action to propagate this change to the clones should be triggered by the customer.

hustodemon:
Thanks for your feedback @moio & @mcalmer !

PR #44, opened 2020-11-12
[WIP] Better Cobbler TFTP Sync

brejoc:
More optimized TFTP syncing for Cobbler proxies. Switching from push to pull.

### DIFF
--- a/accepted/0000-better-cobbler-tftp-sync.md
+++ b/accepted/0000-better-cobbler-tftp-sync.md
@@ -10,7 +10,7 @@ More optimized TFTP syncing for Cobbler proxies. Switching from push to pull.
 # Motivation
 [motivation]: #motivation
 
-For large scale installations Cobbler has proven to be a possible bottleneck. One of the problems is the Cobbler sync that, when using it in conjunction with proxies, also pushes the TFTP folder to the proxies. For installations with just one proxy and around 130 profiles the whole process can easily take up to 15 minutes. Not all of that is due to TFTP sync, but a good chunk of it and might force clients into a reboot, since information are not yet updated.
+For large scale installations Cobbler has proven to be a possible bottleneck. One of the problems is the Cobbler sync that, when using it in conjunction with proxies, also pushes the TFTP folder to the proxies. For installations with just one proxy and around 130 profiles the whole process can easily take up to 15 minutes. Not all of that is due to TFTP sync, but a good chunk of it is. Note also that clients might require unnecessary reboots in case they are booting during the sync, since TFTP information is not updated atomically.
 
 # Detailed design
 [design]: #detailed-design
--- a/accepted/0000-better-cobbler-tftp-sync.md
+++ b/accepted/0000-better-cobbler-tftp-sync.md
@@ -1,0 +1,77 @@
+- Feature Name: Better Cobbler TFTP Sync
+- Start Date: 2020-11-08
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+More optimized TFTP syncing for Cobbler proxies. Switching from push to pull.
+
+# Motivation
+[motivation]: #motivation
+
+For large scale installations Cobbler has proven to be a possible bottleneck. One of the problems is the Cobbler sync that, when using it in conjunction with proxies, also pushes the TFTP folder to the proxies. For installations with just one proxy and around 130 profiles the whole process can easily take up to 15 minutes. Not all of that is due to TFTP sync, but a good chunk of it and might force clients into a reboot, since information are not yet updated.
+
+# Detailed design
+[design]: #detailed-design
+
+The solution of pushing the update files to proxies does not scale well. Especially with large installations and many proxies the time needed for a full sync grows linearly.
+
+The idea is to skip this step completely and move the logic to the proxy. In order to do that, we would need a TFTP server that is capable of dynamically checking if requested files are up-to-date on the file system, or if files need to be fetched from Uyuni. Also certain files need to be modified on the fly, since some files include the hostname of the Uyuni server.
+
+```
+      +----------------+
+      | Client Machine |
+      +-------+--------+
+              |
+            TFTP-
+           Request
+              |
+              v
+         +----+------+
+         |           |
+         |   TFTP    |
+         |           |
+         +-+-------+-+
+           |       |
+
+     Check local disk cache
+      for files. Otherwise
+     pull from Uyuni server.
+           |       |
+           |       |
+           v       v
++----------+-+   +-+------+
+| Local Disk |   | Uyuni  |
+| Cache      |   | Server |
++------------+   +--------+
+```
+
+The scenario would look like this. Once a client requests a file, the TFTP server would check if this file is in the local disk cache. If it's not, then it's immediately fetched via HTTP from Uyuni Server. If the file is present in the local disk cache, then the TFTP server would send a HEAD request to Uyuni server to get the current etag of the file. If the etags are matching, then the local file can be used. If not, then it needs to be fetched from Uyuni server and put into the local disk cache.
+For some files the Uyuni hostname needs to be replaced; this would be done right before delivering it to the client since etags wouldn't match otherwise.
+
+Ideally a proxy would be "warmed up" after a few client requests, since most of the files are then cached on disk. HEAD requests are releatively cheap and reasonably fast. So the first clients would encounter a slight delay, but all of the following requests for profiles with the same kernel and initrd would have nearly the same speed as the current implementation.
+
+Dealing with the thundering herd problem. Since it would be possible to have 300 clients requesting the same not yet synced, or outdated kernel, there needs to be a lock of some sorts on file requests. If there is a request to Uyuni server already running for the same file, there is no additional request allowed. Instead we'd just wait for the initial request to finish.
+
+One possible solution for the TFTP server would be [fbtftp](https://github.com/facebook/fbtftp) from Facebook. They ran into pretty much the same problem and created this dynamic TFTP server framework.
+
+
+# Drawbacks
+
+* This solution moves load and complexity to the proxy.
+* Proxies need to have a permanent connection to the Uyuni server. If that cannot be guaranteed, then this solution won't work.
+* Proxies are pulling files instead of getting files pushed and start slow due ot empty cache.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Improve file transfers with rsync via HTTP. But there is no rsync2 Python wrapper. Not yet tested, but would have the same scaling characteristics as the current solution.
+- Gzip for requests. Tested and only slighly faster with the same scaling characteristics as the current solution. Could be used in combination with rsync.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- A note on _etags_. It would be nice to not have a database to track etag checksums of files. We need to check if etags can be calculated and calculated reasonably fast. Otherwise tracking etags in a sqlite database would be needed.
+
+- There are multiple ways to solve the thundering herde problem, but I have not yet looked into a specific solution. A queue might be reasonable solution, but also adds some complexity.


SchoolGuy:
My two cents which I already had discussed with Jochen internally: I think that this is a good idea since we have a performance bottleneck and it needs to be fixed.

The idea of FB is now for me but I like it very much. I read about it a while back but I never had the chance to try it out. Since FB opensourced it after using it internally it should work pretty good and thus could also work for us. But it will possibly complicate simple setups (which is a drawback I would be fine to take).

accepted/0000-better-cobbler-tftp-sync.md
```diff
@@ -0,0 +1,77 @@
...
+
+- Improve file transfers with rsync via HTTP. But there is no rsync2 Python wrapper. Not yet tested, but would have the same scaling characteristics as the current solution.
+- Gzip for requests. Tested and only slighly faster with the same scaling characteristics as the current solution. Could be used in combination with rsync.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- A note on _etags_. It would be nice to not have a database to track etag checksums of files. We need to check if etags can be calculated and calculated reasonably fast. Otherwise tracking etags in a sqlite database would be needed.
+
+- There are multiple ways to solve the thundering herde problem, but I have not yet looked into a specific solution. A queue might be reasonable solution, but also adds some complexity.
```
brejoc, line 77: Jens Mozdzen pointed out that we have a Squid on the Proxy. So we might not have to solve this problem.

accepted/0000-better-cobbler-tftp-sync.md
```diff
@@ -0,0 +1,77 @@
...
+For some files the Uyuni hostname needs to be replaced; this would be done right before delivering it to the client since etags wouldn't match otherwise.
+
+Ideally a proxy would be "warmed up" after a few client requests, since most of the files are then cached on disk. HEAD requests are releatively cheap and reasonably fast. So the first clients would encounter a slight delay, but all of the following requests for profiles with the same kernel and initrd would have nearly the same speed as the current implementation.
+
+Dealing with the thundering herd problem. Since it would be possible to have 300 clients requesting the same not yet synced, or outdated kernel, there needs to be a lock of some sorts on file requests. If there is a request to Uyuni server already running for the same file, there is no additional request allowed. Instead we'd just wait for the initial request to finish.
+
+One possible solution for the TFTP server would be [fbtftp](https://github.com/facebook/fbtftp) from Facebook. They ran into pretty much the same problem and created this dynamic TFTP server framework.
```
moio, line 57: Cache invalidation is in general a difficult problem, so indeed we should be careful before embarking into the 'roll your own solution' route.

`fbtftp` looks nice, are there any alternatives out there, specifically (ideally) caching proxies?

I suppose a true caching proxy is not really possible as TFTP does not have HTTP-like `Cache:` headers, correct?

accepted/0000-better-cobbler-tftp-sync.md
```diff
@@ -0,0 +1,77 @@
...
+
+
+# Drawbacks
+
+* This solution moves load and complexity to the proxy.
+* Proxies need to have a permanent connection to the Uyuni server. If that cannot be guaranteed, then this solution won't work.
```
moio, line 63:
```suggestion
* Proxies need to have a permanent connection to the Uyuni server. If that cannot be guaranteed, then this solution won't work.
```

Proxies already have that requirement, so adding this solution does not really change things. I think you can remove this from the drawbacks, or at least mention it is not really making the situation any worse :wink:

accepted/0000-better-cobbler-tftp-sync.md
moio: We could leverage the existing Squid as the backend of something like `fbtftp` indeed.

moio (commented):
Hey @brejoc, thanks for the efforts spent around this problem, and sorry for being so late with my review!

I left some questions inline, but I think there's a more fundamental one here.

Question is: no matter if content is pushed or pulled, it needs to be transferred one way or the other - so ultimately the question here is about optimizing the moment when this happens.

If content is "pushed", one gets all the load in a particular point in time (and hopefully to control when), if it's "pulled", it could come at the wrong moment (aka, a thundering herd problem). What I have seen is that for any "pull" architecture, at least in high-scale scenarios, you will also want an additional "cache warming" mechanism, else you get requests when customers discover the thundering herd problem. Substantially, "cache warming" is a polite way to ask a "pull" mechanism in a "push" one on demand. Of course having both and being able to choose which is the best possible world, so the question becomes, is it worth it to actually implement both?

So what I would like to learn from you is, what reasons do you (and the users you interacted with) see to prefer a "pull" model, especially in large scenarios? What are the exact pain points with the current implementation?

Thanks!

accepted/0000-better-cobbler-tftp-sync.md
SchoolGuy: The thing is... In an ideal world we don't want to cache anything on this side because if the Cache is not correct (which is a situation which may occur), then the client would be provisioned with the wrong options. And yes TFTP does not have `Cache:` Headers afaik.

accepted/0000-better-cobbler-tftp-sync.md
```diff
@@ -0,0 +1,77 @@
...
+
+# Summary
+[summary]: #summary
+
+More optimized TFTP syncing for Cobbler proxies. Switching from push to pull.
+
+# Motivation
+[motivation]: #motivation
+
+For large scale installations Cobbler has proven to be a possible bottleneck. One of the problems is the Cobbler sync that, when using it in conjunction with proxies, also pushes the TFTP folder to the proxies. For installations with just one proxy and around 130 profiles the whole process can easily take up to 15 minutes. Not all of that is due to TFTP sync, but a good chunk of it and might force clients into a reboot, since information are not yet updated.
```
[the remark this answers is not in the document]
moio:
```suggestion
For large scale installations Cobbler has proven to be a possible bottleneck. One of the problems is the Cobbler sync that, when using it in conjunction with proxies, also pushes the TFTP folder to the proxies. For installations with just one proxy and around 130 profiles the whole process can easily take up to 15 minutes. Not all of that is due to TFTP sync, but a good chunk of it is. Note also that clients might require unnecessary reboots in case they are booting during the sync, since TFTP information is not updated atomically.
```

Does the above wording keep the original meaning?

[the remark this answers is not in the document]
SchoolGuy: Yes!
A real atomic update will not be possible in my eyes due to the fact that we would need to update the data at the very same moment. When having a few distros in SUMA or Cobbler we are talking about a kernel and initrd for each one, as well as other (smaller) files. But I am open for ideas.

jmozd:
Because of on-going problems with a large Cobbler install using proxies & auto-installation, I decided to pick up on the approach and took it further:

- decided on an individual implementation of a TFTP to HTTP proxy daemon
- no local "file" access is used by the TFTP to HTTP proxy, so no extra sync mechanism required
- used the existing Proxy (as in "product") environment to provide the files via HTTP as a virtual source
- made sure the product-internal mechanisms are used to route requests for these files to the master Server, via the local Squid instance
- made sure that only image files are actually cached (so especially PXE config files are not cached locally)
- provided changes to the master Server to provide files from /srv/tftpboot via https://<masterserver>/tftpboot

Using this setup, I could successfully install PXE clients in the remote location. Using the mechanisms of the Squid cache, image files could be "pre-heated" if desired, OTOH by simply pulling via tftp proxy, only those few images typically required for current installation rounds are pulled across the (potentially slow) uplink (from the master Server).

Installation is as easy as providing the daemon (identically configured on all remotes), a config file to integrate to the local Apache httpd instance and an update to the Squid configuration (to work around issues caused by the way files are served from upstream), plus a config file at the master Server to provide access to /srv/tftpboot. Pre-heating runs on the master Server and is an additional (but separate) feature that either integrates with Cobbler or can be run in off-production hours.

This is a minimum-invasive change to the Proxy, no setup required on the Proxy (except for RPM installation). Individual configuration on the master Server is only required if pre-heating is desired.

accepted/0000-better-cobbler-tftp-sync.md
jmozd: Splitting the architecture into "tftp to http proxy" and "cache" is helpful here. As already mentioned in RocketChat, using the Squid instance on the Proxy machine will provide for a local cache and since we're talking about serving "files" at the upstream server, standard caching mechanisms can be applied.
Looking at the typical content of the (upstream) TFTP server, you'll see either small files with individual configuration (which may change frequently and is each used by only a single PXE client) and large (up to "huge") image files, which will rarely change. Distinguishing between these two at the Squid level keeps the config file always coming from the upstream and the image files from the cache.
Of course, even the config files could be cached, but considering that the over-all workflow of client installation will typically require live access to the upstream server anyhow, and the rather small difference between checking for a new file version or just serving these <5k files, I opt for "no caching of config files".

moio:
@jmozd interesting! Can we see your setup in some form or fashion? (what are the needed changes?)

SchoolGuy:
So I revisited the RFC and I am seeing the following problem:

Cobbler Syncs the whole TFTP and SUMA does not actually wait until it is done, at some point it just ignores it and tries to resync again which is then causing a loop because the sync is never done and thus the TFTP tree is never valid. In this scenario any solution I described would not work since no machine can be installed. This is only a problem for very big instances.

Thus in SUMA we are missing a mechanism of a queue or a lock for autoinstalling systems during this sync. That would solve once and for all any problem of infinite boot loops and it would clearly show times when we actually can autoinstall and when not.

But even in smaller instances a `sync` may happen at the same time taskomatic calls `sync` and then a client would have no data.

Since a `cobbler sync` is only required if a Distribution or Profile changes its content (so an edit is done), we could in theory remove the regular `sync` and change to a on-demand based `sync`. This would remove a lot of the unrequired syncs and thus blocking situations. Maybe (I need to check with the Cobbler implementation) we need a sync on system add but I can make that go away in the future. When adding or removing systems Cobbler automatically does the required things without a `sync`. The `sync` was always thought on the Cobbler side as a `I am stuck and don't know whats wrong, let's start from scratch` mechanism thought. If this is not the case for SUMA it is a Bug in Cobbler. In an ideal world `sync` is never required or only after we edited a Cobbler Item.

So TLDR:

- I am in favor of the current state of this document but would like to extend it.
- Extension 1: Remove the regular `cobbler sync` called by Taskomatic and replace it by a button which can be manually pressed
- Extension 2: Add a locking mechanism which prevents or queues systems to autoinstall after a sync is done

---

And I think this is out of the reach of this RFC but since the Cobbler Server is always on the same system as the SUMA Server, I vote for cutting out the XMLRPC API completely and use the Python API of Cobbler directly. No authentication as a possible error cause, no time required to parse and build XML and no time is spent doing magic (e.g. XMLRPC stuff for None). Debugging get's easier and we remove a lot of code which is running in between to make the communication happen in between. Last but not least we get better return codes in the case of errors or other hickups.

mcalmer:
> And I think this is out of the reach of this RFC but since the Cobbler Server is always on the same system as the SUMA Server, I vote for cutting out the XMLRPC API completely and use the Python API of Cobbler directly. 

We are on the way to Containerize Uyuni which means that every process will run in its own container. At that point we need the API. In addition we call the API from java. How do you want to call python directly from Java. Every interface (swig, etc.) creates extra / new troubles. I would strongly suggest to keep the XMLRPC API.

SchoolGuy:
@mcalmer Thanks then this is from the table. Was just an idea.

jmozd:
> @jmozd interesting! Can we see your setup in some form or fashion? (what are the needed changes?)

Sorry for not getting back to you earlier... our approach didn't pick up momentum and after several updates to our SuMa, I'm not sure what state our integration is in. The required changes were:

- SUSE Manager Server (central) httpd needs to give access to /srv/tftpboot for reuests from at least the SuMa Proxies
- SUSE Manager Proxy needs config updates to route access to i.e. http://127.0.0.1/tftpboot through the local Squid instance and forward to SUSE Manager Server
- SUSE Manager Proxy Squid needed config update to reset cache TTL of upstream files. IIRC, per default every file was reported as non-cachable, which was devastating for image files (but is ideal for the small tftp config files)
- we deactivated TFTP daemon on SuMa Proxy and installed a TFTP-HTTP-Gateway, pointing to http://127.0.0.1/tftpboot

That's basically all that was needed.

If you want a pre-heated Squid cache, you could issue according requests from SuMa server to the proxies' httpd instances, requesting the image file (but remember to limit the requested response, in order to avoid large amount of traffic from Proxy to Server). Pre-heating can be done via simple scripts and even run in parallel without side-effects.

moio:
Quick note after our last meeting:

 - reproduce the scenario and run it through a profiler such as `py-spy`. Quantify what percentage of time we spend on CPU, and what percentage in I/O (copying)
 - repeat the experiment with a SUSE Manager Proxy. Quantify the percentage of time spent in I/O (network transfer)
 - decide which problem to attack first (wasted CPU time, I/O)

CPU time. Go deeper to see why the process is so wasteful. Propose (Python-only) solutions. @moio and Hexagon can help.

I/O. Determine if we are copying/transferring files needlessly (seems so). Evaluate the use of `rsync` or similar "intelligent transfer" mechanisms instead of "wipe and copy from scratch".


Another problem is atomicity of transfers - and it is to be tackled separately. Current strategy is to minimize the window of time in which we serve files that are out of sync and see if that suffices in practice (eg. via directory renaming).

aaannz:
As part of https://github.com/SUSE/spacewalk/issues/17227 I used this `fbtftp` library to implement simple tftp server translating tftp requests to http requests which passes through proxy squid and it works quite nicely. It is noticeable slower then regular tftp server, but not dramatically and IMO this can be accepted. The positives of hustle free configuration, dropping requirements of accessible proxies (cobbler sync requires that) and other limitations of sync like inability to sync only to specific proxies are immense.

Another negative, or rather thing to look at and solve, is making `tftpboot` directory accessible by http. Which may be security concern.

jmozd:
> As part of [SUSE/spacewalk#17227](https://github.com/SUSE/spacewalk/issues/17227) I used this `fbtftp` library to implement simple tftp server translating tftp requests to http requests which passes through proxy squid and it works quite nicely. [...]

I had prototyped such a solution several moons ago via a Go-based tftp2http proxy, to solve the "cobbler sync takes very long for large TFTP environments" problem. Since every TFTP request got proxied to an according HTTP request to **SUSE Manager Server**, I noticed a pretty severe limitation of this simple (and otherwise working really nicely) implementation: While image files can be transfered as-is, the PXE configuration files need extra handling. As part of "cobbler sync", the receiving end modifies these files, before storing them on-disk at the Proxy:

- cobbler@Server generates all config files with mentions of the Server URL within kernel command lines, i.e. for retrieval of autoinstallation profile and turning of PXE boot during installation of a client
- "cobbler sync" transfers these files to all proxies, which replace that Server URL with the corresponding Proxy URL
- atftpd hands out those modified files to the PXE client
- the client interacts with Proxy during installation

When proxying/rerouting the tftp requests to SUMA Server without adopting the PXE config files during transfer, the PXE clients will try to access SUMA Server instead of SUMA Proxy, which in many situations will fail for lack of direct connectivity (firewall etc) or at least contradict data transfer volume reduction intents (bypassing Squid@Proxy for RPM retrieval during installation).
So some extra effort/logic needs to be put into the tftp2http/httpd@Proxy/squid@Proxy chain, to properly adjust the PXE config files.

> Another negative, or rather thing to look at and solve, is making `tftpboot` directory accessible by http. Which may be security concern.

I believe that this can be addressed rather simple, either via configuration or "administrative policy":
- The according URI path (/tftpboot, @ SUMA  Server) could get restricted via httpd ACLs to the IPs of the SUMA Proxies, which are well-known at the server
- Any such request is already "world-readable" via tftp access, both at the Server and at every Proxy. Adding another access path (via HTTP) therefore doesn't really decrease the security level of these files

SchoolGuy:
Just as a note. The URL can easily be overridden via the `server` and `redhat_management_server` attributes. If we would have a subprofile for each proxy we would have first of all a subset of files to be transferred and additionally we wouldn't need to compute this on the proxy. Then only a simple proxy for the files would be enough since every system is assigned to the profile (which is then proxy specific).

SchoolGuy:
Also there is the functionality of `cobbler replicate` which effectively duplicates the content to another Cobbler server. The copying is done via rsync thus the transfer is on a 'as needed' done. The computation of the templates etc would is then done as the last comment said via subprofiles on the Cobbler server sitting on the proxy. The amount of profiles etc that are replicated can be filtered. So we already have a lot of options how to improve here. We just need to know what we want... :)

SchoolGuy:
What I just realized is that we do the copying of files in a linear way. What would already be a big advantage is if we do this in parallel.

moio:
> What I just realized is that we do the copying of files in a linear way. What would already be a big advantage is if we do this in parallel.

Did you test this assumption, ie, are we sure to have unutilized I/O bandwidth?

SchoolGuy:
@moio I will be so blunt and say that copying 3KB files for pxelinux will not consume the whole IO of a drive. In case of the kernels and initrds I am not so sure but most of the files are pxelinux or GRUB config files. However no I did not yet verify that. After Cobbler 3.3.3 was merged into Uyuni I will do measurements.

SchoolGuy:
Rebased the PR.

brejoc:
Do you folks still need this draft? I was just about to delete my fork, since I'm cleaning up my Github repos. Not sure what happens to this PR if I would do that.

SchoolGuy:
@brejoc It is discussed at the moment but my personal opinion is that we do need this. Noone wanted to ever give the green light for any version of this and it is still broken. If it is not too much of a trouble I would love if you could let this fork alive until we merge this RFC. :)

rjmateus:
@brejoc thank you for the heads-up. 
This RFC can be closed since we moved forward with the implementation of [fbtftp](https://github.com/facebook/fbtftp) from Facebook, for at least the containerized proxy.

brejoc:
@SchoolGuy @rjmateus Nice! Godspeed to you and have fun!

PR #45, opened 2020-12-01
WIP: Hub Retail Extension

nadvornik:
https://github.com/SUSE/spacewalk/issues/9594

Render [available here](https://github.com/nadvornik/uyuni-rfc/blob/master/accepted/0000-hub_retail.md)

### DIFF
--- a/accepted/0000-hub_retail.md
+++ b/accepted/0000-hub_retail.md
@@ -19,70 +19,132 @@ Reasons for use of Hub with Retail environment
 - disconnected operation
   - Peripheral server must stay fully functional when the Hub connection is down
 
+See [Hub requirements](https://github.com/SUSE/spacewalk/wiki/Hub-requirements) for details.
 
 Possible configurations
 -----------------------
 
-1. Full
+1. Standard
+```
+    Terminal -> Branch Server -> SUMA Server
+```
+  This is the currently supported configuration.
+  It is suitable for one large department store.
+  For larger deployments there may be problems with the number of terminals or with network outages.
+  The SUMA Server must be reachable in order to perform any change of Terminal configuration.
+
+  Terminal
+    - deployed from OS Image during boot (by applying Saltboot state)
+    - then works as standard SUMA minion
+
+  Branch Server
+    - SUMA Proxy
+    - downloads Images from SUMA (over HTTP) and distributes them to Terminals (over HTTP, TFTP, etc.)
+    - runs services needed by Terminals (DHCP, TFTP, HTTP, ...). The services are configured via Formulas with Forms.
+
+  SUMA Server
+    - standard SUMA operations
+    - holds the configuration for Branch server services in Formulas
+    - controls boot process of each Terminal by applying Saltboot state
+    - triggers update of PXE data on Branch Server
+
+
+2. Full
 ```
     Terminal -> Branch Server -> Peripheral Server -> Hub
 ```
+  Adding Hub on top of previous configuration solves the problems with scalability and offline operations.
+  It is suitable for for large organization with many department stores. It can look like this:
 
-2. Combined SUMA and Branch server 
+  - Branch server on each floor
+  - Peripheral server in each building
+  - Hub in datacenter
+
+
+3. Combined Peripheral and Branch server
 ```
     Terminal -> Branch + Peripheral Server Combined -> Hub
 ```
 
-3. Standard
-```
-    Terminal -> Branch Server -> SUMA Server
-```
+  Customers with many small shops do need Hub for offline operations but do not need
+  separate Branch server and Peripheral Server because the number of connected terminals is low.
+  Running Branch + Peripheral Server on one machine could save resources.
+  The Proxy part of Branch server is not needed then.
+
 
 4. Single server
 ```
     Terminal -> Branch + SUMA Server Combined
 ```
+  Suitable for one small shop.
+  This scenario is not a hard requirement but there were such requests for SLEPOS
+  so it would be nice to have it too.
 
 
 Global Retail data
 ------------------
 
-The following items are typically shared in whole organization, so they should be visible and 
+The following items are typically shared in whole organization, so they should be visible and
 manageable via Hub.
 
   - Images - distributed as large tarballs via http
-  - Image metadata - pillars, list of packages
-  - HWTYPE configurations - group formula
-  
+  - Image metadata - pillars (sls files), list of repos and packages (in DB)
+  - [HWTYPE](https://www.uyuni-project.org/uyuni-docs/uyuni/retail/retail-deploy-terminals.html#_create_a_hardware_type_group) configurations
+
 Local data
 ----------
 
-The following items do not need central management and can stay on Peripheral Server.
+The following items do not need central management and can stay on Peripheral Server (like in the current implementation).
 
   - Branch groups -  groups containing Branch Server and connected terminals
   - Branch formulas - describe configuration of Branch services - network, pxe, dhcp, dns
   - Branch PXE data - entries describing how to boot each terminal
-  
+
 # Detailed design
 [design]: #detailed-design
 
 Images + Image metadata
 -----------------------
 
-Current situation: an image is built on Buildhost, then pushed to SUMA Server htdocs directory. Then SUMA creates a db entry (including package list) and a pillar file.
-The pillar conains URL of image tarball, checksums and additional deployment details. It is used for image synchronization and deployment. These data are extracted 
-on Buildhost at the end of build phase. Some of the data can't be extracted later form the tarball.
+Current situation: an image is built on Buildhost, then pushed to the Server's htdocs directory. Then the Server
+[creates a DB entry](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/utils/SaltUtils.java#L1311)
+(including package list) and
+[a pillar file](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/services/SaltStateGeneratorService.java#L128).
+The pillar conains URL of image tarball, checksums and additional deployment details. It is used for image synchronization and deployment. These data are extracted
+on the build host at the end of build phase. Some of the data are only available on the build host and, not being contained in the image tarball, can't be extracted later.
 
 The SUMA API related to images is rather docker-centered. For Kiwi images it allows to list images and packages and delete an image.
 
 Use cases currently not covered:
 
 - Move image tarball to external HTTP server
-  - In large environments it makes sense to use one or more servers for distribution of images, it should not be the Hub itself.
-- Copy image metadata from one SUMA Server to another 
-  - Hub - copy to all servers
-  - Test in lab, then move the same image to production.
+  In current implementation the image tarballs are stored in htdocs directory and distributed
+  via SUMA http server.
+  Branch Server via [Image Sync Formula](https://github.com/uyuni-project/retail/tree/master/image-sync-formula) downloads the tarball
+  from the URL in pillar, unpacks it and prepares the image for PXE service.
+
+  With Hub we can just extend this schema - tarball will be stored in Hub htdocs directory and copied to Peripheral servers via ISS.
+  This might not be optimal for certain cases (many small shops with combined Peripheral and Branch server).
+
+  So the alternative is to allow use of independent http servers and proxies (set up in customers infrastructure or in public cloud)
+  for distributing the tarballs. The http server structure does not have to strictly follow the Hub - Peripheral Server - Branch Server structure,
+  for example the Branch servers can download tarballs directly from datacenter or cloud.
+
+- Copy image metadata from one SUMA Server to another
+  Most common way of distributing image metadata is from Hub to Peripheral servers - this should be handled by
+  [ISS](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00063-hub-iss.md)
+
+  Another use case, not covered by the above, is this:
+  - Customer has one SUMA server in Lab environment and use it for developing images
+  - Only the images that pass all tests can be moved to production (which means move
+  them to Hub and distribute them)
+  - The approved image must stay the same (so it can't be rebuilt again)
+  - The development SUMA server should be as much separated from Hub as possible,
+  to be sure that the development does not break anything in production. Building
+  images on Hub is quite dangerous from this point of view.
+
 - Adjust deployment details
+  - currently the image pillars are write-only in SUMA, the users might want to change some details
 
 
 Proposed solution:
@@ -92,9 +154,11 @@ Proposed solution:
   - Update Kiwi Image pillar
   - Export Kiwi Image (complete metadata)
   - Import Kiwi Image
+    - The server that exported the image might use differnt repos, activation keys, etc. This call must make sure that
+    everything is consistent.
 
 - Implement a tool that handles the use cases above
-
+- Use the API in ISS
 
 HWTYPE configurations
 ---------------------
@@ -114,53 +178,82 @@ This approach is probably more flexible.
 
 Both alternatives will have to cope with authentization and Hub vs. Peripheral server user credentials.
 
+In long term, alternative 2 is clean solution.
+
+Alternative 1 might be used as a short term fix and also to keep compatibility.
+
+Note:
+The Formula pillar files depend on group IDs. The IDs differs between Hub and Peripheral server so
+direct file-based synchronization of pillars is not possible.
+
+
 YAML configuration
 ------------------
 
 FIXME: optional, related to Alternative 2 in previous section
 
-We have retail_yaml tool for initial configuration of Retail environment from YAML file.
+We have retail_yaml tool for initial configuration of Retail environment from
+[YAML](https://github.com/uyuni-project/retail/blob/master/python-susemanager-retail/example.yml) file.
 [Hub Users RFC](00061-hub-users.md) introduced similar concept, based on Salt.
 Maybe these parts could be merged.
 
 
-Combined SUMA and Branch server
--------------------------------
+Combined Peripheral and Branch server
+-------------------------------------
 
 Branch services are configured by Salt formulas. Also PXE entries for each terminal are generated
 by Salt (currently via reactor).
-With combined SUMA and Branch server, these formulas should be applied on SUMA server itself.
+With combined Peripheral and Branch server, these formulas should be applied on SUMA server itself.
 
 Terminals connect to SUMA directly, SUMA Proxy is not needed.
 
 
 FIXME: pick one alternative
 
-Alternative 1: Hub as a Salt Master for Combined SUMA and Branch server
+Alternative 1: Hub as a Salt Master for Combined Peripheral and Branch server
 
 - this is already used for [Hub Users RFC](00061-hub-users.md)
 
 problems:
-- does not work for PXE entries and disconnected operation
-- mixed environment with Hub, Peripheral Servers with standalone Branch Serveres and 
+- does not work for PXE entries https://github.com/uyuni-project/uyuni-rfc/pull/46
+  The event from each booting Terminal must be stopped on Peripheral server, on
+  Hub it would cause the same scalability problems as before.
+
+- mixed environment with Hub, Peripheral Servers with standalone Branch Serveres and
   Combined Peripheral Servers will have non-optimal configuration
+  For example: A shop uses Combined Peripheral and Branch server, then the number of terminals grows and
+  at some point it is necessary to add standalone Branch Server. Then the configuration of the
+  shop would be split between Hub and Peripheral Server, or the shop would have to be reinstalled
+  with 2 Branch Servers and standalone Peripheral Server
+
 - does not cover the "Single server" setup
 
 
-Alternative 2: Multi-master setup of Combined SUMA and Branch server
+Alternative 2: Multi-master setup of Combined Peripheral and Branch server
 
-- Salt minion running on Combined SUMA and Branch server has 2 masters - Hub and itself
+- Salt minion running on Combined Peripheral and Branch server has 2 masters - Hub and itself
+- the Branch Server configuration via formulas can stay unchanged
 
 problems:
 - SUMA registered to itself cause many problems
   - we already have clients that do not allow certain actions (containers) so this should be fixable
+  - if we allow and test this scenario for Retail, we will know how it behaves for
+  non-Retail use and the most severe problems will be fixed (instead of just saying "we don't know")
+- possible security implications - users will gain access to the server
+  - this is not a problem for Retail. Peripheral server will be managed by Hub admins who can access everything
+  and optionally by local admins who should access for example all machines in one building.
+  There is no need for other roles.
+
 - Salt Multi-master might not be reliable
 
 
 Alternative 3: Use salt-call --local for Branch server part of combined server
 
 - how to represent such setup in GUI and API ?
-- would need significant changes in SUMA code
+- would need significant changes in SUMA code:
+  - use salt-call --local instead of Salt API
+  - make sure that pillars are available
+  - maybe changes in handling salt events
 
 
 Alternative 4: Virtualize Branch server part of combined server
@@ -172,6 +265,8 @@ Alternative 4: Virtualize Branch server part of combined server
 - Branch server as VM could be solved by documentation only - the user might be
 aready using some virtualization setup and we don't want to break it.
 
+Problems:
+- higher HW requirements
 
 
 # Drawbacks
@@ -179,10 +274,7 @@ aready using some virtualization setup and we don't want to break it.
 
 Why should we **not** do this?
 
-  * obscure corner cases
-  * will it impact performance?
-  * what other parts of the product will be affected?
-  * will the solution be hard to maintain in the future?
+TBD
 
 # Alternatives
 [alternatives]: #alternatives
@@ -193,6 +285,9 @@ Images + Image metadata:
 - Implement Kiwi Image Store - HTTP server + API to handle metadata (similar functionality as Docker store)
 
 - Work with Kiwi upstream to include all the data to the image tarball, so the import functionality can be simplified
+  - it will not work for old images build without the metadata,
+  - some metadata are too SUMA-specific - for example activation key.
+  - need to download and unpack the tarball, which can be stored on different machine.
 
 - Use of Bittorrent protocol to distribute images from Hub to branches
   - It is expected that the VPN network topology is star-like, with Hub in the center. In this setup Bittorrent protocol won't help.
@@ -202,12 +297,14 @@ HWTYPE configurations:
 
 TBD
 
-Combined SUMA and Branch server:
+Combined Peripheral and Branch server:
 
 TBD
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
-- What are the unknowns?
-- What can happen if Murphy's law holds true?
+- is nested Hub needed?
+  The use case with one Peripheral Server per shop might exceed the maximum number of Peripheral Servers.
+  This can be decided later, based on feedback.
+
--- a/accepted/0000-hub_retail.md
+++ b/accepted/0000-hub_retail.md
@@ -1,0 +1,213 @@
+- Feature Name: Hub Retail Extension
+- Start Date: 2020-11-30
+- RFC PR: (leave this empty)
+
+# Summary
+[summary]: #summary
+
+This RFC describes an extension of retail features for Hub environment.
+
+# Motivation
+[motivation]: #motivation
+
+Reasons for use of Hub with Retail environment
+----------------------------------------------
+
+- scalability
+  - support large environments with the number of terminal exceeding limits of one SUMA server
+
+- disconnected operation
+  - Peripheral server must stay fully functional when the Hub connection is down
+
+
+Possible configurations
+-----------------------
+
+1. Full
+```
+    Terminal -> Branch Server -> Peripheral Server -> Hub
+```
+
+2. Combined SUMA and Branch server 
+```
+    Terminal -> Branch + Peripheral Server Combined -> Hub
+```
+
+3. Standard
+```
+    Terminal -> Branch Server -> SUMA Server
+```
+
+4. Single server
+```
+    Terminal -> Branch + SUMA Server Combined
+```
+
+
+Global Retail data
+------------------
+
+The following items are typically shared in whole organization, so they should be visible and 
+manageable via Hub.
+
+  - Images - distributed as large tarballs via http
+  - Image metadata - pillars, list of packages
+  - HWTYPE configurations - group formula
+  
+Local data
+----------
+
+The following items do not need central management and can stay on Peripheral Server.
+
+  - Branch groups -  groups containing Branch Server and connected terminals
+  - Branch formulas - describe configuration of Branch services - network, pxe, dhcp, dns
+  - Branch PXE data - entries describing how to boot each terminal
+  
+# Detailed design
+[design]: #detailed-design
+
+Images + Image metadata
+-----------------------
+
+Current situation: an image is built on Buildhost, then pushed to SUMA Server htdocs directory. Then SUMA creates a db entry (including package list) and a pillar file.
+The pillar conains URL of image tarball, checksums and additional deployment details. It is used for image synchronization and deployment. These data are extracted 
+on Buildhost at the end of build phase. Some of the data can't be extracted later form the tarball.
+
+The SUMA API related to images is rather docker-centered. For Kiwi images it allows to list images and packages and delete an image.
+
+Use cases currently not covered:
+
+- Move image tarball to external HTTP server
+  - In large environments it makes sense to use one or more servers for distribution of images, it should not be the Hub itself.
+- Copy image metadata from one SUMA Server to another 
+  - Hub - copy to all servers
+  - Test in lab, then move the same image to production.
+- Adjust deployment details
+
+
+Proposed solution:
+
+- Implement API calls to
+  - Get Kiwi Image pillar
+  - Update Kiwi Image pillar
+  - Export Kiwi Image (complete metadata)
+  - Import Kiwi Image
+
+- Implement a tool that handles the use cases above
+
+
+HWTYPE configurations
+---------------------
+
+This is currently implemented as a formula assigned to a group with special name.
+
+FIXME: pick one alternative
+
+Alternative 1:
+Extend retail_yaml to create the groups and formula data via Hub and Peripheral Server API.
+This is relatively small change to the code.
+
+Alternative 2:
+Central creation of groups is covered in [Hub Users RFC](00061-hub-users.md). The salt states
+can be extended to handle also formula data.
+This approach is probably more flexible.
+
+Both alternatives will have to cope with authentization and Hub vs. Peripheral server user credentials.
+
+YAML configuration
+------------------
+
+FIXME: optional, related to Alternative 2 in previous section
+
+We have retail_yaml tool for initial configuration of Retail environment from YAML file.
+[Hub Users RFC](00061-hub-users.md) introduced similar concept, based on Salt.
+Maybe these parts could be merged.
+
+
+Combined SUMA and Branch server
+-------------------------------
+
+Branch services are configured by Salt formulas. Also PXE entries for each terminal are generated
+by Salt (currently via reactor).
+With combined SUMA and Branch server, these formulas should be applied on SUMA server itself.
+
+Terminals connect to SUMA directly, SUMA Proxy is not needed.
+
+
+FIXME: pick one alternative
+
+Alternative 1: Hub as a Salt Master for Combined SUMA and Branch server
+
+- this is already used for [Hub Users RFC](00061-hub-users.md)
+
+problems:
+- does not work for PXE entries and disconnected operation
+- mixed environment with Hub, Peripheral Servers with standalone Branch Serveres and 
+  Combined Peripheral Servers will have non-optimal configuration
+- does not cover the "Single server" setup
+
+
+Alternative 2: Multi-master setup of Combined SUMA and Branch server
+
+- Salt minion running on Combined SUMA and Branch server has 2 masters - Hub and itself
+
+problems:
+- SUMA registered to itself cause many problems
+  - we already have clients that do not allow certain actions (containers) so this should be fixable
+- Salt Multi-master might not be reliable
+
+
+Alternative 3: Use salt-call --local for Branch server part of combined server
+
+- how to represent such setup in GUI and API ?
+- would need significant changes in SUMA code
+
+
+Alternative 4: Virtualize Branch server part of combined server
+
+- use separate VM / container for the Branch server
+- the minion connects to SUMA on the same physical host
+
+- solves all the problems of other alternatives
+- Branch server as VM could be solved by documentation only - the user might be
+aready using some virtualization setup and we don't want to break it.
+
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we **not** do this?
+
+  * obscure corner cases
+  * will it impact performance?
+  * what other parts of the product will be affected?
+  * will the solution be hard to maintain in the future?
+
+# Alternatives
+[alternatives]: #alternatives
+
+
+Images + Image metadata:
+
+- Implement Kiwi Image Store - HTTP server + API to handle metadata (similar functionality as Docker store)
+
+- Work with Kiwi upstream to include all the data to the image tarball, so the import functionality can be simplified
+
+- Use of Bittorrent protocol to distribute images from Hub to branches
+  - It is expected that the VPN network topology is star-like, with Hub in the center. In this setup Bittorrent protocol won't help.
+
+
+HWTYPE configurations:
+
+TBD
+
+Combined SUMA and Branch server:
+
+TBD
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- What are the unknowns?
+- What can happen if Murphy's law holds true?


accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,188 @@
...
+- Implement a tool that handles the use cases above
+
+
+HWTYPE configurations
+---------------------
+
+This is currently implemented as a formula assigned to a group with special name.
+
+Proposed solution:
+Extend retail_yaml to create the groups and formula data via Hub and Peripheral Server API.
```
aaannz, line 102: `retail_yaml` creates both HWTYPE and branch groups. As stated above HWTYPE is supposed to be hub group, branch is supposed to be system group. Modifying these two can have different set of credentials. This can have an impact on user experience.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,188 @@
...
+
+Alternative:
+Central creation of grups is covered in [Hub Users RFC](00061-hub-users.md). The salt states
+can be extended to handle also formula data.
```
aaannz, line 106: Typo and I think this should be the main solution and then adapting `retail_yaml` to use Hub API

```suggestion
Central creation of groups is covered in [Hub Users RFC](00061-hub-users.md). The salt states
can be extended to handle also formula data.
```

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,188 @@
...
+- mixed environment with Hub, Peripheral Servers with standalone Branch Serveres and 
+  Combined Peripheral Servers will have non-optimal configuration
+- does not cover the "Single server" setup
+
+Alternative 2: Multi-master setup of Combined SUMA and Branch server
+
+- Salt minion running on Combined SUMA and Branch server has 2 masters - Hub and itself
+
+problems:
+- SUMA registered to itself cause many problems, this can't be supported
```
aaannz, line 131: I would move this one as a last one and clearly mark as rejected alternative.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,188 @@
...
+- how to represent such setup in GUI and API ?
+- would need significant changes in SUMA code
+
+
+Alternative 4: Virtualize Branch server part of combined server
+
+- use separate VM / container for the Branch server
+- the minion connects to SUMA on the same physical host
+
+- solves all the problems of other alternatives
```
aaannz, line 145: What needs to be done to support this alternative? Is it just documentation?
Can we prepare OS Image template for branch server image and use virtualization support of Uyuni/SUSE Manager to deploy it on itself? With that comes question as some users run Uyuni in virtualized environment, nested virtualization will need to be allowed.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,188 @@
...
+
+Proposed solution:
+Extend retail_yaml to create the groups and formula data via Hub and Peripheral Server API.
+
+Alternative:
+Central creation of grups is covered in [Hub Users RFC](00061-hub-users.md). The salt states
+can be extended to handle also formula data.
+
+
+Combined SUMA and Branch server
```
aaannz, line 109: I am not sure how much is this relevant to Hub itself. I am missing a chapter discussing disconnected usage itself.

Also when talking about combined Uyuni and Branch server, it should be noted that we are talking about Uyuni taking up Branch responsibilities, but not becoming proxy itself, am I right? That means Uyuni peripheral server will provide necessary pxe, tftp, dhcp, dns and http services usually covered by Branch server proxy

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,188 @@
...
+
+3. Standard
+```
+    Terminal -> Branch Server -> SUMA Server
+```
+
+4. Single server
+```
+    Terminal -> Branch + SUMA Server Combined
+```
```
aaannz, line 41: I would really add something like `(rejected/unsupported)` here as this one cannot be recommended.

accepted/0000-hub_retail.md
nadvornik: We have to pick one alternative and mark the remaining 3 as rejected.

accepted/0000-hub_retail.md
nadvornik: I don't know yet.

accepted/0000-hub_retail.md
nadvornik: It depends on the alternative we pick. We might get it for free.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+# Motivation
+[motivation]: #motivation
+
+Reasons for use of Hub with Retail environment
+----------------------------------------------
+
+- scalability
+  - support large environments with the number of terminal exceeding limits of one SUMA server
+
+- disconnected operation
```
rjmateus, line 19: One thing that I'm missing in here is what are the need for the retail in a large setup scenario, which problem will it face and if the current solution with branch can (or cannot) help.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+
+Images + Image metadata:
+
+- Implement Kiwi Image Store - HTTP server + API to handle metadata (similar functionality as Docker store)
+
+- Work with Kiwi upstream to include all the data to the image tarball, so the import functionality can be simplified
```
rjmateus, line 195: what are the cons of this alternative (if any)

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+
+- Use of Bittorrent protocol to distribute images from Hub to branches
```
rjmateus, line 197: Should branch servers be registered directly to a hub server? Or we are talking about peripheral servers as branch server?

rjmateus (commented):
One question that I feel not totally explored is if we should or not keep the branch server.
One option is exactly what we have now, and add the HUB on top, so we can manage more suse manager server from a centralized place. However, this will not solve the offline of retail.
Merge branch server to suse manager brings other problems also.
Should we have both? Susemanager with branch server capabilities and keep also the current implementation of branch server?

nadvornik:
> One question that I feel not totally explored is if we should or not keep the branch server.
> One option is exactly what we have now, and add the HUB on top, so we can manage more suse manager server from a centralized place. However, this will not solve the offline of retail.
> Merge branch server to suse manager brings other problems also.
> Should we have both? Susemanager with branch server capabilities and keep also the current implementation of branch server?

We have to keep what we have now - to support existing deployment.
Hub will be added on the top. For large organization with many department stores it can look like this:

Branch server on each floor,
Peripheral server in each building,
Hub in datacenter.

Smaller deployment can use the Combined Branch and Peripheral servers if they need Hub for offline operation
or just skip Hub if they are OK with current functionality.

So we have to support all these combinations.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+-----------------------
+
+1. Full
+```
+    Terminal -> Branch Server -> Peripheral Server -> Hub
+```
+
+2. Combined SUMA and Branch server 
+```
+    Terminal -> Branch + Peripheral Server Combined -> Hub
```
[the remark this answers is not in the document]
nadvornik: yes

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+manageable via Hub.
+
+  - Images - distributed as large tarballs via http
+  - Image metadata - pillars, list of packages
+  - HWTYPE configurations - group formula
+  
+Local data
+----------
+
+The following items do not need central management and can stay on Peripheral Server.
```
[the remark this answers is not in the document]
nadvornik: Yes, these data are now on SUMA Server and should stay there even if Hub is added.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+
+Current situation: an image is built on Buildhost, then pushed to SUMA Server htdocs directory. Then SUMA creates a db entry (including package list) and a pillar file.
+The pillar conains URL of image tarball, checksums and additional deployment details. It is used for image synchronization and deployment. These data are extracted 
+on Buildhost at the end of build phase. Some of the data can't be extracted later form the tarball.
+
+The SUMA API related to images is rather docker-centered. For Kiwi images it allows to list images and packages and delete an image.
+
+Use cases currently not covered:
+
+- Move image tarball to external HTTP server
```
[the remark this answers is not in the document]
nadvornik: Branch server downloads the image tarball and then serves it to terminals.
This is about Branch server downloading the tarball.
Now it downloads it from SUMA http server, but any http server can work.
For large environment it can be dedicated http server, or maybe public cloud.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+problems:
+- does not work for PXE entries and disconnected operation
+- mixed environment with Hub, Peripheral Servers with standalone Branch Serveres and 
+  Combined Peripheral Servers will have non-optimal configuration
+- does not cover the "Single server" setup
+
+
+Alternative 2: Multi-master setup of Combined SUMA and Branch server
+
+- Salt minion running on Combined SUMA and Branch server has 2 masters - Hub and itself
```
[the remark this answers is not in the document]
nadvornik: Branch server consist of several services which are configured via Formulas with Forms (via UI or API).
If we merge Branch server with Peripheral server, the formulas have to be applied on the Peripheral server itself - so the Peripheral server minion must be connected to Peripheral server master. At the same time, the minion must connect to the Hub master for Hub functionality.
According to Salt documentation, this should work as long as there are no state conflicts.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+
+problems:
+- SUMA registered to itself cause many problems
+  - we already have clients that do not allow certain actions (containers) so this should be fixable
+- Salt Multi-master might not be reliable
+
+
+Alternative 3: Use salt-call --local for Branch server part of combined server
```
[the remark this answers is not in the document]
nadvornik: This is what I meant by the alternative 1, see the list of problems there.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+aready using some virtualization setup and we don't want to break it.
+
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we **not** do this?
+
+  * obscure corner cases
```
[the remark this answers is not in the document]
nadvornik: These are leftovers from template :-)
It needs to be fixed.

accepted/0000-hub_retail.md
nadvornik: it will not work for old images build without the metadata,
Some metadata are too SUMA-specific - for example activation key.
Need to download and unpack the tarball, which can be stored on different machine.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+on Buildhost at the end of build phase. Some of the data can't be extracted later form the tarball.
+
+The SUMA API related to images is rather docker-centered. For Kiwi images it allows to list images and packages and delete an image.
+
+Use cases currently not covered:
+
+- Move image tarball to external HTTP server
+  - In large environments it makes sense to use one or more servers for distribution of images, it should not be the Hub itself.
+- Copy image metadata from one SUMA Server to another 
+  - Hub - copy to all servers
```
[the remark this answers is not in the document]
nadvornik: The image data are too complex, partially stored in files, partially in DB so it needs to go via API to make sure the result is consistent.
ISS then can use this API.

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
rjmateus: So, If I understand correctly, a suse manager server can live on the same machine with the Branch server?
A peripheral server is just a suse manager server.
Hub server is a suse manager server with some more services runing.
Instead of register the peripheral server to itself, shouldn't we register it to the suse manager hub server, and configure the needed branch formulas from hub?

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
rjmateus: New ISS implementation has access to database and file system. We can add any complex logic we need on there. What would be the benefits of having an API endpoint for it instead of using the ISS mechanism?

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+
+Alternative 1: Hub as a Salt Master for Combined SUMA and Branch server
+
+- this is already used for [Hub Users RFC](00061-hub-users.md)
+
+problems:
+- does not work for PXE entries and disconnected operation
+- mixed environment with Hub, Peripheral Servers with standalone Branch Serveres and 
+  Combined Peripheral Servers will have non-optimal configuration
+- does not cover the "Single server" setup
```
rjmateus, line 147: How does the current use case with one single server is covered now? IIUC the branch server would be instaled in a different machine from suse manager server

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+Both alternatives will have to cope with authentization and Hub vs. Peripheral server user credentials.
+
+YAML configuration
+------------------
+
+FIXME: optional, related to Alternative 2 in previous section
+
+We have retail_yaml tool for initial configuration of Retail environment from YAML file.
+[Hub Users RFC](00061-hub-users.md) introduced similar concept, based on Salt.
+Maybe these parts could be merged.
```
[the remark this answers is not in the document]
aaannz: I agree with this.
Points to consider:
* `retail_yaml` provides some default values for data not provided by customer and also some generic configuration
* `retail_yaml` is already being used by users, we should strive to not break their systems/workflows

Maybe have common backend and have retail_yaml to just fill defaults as now and pass them to this common backend?

accepted/0000-hub_retail.md
aaannz: Currently there is no supported single server scenario. Supported scenario always includes branch server proxy.
I think that will remain the case for non-hub environment, unless we safely figure out how to enable single Uyuni server to manage itself (including applying salt states on itself).

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
aaannz: > Instead of register the peripheral server to itself, shouldn't we register it to the suse manager hub server, and configure the needed branch formulas from hub?

Than this would be scenario 1 I think.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+
+
+FIXME: pick one alternative
+
+Alternative 1: Hub as a Salt Master for Combined SUMA and Branch server
+
+- this is already used for [Hub Users RFC](00061-hub-users.md)
+
+problems:
+- does not work for PXE entries and disconnected operation
```
aaannz, line 144: I am not sure why it would not work for disconnected scenarios? There will need to be an access to hub when configuring services (downloading image, applying states). But after that it should work disconnected.

What is here to solve is pxe generation as for that we need to be able to apply states.

accepted/0000-hub_retail.md
aaannz: I see this as that Hub where images are stored will run bittorrent tracker. Image pillar then can just include magnet link and branches/peripheral servers will download it. As long as other branches/peripheral servers are also seeding the images this can help with lowering the strain on hub to distribute the images. Alternative to http download used by image-sync state.

aaannz:
I remembered one thing not discussing here and that is multi-organizations.
@rjmateus what is the take of Hub on having multiple organizations on Peripheral servers? Are they allowed, supported?

When terminal is being onboarded it can't use any activation key as the initrd is shared between different images thus different HWTYPEs. So to detect to which organization terminal belongs during registration routine proxy server is looked up and the organization of that proxy server is used (see [RegisterMinionEventMessageAction.java#L330](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java#L330)).
As with peripheral server we don't have this kind of relationship and if peripheral server allows multi organizational use, how are we going to figure out correct organization for terminal?

accepted/0000-hub_retail.md
nadvornik: I mean that disconnected scenarios do not work with PXE entries and maybe in the future for generating static DNS and DHCP entries.
Yes, the event must be terminated on peripheral server,  otherwise we would have the same scalability problem as before.

rjmateus:
> I remembered one thing not discussing here and that is multi-organizations.
> @rjmateus what is the take of Hub on having multiple organizations on Peripheral servers? Are they allowed, supported?

Our idea is to support multiple organizations on HUB and peripheral servers. 
In ISS, for example, current implementation allow us to copy data from one organization on master to a different organization on slave suse manager server.

Current approach (an this can/will be changed in future) master and slave organizations will be the same, and we don't allow copy data between different organization on master and slave.

> When terminal is being onboarded it can't use any activation key as the initrd is shared between different images thus different HWTYPEs. So to detect to which organization terminal belongs during registration routine proxy server is looked up and the organization of that proxy server is used (see [RegisterMinionEventMessageAction.java#L330](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java#L330)).
> As with peripheral server we don't have this kind of relationship and if peripheral server allows multi organizational use, how are we going to figure out correct organization for terminal?

This is a good point and I don't have a anwser to it. If will be using suse manager in the same machine with branch server, and terminals registered directed to a suse manager instance, then we should have a new/different way to findout the organization terminal should be part of.

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
nadvornik: The API gives us more flexibility. I am thinking about the scenario where one Peripheral server will be used in lab, for building and testing images. Only the successfully tested images should be distributed via Hub to other Peripheral servers.
I am not sure if ISS can handle this.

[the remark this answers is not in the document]
rjmateus: In the HUB initial idea, all content should be done in HUB, and then be sync to peripheral servers.
We should be able to build images in HUB, sync them to a lab peripheral server, test them in there. If we customer is satisfied, then they can sync them to all other existing peripheral servers.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+- Implement API calls to
+  - Get Kiwi Image pillar
+  - Update Kiwi Image pillar
+  - Export Kiwi Image (complete metadata)
+  - Import Kiwi Image
+
+- Implement a tool that handles the use cases above
+
+
+HWTYPE configurations
```
moio, line 99: Where can I read about HWTYPE configurations in general?

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
moio: Can you point me to the current `retail_yaml` code to get a better idea on what it does in detail, please?

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+  - Images - distributed as large tarballs via http
+  - Image metadata - pillars, list of packages
+  - HWTYPE configurations - group formula
+  
+Local data
+----------
+
+The following items do not need central management and can stay on Peripheral Server.
+
+  - Branch groups -  groups containing Branch Server and connected terminals
```
rjmateus, line 62: IIUC those are regular system groups. They can be created on a peripheral server from the HUB user the uyuni configuration formula. https://github.com/SUSE/salt-formulas/tree/master/uyuni-config-formula

aaannz: Yes, regular groups. Purpose is to help with navigating large number of terminals. Also they are mandatory for successful onboarding of minion (existence is checked during minion registration)

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+  - Branch formulas - describe configuration of Branch services - network, pxe, dhcp, dns
+  - Branch PXE data - entries describing how to boot each terminal
+  
+# Detailed design
+[design]: #detailed-design
+
+Images + Image metadata
+-----------------------
+
+Current situation: an image is built on Buildhost, then pushed to SUMA Server htdocs directory. Then SUMA creates a db entry (including package list) and a pillar file.
```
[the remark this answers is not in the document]
aaannz: Pillar generated in [SaltStateGeneratorService.java#L128](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/services/SaltStateGeneratorService.java#L128)
DB entries prepared here I think [SaltUtils.java#L1311](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/utils/SaltUtils.java#L1311)

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+The pillar conains URL of image tarball, checksums and additional deployment details. It is used for image synchronization and deployment. These data are extracted 
+on Buildhost at the end of build phase. Some of the data can't be extracted later form the tarball.
```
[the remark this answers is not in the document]
aaannz: Not all data from the build are copied over from the build host. Kiwi creates python pickle from where some data are extracted after image has been built and for this to function there must be Kiwi python files present. Kiwi is currently not installed directly on Uyuni server.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+  - Test in lab, then move the same image to production.
+- Adjust deployment details
+
+
+Proposed solution:
+
+- Implement API calls to
+  - Get Kiwi Image pillar
+  - Update Kiwi Image pillar
+  - Export Kiwi Image (complete metadata)
```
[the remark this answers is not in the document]
aaannz: There is one use case which is not directly related to Hub but Vladimir keeps in mind as well.
We have a user who have multiple SUSE Manager servers. One for dev, second for QA and third for prod. They build image on dev machine and want the very same image to be tested by QA and then moved to production. No modification to image allowed. For this scenario it would be great to have image import/export and this IMO can help with Hub as well.

Imagine in Hub architecture one Uyuni server to be dedicated to Development. When work is done here then using some mechanism import image to Hub server and then we can use whatever mechanism to distribute the images.

accepted/0000-hub_retail.md
aaannz: [Retail documentation](https://www.uyuni-project.org/uyuni-docs/uyuni/retail/retail-deploy-terminals.html#_create_a_hardware_type_group)

They are just regular server groups only following specific naming and have saltboot formula assigned to them.

accepted/0000-hub_retail.md
admd: So if I understood correctly with the above discussion. Hub in retail scenario will still be an optional component and if an existing customer wants to keep the original setup with branch servers, they can keep it.

What I don't get is that if customer decide for hub and we decide to merge the branch server with peripheral server(susemanager). Will we have 2 different version of prephiral servers?

-  peripheral servers with possible branch server capabilities 
-  peripheral server(suse manager in current shape)

admd:
> > When terminal is being onboarded it can't use any activation key as the initrd is shared between different images thus different HWTYPEs. So to detect to which organization terminal belongs during registration routine proxy server is looked up and the organization of that proxy server is used (see [RegisterMinionEventMessageAction.java#L330](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java#L330)).
> > As with peripheral server we don't have this kind of relationship and if peripheral server allows multi organizational use, how are we going to figure out correct organization for terminal?
> 
> This is a good point and I don't have a anwser to it. If will be using suse manager in the same machine with branch server, and terminals registered directed to a suse manager instance, then we should have a new/different way to findout the organization terminal should be part of.

IIUC, this should be covered already with current implementation unless I am missing something about the onboarding process of the termians. If AK cannot be used, we will try to get from user's who is onborading and in last resort the default orgnization of the server. Just like as defined here [RegisterMinionEventMessageAction.java#L330](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java#L330)

accepted/0000-hub_retail.md
aaannz: > peripheral servers with possible branch server capabilities

Don't look at this like something special. It is just that this peripheral server will have enabled some retail formula (branch network, pxe, tftp and maybe dhcp and dns) and in some cases more than one network interface.
Once we figure out the problem with PXE entries I expect that solution to be used in all configuration.

rjmateus: I would say it would be expected to have both.
Peripheral servers with branch server capabilities, which can be deployed in a store and manage terminal directly.
Also, a normal suse manager server which is responsible to manage proxy/branch servers deployed in several different stores.
And probably both coexisting in the same deployment infrastrucure.

aaannz:
> IIUC, this should be covered already with current implementation unless I am missing something about the onboarding process of the termians. If AK cannot be used, we will try to get from user's who is onborading and in last resort the default orgnization of the server. Just like as defined here [RegisterMinionEventMessageAction.java#L330](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/reactor/messaging/RegisterMinionEventMessageAction.java#L330)

Who is the user doing onboarding when you simply point salt-minion to salt master? Who is the user doing onboarding when salt keys autoaccept is enabled?
Default of server is ok if we do not allow multiple organizations on peripheral server.

I added the third option about the proxy organization specifically for this case. We have a user who use multiorganizations in retail case and they were having issues terminals registering to default organization instead of the organization they used for retail configuration.

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
moio: ISS is designed to move any content from any Server to any Server. Its main use case will be from Hub to peripheral servers, but nothing stops an enterprising user to come up with more elaborated schemes.

This one in particular looks pretty convoluted to me, but I guess we can (and actually should) leave this particular door open.

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
moio: Thanks! Including in text for reference:
```suggestion
Current situation: an image is built on Buildhost, then pushed to the Server's htdocs directory. Then the Server [creates a DB entry](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/utils/SaltUtils.java#L1311) (including package list) and [a pillar file](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/services/SaltStateGeneratorService.java#L128).
```

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
moio:
```suggestion
on the build host at the end of build phase. Some of the data are only available on the build host and, not being contained in the image tarball, can't be extracted later.
```

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+Current situation: an image is built on Buildhost, then pushed to SUMA Server htdocs directory. Then SUMA creates a db entry (including package list) and a pillar file.
+The pillar conains URL of image tarball, checksums and additional deployment details. It is used for image synchronization and deployment. These data are extracted 
+on Buildhost at the end of build phase. Some of the data can't be extracted later form the tarball.
+
+The SUMA API related to images is rather docker-centered. For Kiwi images it allows to list images and packages and delete an image.
+
+Use cases currently not covered:
+
+- Move image tarball to external HTTP server
+  - In large environments it makes sense to use one or more servers for distribution of images, it should not be the Hub itself.
```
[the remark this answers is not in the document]
moio: Idea in the new ISS implementation is that it acts like an "exporter/importer".

On one end, ISS "exports" data from a Server's (filesystem and DB) to a directory, then on the other end it is "imported" into another Server. At the moment the mechanism to transfer the directory is still not decided.

For simple (low-scale) cases we could just use Salt's transport or plain http, while in more complicated cases any kind of "CDN", including possibly public clouds, may be used.

accepted/0000-hub_retail.md
moio:
```suggestion
[HWTYPE](https://www.uyuni-project.org/uyuni-docs/uyuni/retail/retail-deploy-terminals.html#_create_a_hardware_type_group) configurations
```

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
moio: From a quick look it seems to me that `retail_yaml` is Python code that does configuration, which by definition is very similar to a Salt module. I guess retail_yaml could be broken into one or more Salt modules and, at least in principle, be substitured or become a wrapper to `salt minion state.apply retail` or something similar.

Am I getting this right?

accepted/0000-hub_retail.md
moio: > I am not sure how much is this relevant to Hub itself.

In my understanding, it is in principle not relevant and could be tackled by a completely separate effort/RFC.

But once the capability of managing clients without a Proxy/RBS exists, then it's much easier to implement scenario 2 listed above ("just add a hub on top").

Am I on the right track here?

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+Branch services are configured by Salt formulas. Also PXE entries for each terminal are generated
+by Salt (currently via reactor).
+With combined SUMA and Branch server, these formulas should be applied on SUMA server itself.
+
+Terminals connect to SUMA directly, SUMA Proxy is not needed.
+
+
+FIXME: pick one alternative
+
+Alternative 1: Hub as a Salt Master for Combined SUMA and Branch server
```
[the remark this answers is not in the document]
moio: I think what @nadvornik means here is:

Hub -> Peripheral Server + RBS services/components -> POS terminals

Hub will not have Proxies/RBSs attached.

accepted/0000-hub_retail.md
moio: Where are PXE entries supposed to be set and when, and by what component?

accepted/0000-hub_retail.md
moio: > enable single Uyuni server to manage itself (including applying salt states on itself).

What management features would be needed? Registering a Server onto itself is not an option at the moment.

But applying Salt states is very easy, we can use `salt-call --local` especially if it's a one-shot setup only.

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
moio: Registering a SUSE Manager onto itself (or having a salt-minion attached to a salt-master on the same machine, for that matter) opens up a considerable can of worms and we have to have strong reasons to dig into that direction.

Among other things, consider that while salt-master is by and large an unprivileged service, salt-minion runs as root, so any user accessing the machine can potentially use the salt-master and get root access to the machine itself as well. Arguably that's not a whole lot better than accessing all minions in any case, but still it's not a nice proposition from a security perspective.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,213 @@
...
+problems:
+- SUMA registered to itself cause many problems
+  - we already have clients that do not allow certain actions (containers) so this should be fixable
+- Salt Multi-master might not be reliable
+
+
+Alternative 3: Use salt-call --local for Branch server part of combined server
+
+- how to represent such setup in GUI and API ?
+- would need significant changes in SUMA code
```
moio, line 163: Are the changes only about allowing to apply formulas (optionally with forms) via `--local` or do you see other changes? Which?

accepted/0000-hub_retail.md
moio: Containerizing the Proxy is something that was imagined and even attempted in the past. Solving this case might be an additional good side-effect, although I am not yet sold on throwing in containerization yet.

accepted/0000-hub_retail.md
moio: I'd leave the transport choice up to the ISS implementation and yes, I also see Bittorrent as a possibility.

accepted/0000-hub_retail.md
aaannz: During saltboot process terminal sends event `pxe_update`, this event is processed by reactor and applies `pxe/terminal_entry` state on branch server.

This happens after terminal is deployed, but before switching from initrd to system image. In case kernel of initrd does not match kernel of system, terminal is tried to kexec to new kernel. However if kexec is not possible or fails, terminal is rebooted. For this pxe entry needs to be already prepared as it specifies correct kernel and initrd to use for the machine.

accepted/0000-hub_retail.md
aaannz: Talking about non-hub usage it would need to read formula pillars and apply states (formula states and pxe entry)

In hub environment, there is only issue about pxe entry state as the configuration of formulas can be done from hub.

accepted/0000-hub_retail.md
nadvornik: Setup  without RBS is requested in the issue https://github.com/SUSE/spacewalk/issues/9594
I think it was in the original Hub requirements, but he link in the issue is now dead.

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
nadvornik: yes

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
nadvornik: The Hub users and groups implementation use XMLRPC API called from salt, so this should be already solved. 

In long term we should use salt, the only question is how to handle the transition.
Fixing retail_yaml to work with Hub API is probably easier than rewriting everything with salt and we probably have to keep the original tool with original syntax for compatibility.

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
nadvornik: I think this is not a problem for Retail.
Peripheral server will be managed by Hub admins who can access everything 
and optionally by local admins who should access for example all machines in one building.
There is no need for other roles.

accepted/0000-hub_retail.md
nadvornik: I think it would need changes in code to apply formulas and supply pillars for them, maybe also in handling of salt events.

accepted/0000-hub_retail.md
nadvornik: Proxy is not needed, terminals can connect to Peripheral servers directly.

This is about containerizing salt minion and the Branch services.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,310 @@
...
+
+Possible configurations
+-----------------------
+
+1. Standard
+```
+    Terminal -> Branch Server -> SUMA Server
+```
+  This is the currently supported configuration.
+  It is suitable for one large department store.
```
aaannz, line 32: > It is suitable for one large department store.

Really? Maybe I have different definition of department store, but we have users operating multiple department stores. This is a nitpick, but IMO this makes disservice to the solution.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,310 @@
...
+  - Customer has one SUMA server in Lab environment and use it for developing images
+  - Only the images that pass all tests can be moved to production (which means move
+  them to Hub and distribute them)
+  - The approved image must stay the same (so it can't be rebuilt again)
+  - The development SUMA server should be as much separated from Hub as possible,
+  to be sure that the development does not break anything in production. Building
+  images on Hub is quite dangerous from this point of view.
+
+- Adjust deployment details
+  - currently the image pillars are write-only in SUMA, the users might want to change some details
```
aaannz, line 147: > currently the image pillars are write-only in SUMA

I don't follow. What do you mean by `write-only`? Do you mean read only?

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,310 @@
...
+
+problems:
+- SUMA registered to itself cause many problems
+  - we already have clients that do not allow certain actions (containers) so this should be fixable
+  - if we allow and test this scenario for Retail, we will know how it behaves for
+  non-Retail use and the most severe problems will be fixed (instead of just saying "we don't know")
+- possible security implications - users will gain access to the server
+  - this is not a problem for Retail. Peripheral server will be managed by Hub admins who can access everything
+  and optionally by local admins who should access for example all machines in one building.
+  There is no need for other roles.
```
aaannz, line 245: This is an assumption that we know how our users use the product with which I can't agree.

Given the possibilities of Uyuni I am pretty sure users would want to use peripheral server for more then just 'offline capable proxy', so management of terminals, but also management of servers running local database instance for retail app and what not.
For that it is easy to imagine they have different users with different roles, e.g. one junior admin assigned to branch group maintaining terminals with low level of risk. Then different admin, more senior posion, assigned to group maintaining backend services.
Now if this user follow recommendation wrt proxy we currently have (and retail_yaml does automatically), they may be tempted to add this peripheral server to the group with terminals as we do with branch proxy. At this moment the junior admin maintaining terminals will get root access to the peripheral server itself and thus root access to backend servers as well.

accepted/0000-hub_retail.md
nadvornik: As I wrote on the next line, larger deployments have the problems that we are trying to solve by this RFC.

accepted/0000-hub_retail.md
nadvornik: Java code just writes the pillars and forgets about them. There is no support for reading back and modifying.

accepted/0000-hub_retail.md
nadvornik: I don't think that one small shop (which needs this combined server to save resources) will need all these admin roles.

aaannz: I am sorry, but this is not just one small shop. This is scenario with hub and combined peripheral server. This would be every user who want to have offline functionality.

But at the end all of this is speculation. I just really don't like statements 'There is no need for other roles.' IT is plenty of similar statements only to be proved wrong. Thus I don't agree with using these statements as a reasoning for some decision.

accepted/0000-hub_retail.md
aaannz: Ah, so by SUMA you meant just Java part. By SUMA I am always assuming the whole product including salt.

accepted/0000-hub_retail.md
moio: > Talking about non-hub usage it would need to read formula pillars and apply states (formula states and pxe entry)

OK, so IIUC in this scenario using `salt-call --local` could be sufficient. We would "only" need to allow formulas with forms to be applied to the Server itself optionally, correct?

> In hub environment, there is only issue about pxe entry state as the configuration of formulas can be done from hub.

IIUC that problem would be solved via https://github.com/uyuni-project/uyuni-rfc/pull/46, or is there something else?

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,310 @@
...
+Alternative 2: Multi-master setup of Combined Peripheral and Branch server
+
+- Salt minion running on Combined Peripheral and Branch server has 2 masters - Hub and itself
+- the Branch Server configuration via formulas can stay unchanged
+
+problems:
+- SUMA registered to itself cause many problems
+  - we already have clients that do not allow certain actions (containers) so this should be fixable
+  - if we allow and test this scenario for Retail, we will know how it behaves for
+  non-Retail use and the most severe problems will be fixed (instead of just saying "we don't know")
```
moio, line 241: A few problems are known, and I can explain them.

Most important is the Server package upgrade problem. Registering a minion against a Server implies deactivating all of the normal, plain http zypper repos that are configured and replacing them with the Server's assigned channels. Those are served dynamically via Java code.

Problem is: once a Server is registered to itself, how to upgrade it? In order to upgrade Server packages you have to take Server services down first, and at that point your repos will not be available any more. One option would be to keep plain http repos around or the Server won't be upgradable. This is hairy but could be fixed in principle.

If you do that, you will bump into the next problem: the availability of updates on Server would not be displayed correctly. That follows because it's only possible if the repos are really channels, and not plain http repos. That's why we generally disable those upon registration/highstate, and I do not have a good solution in mind for now.

At leas when talking about the Salt stack we are willing to improve it, but we get another problem considering the traditional stack - there we do not want to add any new code per policy. And we know that the traditional stack completely breaks down if a Server is registered onto itself. So we would also need to selectively disallow traditional and allow only Salt for self-registration only.

Then we get to fix subscription matching, as it would count one extra subscription for the minion it currently expects not to be there.

And then, of course, we will have to make sure that no future configuration modification to minions interferes with the master or Server code, which is unexplored territory at this point.

As you can see it's not trivial, and I would be hesitant in considering this use case important enough to change the design - unless it becomes terribly complicated or fragile to do otherwise of course.

accepted/0000-hub_retail.md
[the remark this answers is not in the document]
nadvornik: I am not sure if the "ISS only" mechanism fits the use case with Lab server
https://github.com/uyuni-project/uyuni-rfc/pull/45/files#diff-1d42c41712a386afc3dd36a0fd7ce312404a01ccd1a535b4e994cdfeb1a3927cR134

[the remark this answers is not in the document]
moio: @nadvornik I would expect that case to be something we can cover as well. Just as we move software channels by copying RPMs and their associated database metadata around without changing them, we should be able to move images and their associated metadata as well.

nadvornik:
I have significantly rewritten this RFC. Most of the discussion here is no longer relevant. I will start new PR.

PR #46, opened 2020-12-02
Retail PXE database

aaannz:
Based on https://github.com/SUSE/spacewalk/issues/7773

Render [available here](https://github.com/uyuni-project/uyuni-rfc/blob/pxe-database/accepted/0000-retail-pxe-database.md)


nadvornik:
I don't like introducing new source of pillar data - this is inconsistent with existing pillars and (as you wrote) it can have negative impact on performance.

Do we need any API/UI to maintain the data - list, delete, modify?

How you want to identify the branch server? Current salt-only implementation use addressing by pillar, I am not sure if this can be used from Java.

aaannz:
> Do we need any API/UI to maintain the data - list, delete, modify?
> 
Sure, that can be done. But the idea is that handler of `suse/manager/pxe_update` will update data based on the data sent in the event (partitions uuids, branch id..). Deletion is done automatically when server entry is deleted. Listing done in salt pillar.

> How you want to identify the branch server? Current salt-only implementation use addressing by pillar, I am not sure if this can be used from Java.
Event handler finds branch by looking for the proxy terminal is connected through. State maintaining pxe entries will get the branch id from the branch pillars.
But this is good remark as this will not work when terminal is connected directly to Uyuni/SUSE Manager. I will need to figure this option out.

aaannz:
I added notes about API and detecting PXE server id in https://github.com/uyuni-project/uyuni-rfc/pull/46/commits/5fb343455b27465e39508e418f39849137e4ddbe

With proposed API it may be even best generally not to expose these entries in salt pillars and only pass them directly when scheduling state application. WDYT?

aaannz:
I added flowchart about the detection of PXE server id and dealing with it, added TBD items to unresolved questions, extended alternatives also reworded and clarified some bits.

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+
+
+Keeping data hidden in database has a positive side by not cluttering pillar space with rarely used data.
+To allow user to manually trigger PXE entries regeneration it is needed to develop UI or XMLRPC API endpoint, proposed below.
+
+Using these data `pxe/terminal_entry` state would need to be adapted to look for pillar data in correct places and also to support generating single pxe entry or all pxe entries depending on the values passed to the state.
+
+## Generating PXE entries
+
+Update to routine is needed to support terminals connected directly to Uyuni server. Mainly calling `pxe/terminal_entry` state using salt runner or other means locally on Uyuni.
```
nadvornik, line 159: `pxe/terminal_entry` is a salt state, runner modules are different from salt states, AFAIK.

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+
+### Detecting pxeserver_id
+
+In scenarios where terminals are connected to the branch server proxy, `pxeserver_id` is currently assumed to be branch server proxy and can be looked up in database by event handler. However architectures where proxy server is not involved such detection will not work (for example Hub scenarios where peripheral server is used instead of proxy server).
+
+We propose to solve this by **adapting saltboot to send FQDN of `next-server`**, that means server which provided initrd and kernel during PXE boot. This can be obtained by parsing DHCP response for `next-server` field. DHCP response is stored and available in dracut environment where saltboot is operating. Mandatory requirement in this case would be that **reverse lookup of IP address must be correctly configured** as `next-server` fields is IP address.
+
+Alternative is to use DNS service to lookup `tftp` CNAME, this is discussed in alternatives below.
+
+When the FQDN of `next-server` is not in the Uyuni database, such is the case when terminal is connected directly to the Uyuni, then `pxeserver_id` would be set to `null`.
```
nadvornik, line 109: Terminal connected directly to Uyuni can be dectected more reliably - Uyuni server should know it's own FQDN.

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+
+- Detection of `pxeserver_id` can also be done similarly as `salt` CNAME is being resolved by saltboot. That means adding `tftp` CNAME as hard requirement for DNS resolution. It is already documented to do so, however it wasn't used for anything and given number of users use their DNS servers instead of retail provided one this can introduce a regression in users scenarios.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Is presented detection of `pxeserver_id` acceptable?
+
+- `pxeserver_id` when terminal is not connected to the proxy, but directly to the Uyuni/SUSE Manager is `null` and has a functional meaning. However the same state can be achieved by removing branch server server entry. How big problem is it?
```
nadvornik, line 255: Detection from DHCP next-server is a good idea.

As I wrote above, a terminal connected directly can be determined more reliably, it does not have to be just a fallback.
Also in some alternatives in Retail Hub RFC this is not needed because Branch server will be allways connected as minion.

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+
+- UI for pxe generation. Is it needed, or is XMLRPC API enough?
+
+- Do we need to export pillar data? If so, what mechanism to use to export pillar data?
```
nadvornik, line 259: I think it does not have to be in pillar. I am against introducing new pillar source.
It can be added when https://github.com/SUSE/spacewalk/issues/10679 is solved.

accepted/0000-retail-pxe-database.md
aaannz: True, but the database entry is a reference to server id and AFAIK that would be populated only if uyuni is registered to itself. So it can be detected that fqdn is uyuni or something else but will result it same `pxeserver_id` entry.

accepted/0000-retail-pxe-database.md
aaannz: I agree. With the API and passing these data directly to applying state there is really no point in having these data in minion pillars.

accepted/0000-retail-pxe-database.md
aaannz: That is true. Idea is to use [runners.salt.cmd](https://docs.saltstack.com/en/latest/ref/runners/all/salt.runners.salt.html#salt.runners.salt.cmd) to run [modules.state.single](https://docs.saltstack.com/en/latest/ref/modules/all/salt.modules.state.html#salt.modules.state.single) to apply `pxe/terminal_entry` on the master. My goal is to have the same states for both master and minion.

Another option would be to split `pxe/terminal_entry` to very simple state and execution module and then calling execution module using the same runners.salt.cmd on master.

accepted/0000-retail-pxe-database.md
moio: Where would we save such an FQDN, assuming `pxeserver_id` is `NULL`?

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+
+Alternative is to use DNS service to lookup `tftp` CNAME, this is discussed in alternatives below.
+
+When the FQDN of `next-server` is not in the Uyuni database, such is the case when terminal is connected directly to the Uyuni, then `pxeserver_id` would be set to `null`.
+
+This is related to the [Hub Retail extension](https://github.com/uyuni-project/uyuni-rfc/pull/45) discussion about how should Retail architecture look like in Hub environment.
+
+## Storing data in database
+
+As data are provided within salt event, implementing new event handler `SaltbootPxeUpdateEvent` with accompanying `SaltbootPxeUpdateMessageAction` should be sufficient.
```
moio, line 115: Implementation detail: remember to add such an event to the whitelist in:

https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/modules/engines/mgr_events.py#L130

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+                (SELECT * FROM suseRetailMinionInfo as RMI, rhnServer as S WHERE RMI.pxeserver_id = S.id AND S.name LIKE %s) AS I JOIN rhnServer AS S ON I.server_id = S.id"
+         as_list: False
+         depth: 1
+```
+
+or by writing pillar files for given PXE server using already existing Java methods provided by `MinionPillarFileManager`.
+
+
+Keeping data hidden in database has a positive side by not cluttering pillar space with rarely used data.
+To allow user to manually trigger PXE entries regeneration it is needed to develop UI or XMLRPC API endpoint, proposed below.
```
moio, line 153: Not sure I got it all right up to this point - I'll try to rephrase and you tell me if I'm wrong :wink: 

Currently, relevant data is contained in the `suse/manager/pxe_update` event and it is saved to appropriate plain files (`01-mac_address_dash_separated`, `01:mac_address_colon_separated.grub2.cfg`)  via a Salt state that reacts to it.

What we want to change the role of those plain files - they will still be needed but they will not be the canonical source of information (originally in the event) but just a copy, a "cache" of it. We want the event to processed by Java code and the event handler to save event data onto the DB, so that it becomes the canonical source. So far so good.

IIUC, those plain files will still be needed, and should be regegerated starting from data in the DB. Which opens the problem of how to transfer such data from Server to Branch/Proxy.

We have two choices: pillars or XMLRPC API.

Is all the above correct to this point?

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+In current implementation, consumer of the `pxe_update` event data is `pxe/terminal_entry` state which is part of PXE formula.
+
+In case of keeping the final action in salt space, there is a need to provide pillar data to the state:
+
+  * by supplying directly to state when scheduling pxe generating action.
+  * by exposing database data to salt pillar system.
+
+Exposing PXE data is not hard requirement but may have some benefit like enabling user to manually apply `pxe/terminal_entry` or similar state to (re)generate pxe entries. This option can save resources by making UI/XML-RPC API changes non-essential.
+
+Database can be exposed to pillar system by external pillar mechanism connecting to the database using postgresql external pillar:
```
moio, line 130: I think we will need a solution to avoid file pillars at some point, but should be addressed in a separate PR.

I do not see a compelling reason to introduce a postgres external pillar only for the use case under discussion - rather, all generated pillars should be moved off the filesystem and onto the database in one shot, as part of an ad-hoc effort.

So for now we are only left with either generating file-pased pillars or exposing an XMLRPC endpoint.

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+
+Why should we **not** do this?
+
+  * obscure corner cases
+  
+    Currently unknown
+  
+  * will it impact performance?
+  
+    Multiple terminals booting at the same time - current state is handled solely by salt, proposed solution adds communication with database - increasing latency and some load on database
```
moio, line 221: How many is multiple worst case?

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+- Automatic detection of partition where salt configuration is stored
+
+  Draft PR [is here](https://github.com/uyuni-project/retail/pull/71). This removes the need to reregister terminal in case of lost PXE entry. However this does not preserve terminal-specific kernel command line options and always boot default initrd and kernel.
+
+- TPM as storage for salt configuration
+
+  Trusted Platform Module is required to be available on every hardware certified to run MS Windows 10, essentially every new hardware. It can be used as a secure storage for salt keys and minion id instead of storing it on hard drive. This would solve the same problem as automatic detection of salt configuration with the same drawbacks. This would also need some more time for research and packaging ensuring tpm2-tss is available in saltboot initrd.
```
moio, line 241: Given information we have about current users, is this something we can just assume, or is there lots of older hardware we will still be asked to support?

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+
+- Storage of pillar data in pillar files instead of database
+
+  This would allow to reuse of current external pillar mechanism, but suffer from the same issues like existing implementation - maintaining consistency of data and others. (See [overall issue](https://github.com/SUSE/spacewalk/issues/10679))
```
moio, line 245: Would it work if we accepted pillar files now, and refactor all pillars to come from the DB later? How good/bad would such an end result look like?

accepted/0000-retail-pxe-database.md
aaannz: I didn't do any measurement. I am just stating the obvious that java event handler and database communication enters the process of generating  pxe entries.

In case of salt-based system we are limited by salt resources (worker_threads), in this case we are also limited by (IIUC) tomcat and salt.

accepted/0000-retail-pxe-database.md
aaannz: I have no information about this and would need some form of survey among our users. But I think this would need to be done in addition to some fallback.

accepted/0000-retail-pxe-database.md
aaannz: Yes.

I also think that we don't need pillars files nor postgres pillar at the end. If XMLRPC API allows user to trigger pxe entries generation, then required data can be passed to the called states directly.

accepted/0000-retail-pxe-database.md
moio: > I am assuming FQDN of Uyuni server is generally known.

That assumption is correct, although the implementation is not ideal (we have a fragility problem there)

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+
+`kernel_parameters` are taken from HWTYPE group given terminal is member of or from defaults passed on kernel command line. It is sent in event, because pxe server, where PXE entries are to be generated, does not have access to pillars from saltboot formula of HWTYPE group.
+
+`pxeserver_id` column references PXE server (usually branch server) which provides PXE/UEFI booting data (including UEFI HTTP boot if configured).
+
+Column `mac_address` specify MAC address of the terminal. Similar column is also present in `rhnServerNetInterface` table, however that information is populated only after hardware refresh is scheduled, which is done only after successful deployment. Whereas pxe entries needs to be available before full deployment because of cases when kernel version of inird is different to kernel version in image and terminal requires reboot to correct kernel version.
+
+Another option is that event handler creates the correct entry in `rhnServerNetInterface` table.
```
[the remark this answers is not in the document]
moio: Good, please add that note to the text.

accepted/0000-retail-pxe-database.md
moio: Then one problem I see with the XMLRPC route is authentication and authorization. What credentials should the Branch Server use to get the XMLRPC session key?

Additional question: would such XMLRPC endpoints make sense outside of this scope, for any other general use?

aaannz: There is some misunderstanding. I am not proposing to use XMLRPL API to transfer data from server to branch. I still expect salt state to take care of generating the pxe entry. XMLRPC API is there for users to be able to regenerate pxe entries if needed.

The data are being passed using pillars. But not by writing them to the pillar file and exposing, but by passing them to the state application directly (e.g. like calling `salt $branch state.apply pxe/terminal_entry boot_image='...' root='...' ...`. We can do this by using proper LocalCall from salt netapi.)

accepted/0000-retail-pxe-database.md
moio: Please then move this section to Alternatives, reporting the reason why it's not tackled this way for now.

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,259 @@
...
+
+Complete routine should look like:
+
+![Generating PXE entries flowchart](images/0000-retail-pxe-db-workflow.png)
+
+## Updating PXE entries
+
+Moving terminal between pxe servers happen. This means event handler needs to be aware of this possibility and adapt `pxeserver_id` field as needed.
+  
+Removal of the PXE server (branch server) system profile would set `pxeserver_id` to `null`. To update the entry terminal will need to be rebooted under new PXE server or user would need to manually update using API call.
```
moio, line 169: I feel we will have to give a `NULL` `pxeserver_id` a precise meaning - I can see we run into the situation where it's not immediately clear from logs whether a terminal is served by an unknown PXE server from one which is served by the Server itself.

What benefits do you see in using a system ID rather than, say, the fully-qualified domain name as a string?

accepted/0000-retail-pxe-database.md
moio: > The data are being passed using pillars. But not by writing them to the pillar file and exposing, but by passing them to the state application directly

Thanks for clarifying, I indeed had a confused understanding.

In the hypothesis pillar data was saved in the database instead of files, would it be an option to just store as a pillar, thus avoiding the need of a new database table and set of APIs?

In other words, how would this RFC change if we had solved the problem of pillar storage already?

aaannz: If data are just in pillar files and not in database:
- need to add/adapt mechanism to maintain sync between system entries and pillars with pxe entries
- no need for xmlrpc api, only documentation how to manually trigger salt state or edit them
- no need for database

We would still need:
- java based event handler to write pillar files

The biggest drawback is the need to maintain coherence between system entries and pillar files which may not be completely straightforward as the pillar files are for branch servers, however tracing terminal system entries.

accepted/0000-retail-pxe-database.md
aaannz: If we use FQDN passed from terminal, we will sooner or later need to determine if it is known system, uyuni server or something else anyway. Because we want to know if we should schedule salt state, salt-call or nothing or something else. That boils to question what we want to do in case of not managed pxe server?

I see it that upon receiving the event we do server lookup based on FQDN received in the event. If we find it - assign `pxeserver_id`, if not we will check if it is uyuni server FQDN and set `NULL`. If it is not even uyuni server FQDN, report warning in the log and set `NULL`

Then when `pxeserver_id` is `NULL` we will always generate pxe entries on Uyuni server. This may be documented and users can use this even in case when using non managed pxe server and copy the files over there.

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,257 @@
...
+
+Later in the text I introduced term `pxe server`. This is usually branch server proxy, however design should be flexible enough to allow any salt client to become pxe server, particularly in branchless scenarios with e.g. peripheral hub servers instead of branch server proxy.
+
+# Detailed design
+[design]: #detailed-design
+
+
+## Workflow description
+
+On each boot (first boot and all next boots) terminal sends an event `suse/manager/pxe_update` together with following data:
```
moio, line 33: How is this data computed? Some time ago we added the possibility of adding arbitrary grains to the minion start event via:

https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/bootstrap/susemanager.conf#L11

If all this data can be rendered as (custom) grains, then one possibility could be not to use an event at all and just piggyback required data to the minion start event. That would save a round-trip to the client per boot.

accepted/0000-retail-pxe-database.md
```diff
@@ -0,0 +1,257 @@
...
+}
+```
+
+The data send in the `pxe_update` event are not stored anywhere on the Uyuni/SUSE Manager and upon receiving the event, salt reactor applies `pxe/terminal_entry` state on the branch server which will create the PXE entry for given terminal.
+
+
+To be able to recreate PXE entries at later time, it is clear that the data send in the `pxe_update` event should be stored somewhere on the Uyuni/SUSE Manager and be accessible to branch servers.
+
+
+## Storage of the data
```
moio, line 65: Alternative idea:
 - we provide a mechanism to save arbitrary pillar data (eg. an arbitrary JSON blob) into the database, per minion. This comes with Java methods to read and write such data
 - we provide a way (likely implemented with some new API) to consume such DB-backed pillars (via a new [external pillar](https://docs.saltstack.com/en/latest/topics/development/modules/external_pillars.html) module)
 - code is added to the Java app to react to `suse/manager/pxe_update`. This code will save data in the event as pillar data into the database
 - a new state is written that reads such pillar data from the external pillar (thus the DB) and writes PXE/GRUB entries. This can be triggered on highstate, by reactor on `suse/manager/pxe_update` or manually via normal means

How does that sound?

accepted/0000-retail-pxe-database.md
aaannz: Startup grains are already used to pass `saltboot_initrd` grain to indicate we are in saltboot initrd mode. However it cannot be used for all pxe data, because terminal will know what boot image to use only after obtaining saltboot and image pillars.

accepted/0000-retail-pxe-database.md
aaannz: This sounds good 👍🏾

aaannz: I'm preparing JSON blob storage as a separate RFC draft and there are some pain points wrt pxe entries.

Pxe entries pillars are meant to be available to branch proxy.
What exactly did you mean by `have arbitrary pillar data into the database, per minion`?
* at most one entry per minion? In this case we would need to do JSON preprocessing to concatenate/remove entries as needed when modifying it.
* there can be multiple blobs per minion? Removing/modifying entries would be easier, merging would be done in external pillar/salt.

Next these pxe entries are closely linked with the terminal minions, however they are not meant for them. So if terminal minion is removed, its pxe entry should be removed as well (not a hard requirement, but under some circumstances can lead to error during deployment).
To support this scenario we can either:
- have an owner of the JSON blob for whom the pillars are meant and then a list of systems on which it depends. I don't know if there is any other use case for this architecture and this would require to have multiple blobs per minion
- adapt system entry removal procedure to check if it is a retail minion (e.g. member of TERMINAL groups), find its associated proxy and modify branch proxy pxe entries JSON blob.

moio: > I'm preparing JSON blob storage as a separate RFC draft and there are some pain points wrt pxe entries.
> 
> Pxe entries pillars are meant to be available to branch proxy.
> What exactly did you mean by `have arbitrary pillar data into the database, per minion`?

Just to make sure we are on the same page: in Salt, [a pillar](https://docs.saltstack.com/en/latest/topics/tutorials/pillar.html#pillar-walkthrough) is a "JSON-like" arbitrary data structure _relative to a certain minion ID_. "JSON-like" here means it can basically be strings, integers, booleans, none plus lists and dictionaries (however nested). Effectively, any pillar can be serialized into JSON, and it's typically represented in YAML (which is a superset of JSON).

An "external pillar" is a source of pillar data that comes from a third party. Could be a database, LDAP, Amazon S3, or anything really. All one needs to do in order to implement such an "external pillar" is to define [a function that returns the data given a minion ID](https://docs.saltstack.com/en/latest/topics/development/modules/external_pillars.html#ext-pillar) (and a few other secondary implementation details).

So the idea above is really to provide an Uyuni-backed external pillar, where pillar data is stored inside of the database (for each minion).

>     * at most one entry per minion? In this case we would need to do JSON preprocessing to concatenate/remove entries as needed when modifying it.
>     * there can be multiple blobs per minion? Removing/modifying entries would be easier, merging would be done in external pillar/salt.

The easiest implementation I can think of is to have the JSON blob in the database mapped to Java objects, so that any manipulation happens in Java code where it's easiest to write/debug/test. Basically, Java would load the blob as-is, alter it, and then re-save it to the DB as a whole. In that case, the first option seems straightforward.

Of course that could bring to performance problems, in case we need to manipulate pillars of very many minions, or in case blobs become very big. In that case, maybe we could do the manipulation [directly in SQL](https://www.postgresql.org/docs/11/functions-json.html), such as [removing a key](https://stackoverflow.com/questions/23490965/postgresql-remove-attribute-from-json-column) or setting one, even deep in the structure. This could be done as an incremental improvement to the easiest implementation above, even at a later point in time.

So I do not see big advantages with a 1 minion to N blobs approach - but maybe I'm missing something.

> Next these pxe entries are closely linked with the terminal minions, however they are not meant for them. So if terminal minion is removed, its pxe entry should be removed as well (not a hard requirement, but under some circumstances can lead to error during deployment).

How is the cleanup performed in the current implementation?

> To support this scenario we can either:
> 
>     * have an owner of the JSON blob for whom the pillars are meant and then a list of systems on which it depends. I don't know if there is any other use case for this architecture and this would require to have multiple blobs per minion

I cannot presently think of other possible cases, so I would be more inclined to call this a Retail-specific mechanism until we find a second user. Only then I would market it as general.

>     * adapt system entry removal procedure to check if it is a retail minion (e.g. member of TERMINAL groups), find its associated proxy and modify branch proxy pxe entries JSON blob.

Idea: when a minion is deleted, its minion key must be deleted, and Salt sends out an event in that case. Would it be an option to react to such event to do the cleanup, just as we react to `suse/manager/pxe_update` to start this process? (as event reaction is idiomatic in Salt, I guess this would be the idiomatic way to resolve the problem in the Salt world)

Yet another alternative: how about a Taskomatic cleanup job that checks all the Retail Proxies and cleans up any pillars belonging to clients that are not there any more? (As garbage collection in the database via Taskomatic is idiomatic in Spacewalk, I guess this would be the idiomatic way to resolve the problem in the Spacewalk world)

aaannz: > Just to make sure we are on the same page: in Salt, [a pillar](https://docs.saltstack.com/en/latest/topics/tutorials/pillar.html#pillar-walkthrough) is a "JSON-like" arbitrary data structure _relative to a certain minion ID_. "JSON-like" here means it can basically be strings, integers, booleans, none plus lists and dictionaries (however nested). Effectively, any pillar can be serialized into JSON, and it's typically represented in YAML (which is a superset of JSON).
> 
> An "external pillar" is a source of pillar data that comes from a third party. Could be a database, LDAP, Amazon S3, or anything really. All one needs to do in order to implement such an "external pillar" is to define [a function that returns the data given a minion ID](https://docs.saltstack.com/en/latest/topics/development/modules/external_pillars.html#ext-pillar) (and a few other secondary implementation details).

That is how I understood it.

> > ```
> > * at most one entry per minion? In this case we would need to do JSON preprocessing to concatenate/remove entries as needed when modifying it.
> > * there can be multiple blobs per minion? Removing/modifying entries would be easier, merging would be done in external pillar/salt.
> > ```
> 
> The easiest implementation I can think of is to have the JSON blob in the database mapped to Java objects, so that any manipulation happens in Java code where it's easiest to write/debug/test. Basically, Java would load the blob as-is, alter it, and then re-save it to the DB as a whole. In that case, the first option seems straightforward.
> 
> Of course that could bring to performance problems, in case we need to manipulate pillars of very many minions, or in case blobs become very big. In that case, maybe we could do the manipulation [directly in SQL](https://www.postgresql.org/docs/11/functions-json.html), such as [removing a key](https://stackoverflow.com/questions/23490965/postgresql-remove-attribute-from-json-column) or setting one, even deep in the structure. This could be done as an incremental improvement to the easiest implementation above, even at a later point in time.

👍🏾 
 
> > Next these pxe entries are closely linked with the terminal minions, however they are not meant for them. So if terminal minion is removed, its pxe entry should be removed as well (not a hard requirement, but under some circumstances can lead to error during deployment).
> 
> How is the cleanup performed in the current implementation?

It is not done curently, something I would like to fix with this development as well.

> Idea: when a minion is deleted, its minion key must be deleted, and Salt sends out an event in that case. Would it be an option to react to such event to do the cleanup, just as we react to `suse/manager/pxe_update` to start this process? (as event reaction is idiomatic in Salt, I guess this would be the idiomatic way to resolve the problem in the Salt world)
> 
> Yet another alternative: how about a Taskomatic cleanup job that checks all the Retail Proxies and cleans up any pillars belonging to clients that are not there any more? (As garbage collection in the database via Taskomatic is idiomatic in Spacewalk, I guess this would be the idiomatic way to resolve the problem in the Spacewalk world)

I wasn't aware salt sends out an event on minion key deletion. That sounds interesting and good way to trigger this deletion on branch proxy when minion is deleted.

aaannz:
Will not happen this way. Closing

PR #48, opened 2020-12-16
Live patching enhancements (PM-2282)

cbbayburt:
Rendered documents for review:
- [Live Patching with CLM](https://github.com/cbbayburt/uyuni-rfc/blob/live-patching-eol/accepted/00075-live-patching-with-clm.md)
- [Live Patching EOL Date Information](https://github.com/cbbayburt/uyuni-rfc/blob/live-patching-eol/accepted/00076-live-patching-eol-data.md)


accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+ - Unpacking must be done during reposync, adding complextion
+
+## 2. Live patching CLM workflow
+
+SLES channels can be prepared for live patching applications using the Content Lifecycle Management feature in Uyuni. This is done by adding a set of filters that exclude the regular kernel package updates and include live patch versions of those. However, the UI doesn't provide any special workflow for live patching and this process must be done manually by the user.
+
+Below are some examples of such sets of filters:
+
+Live patching project with a single filter to exclude regular kernel packages newer than a specific version:
+ - `DENY erratum contains package with epoch/version/release greater than kernel-default <EVR> (package_nevr)`
```
cbbayburt, line 92: @dvosburg is this a valid alternative to your project filters?

mcalmer: This is what we use on manager.mgr.suse.de for our hardware server. The only challenge is, to get the correct data to create the filter. This is where a UI could help.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+
+Live patching project using `ALLOW` overrides:
```
mcalmer, line 94:
```suggestion
Alternative way for live patching project filters using `ALLOW` overrides:
```
Call it "alternative" to make it clear that this is not required together with the filter above.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+ - `DENY errata containing "update for the Linux Kernel" in synopsis`
```
mcalmer, line 95: I think this was a DENY of all patches which suggest a reboot.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+ 2. Select a SLES client from a combobox
+ 3. Select a kernel version from a dropdown (current kernel version of the system is preselected)
+
+The resulting filters will set up the project for live patching from the specified kernel version.
+If the wizard is accessed from the project view, all the created filters will be automatically added to the project.
+
+#### Live patching for a product
+
+ 1. Input a label prefix for the resulting filters. If the wizard is accessed from the project view, prefill this field with the project label
+ 2. Select a synced SLES product from a dropdown
```
mcalmer, line 125: I think `2.` is not needed. When you select a Project, you have already channels in that project which defines which kernel versions are available. So we can just show what is available.

Or is the idea to add also channels? But than we talk about a CLM project wizard and not about a Filter Wizard.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+ - `DENY erratum contains package with epoch/version/release greater than kernel-default <EVR> (package_nevr)`
+
+Live patching project using `ALLOW` overrides:
+ - `DENY errata containing "update for the Linux Kernel" in synopsis`
+ - `ALLOW errata containing "Live Patch" in synopsis` *- an override to the first rule*
+ - `ALLOW a specific patch containing current kernel version (via advisory name)` *- an override to the first rule*
+
+### CLM project wizards
+
+The process of live patching can be made more user-friendly using "CLM project wizards". The wizards are alternatives to the regular filter creation dialogs that provide a tailored interaction to create multiple filters that serve a single purpose. These dialogs can be accessed via the "Use wizard" button in the filter creation dialog. This concept can be extended to other use cases and workflows in the future.
```
mcalmer, line 101: This is a bit unclear. When we talk about a "Project Wizard" it might create the whole project incl. Channels, Environments, Filters, etc. . 
But when the wizard should only create filters, we should name it "Filter Wizard".

This become important below.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+If the wizard is accessed from the project view, all the created filters will be automatically added to the project.
+
+#### Additional enhancements
+
+##### CLM filter list
+
+CLM project wizards is aimed to make the project creation easier by automatically adding multiple filters. As a consequence, the list of filters in Uyuni can grow very quickly. To handle this problem, the following improvements can be made in the CLM filter list page:
+
+ - Add selection checkboxes to delete multiple filters at once
+ - Add "Delete unused/Select unused" shortcuts to delete or select all the filters that are not currently attached to any project
```
mcalmer, line 138: - Filter by "Projects in use"
- Sort table by "Project in use"

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+[alternatives]: #alternatives
+
+Alternative implementations for EOL data retrieval are mentioned in the [data retrieval](#data-retrieval) section.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+ - Which approach shall be used for [EOL data retrieval](#data-retrieval)
+ - What is the roadmap for `zypper lifecycle` and the lifecycle data format
+ - What is the best way to define a live patching project with the existing filters (see [Live patching CLM workflow](#live-patching-clm-workflow))
```
mcalmer, line 165: - Do we need a "Project Wizard" or a "Filter Wizard" ?

accepted/00075-live-patching-enhancements.md
moio: What would be pros and cons of the two alternative approaches? Is now (RFC time) the right time to decide which path to take?

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+
+This section evaluates three alternative implementations for this feature.
+
+#### Call 'zypper lifecycle' on client
+
+Uyuni calls `zypper lifecycle` on the client and collects the data via Salt during package profile update. The data is then stored in the database per client along with the kernel live version data.
+
+**Pros:**
+ - `zypper lifecycle` is publicly supported
+ - Can be attached to profile update data with an existing process
```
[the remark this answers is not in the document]
cbbayburt: > Does that imply that correct channel assignment should be there for the Server to be able to know correct EOL data? What would happen if critical channels are missing for example?

`zypper lifecycle` seems to work independently from the assigned channels, so as long as the live patching product (the release package) is installed, it will report correctly.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+
+**Cons:**
+ - `zypper lifecycle` currently provides no machine-readable format
```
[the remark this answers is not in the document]
cbbayburt: Thanks, I was not aware of this. `--save` seems to work only if no package name is given, though.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+ - Extra packages need to be installed on the client:
+   - `zypper-lifecycle-plugin`
+   - `lifecycle-data-sle-module-live-patching`
+
+#### Read the lifecycle data file on client
+
+Uyuni runs a Salt state to read the lifecycle data file provided with the package `lifecycle-data-sle-module-live-patching` on the client and collects the data during package profile update. The data is then stored in the database per client along with the kernel live version data.
+
+**Pros:**
+ - Easy to implement
```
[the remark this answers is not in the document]
cbbayburt: Both approaches use Salt mechanisms to get the data from the client. The difference is;
1. `zypper lifecycle` way: Salt runs `zypper lifecycle` with `--save` arg on client, then we read and parse the output XML
2. CSV file way: We just read and parse the CSV

> I think we would need to re-implement the whole logic of the lifecycle plugin. This means reading all data files in that directory and check for a specific package name.

I guess we will end up with this in the end. We might even consider extending this to EOL reports for all products.

My actual issue with the first two approaches is that the EOL data is irrelevant to any client in any way. It just feels more natural to handle this all at the server-side.

accepted/00075-live-patching-enhancements.md
cbbayburt: These are just meant to be 2 different examples to solve the problem with the current CLM manually.

> Is now (RFC time) the right time to decide which path to take?

I believe not. It's an implementation detail and can be shaped by testing different combinations.

accepted/00075-live-patching-enhancements.md
cbbayburt: IIRC, Don demoed two different setups, one like this, and one using the newer reboot filter.

accepted/00075-live-patching-enhancements.md
cbbayburt: I agree. "Filter wizard" sounds more suitable for what I had in mind. I'll change it accordingly.

accepted/00075-live-patching-enhancements.md
cbbayburt: Similar to what we experienced with AppStream filters, the order in which the user adds different entities is not enforced, so the user may start with adding filters before channels or anything else. Moreover, the design allows the wizard to be accessed from the filters page independent from any project.

accepted/00075-live-patching-enhancements.md
cbbayburt: Filter wizard :mage_man: :+1:

cbbayburt:
@moio @mcalmer thank you for the review. I've tried to answer all the questions and updated the text per your feedback.

One important thing I initially missed is that the lifecycle data packages are required by the related release packages so we would expect it to be already installed on clients.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,171 @@
...
+Uyuni calls `zypper lifecycle` on the client and collects the data via Salt during package profile update. The data is then stored in the database per client along with the kernel live version data.
+
+**Pros:**
+ - `zypper lifecycle` is publicly supported
+ - Can be called directly for a package name (a specific live-patch package)
+ - Can be attached to profile update data with an existing process
+
+**Cons:**
+ - Is a 2-step process that requires `zypper lifecycle` to output to a file
+ - `zypper-lifecycle-plugin` needs to be installed on the client
```
moio, line 57:
```suggestion
 - `zypper-lifecycle-plugin` needs to be installed on the client
 - it requires a round-trip to the client, per client
```

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,170 @@
...
+The lifecycle information is provided in the `lifecycle-data-sle-module-live-patching` package in the Live Patching product. Uyuni unpacks this RPM and reads the data from the extracted CSV file during reposync of this product. The data is stored in the database per live patch package.
+
+**Pros:**
+ - Independent, no impositions on clients
+ - Gathered data is always available and can be used for UI enhancements such as displaying EOL info for all live patch packages displayed in the UI
+
+**Cons:**
+ - RPM format can change in the future
+ - Uyuni doesn't have any existing functionality that can be utilized
+ - Unpacking must be done during reposync, adding complextion
```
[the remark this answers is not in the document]
moio:
```suggestion
 - Unpacking must be done during reposync, adding complication
```

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,171 @@
...
+ - Is a 2-step process that requires `zypper lifecycle` to output to a file
+ - `zypper-lifecycle-plugin` needs to be installed on the client
+
+#### Read the lifecycle data file on client
+
+Uyuni runs a Salt state to read the lifecycle data file provided with the package `lifecycle-data-sle-module-live-patching` on the client and collects the data during package profile update. The data is then stored in the database per client along with the kernel live version data.
+
+**Pros:**
+ - Easy to implement
+ - Can be attached to profile update data with an existing process
```
moio, line 65:
```suggestion
 - it requires a round-trip to the client, per client
```

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,171 @@
...
+
+**Cons:**
+ - File name or format can change in the future
+
+#### Unpack and read the lifecycle data RPM on Uyuni server
+
+The lifecycle information is provided in the `lifecycle-data-sle-module-live-patching` package in the Live Patching product. Uyuni unpacks this RPM and reads the data from the extracted CSV file during reposync of this product. The data is stored in the database per live patch package.
```
moio, line 72: How complex is such a CSV, can you post an example as snippet?

moio (commented):
@cbbayburt I left a couple more questions.

The more I think about it, the more I think I like the idea of processing CSV data Server-side: it might be a more complicated piece of software, but at least it's not a distributed piece of software!

accepted/00075-live-patching-enhancements.md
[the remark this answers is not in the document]
mcalmer: > `zypper lifecycle` seems to work independently from the assigned channels, so as long as the live patching product (the release package) is installed, it will report correctly.

Depends. You need to install the package to get the data. After the package is installed you can also run this without the channel assigned. The problem is, that you will not get package updates and therefore no update of the lifecylce data when the channel is not assigned.
As sometimes packages will be limited later as it depends on upstream supportability you may think that you have support, but you just have no up to date data.
For Livepatching Lifecycle data this should not happen as long as you update the lifecycle package with the kernel lifepatch you should be fine. And AFAIK both comes from the same channel.

accepted/00075-live-patching-enhancements.md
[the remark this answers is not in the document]
mcalmer: I think this is by design. You get a "Report" of all data. The tool can parse it and when it knows the package name it can easily find the searched data.
Especially when you need to lookup multiple packages this is very usefull as you need to generate the report only 1 time.

accepted/00075-live-patching-enhancements.md
mcalmer: I think we should decide during the RFC process. AFAIK we documented option 1 in our manuals, while option 2 was choosen by @dvosburg . If we implement one option, but sales do not like it we do not win anything.

accepted/00075-live-patching-enhancements.md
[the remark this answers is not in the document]
mcalmer: I think the main difference between XML and CSV is, that CSV data are split into multiple files and we need to read all or we need to know which one we need. The XML file is a report of all and we just need to parse this one file.

accepted/00075-live-patching-enhancements.md
mcalmer:
```
cokerunner:~ $> cat /usr/share/lifecycle/data/sle-module-live-patching.lifecycle
# Last updated: 2020-09-22T13:14:02+00:00
# format: name,version,date
# version can contain wildcard character *
kernel-livepatch-4_12_14-23-default,*,2019-06-16
kernel-livepatch-4_12_14-25_13-default,*,2019-08-16
kernel-livepatch-4_12_14-25_16-default,*,2019-08-28
kernel-livepatch-4_12_14-25_19-default,*,2019-10-02
kernel-livepatch-4_12_14-25_22-default,*,2019-10-16
kernel-livepatch-4_12_14-25_25-default,*,2019-10-31
kernel-livepatch-4_12_14-25_3-default,*,2019-07-27
kernel-livepatch-4_12_14-25_6-default,*,2019-08-06
kernel-livepatch-4_12_14-25_28-default,*,2020-02-01
kernel-livepatch-4_12_14-150_14-default,*,2020-03-28
kernel-livepatch-4_12_14-150_17-default,*,2020-05-14
kernel-livepatch-4_12_14-150_22-default,*,2020-06-17
kernel-livepatch-4_12_14-150_27-default,*,2020-07-15
kernel-livepatch-4_12_14-150_32-default,*,2020-08-06
kernel-livepatch-4_12_14-150_35-default,*,2020-09-20
kernel-livepatch-4_12_14-150_38-default,*,2020-10-17
kernel-livepatch-4_12_14-150_41-default,*,2020-11-12
kernel-livepatch-4_12_14-150_47-default,*,2020-12-23
kernel-livepatch-4_12_14-150_52-default,*,2021-06-18
kernel-livepatch-4_12_14-150_55-default,*,2021-08-03
kernel-livepatch-4_12_14-150_58-default,*,2021-09-11
kernel-livepatch-4_12_14-195-default,*,2020-06-24
kernel-livepatch-4_12_14-197_4-default,*,2020-06-24
kernel-livepatch-4_12_14-197_7-default,*,2020-07-04
kernel-livepatch-4_12_14-197_10-default,*,2020-07-15
kernel-livepatch-4_12_14-197_15-default,*,2020-08-06
kernel-livepatch-4_12_14-197_18-default,*,2020-09-20
kernel-livepatch-4_12_14-197_21-default,*,2020-10-18
kernel-livepatch-4_12_14-197_26-default,*,2020-11-12
kernel-livepatch-4_12_14-197_29-default,*,2020-12-17
kernel-livepatch-4_12_14-197_34-default,*,2021-03-02
kernel-livepatch-4_12_14-197_37-default,*,2021-03-31
kernel-livepatch-4_12_14-197_40-default,*,2021-04-29
kernel-livepatch-4_12_14-197_45-default,*,2021-06-10
kernel-livepatch-4_12_14-197_48-default,*,2021-08-03
kernel-livepatch-4_12_14-197_51-default,*,2021-08-07
kernel-livepatch-4_12_14-197_56-default,*,2021-09-09
kernel-livepatch-5_3_18-22-default,*,2021-07-21
kernel-livepatch-5_3_18-24_9-default,*,2021-08-03
kernel-livepatch-5_3_18-24_12-default,*,2021-09-03
kernel-livepatch-5_3_18-24_15-default,*,2021-09-09
```

accepted/00075-live-patching-enhancements.md
mcalmer: Btw: Example B is potentionally dangerous as we create 2 Filters which depend on the Patch Synopsis. This is a free text field and we cannot rely on Maintenance Team always stick to the same naming conventions. Especially as this synopsis is autogenerated from the buildservice and manually updated by them. When they forget it of do a typo our filter might not match anymore.

accepted/00075-live-patching-enhancements.md
moio: Seems pretty straightforward format wise, thanks

accepted/00075-live-patching-enhancements.md
moio: Did I get it correctly that with the new metadata we should be able to (optionally automatically) create filters that do not depend on relatively-fragile synopsis like the one described here?

mcalmer: No. The new metadata are for knowing when there will be no live-patches anymore for the installed kernel and it is time to switch to the new kernel.
These filters are for general use of life-patching feature. Stuff we already have.
But we want to add a new UI to make it easier to get the filters. But for this we need to know which filters we want to set. There are multiple ways to achive the wanted result. We described Example A in the manuals, but Don took a different approach (Example B).

What I want to say is: it does not make sense if we implement a UI for Example A but Customers like more Example B and do not use our new shiny UI. 
So we need to find out why people like Example B and if we can convince them to use Example A or modify it and make an Example C which please them all :-)

cbbayburt: I'll start an email thread to discuss all the different options with supporters

cbbayburt:
> The more I think about it, the more I think I like the idea of processing CSV data Server-side: it might be a more complicated piece of software, but at least it's not a distributed piece of software!

@moio I completely agree. It should be independent of any client by nature.

accepted/00075-live-patching-enhancements.md
cbbayburt: and here's one for SLES12:
```
cb-min2:~ # cat /var/lib/lifecycle/data/sle-live-patching.lifecycle
# Last updated: 2020-12-16T16:07:48+00:00
# format: name,version,date
# version can contain wildcard character *
kgraft-patch-3_12_32-33-default,*,2016-01-16
kgraft-patch-3_12_32-33-xen,*,2016-01-16
kgraft-patch-3_12_36-38-default,*,2016-02-10
kgraft-patch-3_12_36-38-xen,*,2016-02-10
kgraft-patch-3_12_38-44-default,*,2016-03-18
kgraft-patch-3_12_38-44-xen,*,2016-03-18
kgraft-patch-3_12_39-47-default,*,2016-04-02
kgraft-patch-3_12_39-47-xen,*,2016-04-02
kgraft-patch-3_12_43-52_6-default,*,2016-06-16
kgraft-patch-3_12_43-52_6-xen,*,2016-06-16
...
```

cbbayburt: For this option, a high-level algorithm can be:

(at reposync or as a separate task after sync)
```
for all RPMs that start with `lifecycle-data-*`:
  unpack RPM and loop through `*.lifecycle` files:
    read `name,version,date` in file
    if name in [`kgraft-patch-*`, `kernel-livepatch-*`, ...]:
       insert into DB table `susePackageLifecycleData`
    delete temp data
```

cbbayburt:
> The more I think about it, the more I think I like the idea of processing CSV data Server-side: it might be a more complicated piece of software, but at least it's not a distributed piece of software!

@moio assuming we go this way, when and where do you think it would be ideal? at reposync or as a separate task? in reposync code, in Java, or as a standalone module?

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,189 @@
...
+
+A high-level algorithm of the process is as follows:
+```
+# At reposync or as a separate task after sync
+
+for all RPMs that start with `lifecycle-data-*`:
```
mcalmer, line 80: This is bad, what happens when they decide to change the file name or for a specific product put it into a different package. We should not rely on the package name.

When we parse a package, we extract the filelist from the metadata and put every file in the database.
So we could simply query every file which is in a directory and find the corresponding package.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,189 @@
...
+  unpack RPM and loop through `*.lifecycle` files:
```
mcalmer, line 81: which one? We need to find the "latest" older packages will provide old data.

accepted/00075-live-patching-enhancements.md
cbbayburt: We could parse every package and look for `*.lifecycle` files everywhere, but I'm not sure if it will have some significant performance burden

accepted/00075-live-patching-enhancements.md
cbbayburt: Good point. Maybe on the first run we can just keep a list of packages including lifecycle data, and then pick the latest EVR of each such package.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,189 @@
...
+
+# Alternatives
+[alternatives]: #alternatives
+
+Alternative implementations for EOL data retrieval are mentioned in the [data retrieval](#data-retrieval) section.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+ - Which approach shall be used for [EOL data retrieval](#data-retrieval)?
```
paususe, line 182: Third-one seems to be more complex but IMHO it makes the most sense and will be the most performant.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,189 @@
...
+ - What is the roadmap for `zypper lifecycle` and the lifecycle data format?
+ - What is the best way to define a live patching project with the existing filters (see [Live patching CLM workflow](#live-patching-clm-workflow))?
+ - Do we plan to extend the EOL reporting to other products in the future?
```
paususe, line 185: What are the other products we may want to extend EOL reporting? Is there such information available for them in our channels?

paususe (commented):
Looks quite good to me, just a few questions.

moio:
> @moio assuming we go this way, when and where do you think it would be ideal? at reposync or as a separate task? in reposync code, in Java, or as a standalone module?

### reposync

Pros:
 - reposync is "the right place" architecturally. reposync is for loading rpm/repo data into the database, and that's just what we need here
 - it should be relatively straightforward to get RPM contents from Python, as there's libraries for that already (please double check)

Cons:
 - Python lacks very good tests and type checking (partially), so it is less robust of a base to build upon
 - reposync surrounding code especially so

### Taskomatic task

Pros:
 - Java and the particular ecosystem is more robust
 - unpacking RPMs in Java is possible

Cons:
 - more clunky. Users have to wait until reposync is done and the task is also done before UI/APIs return correct results
 - unpacking RPMs in Java is [possible](https://github.com/moio/pase/blob/94e8d006b932ef3c691c76d9962601a516bc3c38/src/main/java/com/suse/pase/walkers/ArchiveWalker.java#L59-L62) but [requires a little-known library](https://github.com/moio/pase/blob/94e8d006b932ef3c691c76d9962601a516bc3c38/pom.xml#L101-L105)

### Verdict

It's not easy to make a call here. I would evaluate the complexity: I'd go with Python only if we expect something really straightforward (100-200 lines or so). If this smells complex, bite the bullet and build there where it's safest.

@mcalmer do you have any feeling on how complex this can be in Pythonland?

accepted/00075-live-patching-enhancements.md
moio: @cbbayburt we already extract the whole filelist and save it into the DB, so just querying it after the fact will be fine.

mcalmer:
> @mcalmer do you have any feeling on how complex this can be in Pythonland?

In the end I would do the unpack in both cases using a script or call `unrpm` from the build package.

- create a tmp dir
- call unrpm in the tmp dir
- parse the file
- insert the data into the DB (do we have a schema idea for it?)
- remove the tmp dir

The only problem might be the `unrpm` is part of the `build` package which is in development tools module.
But as build is only a (very small) shell script we can just write our own.

```
input=$1
CPIO_OPTS="--extract --unconditional --preserve-modification-time --make-directories"
rpm2cpio $input | cpio ${CPIO_OPTS}
```

`rpm2cpio` is part of "rpm" and `cpio` is in "cpio". Both are in the Basesystem Module.

As Filesystem operations are easier in python than in Java we can do it there. If we want to use the DB for the info which packages to unpack we only need to take care that we commit before. 

But the main question we need to answer is, how to store the data?

When we call every system to return the data, we can store it directly at the system info and new clients would just add there info there.
But when we parse RPMs, we need a new structure and we need an algorithm similar to ServerNeededCache which calculate lifecycle Info for every System.

accepted/00075-live-patching-enhancements.md
```diff
@@ -0,0 +1,189 @@
...
+
+A wizard can be applied independently from a project. In that case, the filters will be created without being added to a specific project. These filters can be used in any project later on.
+
+In scope of this RFC, two different wizards are proposed. The process of creating the filters using these wizards are described below.
+
+#### Live patching for a system
+
+ 1. Input a label prefix for the resulting filters. If the wizard is accessed from the project view, prefill this field with the project label
+ 2. Select a SLES client from a combobox
+ 3. Select a kernel version from a dropdown (current kernel version of the system is preselected)
```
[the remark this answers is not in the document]
cbbayburt: The purpose of the system selection here is just to provide the user a way to say "I want my live patching based on this system's kernel version" (as opposed to looking for the version and adding the filters manually). It's useful because one of the most common use cases is that the user wants to setup live patching based on the currently installed kernel version of some of their systems.

> How does this look when there are multiple kernel version available in the same channel? I. e. are we sure this will not define filters to be used with a different kernel version?

The correct approach for the user would be to make sure all the systems assigned to the project have the same kernel version installed. If they have systems with older versions, they would still see the updates up to the picked version. On the other hand, if there are systems with newer versions than the one chosen, **those ones will have neither real updates nor the live-patching updates.**

Basically, a live-patching project will target a specific kernel version and therefore all the systems assigned must respect that.

Imagine a distribution that has kernel versions 1, 2 and 3. When you setup a live-patching project targeting version 2, you'll have the following packages in the channels:
File_000

accepted/00075-live-patching-enhancements.md
cbbayburt: Every product (hopefully) has a corresponding `lifecycle` package in the same channel as the product and they are also "required" by the product's release package.

cbbayburt:
> But when we parse RPMs, we need a new structure and we need an algorithm similar to ServerNeededCache which calculate lifecycle Info for every System.

Why do we need to calculate for every system? The lifecycle data is static and doesn't depend on the system. So basically we can get the package name of the currently installed live patch and query the lifecycle information of this package from the DB.

cbbayburt:
> It's not easy to make a call here. I would evaluate the complexity: I'd go with Python only if we expect something really straightforward (100-200 lines or so). If this smells complex, bite the bullet and build there where it's safest.

In that case, I'd say let's go with Python, but instead of patching it into the existing code, let's structure it into a separate module and call it from the reposync.

mcalmer:
> > But when we parse RPMs, we need a new structure and we need an algorithm similar to ServerNeededCache which calculate lifecycle Info for every System.
> 
> Why do we need to calculate for every system? The lifecycle data is static and doesn't depend on the system. So basically we can get the package name of the currently installed live patch and query the lifecycle information of this package from the DB.

Well, we do not store the data at the system directly, but more general. So 

Q1: how do we store it? New table or in rhnPackage?
Q2: where do we need to show that a package reached EOL? Can we get this info fast enough with a new query (performance)? 

I think we might not need a "needed cache" table, but how we store the data influence how fast we can get it.
Do we need it for multiple system? What about the daily email with important info for all managed systems per admin?

cbbayburt:
We'll store the data in the most agnostic way.
Basically, in a new table such as `susePackageLifecycle`:
`package_id, eol_date`

To show it in some UI, we'll query the table with the package ID of the installed live-patch package.
Finding the package ID should be as straightforward as querying the `rhnServerPackage` to get the latest.

In the end, it will be a query with some join of `rhnServerPackage`, `rhnPackage` and `susePackageLifecycle`.

cbbayburt:
And for a report of EOL (notifications or email) we can go the other way around and start with the `susePackageLifecycle` table.
We'll get the package ID's that expire soon, find which systems have those packages installed as latest. A view might be handy for this.

cbbayburt:
@mcalmer, @moio, @paususe, thank you for the continuous feedback so far!

I've finalized the design and split the two distinct topics into separate documents. Please feel free to take a final look and if we have the green light, I'll go ahead and merge them.

- [Live Patching with CLM](https://github.com/cbbayburt/uyuni-rfc/blob/live-patching-eol/accepted/00075-live-patching-with-clm.md)
- [Live Patching EOL Date Information](https://github.com/cbbayburt/uyuni-rfc/blob/live-patching-eol/accepted/00076-live-patching-eol-data.md)

moio (approved):
I do not have big objections to this text.

As always, I recommend proceeding with implementation in as-small-as-possible steps, to be ideally delivered in Uyuni bit by bit :wink: 

Thanks for all the good work in here!

accepted/00076-live-patching-eol-data.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Live patching EOL information
+
+SUSE provides lifecycle information on any product to users via 'zypper lifecycle' plugin. The actual data about a product is provided in a plain CSV file in a supplementary package per product. In case of live patching, this package is called `lifecycle-data-sle-live-patching` and is installed together with the product. Zypper lifecycle plugin reads this data to display human-friendly information to users.
+
+Uyuni shall use this data to provide the following:
+ - Display EOL dates for the current live patch in system details
+ - Display an out-of-support systems list in the home page
+ - Use the notifications system and/or send emails to report live patched systems approaching/past EOL
+ - The notifications/emails will be sent in advance according to a time value configurable from the UI
```
paususe, line 33: - Provide EOL information via API (expand to what namespaces, probably later, after the SQL)

paususe (approved):
The only thing I'm missing is explicitly mentioning EOL information will also be available from the API. Other than that, LGTM.

accepted/00075-live-patching-with-clm.md
```diff
@@ -0,0 +1,97 @@
...
+**Example C:** Live patching project using reboot flag
+ - `DENY errata containing reboot_suggested (keyword)`
+ - `ALLOW errata containing "Live Patch" in synopsis` *- an override to the first rule*
+ - `ALLOW a specific patch containing current kernel version (via advisory name)` *- an override to the first rule*
+
+## CLM filter wizards
+
+The process of live patching can be made more user-friendly using "CLM filter wizards". The wizards are alternatives to the regular filter creation dialogs that provide a tailored interaction to create multiple filters that serve a single purpose. These dialogs can be accessed via the "Use wizard" button in the filter creation dialog. This concept can be extended to other use cases and workflows in the future.
+
+When completed, the wizards will create multiple filters that look like the examples mentioned in the section above to set up a project with a specific purpose. Since live patching channels can be composed using the existing filter types, no additional filter type needs to be implemented.
```
mcalmer, line 45: We need to agree on what exact filter we want to setup. Having "options" in an RFC is ok for discussion, but the final version must describe 1 way we want to go. Otherwise the development team has again to make a decission.

accepted/00076-live-patching-eol-data.md
```diff
@@ -0,0 +1,119 @@
...
+ - `eol_date` of `DATE` type
+
+ The data stored in this table is available to be queried and consumed in the UI.
+
+ *For other methods of data retrieval that are evaluated in the design, see [Alternatives](#alternatives).*
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+ - Unpacking must be done during reposync, adding complication
```
mcalmer, line 81: An option would be to do it when we generate the metadata for a new synced channel.
Of cause we need to call an external program from the taskomatic job.

But during reposync is also ok.

accepted/00076-live-patching-eol-data.md
```diff
@@ -0,0 +1,119 @@
...
+WHERE
+    --Select packages that have '*.lifecycle' files in them
+    c.name LIKE '%.lifecycle' AND
+    e.evr = (
+        --Select latest EVR for a package name
+        SELECT MAX(evr)
+        FROM rhnPackageEvr se
+            JOIN rhnPackage sp ON se.id = sp.evr_id
+        WHERE sp.name_id = p.name_id
+    );
```
mcalmer, line 59: We should limit this to the channel we synced. Otherwise we call this for all channels again and again and also unpack RPM packages when we already have the data.

Or even better we return the package ID of the lifecycle package and store that ID also in the new table.
With this we can exclude lifecycle data packages from unpacking, when the data were already processed

mcalmer (changes requested):
We must decide what filters we want to set

accepted/00075-live-patching-with-clm.md
cbbayburt: Since there weren't any concrete objections and as the most sensible choice at this stage, I went with Example A

accepted/00076-live-patching-eol-data.md
cbbayburt: I think when implemented properly with enough logging, reposync is the most natural choice.

accepted/00076-live-patching-eol-data.md
cbbayburt: Good point. I've updated the SQL example accordingly.

mcalmer (approved):

PR #49, opened 2020-12-17
System assignments visibility of configuration state channels

ncounter:


accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,81 @@
...
+
+### Why are we doing this?
+
+At the moment there is no way to figure out Configuration and State Channels connections other than the ones directly in place to a specific System. Such a type of channel can exist and be difined at:
+
+- Organization level
+- System Group level
+- System level
+
+but only the third level is visible for each System. This means that once Salt States are applied, there will be more than what is visible: Salt States are a set of defined states from
```
hustodemon, line 21: - There is some repeated information in this section (e.g. the fact that only direct channel-system association is visible).
- If I didn't know the topic up to some point, I'd be a bit confused (for instance "This means that once Salt States are applied, there will be more than what is visible" sentence).

I suggest that we re-phrase it in the following way:

```suggestion
At the moment there is no way to figure out which Configuration/State Channels are assigned to a System, other than the directly assigned ones.

Apart from this direct assignment, the System inherits the Configuration/State channels defined on the Organization level and the System Group level, but this inheritance is not visible in the Web UI/XML RPC.

To sum it up: The Configuration/State channels effective for a system are composed based on:
```

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,81 @@
...
+8. create a State Channel C3 directly at a System S1 detail level
+9. go to System Detail > States > Configuration Channels > System
+10. the C3 channel is listed as an assignment to S1
+11. there is no info about C1 nor C2 channels, but because of step 4. 5. 6. 7.  in reality S1 will receive not only C3 but also C1 and C2 as inheritance of its own setup, being part of O1 and G1
+12. the reverse logic is valid as well, there is no information about systems assigned to channels C1 or C2 in Configuration > Channels > Channel Overview > Systems because no system is directly assigned to them, but in the end they are.
+
+
+### What is the expected outcome?
+
+The fact that with Salt *black magic* there is no way to figure out what will override others in the applied states, we cannot provide a reliable hierarchy.
```
hustodemon, line 45: I'd suggest hiding "black magic" behind a somewhat "weaker" term :rofl: 

```suggestion
Because of the way the Salt highstate is composed, it is hard to show the order of the application of the states in Uyuni.
```

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,81 @@
...
+What we can provide is a full set of assignments, no matter if directly assigned or inherited by other assignments.
```
hustodemon, line 46: Too much `assign.*` in the sentence.

```suggestion
What we can provide is a representation of all State/Configuration channels assigned to a system: both direct and inherited (from System Group and Organization).
```

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,81 @@
...
+At the moment there is no way to figure out Configuration and State Channels connections other than the ones directly in place to a specific System. Such a type of channel can exist and be difined at:
+
+- Organization level
+- System Group level
+- System level
+
+but only the third level is visible for each System. This means that once Salt States are applied, there will be more than what is visible: Salt States are a set of defined states from
+
+```Organization Configuration Channels **+** System Group Configuration Channels **+** System Configuration Channels```
+
```
hustodemon, line 24: I'd suggest mentioning traditional clients here.

```suggestion

> **NOTE** All of the above applies solely to the Salt clients. Traditional clients do not inherit Configuration channel assignments from System Groups/Organizations.


```

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,81 @@
...
+TODO
+
+{
+  - data are already available in the database
+  - identify the proper way to show them
+  - how to interact, if any interaction is expected
+
+}
+
```
hustodemon, line 59: Adding an initial list. We might get more from @mbrookhuis and we should also do some deeper research ourselves.

```suggestion
## Changes in the user-facing parts
We need to change the following places in the Web UI/XMLRPC. As stated above, we target the salt minions only.

### Web UI
- _System detail -> Configuration -> Overview_: Show inherited channels in the _configuration overview_
- _System detail -> Configuration -> Manager Configuration Channels_: Show the inherited channels (with *no* possibility to unsubscribe from it) .TODO: how? same table? a separate one?
- _Configuration/State Channel detail -> Systems -> Subscribed systems_: show the information about *minions* that inherit this channel
- TODO: _Configuration File detail -> deploy file_: show inherited *minions* too?

### XMLRPC
- _ServerConfigHandler.listFiles_
- _ServerConfigHandler.listChannels_
- _ConfigChannelHandler.listGlobals_ contains `systemCount`. TODO: Should this reflect minions that inherit the channel?
- _ConfigChannelHandler.listSubscribedSystems_ TODO: Should contain inherited systems?
```

hustodemon (commented):
Hi @ncounter ! Thanks for the draft, pushing an initial set of comments/suggestions.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,81 @@
+- Feature Name: Enhance visibility of connected Configuration/State Channels
+- Start Date: 2020-12-17
+- RFC PR: 
+
+# Summary
+[summary]: #summary
+
+Provide visibility and overview of **inherited** Configuration and State Channels connections at a System Details level.
```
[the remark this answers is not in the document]
paususe: Salt minions? Or Salt clients? Please don't forget we also support salt-ssh clients

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,81 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+### Why are we doing this?
+
+At the moment there is no way to figure out Configuration and State Channels connections other than the ones directly in place to a specific System. Such a type of channel can exist and be difined at:
```
paususe, line 15: Typo: "difined"

accepted/00076-configuration-channels-visibility.md
paususe: Please don't forget about the Uyuni level (global), and the formulas with forms level.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,81 @@
...
+3. onboard a System S1 as a SUSE Manager client
+4. have the System S1 part of the Organization O1
+5. have the System S1 part of the System Group G1
+6. create a State Channel C1 at an Organization O1 level
+7. create a State Channel C2 at a System Group G1 level
+8. create a State Channel C3 directly at a System S1 detail level
+9. go to System Detail > States > Configuration Channels > System
+10. the C3 channel is listed as an assignment to S1
+11. there is no info about C1 nor C2 channels, but because of step 4. 5. 6. 7.  in reality S1 will receive not only C3 but also C1 and C2 as inheritance of its own setup, being part of O1 and G1
+12. the reverse logic is valid as well, there is no information about systems assigned to channels C1 or C2 in Configuration > Channels > Channel Overview > Systems because no system is directly assigned to them, but in the end they are.
```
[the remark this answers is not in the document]
paususe: Please remember a system may belong to several system groups. We need to deal with that case.

accepted/00076-configuration-channels-visibility.md
paususe: We should represent the whole list and origin of assigned channels:

- Uyuni
- Organization
- System group(s)
- System
  - Channel(s)
  - Formula(s)

accepted/00076-configuration-channels-visibility.md
paususe: What about System detail -> States -> Configuration channels ?

I am a bit confused by the names used here btw, as they do not match the menu options. Can we please use exactly the same wording we have in the WebUI?

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,84 @@
...
+
+At the moment there is no way to figure out Configuration and State Channels connections other than the ones directly in place to a specific System. Such a type of channel can exist and be defined at:
+
+- Organization level
+- System Group level
+- System level
+
+but only the third level is visible for each System. This means that once Salt States are applied, there will be more than what is visible: Salt States are a set of defined states from
+
+Organization Configuration Channels **+** System Group Configuration Channels **+** System Configuration Channels
```
moio, line 26: Trying to reword in order to see if I got the concept right:

```suggestion
At the moment it is possible to assign a Configuration or State channel to:

- all clients in an Organization
- all clients in a System Group
- a single client (direct assignment)

Direct assignments are visible in the UI, but there is no representation out which Configuration and State Channels apply to which client from the Organization and System Group assignments.

This means that at highstate application time, it is not evident which State comes from which "assignment type" (default, Organization, System Group or direct).

This RFC is about giving users a way to see which State comes from which assignment before application.
```

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,84 @@
...
+3. onboard a System S1 as a Salt client in the Organization O1 and System Group G1
+4. create a State Channel C1 at an Organization O1 level
+5. create a State Channel C2 at a System Group G1 level
+6. create a State Channel C3 directly at a System S1 detail level
+7. go to System Detail > States > Configuration Channels > System
+8. the C3 channel is listed as an assignment to S1
+9. there is no info about C1 nor C2 channels, but because of step 4. 5. in reality S1 will receive not only C3 but also C1 and C2 as inheritance of its own setup, being part of O1 and G1
+10. the reverse logic is valid as well, there is no information about systems assigned to channels C1 or C2 in Configuration > Channels > Channel Overview > Systems because. No system is directly assigned to them, but S1 inherits them.
+
+Additional use case: a system can be part of multiple groups, therefore **inheritance** can become wide and complex really fast.
```
moio, line 43: Rather than qualitatively explained ("become wide and complex really fast") such complexity should be explained here IMO.

How is inheritance defined in this context (pointer to existing documentation if any)? What are the elements of such complexity? What does it mean it increases in a "really fast" way?

Remember that the aim of the RFC is not to sell the feature - rather, to explain the problem as an introduction to the solution, which is the part under scrutiny here. The intent is to submit the solution to peers before going to implementation to reduce the risk of "bad surprises" at PR time.

accepted/00076-configuration-channels-visibility.md
moio: I'd also put it last, as it is a limitation to the title question "What is the expected outcome?"

I found it confusing on a first read that the first point in the response is what we _won't_ do :wink:

accepted/00076-configuration-channels-visibility.md
moio: I agree with @paususe that, from a user perspective, it should be more useful to frame the overall question in terms of States: where does a State that applies to a certain client come from?

Note that this is an extension of the question: what _Channel provided_ State applies to a certain client, because it involves at least:
 - default states applied to all clients
 - states from assigned formulas

Do we intend to tackle those two, at this time or eventually? Do we have ideas how?

moio (commented):
I have added a clarification and a bigger question for now.

I plan to review again at a later point when those are addressed.

Thanks for the efforts so far!

accepted/00076-configuration-channels-visibility.md
ncounter: > * If I didn't know the topic up to some point, I'd be a bit confused (for instance "This means that once Salt States are applied, there will be more than what is visible" sentence).

I see the point, but the RFC is not meant to explain the feature from scratch, otherwise it would be much longer. And the focus would diverge to something which is not intended to be discussed.

In agreement with your comment and others, I am rephrasing using words from all of you.

accepted/00076-configuration-channels-visibility.md
ncounter: > Rather than qualitatively explained ("become wide and complex really fast") such complexity should be explained here IMO.
> 
> How is inheritance defined in this context (pointer to existing documentation if any)? What are the elements of such complexity? What does it mean it increases in a "really fast" way?

Probably "complexity" is not the right word then. I left out the detailed explanation of such a "wide and complex scenario" because it is still the original problem explained. It is only about the same **inheritance invisibility**, but multiplied for every group a system can be part of.
If you are part of one group, you inherit from that group only: in such a case you can even deduct where do they comes from. If they are directly assigned, you got them. If they are not directly assigned, well they come from the single group you are in.
But if you are part of 5 different groups, and you inherit from all of them, the resulting set will be uncategorized and you cannot distinguish what comes from group G1, group G2, etc.

I'm adding here two pictures as they usually give more insight than many many words. I plan to inject them into the RFC, anything against or anything you would like to change about it?

### Visible inheritance
visible-inheritance

### Hidden inheritance result
uncategorized-result




> 
> Remember that the aim of the RFC is not to sell the feature - 

JFYI, I don't see any marketing slogan here to sell the feature actually :/

moio: As per our meeting, this PR is to be considered WIP for now, so let's just suspend any reviewing until it's ready.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+
+After:
+![Configuration Channels AFTER](images/configuration-channels-visibility/Configuration_Channels_AFTER.png)
+
+### `Configuration > Channels > Details > Overview`
+Before:
+![Configuration Channels Detail Overview BEFORE](images/configuration-channels-visibility/Configuration_Channels_Details_Overview_BEFORE.png)
+
+After:
+![Configuration Channels Detail Overview AFTER](images/configuration-channels-visibility/Configuration_Channels_Details_Overview_AFTER.png)
```
cbbayburt, line 149: Question: Can the channels be shared across organizations in the current implementation? Is it going to change with the RFC?

cbbayburt (commented):
Thanks @ncounter, screenshots make it really easy to grasp.

A couple of comments below:

- One thing I couldn't find is the `System Details > Configuration > *` pages. They're almost identical to mentioned SSM pages and probably need the same enhancements.
- Another thing is, maybe it'll be useful to describe the implementation plan a bit, without going too much into detail. For example, where do we have all this inheritance data and how do we get it?

accepted/00076-configuration-channels-visibility.md
ncounter: AFAIK they are not shared across organizations (`rhnconfigchannel` has an `org_id` parameter), and there is no plan or intention to change it at all, not in this RFC scope and not in general.

ncounter:
>     * One thing I couldn't find is the `System Details > Configuration > *` pages. They're almost identical to mentioned SSM pages and probably need the same enhancements.

@cbbayburt That's a good one I evaluated but I have some concerns about, thanks for asking. At the moment we have the following:
- in `System Details` we have a split:
  - `System Details > Configuration > *`
  - `System Details > State > Configuration Channels`

and reality is that in the first section you can play with managed files, in the second one you can play with states, so to me it is a good separation of the two concepts, and keeping them like it is it makes sense. Therefore my proposal is to apply changes only to `System Details > State > Configuration Channels` because we are in the context of Configuration State Channels only and for salt minion systems only. A note aside is that, following my point, we have a bug in both the lists because we always list all the Configuration Channels types without distinguishing them. From my POV we should instead filter out State Channels in the `System Details > Configuration > *` and the other way around for the other section.

All that said, we don't have the same behavior in SSM: in there we simply have the `SSM > Configuration` section where we can subscribe/unsubscribe systems to Configuration Channels and Configuration State Channels, and we have a separate section for `SSM > States` where we only have the Highstate management. Now, I cannot omit I feel this should be something to align across all the named pages, but it should also be seen as out of scope of the current RFC. The only option to include all of the alignment would be to go for the **Alternative** section described in the RFC, but it requires more in terms of time, effort, research and plan.

ncounter:
>     * Another thing is, maybe it'll be useful to describe the implementation plan a bit, without going too much into detail. For example, where do we have all this inheritance data and how do we get it?

@cbbayburt point taken. It should all be available in the database, more a matter of fetching and putting data together where needed, but yeah, I'll go through it and add something in that regard.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+
+### `Configuration > Channels > Details > Systems > Inheriting Systems`
+This is a new page where systems are listed only if not subscribed to the channel directly, but inheriting it through Organization or System Group assignment.
```
hustodemon, line 152: Maybe it'd be nice to also display the system group(s)/org in the inherited systems list, so the table would something like this:
```

|----------------+-----------------|
| System name    | Inherited from  |
|----------------+-----------------|
| test01.suse.de | org1, my group1 |
| test02.suse.de | my group2       |
|----------------+-----------------|
```

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+
+Before:
+![Configuration Channels Detail Systems BEFORE](images/configuration-channels-visibility/Configuration_Channels_Details_Systems_BEFORE.png)
+
+After:
+![Configuration Channels Detail Systems AFTER](images/configuration-channels-visibility/Configuration_Channels_Details_Systems_AFTER.png)
+
+
+### `Configuration > Channels > Details > System Groups`
```
hustodemon, line 161: What would be the plan for implementation? Write in it ReactJS? Or generalizing the struts one somehow?

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+This page reproduces the same behavior of the `Configuration > Channels > Details > Systems` tab, but for System Groups.
+
+This **new** page is pointed by:
+- the `Channel information` box link of the system groups counting in the `Configuration > Channels > Details > Overview` page: the link will land on the new page at the `Subscribed System Groups` tab
+
+### `Configuration > Channels > System Groups`
+This page represents a complete list of pairs value `Configuration State Channel - System Group`: every value is a link to the specific entity detail (channel or system group).
```
hustodemon, line 168: I'm not sure if the added value of such a über-overview page is bigger than the possible confusion in brings. If the user has a lot of groups assigned to a lot of channels, then the list grows pretty well: if you have 5 groups and each one has 10 channels assigned, then the list will have 50 items. Then the user would need to have some kind of filtering in order to focus on needed groups/channels. But then again: they could use the channel details or system group details pages for getting the same information.
But maybe there is a use case that I overlooked.

hustodemon (commented):
Thanks for the RFC, @ncounter ! I focused on the factual side of it in my review. I still need some clarification for some things, so I posted some question. I also added a couple of suggestions, that could be discussed further (like the "columns consolidation").

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+
+And here it is the problem this RFC is trying to describe to solve: apart from the direct assignment, the System inherits channels (thus the content) from the Organization it is part of, and the same goes for the System Groups he is part of, but this inheritance is not visible in the Web UI nor in the XML RPC API.
+
+Note: in addition to what has been described so far, for what concerns the Configuration State Channels case, behind all the Salt States collected from the channels, there are also a bunch of default Salt States, the ones the server has to supply during the registration of a salt minion client. Those are not *assigned* in any way because they are not part of any Configuration State Channel, they do just exists due to the logic implementation of a client registration using Salt. In the end, those are not visible Salt States anyway, and the System will be affected by them during the registration, so from the System perspective an additional source of Salt States could be defined as directly assigned: a non-manageable source behaving as a Configuration State Channel that will be named *registration-states*.
```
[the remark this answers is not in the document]
ncounter: Initially I left part of your point in the _unresolved questions_ section, as I don't yet see a *strong* need to represent these states for two reasons:
1- we don't know in advance what can be part of such a bunch of states
2- we would need to just represent statically a generic `internal management states` entity that would be valid for all the registered system, ever.

Therefore, if we agree on this, I'll adapt the RFC removing the _unresolved question_ point and adding to the design the fact that this information will be statically presented in a Note in `Systems > Details > States` page: that is the relevant place where it makes sense to remember _Hey, keep in mind this systems has also a couple of internal states due to internal logic management, ..._ . By doing so we also keep it open for the future the possibility to implement from there a detailed view for those internal states only, a view we don't have and we don't need at the moment.

Bottom line, I fully agree it makes much more sense a generic `internal states` instead of `registration states`. Thanks for your point :+1:

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
+- Feature Name: Enhance visibility of connected Configuration State Channels
+- Start Date: 2020-12-17
+- RFC PR:
+
+# Summary
+[summary]: #summary
+
+Provide visibility of **inherited** Configuration State Channels connections at a System Details level. The reverse information as well should be provided by giving visibility of those systems they are connected (directly or by **inheritance**) to a certain Configuration State Channel at a Channel Details level.
```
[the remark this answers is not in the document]
ncounter: I see your point, but for consistency and clear matching I actually used the naming we have all across the UI.
There is no mentioning of `State Channels` only, it is always `Configuration State Channels`, or more generically `Configuration Channels` with the `state` as the `type` of the channel (it's just a property).

The entire RFC scope is for `State Channels` only.
Do you feel I should adapt the RFC removing the `Configuration` prefix, or an initial clarification would be enough?

Anyway, your point is completely understandable to me, and it empowers the need to distinguish more the concept and the UX across this feature. Something I consider not part of this RFC though.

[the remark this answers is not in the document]
moio: It seems to me we call them "state channels" in the documentation:

https://documentation.suse.com/external-tree/en-us/suma/4.1/suse-manager/reference/configuration/config-channels.html

And, as you mentioned, we basically do not have a name for them in the UI. I would suggest to align to the documentation jargon and, if you feel there might still be any ambiguity, add a small glossary item.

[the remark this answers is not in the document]
ncounter: Fully agreed. I'll rename it here to clarify what we are talking about, no reason to add more clarifications.
About the UI, I think we should fix it too. Out of scope of this RFC output, but maybe we can let it sneak into.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+# Detailed design
+[design]: #detailed-design
+
+### `Home > My Organization > Configuration Channels`
+This page represents the assignment connection between an organization and configuration channels. All the systems in this organization will inherit the selection set of channels here. This should be only a list of Configuration Channels with the possibility of simply select/deselect channels and Apply changes. If there is a need to show only the channel selected, a filter to the list page should be added instead.
+
+Before:
+![My Organization BEFORE](images/configuration-channels-visibility/Home_MyOrganization_ConfigurationChannels_BEFORE.png)
+
+After:
```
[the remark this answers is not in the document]
ncounter: That will be part of the task, yeah: one tab means no tab in the end :wink:

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+![My Organization AFTER](images/configuration-channels-visibility/Home_MyOrganization_ConfigurationChannels_AFTER.png)
+
+### `Systems > Details > States > Configuration Channels`
+Before:
+![System Details BEFORE](images/configuration-channels-visibility/Systems_Details_States_ConfigurationChannels_BEFORE.png)
+
+After:
+![System Details AFTER 1](images/configuration-channels-visibility/Systems_Details_States_ConfigurationChannels_AFTER_1.png)
+![System Details AFTER 2](images/configuration-channels-visibility/Systems_Details_States_ConfigurationChannels_AFTER_2.png)
```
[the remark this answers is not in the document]
ncounter: You're right, I think we can do the same here. In the end the _three tabs_ pattern is useful when it comes to packages (`Systes > Details > States > Packages`) because in there it is nice to see changes and what is currently on the system. But in this case we just need one list where the user can check/uncheck channels, confirm, reorder and save. I'll update the screenshot accordingly. :+1:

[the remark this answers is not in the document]
ncounter: And same as before, the one tab left will be no-tab in the end.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+### `Systems > System Set Manager > Overview > Configuration > Subscribe to Channels`
+Before:
+![SSM Subscribe To Channels BEFORE](images/configuration-channels-visibility/Systems_SSM_Configuration_SubscribeToChannels_BEFORE.png)
+
+After:
+![SSM Subscribe To Channels AFTER](images/configuration-channels-visibility/Systems_SSM_Configuration_SubscribeToChannels_AFTER.png)
+
+### `Systems > System Set Manager > Overview > Configuration > Unsubscribe to Channels`
+Before:
+![SSM Unsubscribe From Channels BEFORE](images/configuration-channels-visibility/Systems_SSM_Configuration_UnsbscribeFromChannels_BEFORE.png)
```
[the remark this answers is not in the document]
ncounter: There was a typo in the image name link. Fixed now

accepted/00076-configuration-channels-visibility.md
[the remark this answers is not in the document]
paususe: Think of the future Ansible integration. Users managing their systems mainly with Ansible, where Salt will be only used for those `internal states`, definitely want to know what those states are to avoid the Ansible playbooks and Uyuni internal states conflicting.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+## What other designs/options have been considered?
+
+Without investigating deeply, the alternative would be redesigning the feature completely from both the design and technical perspectives: the technical refactoring should get rid of the struts stack and redefine the concept of Configuration Channel and Configuration State Channel used by traditional or salt minion clients. By addressing the technical part, also the workflow could be re-drawn, dropping some pages and instead of adding separated tabs of tables it would be preferrable to have unified modern lists of filterable entries. The drawback of this alternative is that it would take much longer (from 3 to 5 times roughly) than the proposed official solution.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+## What are the unknowns?
+1. There is a real need for representing `registration-states` (as mentioned before) or they can still stay behind the scene?
```
paususe, line 196: I think so, yes, especially in the future with Ansible in the picture.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+2. Salt States coming from Formulas are not yet considered as part of this RFC
```
paususe, line 197: Why? Can we assimilate them to states assigned directly to a system?

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+3. This RFC is considering presenting all the sources of the Salt States for a given system. The RFC is **not** considering instead to implement a reverse-engineering to pick every Salt State from a given HighState and build the history of the specific source where it came from.
```
paususe, line 198: Agreed.

I would also add: this RFC is no about tracing each line in the highstate to one of the states assigned directly/indirectly to a system.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+No real reason for not enhancing the feature adding the missing visibility.
+
+The concrete drawback is: the solution is adding information to a tech-debt affected architecture, spreading the same information in many places (see how many pages/screenshots). Maintaining should not be complicated, but future refactoring of the feature workflow could be more difficult.
+
+# Alternatives
+[alternatives]: #alternatives
+
+## What other designs/options have been considered?
+
+Without investigating deeply, the alternative would be redesigning the feature completely from both the design and technical perspectives: the technical refactoring should get rid of the struts stack and redefine the concept of Configuration Channel and Configuration State Channel used by traditional or salt minion clients. By addressing the technical part, also the workflow could be re-drawn, dropping some pages and instead of adding separated tabs of tables it would be preferrable to have unified modern lists of filterable entries. The drawback of this alternative is that it would take much longer (from 3 to 5 times roughly) than the proposed official solution.
```
paususe, line 189: I think we can leave the big refactoring for the day the traditional stack will be gone.

accepted/00076-configuration-channels-visibility.md
```diff
@@ -0,0 +1,201 @@
...
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## Why should we **not** do this?
+
+No real reason for not enhancing the feature adding the missing visibility.
+
+The concrete drawback is: the solution is adding information to a tech-debt affected architecture, spreading the same information in many places (see how many pages/screenshots). Maintaining should not be complicated, but future refactoring of the feature workflow could be more difficult.
```
paususe, line 182: Is the tech debt you are referring to Struts? How much more Struts will we be adding? It seems to me rewriting every page we need to touch in ReactJS is rewriting a good portion of the UI :-|

paususe (approved):
Thank you for the RFC. Mockups definitely make it easier to understand. It looks generally OK to me.

accepted/00076-configuration-channels-visibility.md
ncounter: Ack, I didn't gave it a though about the Ansible in the picture actually, thanks for the heads-up. I'll adapt the RFC including the `internal-states` piece.

accepted/00076-configuration-channels-visibility.md
ncounter: The tech debt part is about Struts yes, but it touches it directly and indirectly. One reason is the direct way of introducing data into a page: when it is about Struts we need to fetch and inject data at the same time because there is not a real separation of MVC so we will have to edit both the `ActionController` and the `JSP` to render the information required. But all of this implementation we cannot reuse it because in the ReactJS stack we have a better separation of context where we can fetch and serve data at a specified endpoint, then different pages can consume the information independently and reusing such an endpoint. So in the end we are adding the double of the code to maintain because we touch both the stacks for the same thing.
The indirect point instead is the fact that the UX cannot be changed that much (mainly for consistency) due to existing pages and sub-tabs, so we need to keep the same behavior where we present a link to a different page, instead of rethinking to the entire workflow, with possible dynamic interactions inside the same page, by swapping content or adding boxes, popup frame list or so. This is not possible, or at least it does not worth the effort to implement it in the Struts/JSP stack.

From an Usability perspective, rewriting a good portion of the UI in this case would be something good because the UX is quite old and cumbersome (many tabs and sub tabs pages, lots of tables to touch, different places to present partial information), it could be simplified and smoothed in the end. But of course it would require more time and effort to achieve.

ncounter: Probably it can be seen as part of the `## What other designs/options have been considered?` section. Therefore we will tackle it during a bigger refactoring, even if I am not yet sure when dropping the traditional stack will come.

accepted/00076-configuration-channels-visibility.md
ncounter: Thanks for raising this up: your last point is exactly what this page would be there to serve. At the moment in case you want look for an entity, you need to go **either** to the channel details page **or** the system group details one. While it is true that the most common (probably the only one) use case is to look for one entry at a time, at the same time the UX would be a PITA if you want to look first for more than one channel (you would have to go back and forth from channel details to the right tab and back to the list and forth again), or maybe you first want to look for a channel because you want to know all the groups connected, then for a system group because you want also to check about all the channels connected to it, just to make sure you have everything under your radar. In order to achieve that, you would have to go crazy around pages, while with a single page you could have all, just a matter of filtering. For sure filters are meant to be there for both the channels and the system groups.

Overall, this is one of a generic issue that exists in the product for the most of the features: doing slightly the same (but not exactly the same) in many different places, but no place to do them all.

There is one possible drawback, and I'd like to hear your POV @hustodemon : could it be a slow performing page presenting a huge amount of data? I don't really see this as an issue because it's just a list of pair values, but maybe I am missing something in the picture.

accepted/00076-configuration-channels-visibility.md
ncounter: That's a really good one to add here as well: adding not only the fact that the system in inheriting this channel, but also from where. :+1:
I'll update accordingly, thank you.

accepted/00076-configuration-channels-visibility.md
hustodemon: Ok, I agree, that having a single page with a responsive filtering, where the user could just punch in the channel name or the system group name would be a nice feature.

I suppose this will be implemented as a eager fetch from the backend and then client-side filtering (afaik, we still don't have a server-side filtering for the react tables). In this case, the data set would need to be really big to exhibit low performance. The main problem I have is more general: it's the absence of the server-side filtering "framework" for the new components. What we're doing is that we are piling new components that work ok on reasonable data sets to Uyuni. Then one day, when we discover we need to rewrite them all, it can be to much. So in the end, this is a tech debt problem and we should talk about it outside this RFC.

accepted/00076-state-channels-visibility.md
```diff
@@ -0,0 +1,218 @@
...
+Provide visibility of **inherited** State Channels connections at a System Details level. The reverse information as well should be provided by giving visibility of those systems they are connected (directly or by **inheritance**) to a certain State Channel at a Channel Details level.
+
+*Note: everything discussed in this RFC applies to Salt clients only. Traditional clients do not inherit State Channels.*
+
+# Motivation
+[motivation]: #motivation
+
+## Why are we doing this?
+
+By definition a State Channel is a pool of salt states resources. Once a channel of this type is created, it is possible to subscribe systems to it in order to apply the content of that channel to them. Alternatively, channels can also be assigned to System Groups and/or to Organizations.
```
moio, line 17:
```suggestion
By definition a State Channel is a pool of Salt states. Once a channel of this type is created, it is possible to subscribe systems to it in order to apply the content of that channel to them. Alternatively, channels can also be assigned to System Groups and/or to Organizations.
```

accepted/00076-state-channels-visibility.md
```diff
@@ -0,0 +1,218 @@
+- Feature Name: Enhance visibility of connected State Channels
+- Start Date: 2020-12-17
+- RFC PR:
+
+# Summary
+[summary]: #summary
+
+Provide visibility of **inherited** State Channels connections at a System Details level. The reverse information as well should be provided by giving visibility of those systems they are connected (directly or by **inheritance**) to a certain State Channel at a Channel Details level.
```
moio, line 8:
```suggestion
Provide better visibility, in the UI, of the relationships between State Channels, System Groups, Organizations and Systems - especially indirect ones.

The goal is to clarify which system **inherits** which State Channels, and why - and conversely, which State Channel are ultimately associated to which Groups, Organizations and Systems.
```

accepted/00076-state-channels-visibility.md
```diff
@@ -0,0 +1,218 @@
...
+It turns out that once the Salt States are collected from the various source States Channels, they are handled by the Salt engine itself which builds up the Salt highstate, and this one has no tracking information of the order or the source of the states origin.
+
+The proposed solution of this RFC is not including an end-to-end description of where each single Salt State comes from where at the moment, the initial improvement is to add the visibility of the set of State Channels every system inherits from other entities (System Groups and Organizations as described before).
+
+
+# Detailed design
+[design]: #detailed-design
+
+### `Home > My Organization > Configuration Channels`
+This page represents the assignment connection between an organization and both configuration channels and state channels. All the systems in this organization will inherit the selection set of channels here. This should be only a list of Configuration Channels and State Channels with the possibility of simply select/deselect channels and Apply changes. If there is a need to show only the channel selected, a filter to the list page should be added instead.
```
moio, line 91:
```suggestion
This page presents associations between an organization and both Configuration Channels and State Channels. All systems in this organization will inherit Channels selected here. This is a list of Configuration Channels and State Channels with the possibility of selecting/deselecting channels and Applying any changes.

Proposal is to add a filter to the list page for convenience.
```

accepted/00076-state-channels-visibility.md
```diff
@@ -0,0 +1,218 @@
...
+### `Systems > Details > States > Configuration Channels`
+Before:
+![System Details BEFORE](images/state-channels-visibility/Systems_Details_States_ConfigurationChannels_BEFORE.png)
+
+After:
+![System Details AFTER 1](images/state-channels-visibility/Systems_Details_States_ConfigurationChannels_AFTER_1.png)
+![System Details AFTER 2](images/state-channels-visibility/Systems_Details_States_ConfigurationChannels_AFTER_2.png)
+
+### `Systems > System Groups > Details > States > Configuration Channels`
+This page represents the assignment connection between a system group and state channels. All the systems in this system group will inherit the selection set of channels here. This should be only a list of State Channels with the possibility of simply select/deselect channels and Apply changes. If there is a need to show only the channel selected, a filter to the list page should be added instead.
```
moio, line 115:
```suggestion
This page presents associations between a system group and State Channels. All systems in this system group will inherit the selected set of channels.

Proposal is to add a filter to the list page for convenience.
```

accepted/00076-state-channels-visibility.md
```diff
@@ -0,0 +1,218 @@
...
+![Configuration Channels AFTER](images/state-channels-visibility/Configuration_Channels_AFTER.png)
+
+### `Configuration > Channels > Details > Overview`
+Before:
+![Configuration Channels Detail Overview BEFORE](images/state-channels-visibility/Configuration_Channels_Details_Overview_BEFORE.png)
+
+After:
+![Configuration Channels Detail Overview AFTER](images/state-channels-visibility/Configuration_Channels_Details_Overview_AFTER.png)
+
+### `Configuration > Channels > Details > Systems > Inheriting Systems`
```
moio, line 158:
```suggestion
### `Configuration > Channels > Details > Systems > Inherited Systems`
```

Please check with native English speakers if "inherited" is a good term here - we might want to prefer "indirectly assigned" or something similar. I just do not feel qualified enough to be authoritative.

accepted/00076-state-channels-visibility.md
```diff
@@ -0,0 +1,218 @@
...
+This is a **new** page where systems are listed only if not subscribed to the channel directly, but inheriting it through Organization or System Group assignment. The list will contain two columns: one for the system name and one for the source the system got the channel inherited from.
```
moio, line 159:
```suggestion
This is a **new** page where systems are listed only if not subscribed to a given channel directly, but inheriting it through an Organization or System Group assignment. The list will contain two columns: one for the system name and one for the source the system got the channel inherited from.
```

accepted/00076-state-channels-visibility.md
```diff
@@ -0,0 +1,218 @@
...
+## What other designs/options have been considered?
+
+Without investigating deeply, the alternative would be redesigning the feature completely from both the design and technical perspectives: the technical refactoring should get rid of the struts stack and redefine the concept of Configuration Channel and State Channel used by traditional or salt minion clients. By addressing the technical part, also the workflow could be re-drawn, dropping some pages and instead of adding separated tabs of tables it would be preferrable to have unified modern lists of filterable entries. The drawback of this alternative is that it would take much longer (from 3 to 5 times roughly) than the proposed official solution.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+## What are the unknowns?
+1. Salt States coming from Formulas are not yet considered as part of this RFC
```
moio, line 214: At this point, this is not an unknown, it should rather be presented as a known limitation of the proposed approach.

accepted/00076-state-channels-visibility.md
```diff
@@ -0,0 +1,218 @@
...
+2. This RFC is considering presenting all the sources of the Salt States for a given system. The RFC is **not** considering instead to implement a reverse-engineering to pick every Salt State from a given HighState and build the history of the specific source where it came from. This RFC is also **not** about tracing each line in the highstate to one of the states assigned to a system directly and/or indirectly from their sources.
```
moio, line 215: In my understanding this is covered by the "Limitation" paragraph above and could be removed.

moio (approved):
Substance is good FMPOV, thanks for this contribution!

I took the liberty to fine tune the language here and there, more to provide feedback to improve upcoming proposals than out of a real need to perfect this one. Feel free to accept completely or in part, or take inspiration as you see fit.

PR #50, opened 2020-12-18
Maintenance Windows Web Calendar

parlt91:


paususe:
Why React Big Calendar or Full Calendar and not e. g. Toast UI React Calendar or any other widget?

https://github.com/nhn/toast-ui.react-calendar

BTW, the same NHN people seem to have other interesting widgets:
https://github.com/nhn/tui.date-picker
https://github.com/nhn/tui.time-picker
https://github.com/nhn/tui.chart
https://github.com/nhn/tui.grid
https://github.com/nhn/tui.editor/tree/master/apps/react-editor
https://github.com/nhn/tui.tree
https://github.com/nhn/toast-ui.react-image-editor

accepted/00078-maintenance-windows-web-calendar.md
```diff
@@ -0,0 +1,87 @@
...
+1. Head to `Schedule` -> `Maintenance Windows` -> `Calendars` -> `Create`
+2. Follow the steps to create a new calendar
+3. Go to `Schedules` to create a new maintenance schedule
+4. Use the information provided in the created calendars details view to pick a maintenance window
+5. Assign systems to the newly created maintenance schedule
+
+### What is the expected outcome?
+
+Replace the plain ical text in the calendar details and creation menu with a visual representation of the maintenance windows contained in the uploaded ical file.
+Use the created calendar view to also display information about a maintenance window in the maintenance schedule details and creation menu.
```
moio, line 31: Is it correct to say that this RFC is just about the visualization of maintenance windows, and that the creation will still happen via upload of an iCal file?

accepted/00078-maintenance-windows-web-calendar.md
```diff
@@ -0,0 +1,87 @@
...
+Use the created calendar view to also display information about a maintenance window in the maintenance schedule details and creation menu. 
+
+# Detailed design
+[design]: #detailed-design
+
+### UI
+
+There seem to be two solutions suited for our use case:
+1. [React Big Calendar](https://github.com/jquense/react-big-calendar)
+2. [Full Calendar](https://github.com/fullcalendar/fullcalendar)
```
moio, line 40: Since we have to make a choice at some point (between these two and perhaps others as @paususe also suggested) I think it would be good to start exploring pros/cons of each at this stage. Of course hacking some quick experiments is a perfectly valid way to explore.

Ideally, we would have a list of pros and cons for each alternative here. Some relevant criteria that come to my mind:
  - **risk**: how easy will it be to swap a given library with a different one later on? If many options exist and switching from one to the other is cheap, following questions are less important and a choice can be made with less detailed data, as in case of trouble we can always switch. On the other end of the spectrum, if changing our minds later will require substantial work, it becomes more important to try to get it right from the start
  - **fitness to purpose**: do solutions have all the functionality we need out of the box or not? If not, "how much" do we have to add to which solutions?
 - **integration burden**: are all solutions equally easy to integrate in the current codebase, or do some require more work than others (frontend/backend)?
 - **long-term maintenance burden**: which solutions are expected to cause less problems down the road?
   - how big is the codebase (we will have to dig through in case of L3s)? How big is the dependency tree?
   - how active is the project? (eg. are there signs of recent activity? How big is the the community? Is the codebase de facto controlled by one single company? How receptive is the project about new issues or pull requests?)
   - how stable is the project? (do we have any idea if the API changes often, or if we can expect "drop-in" updates in the future?)
   - how good is the codebase? (does it cause shivers to read a random file? are there tests? is it documented well?)

I understand I am asking you to collect quite some data points to try to answer those questions, and that takes time. Nevertheless I think such time is well spent and will definitely pay off!

moio (commented):
Commented inline, including @paususe's point above

accepted/00078-maintenance-windows-web-calendar.md
paususe: Please also consider future/nice to have features, e. g. creation of calendars from Uyuni itself (stored in Uyuni DB or git or timezone selection (show current calendar in a different timezone).

accepted/00078-maintenance-windows-web-calendar.md
parlt91: For now it will. Current scope is to only change the way Maintenance schedules/calendar are displayed to the user. The creation is something that can be added in the future. I put into consideration if the investigated component option would be suitable for that as well.

accepted/00078-maintenance-windows-web-calendar.md
parlt91: Thank you for your valuable feedback. This is my first time working on an rfc so I expected that there will be some rough edges. I tried out and evaluated each of the 3 options (including the one mentioned by Pau) using your suggestions and updated the rfc with my findings.

accepted/00078-maintenance-windows-web-calendar.md
```diff
@@ -0,0 +1,181 @@
...
+From my point of view 1 is out. 3 would be the preferred way, but 2 is faster to implement and probably good enough.
+
+# Alternatives
+[alternatives]: #alternatives
+
+There seem to be a couple of solutions potentially suited for our use case:
+
+1. [Full Calendar](https://github.com/fullcalendar/fullcalendar)
+2. [React Big Calendar](https://github.com/jquense/react-big-calendar)
+3. [TOAST UI Calendar for React](https://github.com/nhn/toast-ui.react-calendar)
```
moio, line 71: And now the million-dollar question: as a designer and given the evidence you've accumulated out of the aspects you've evaluated, which solution would you pick - aka which criteria do you judge most important and which of the above projects satisfies them best?

That choice, and especially its motivation, could well represent the biggest value of this RFC!  

As a reviewer of course I have my idea but would like to hear your first, without influencing it too much. In the end, it will be you and your colleagues implementing and maintaining this, so your point of view is surely the richest.

Once the RFC contains the choice, if anybody has any concerns over the particular pick, the criteria or the research he/she can point out any doubt areas so that an informed choice is ultimately made.

Note that there will always be some risks and unknowns - there is no correct answer strictly speaking - unless we complete a full implementation of all possible options and run them all for a few years, we will never have perfect information! But that's fine, and that's why we like having research sessions, RFCs and constructive design conversations - to reduce some risks via an appropriately sized upfront time investment.

That's why we need (to form) good designers, that's what designs are for.

:+1: for all the work done so far

moio (approved):

Etheryte:
As we discussed in the morning call, this looks good to me.

The only thing I'm unsure about is using the iCal format as the API exchange format in the long run, but that's more of an implementation detail.

Etheryte (approved):

accepted/00078-maintenance-windows-web-calendar.md
```diff
@@ -0,0 +1,185 @@
...
+### Getting data from the backend
+
+A new function to calculate the maintenance windows contained in a maintenance Calendar needs to be implemented. It looks like `calculateUpcomingMaintenanceWindows` in `IcalUtils` can be reused/adapted in this case.
+
+I see 3 ways of calculating the maintenance windows displayed in the calendar
+1. Stick to the current way and calculate the next x (e.g. 10) maintenance windows
+2. Calculate all the maintenance windows within a given period (e.g. one year) and limit the months displayed in the calendar.
+3. Make a async call to the backend to retrieve all the maintenance windows of a month when clicking the `Back` and `Next` buttons. (basically server side pagination)
+
+From my point of view 1 is out. 3 would be the preferred way, but 2 is faster to implement and probably good enough.
```
ncounter, line 62: I agree that 1 is out of the game. I'd still prefer 3 even if it gets a bit harder to achieve, because a fixed period will never satisfy all, and there will always be the corner case where one will have problems: say I have a maintenance window of 30 minutes every day and you fix the given period to 1 year, you'll have to fetch 365 maintenance windows and prepare them to be displayed. I believe a single async fetch of the related data view should be the best approach.

Etheryte: I agree with Dario, and this is in part why I brought up the ical format above: ical is nice if people want to import it into their own calendar apps etc, but as an API endpoint format, it doesn't really cover our needs well.

accepted/00078-maintenance-windows-web-calendar.md
cbbayburt: Thanks, @parlt91. The comparison is really helpful

cbbayburt (approved):
LGTM :+1:

We should also think about user roles / access to calendars and user timezone preferences from the beginning

ncounter (approved):

accepted/00078-maintenance-windows-web-calendar.md
```diff
@@ -0,0 +1,185 @@
...
+The calendar should have a `read-only` mode to prevent manipulation of displayed events but also support it for potential future use.
+
+### Getting data from the backend
+
+A new function to calculate the maintenance windows contained in a maintenance Calendar needs to be implemented. It looks like `calculateUpcomingMaintenanceWindows` in `IcalUtils` can be reused/adapted in this case.
+
+I see 3 ways of calculating the maintenance windows displayed in the calendar
+1. Stick to the current way and calculate the next x (e.g. 10) maintenance windows
+2. Calculate all the maintenance windows within a given period (e.g. one year) and limit the months displayed in the calendar.
+3. Make a async call to the backend to retrieve all the maintenance windows of a month when clicking the `Back` and `Next` buttons. (basically server side pagination)
```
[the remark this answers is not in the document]
parlt91: I agree and added a comment to the rfc

accepted/00078-maintenance-windows-web-calendar.md
parlt91: Agreed. Let's go with option 3. I updated the RFC accordingly

hustodemon (approved):
Left a couple of suggestions about the terminology (might be wrong about them, so please check them) and some other small comments. Otherwise it looks great to me, a big :+1: !

accepted/00078-maintenance-windows-web-calendar.md
```diff
@@ -0,0 +1,204 @@
...
+
+It allows to create custom buttons to be displayed in the header toolbar to allow navigating between dates/months containing maintenance window relevant data.
+
+Supports using the calendar in `read-only` mode
+
+Coloring events differently is supported as well.
+
+Supports displaying events in different timezones.
+
+As a bonus Full Calendar is able to read its events directly from ical files. This allows us to potentially cut back on backend work.
```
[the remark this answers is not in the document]
parlt91: Right in the end we decided against this. See the conclusion below (you probably already read it after creating the comment)

accepted/00078-maintenance-windows-web-calendar.md
```diff
@@ -0,0 +1,204 @@
...
+Generally switching between calendar components once settled for one won't be straitforward. Whilst the backend part should be salvageable between multiple components the frontend obviously won't. The components presented have complety different APIs, hence switching calendar components will require the frontend part to be reimplemented. 
+
+# Future Improvements
+The current scope includes only the display of maintenance schedules/calendars. In the future we might expand this to support the creation of Maintenance Windows within our webui. All off the options discussed above support the creation of events.
+Reworking the Maintenance Window picker (part of the Action picker if a maintenance window is assigned) will require our struts/jsp action pickers to be migrated to react first and thus will be handled separately.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should the display of the next 10 upcoming MWs be replaced or displayed alongside the calendar?
```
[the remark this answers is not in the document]
parlt91: Maybe not necessary but, it would be nice to have. If there are many months between the Maintenance Windows. I even thought about adding them in addition the the regular previous and next buttons. But we'll need to see during actual implementation how it work out.

parlt91:
@paususe Could you have a look at the RFC again? Are there any unresolved questions for you?

paususe (approved):
LGTM. Thank you and congratulations on your first approved RFC! :+1:

PR #51, opened 2021-01-22
DB based JSON pillar storage proposal

aaannz:
Render [available here](https://github.com/uyuni-project/uyuni-rfc/blob/json_pillars/accepted/00080-jsonb-pillar-db-storage.md)


accepted/00000-jsonb-pillar-db-storage.md
```diff
@@ -0,0 +1,153 @@
...
+# Summary
+[summary]: #summary
+
+Use database as a storage for Uyuni salt pillars instead of plain files.
+
+# Motivation
+[motivation]: #motivation
+
+Pillars are used for passing reliable and trustworthy data to salt clients. Salt integration into Uyuni generates pillar files which are then loaded to the salt ecosystem by Uyuni external pillar.
+Advantage of this solution is rather easy implementation and quick visibility of the data when inspecting in filesystem, however there are disadvantages to consider. Data consistency [issue#5678](https://github.com/SUSE/spacewalk/issues/5678) and [issue#10679](https://github.com/SUSE/spacewalk/issues/10679), [filesystem performance](https://github.com/SUSE/spacewalk/pull/10391#discussion_r375147529), backup procedure must consider these files as well.
```
mcalmer, line 14: The file approach was choosen as the original plan was, to use salt-syndic on the Proxies. One requirement of syndic is, that all states and pillars are availabel on the syndic. Transfer files or use a network storage would be required for states anyway. So pillar as files was just the same problem and easier then opening database ports to the world and dealing with a second access way.

As the syndic story seems not to become reality, putting pillar into the DB can make things easier in some areas.

paususe (approved):
LGTM

accepted/00000-jsonb-pillar-db-storage.md
```diff
@@ -0,0 +1,153 @@
...
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we **not** do this?
+
+* will it impact performance?
+  
+  TBD, I suspect better performance then with thousands of files in one directory
```
moio, line 132: Is it possible to test this? In particular, I wonder what happens if a lot of pillars are requested in a short amount of time (eg. highstate application on many minions concurrently). Will each pillar request need an opening of a new DB connection, or are those pooled somehow? Is there a caching mechanism of any sort?

aaannz: I don't think there is caching of pillars for highstate and other states applications (see [in-memory vs on-demand pillars](https://docs.saltproject.io/en/latest/topics/pillar/#in-memory-pillar-data-vs-on-demand-pillar-data)). Also I didn't see any batching in the code of external pillar and seems it is all designed for per-minion queries.
As I proposed in the external pillar settings, it is 3 DB queries for a minion so in case of e.g. 10000 minions, we will get 30000 queries concurrently + other queries resulting from state applications.

Compare to current state of `suma-minion.py` external pillar. That means 10000 executions of external pillar, accessing from 2 to many files (formula, image, virtualization, custom info pillars).

To be fair, in the initial implementation I plan to migrate pillars generated by `MinionPillarManager` and `SaltStateGeneratorService::generateImagePillar`. This does not cover formulas yet as there it is a little more work. So `suma-minion.py` will need to stay for time being until formulas are migrated as well.

As for testing, I should be able to gather some performance data using evil minions. I will look into it.

aaannz: I did a quick test with evil minions:

2 regular minions, 100 evil minions

Command:
time bash -c 'for i in `seq 1 100`;do salt "*" saltutil.pillar_refresh > /dev/null; done'

* for 102 minions and suma-minion.py only:

```
real    4m37,656s
user    1m53,964s
sys     0m10,080s
```


* for 102 minions and psql pillar only:

```
real    4m39,619s
user    1m54,091s
sys     0m10,101s
```


* for 102 minions and both suma-minion.py and psql pillar:

```
real    4m40,897s
user    1m52,634s
sys     0m10,143s
```

I was monitoring CPU load on the server and majority was taken by salt, (~30%). For postgresql this wasn't difficult at all steadily remaining under 1%.

accepted/00000-jsonb-pillar-db-storage.md
```diff
@@ -0,0 +1,219 @@
...
+
+* when only DB pillar was active:
+
+```
+real    3m21,272s
+user    0m39,661s
+sys     0m2,320s
+```
+
+In conclusion I do not see very notable impact in this isolated scenario.
```
moio, line 191: That's excellent. One important point: how many Postgres connections were used during these tests?

It would be a problem if the answer is "one per concurrent minion access".

accepted/00000-jsonb-pillar-db-storage.md
```diff
@@ -0,0 +1,219 @@
...
+postgres:
+  db: susemanager
+  host: localhost
+  pass: spacewalk
+  port: 5432
+  user: spacewalk
+
+ext_pillar:
+  - postgres:
+      - query: "SELECT pillar from suseSaltPillars WHERE target = '*'"
```
rjmateus, line 63: It would have some performance impact have just one query which collects all data instead of these 3 separated queries?
How will the merge happen in the case we use just one query?

accepted/00000-jsonb-pillar-db-storage.md
```diff
@@ -0,0 +1,219 @@
...
+* will the solution be hard to maintain in the future?
+  
+  I expect easier maintaining then with file generators.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Use single pillar row for each minion and without `category` column.
+
+This was the original idea how to implement this, however during prototype implementation having separate rows for separate pillar data made implementation better match how it is done now. I am not opposed using single row per minion, but I also do not see disadvantages for using multiple rows per minion.
```
rjmateus, line 211: IMHO One disadvantage of this alternative approach it will be harder to update the json data in the database

rjmateus (approved):

mcalmer (approved):

accepted/00000-jsonb-pillar-db-storage.md
aaannz: I fear that this is indeed the case. SQL external pillar itself doesn't have any caching or maintaining of single connection. Instead it for each pillar request not handled by salt pillar in-memory cache opens a new connection.
However I did not see failures due to reaching max connections. Postgres configured to 400 max, having 1001 minions.

moio: This is a bit strange. One possibility is that the requests were done quicker than the connections were exhausted - but still this is an important potential problems, as we know Apache and possibly Taskomatic will have problems if they starve out of connections.

Could you try again and really monitor the number of open TCP sockets, to see what is actually going on?

How difficult would it be to solve this on the Salt side? Another possibility would be to add a whole-system connection pooler, which is an option we had thought about before but is kind of yet-another-separate-effort.

aaannz: I will redo the test. I suppose that salt worker threads limits the workload here - by default we have 8 so I suppose 8 threads are handling `pillar_refresh` events and thus even though there will be new connection per minion only `worker_threads`# concurrently.

> How difficult would it be to solve this on the Salt side? Another possibility would be to add a whole-system connection pooler, which is an option we had thought about before but is kind of yet-another-separate-effort.

I am not sure, I don't have enough knowledge about implementing external pillars. I will look into other pillar modules, maybe it is already solved somewhere else.

aaannz: I redone the tests, noticed that `saltutil.refresh_pillars` do not open the connection to database, which is surprising to me, however `pillar.items` does as it ignores inmemory cache by design. So I redone all tests using `pillar.items`.

Wrt. connections I saw:
tcp: 2 - 14 (SUMA was shutdown, only salt master was running), 
`numbackends`: 2 - 7  (from `pg_stat_database`)

Hypothesis is that it is indeed limited to configured salt worker threads.

I increased salt `worker_threads` from 8 to 28 and I saw increase of tcp connections to postgres and `numbackends` as well:
tcp: 2 - 44
`numbackends`: 2 - 14

I will update RFC with these results shortly.

accepted/00000-jsonb-pillar-db-storage.md
moio: Would something like this work?

```sql
SELECT pillar from suseSaltPillars WHERE target = '*'
  UNION (
    SELECT pillar from suseSaltPillars WHERE target = (SELECT CONCAT('org_id:', org_id::VARCHAR) FROM rhnServer AS S LEFT JOIN suseMinionInfo AS M on S.id = M.server_id WHERE M.minion_id = %s)
  )
  UNION (
     ...
  )
...
```

accepted/00000-jsonb-pillar-db-storage.md
```diff
@@ -0,0 +1,225 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Pillars are used for passing reliable and trustworthy data to salt clients. Salt integration into Uyuni generates pillar files which are then loaded to the salt ecosystem by Uyuni external pillar.
+Advantage of this solution is rather easy implementation and quick visibility of the data when inspecting in filesystem, however there are disadvantages to consider. Data consistency [issue#5678](https://github.com/SUSE/spacewalk/issues/5678) and [issue#10679](https://github.com/SUSE/spacewalk/issues/10679), [filesystem performance](https://github.com/SUSE/spacewalk/pull/10391#discussion_r375147529), backup procedure must consider these files as well.
+
+This RFC proposes to instead of plain files, database based storage should be used for all pillars produced by Uyuni.
+
+This is expected to help with maintaining consistency, help with backup procedure and with interserver sync as well, potential performance improvement (TBD - filesystem access vs. database indexes).
```
moio, line 18: Do you intend to actually measure expected performance improvement over the base filesystem case as part of this RFC?

accepted/00000-jsonb-pillar-db-storage.md
```diff
@@ -0,0 +1,225 @@
...
+
+`target` column is mainly to contain `minion_id` of salt client, however to support universal and system group pillars I opted to use VARCHAR with limit of 256 (as is the limit of minion id, group name is limited to 64 characters). For universal pillar I propose to use `*` character.
+
+Added `category` VARCHAR is there to provide easer manipulation and understanding of stored data. Currently all Uyuni pillar data are written using `com.suse.manager.webui.services.pillar.MinionPillarManager` (assuming [PR#3093](https://github.com/uyuni-project/uyuni/pull/3093) is merged) and each type of pillars have their own respective [generator class](https://github.com/nadvornik/uyuni/tree/custominfo/java/code/src/com/suse/manager/webui/services/pillar). To maintain some form of distinction I propose this `category` column to allow to have multiple pillar entries per salt client.
```
moio, line 46: You might want to also add an index for the category - not necessary to specify here in the RFC, just a reminder for implementation time :wink:

accepted/00000-jsonb-pillar-db-storage.md
```diff
@@ -0,0 +1,225 @@
...
+                 # Should we and will we have a list at the end?
+                 if ((self.as_list and (ret[nk] in crd)) or
+```
+
+Tracked in [PR#59777](https://github.com/saltstack/salt/pull/59777).
+
+## Maintaining consistency
+
+Responsibility to provide up to date data remains, as is the case today, on the component changing the data. That means, when information about minion is changed, then relevant pillar generator must be called to refresh the pillar in database.
+However with the pillar in database and loosely (no foreign key relation) connected to minions and groups, database trigger which can remove pillar entries when minion or group is removed:
```
moio, line 107: What advantages and disadvantages you see with using triggers rather than enhancing Java code to maintain consistency when, e.g., a minion is removed from a group?

accepted/00000-jsonb-pillar-db-storage.md
```diff
@@ -0,0 +1,225 @@
...
+user    0m47,978s
+sys     0m3,124s
+```
+
+  Connections to postgres up to 44
+
+  Postgres backends up to 14
+
+
+In conclusion I do not see very notable impact in this isolated scenario. Number of opened connections is correlated with number of salt worker threads, this needs to be taking into account in case of tuning individual installations.
```
moio, line 193: Remark for implementation time: update "The LSD" https://github.com/uyuni-project/uyuni-docs/blob/master/modules/large-deployments/pages/tuning.adoc

I can help with that, just ping me when the time comes

moio (approved):
I like this proposal very much and once again I need to apologize for being late with my review.

accepted/00000-jsonb-pillar-db-storage.md
aaannz: Righ. The most I did is to do simple E2E performance test when I focused if there is noticable performance hit.
I reevaluated performance when I added one query option in form of postgres function.
Testing criteria are as follows:
* 1001 minions (one real, 1000 evil minions)
* 10 `pillar.items` repetitions for all minions
* 4000 rows in pillar database (1000 rows per organization, group, minion and global)
* regular Uyuni pillar files for one registered minion
* default Uyuni salt worker settings (8 worker threads)
* uptodate SUMA4.1 VM one machine (4vCPU, 4GiB RAM) and minions on real HW on different machine (8CPU, 24GiB RAM)

Command:

```bash
time bash -c 'for i in `seq 1 10`;do time salt "*" pillar.items >/dev/null;salt "*" test.ping > /dev/null;echo "== Round $i done"; done'       
```

And results are:
performance-pillars

These results may be skewed by the fact that with the default settings I was occasionally hitting salt timeouts on some evil-minions. But these timeouts appeared on all of methods tested.

The best results were for single query and not using our external `suma_minion` pillar, next best was combined 1-query and `suma_minion` pillar, which is interesting that it was in average better then just using `suma_minion` pillar.

accepted/00000-jsonb-pillar-db-storage.md
aaannz: Added to the proposal.

accepted/00000-jsonb-pillar-db-storage.md
moio: Excellent, please transcribe some of this in the RFC's text - just as a reminder to posterity that proper benchmarking was conducted :-)

accepted/00000-jsonb-pillar-db-storage.md
moio: I understand and agree about the strong consistency guarantee argument, although I am also reminded of this short story:

http://thecodelesscode.com/case/42

which is the reason I normally prefer code-level enforcing (where refactorings are typically eased by IDEs).

That said, I fully trust you on having the last word and ultimately decide on this aspect.

moio:
Excellent work @aaannz, looking forward to see this implemented!

aaannz:
Updated to include performance results in the rfc itself, renamed to 00080- prefix. All reviews accepted, merging the PR.

PR #52, opened 2021-02-12
Add Prometheus TLS support

witekest:


accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,238 @@
...
+Prometheus exporters in the coming months and the other projects, such as
+Prometheus, Alertmanager, Pushgateway [^5]. The functionality is implemented as
+part of Prometheus Exporter Toolkit [^11] and can be added to other exporters.
+
+The TLS support for SUSE Manager distributed unofficial exporters at the time of
+writing:
+
+* Apache exporter (no TLS support)
+* PostgreSQL exporter (no TLS support)
+* Exporter Exporter (TLS support provided)
```
paususe, line 108: Does this mean if I want to hide my TLS-enabled node exporter behind Exporter Exporter, it's not possible?

accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,238 @@
...
+implementing following default configuration:
+
+* Upgrade Prometheus to version >= 2.24.1.
+* Create and deploy CA signed Prometheus server certificate.
+* Configure `tls_server_config` using this certificate.
+* Configure basic authentication accounts to map SUSE Manager credentials.
+
+All these steps should be implemented in Salt prometheus-formula. Alternatively
+certificate generation and deployment could be moved to a new dedicated
+certificates formula.
```
paususe, line 61: That certificates formula could also come handy for the OpenVPN formula.

accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,238 @@
...
+```
+
+
+# Drawbacks
+
+[drawbacks]: #drawbacks
+
+One important drawback is the increased complexity of deployment. In
+particular changing client hostnames or renewing CA certificate will require
+updating the certificates on clients.
```
paususe, line 203: Maybe we can leverage the cert-formula for deployment? https://github.com/saltstack-formulas/cert-formula
Given Uyuni is a systems management solution, it should manage :-) (although I agree certificates are usually generated by different departments which may have their own means for distribution)

accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,238 @@
...
+described in official documentation [^3]. Basic authentication credentials are
+stored in the configuration file. Passwords are hashed with bcrypt.
+
+To follow current SUSE Manager security model [^4] we should consider
+implementing following default configuration:
+
+* Upgrade Prometheus to version >= 2.24.1.
+* Create and deploy CA signed Prometheus server certificate.
+* Configure `tls_server_config` using this certificate.
+* Configure basic authentication accounts to map SUSE Manager credentials.
```
moio, line 57: Can you clarify the scope of this RFC more exactly? Looking at the title, I had the impression this was just about encryption, but here you are talking about authentication, leaving me to wonder whether the intention is actually to provide fully secured communication channels. Please adapt the one or the other :wink:

accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,238 @@
+- Feature Name: Prometheus TLS support
+- Start Date: 2021-02-11
+- RFC PR:
+
+# Summary
+
+[summary]: #summary
+
+Secure connection to Prometheus server and exporters.
```
moio, line 9: Is the idea to provide optional encryption or mandatory?

accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,238 @@
...
+Prometheus 2.24.0 has introduced TLS encryption as well as basic and
+client-certificate authentication [^2]. All available configuration options are
+described in official documentation [^3]. Basic authentication credentials are
+stored in the configuration file. Passwords are hashed with bcrypt.
+
+To follow current SUSE Manager security model [^4] we should consider
+implementing following default configuration:
+
+* Upgrade Prometheus to version >= 2.24.1.
+* Create and deploy CA signed Prometheus server certificate.
```
moio, line 55: Are we contemplating the case of users bringing their own CA? That is a possibility for general Uyuni.

Do we intend to offer generation as the only possibility, or as an option, or we plan to add "bring your own CA" at a later point?

accepted/0000-prometheus-tls.md
witekest: It is possible. I've clarified the description.

accepted/0000-prometheus-tls.md
witekest: Added comment on that and the repository reference.

accepted/0000-prometheus-tls.md
witekest: Added to the references list.

accepted/0000-prometheus-tls.md
witekest: Agree. Updated the summary.

accepted/0000-prometheus-tls.md
witekest: Updated.

accepted/0000-prometheus-tls.md
witekest: I've reworded that and marked as optional step.

accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,255 @@
...
+Prometheus exporters in the coming months and the other projects, such as
+Prometheus, Alertmanager, Pushgateway [[5]](#5). The functionality is implemented as
+part of Prometheus Exporter Toolkit [[11]](#11) and can be added to other exporters.
+
+The state of TLS support for SUSE Manager distributed unofficial exporters at
+the time of writing:
+
+### TLS support provided
+
+* Exporter Exporter
```
mcalmer, line 109: To make things easier: 

Why not define that we only support "Exporter Exporter" with TLS and all other exporters are just configured localhost and served via ExEx?

I think this would solve all problems with has TLS support or not.
I am right, or did I miss something?

accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,255 @@
...
+particular changing client hostnames or renewing CA certificate will require
+updating the certificates on clients.
+
+# Alternatives
+
+[alternatives]: #alternatives
+
+One alternative would be to outsource certificates generation and let users
+manage them. Then we would only need to provide configuration options to
+point to user generated certificates.
```
mcalmer, line 221: We need to provide this option anyway. We have a lot of customers with own PKIs and they need to have a way to just define CA, Server Cert, Server Key and this gets deployed and used.

An own mechanism will make it easy to use, but I think 50% have to take the other way.

mcalmer (commented):
I would also suggest to use a different CA for monitoring then for SUMA Server/Proxy.
Possible would be, to use a sub ca only used for monitoring.

witekest:
> I would also suggest to use a different CA for monitoring then for SUMA Server/Proxy.
> Possible would be, to use a sub ca only used for monitoring.

What would be the benefits of such approach?

accepted/0000-prometheus-tls.md
witekest: _Exporter Exporter_ is not the standard way or running Prometheus exporters. I can imagine some customers being reluctant to using it.

On the other hand TLS support in exporters is getting a well established standard. Also contributing to the upstream code is pretty straight forward using the provided toolkit library.

accepted/0000-prometheus-tls.md
witekest: Agree, I've listed certificates generation step as optional. The alternative described here would mean to drop that step completely and ask users to care about certificates.

paususe:
> > I would also suggest to use a different CA for monitoring then for SUMA Server/Proxy.
> > Possible would be, to use a sub ca only used for monitoring.
> 
> What would be the benefits of such approach?

It's common in enterprise environments.

The main IT organization owns the root CA, and then they create sub-CAs for different departments, uses, etc. That allows them to invalidate a sub-CA (e. g. in case of a security penetration, or some department being carved out from the parent company, etc) without invalidating the whole root CA.

witekest:
> I would also suggest to use a different CA for monitoring then for SUMA Server/Proxy.
> Possible would be, to use a sub ca only used for monitoring.

Added new optional step for creating intermediate monitoring CA.

mcalmer (approved):

accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,265 @@
...
+stored in the configuration file. Passwords are hashed with _bcrypt_.
+
+To follow current SUSE Manager security model [[4]](#4) we should consider
+implementing following default configuration:
+
+* Upgrade Prometheus to version >= 2.24.1.
+* [Optional] Create intermediate certificate authority (CA) dedicated for monitoring.
+* [Optional] Create and deploy Prometheus server certificate signed by this CA.
+* Configure `tls_server_config` using this certificate.
+* Configure basic authentication accounts to map SUSE Manager credentials.
```
moio, line 58: Does this mean that a static configuration file with all credential pairs would be generated at formula application time, based on accounts known from the DB, and regenerated at highstate time?

witekest: Thanks for catching this. I think that would be too complicated and error prone. I've changed that point and proposed setting up a dedicated administrator account.

paususe (approved):

accepted/0000-prometheus-tls.md
```diff
@@ -0,0 +1,268 @@
...
+It would involve following configuration steps:
+
+* [Optional] Create and deploy CA signed server certificates for all monitored minions.
+* Configure `tls_server_config` in node exporter using these certificates.
+* Configure `tls_config.ca_file` in Prometheus job configuration to use server
+  certificate [[7]](#7).
+* Configure `tls_config` in Prometheus job configuration to use Prometheus
+  server certificate and key files for client cert authentication.
+
+Alternatively to last point basic authentication can be configured. Please note
```
moio, line 86: Do we plan to support cert authentication, basic authentication or both when it comes to the node exporter?

moio (approved):
LGTM, I did not see evident flaws.

That said I am not a great experts in SSL, and trust @mcalmer's review on the certificate handling aspects!

Thanks @witekest

accepted/0000-prometheus-tls.md
witekest: The plan is to go with cert authentication for _node exporter_. Basic authentication is mentioned as an alternative which we could implement if it turns out that it's requested by users or if handling of certificates turns out to be too complex.

PR #53, opened 2021-02-26
Ansible Integration

meaksh:
Read here: https://github.com/uyuni-project/uyuni-rfc/blob/master-ansible-gate/accepted/00000-ansible-integration.md


accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,255 @@
...
+NOTE: The profile for a system with registered in Uyuni as "Foreign/ANSIBLE" does not look like as the one Salt/Traditional system, since most of the Uyuni features (channels, software profile, etc) are not implemented for neither "Foreign" nor "ANSIBLE" entitlements. In order to get the entire portfolio of Uyuni features for Salt minion, this "Foreign/ANSIBLE" systems would need to be transitioned to "Salt" minion.
+
+Notice also that the Ansible controller is not listed as systems list page, since there is not system entry for it, it would be shown in the respective new UI page for managing Ansible Controller.
+
+#### New UI tabs & pages
+
+* Ansible
+* Ansible / Ansible Controllers list
+* Ansible / Add new Ansible controller
+* Ansible / View Ansible controller information & and visualize playbooks (would also allow triggering playbooks via "salt-ssh" in the controller)
```
paususe, line 108: Instead of adding a top-level "Ansible", what about renaming the current "Salt" to "Automation", and then having "Salt" and "Ansible" as children?

Something like this but well-implemented:
image

We can even think of putting "Remote commands" at the same level as "Ansible" and "Salt", since that will be possible with both Ansible and Salt.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,255 @@
...
+[alternatives]: #alternatives
+
+* Interfacing AWX.
+
+An alternative would be to use the API of AWX. While this would mean that all options of AWX would be available, it would also mean that we would need to make sure to always stay compatible with all versions used by users, breaking changes could be introduced by AWX and some features might even need to be implemented in AWX first before being able to use them in Uyuni.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+* Where should we move from here? Full integration of Ansible features? Moving Ansible integration to Spacewalk core? Fully interfacing AWX?
```
paususe, line 255: Objective of this RFC is only running Ansible playbooks form the Uyuni Server. We are not trying to replace Salt as the foundation of Uyuni but only adding Ansible as a sidecar.

In fact, I like that name for this feature: "Ansible sidecar" :-D

paususe (changes requested):
Some questions added.

I think there is another case, which might be even easier to implement, and is not explicitly tackled by this RFC: running Ansible playbooks directly from the Uyuni Server, i. e. not integrating with an existing Ansible controller but just providing a UI to AnsibleGate. That looks like a common case, where your systems are managed as Salt minions (or salt-ssh), and now you get some Ansible playbook to set up some application, secure it or whatever. You do not really want all this complex integration with a controler, or AWX or anything fancy: you just want to run your Ansible playbook on a Salt minion.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,255 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+There are two main motivations for this:
+
+1. A user might have some investment in Ansible in the past but wants to switch to Uyuni now.
+This RFC would offer a transition path for that. The user can import the Ansible systems, start with the already existing playbooks, get familiar with Uyuni and then switch over. It is also be possible to manage clients with Salt and Ansible in parallel.
+2. Managing parts of the infrastructure in Ansible.
+If it is not possible or not wanted to move everything to Uyuni, it is possible to just keep using Ansible for the few clients that cannot be transitioned.
```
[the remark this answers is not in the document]
meaksh: Ok, it's "move everything to Salt". fixing it

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,255 @@
...
+Of course, we need some new tables in the database, at least something like:
+
+```
+"suseAnsibleController" (label, type, org_1)
+"suseAnsibleControllerConfig" (controller_id, parameter, value)
+"suseAnsibleControllerSystem" (system_id, controller_id)
+"suseAnsibleControllerPlaybook" (controller_id, local_path, content?)
+```
+
+And also create the new "ANSIBLE" entitlement, which should be an addon-entitlement compatible with Salt, Foreign and maybe also traditional entitlements. Here is the [SQL script](https://github.com/meaksh/uyuni-hacks/blob/master/scripts/ansible/add_ansible_entitleme.sql) used for PoC demo. It's also needed to adapt the rhn databaste functions implemented in the DB.
```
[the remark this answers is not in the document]
meaksh: Yes, exactly. I've clarified this a bit more.

accepted/00000-ansible-integration.md
meaksh: Hmmm, that sounds like a good idea, but at the same time not sure if that would create the impression that you can do on Ansible mostly what you can do with Salt.

paususe: Who knows where we will arrive. Especially if the community weighs in.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,293 @@
...
+
+Assuming the Ansible inventory is properly created and SSH keys are deployed, the user can operate it from the command line, but Uyuni can also operate it via Salt runner call (since we're on the Uyuni server and not in a minion). Example:
+
+```console
+# salt-run salt.cmd ansible.playbooks name=example_playbook.yml
+```
+
+This means, playbooks can be triggered via CLI by the user, but also via "salt-api" so Uyuni is also easily able to trigger playbooks executions.
+
+Once we have a "Playbook catalog" in the UI, and Uyuni would trigger playbook executions we would need to implemente of course a new type of action: `ApplyPlaybook", which should expose the playbook to the Salt file_roots so it's available for Ansible when running.
```
agraul, line 49: Found a little typo
```suggestion
Once we have a "Playbook catalog" in the UI, and Uyuni would trigger playbook executions we would need to implemente of course a new type of action: `ApplyPlaybook`, which should expose the playbook to the Salt file_roots so it's available for Ansible when running.
```

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,293 @@
...
+This RFC proposes how to integrate and manage your existing Ansible nodes in Uyuni with the help of Salt and its "Ansible-Gate" module. This means that you would be able to import Ansible inventories, their systems and also apply playbooks to these nodes using Uyuni.
+
+
+# Motivation
+[motivation]: #motivation
+
+There are two main motivations for this:
+
+1. A user might have some investment in Ansible in the past but wants to switch to Uyuni now.
+This RFC would offer a transition path for that. The user can import the Ansible systems, start with the already existing playbooks, get familiar with Uyuni and then switch over. It is also be possible to manage clients with Salt and Ansible in parallel.
```
agraul, line 17:
```suggestion
This RFC would offer a transition path for that. The user can import the Ansible systems, start with the already existing playbooks, get familiar with Uyuni and then switch over. It is also possible to manage clients with Salt and Ansible in parallel.
```

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,293 @@
...
+
+
+2. Adding external Ansible controllers:
+
+The premise here is that there is already an existing Ansible infrastructure somewhere and we want to import it into Uyuni. here we're **not** primarely focused on building a new Ansible managed infrastructure from scratch using Uyuni.
+
+Therefore, there might be more than one Ansible controller host, or even the Uyuni server could act at some point as an Ansible controller. Since data sources, and type of sources (i.a. single host, AWX api) might be multiple, this RFC proposes an approach similarly to what Uyuni does for handling "Virtual Host Managers (VHM)". This means, using a Python tool, in this case called something like "ansible-gatherer", which is plugin-based (so easily allow implementing different sources of Ansible inventores). This tool would be called via an Uyuni Java schedule, like the "virtual-host-gatherer", passing the necessary information (parameters, type of host, etc) to reach the Ansible controller and collect the necessary information.
```
agraul, line 56:
```suggestion
Therefore, there might be more than one Ansible controller host, or even the Uyuni server could act at some point as an Ansible controller. Since data sources, and type of sources (e.g. single host, AWX API) might be multiple, this RFC proposes an approach similarly to what Uyuni does for handling "Virtual Host Managers (VHM)". This means, using a Python tool, in this case called something like "ansible-gatherer", which is plugin-based (so easily allows implementing different sources of Ansible inventories). This tool would be called via an Uyuni Java schedule, like the "virtual-host-gatherer", passing the necessary information (parameters, type of host, etc) to reach the Ansible controller and collect the necessary information.
```

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,293 @@
...
+      state: started
+```
+
+As you can see, this playbook is defining different tasks to execute on different "hosts" groups. Those groups are of course defined in the Ansible inventory (default at /etc/ansible/hosts).
+
+The execution of a given playbooks is done in the corresponding Ansible controller, which contains the inventory and the hosts definition.
+
+The playbooks available to apply on each Ansible controller, are the ones exposed by "ansible-gatherer" and ultimately, in the future, playbooks created and maintained in Uyuni (via future Ansible playbook catalog) that would be then pushed to the controller node.
+
+At the time of running a playbook from the Ansible controller, Uyuni is able to do it by reaching the controller and operating it using "salt-ssh". There are different approaches here:
```
agraul, line 173: Why is operating Ansible the Ansible Controller node limited to `salt-ssh`? In other words, why is `salt my-ansible-controller ...` not an alternative?

accepted/00000-ansible-integration.md
meaksh: I've added this suggestion to the RFC

accepted/00000-ansible-integration.md
meaksh: Thanks!

accepted/00000-ansible-integration.md
meaksh: Thanks for the feedback @agraul ! :+1:

I've added few notes that I think clarify this part of the text a little bit.

I mentioned using "salt-ssh" because at this point, it was not required that the ansible controller is a registered system, so in this case, Uyuni would use the "ssh" credentials provided to "ansible-gatherer" in order to reach the controller.

Of course, if the Ansible controller is a registered system, we do not need to do a Salt SSH, simply trigger the playbook execution via normal Salt minion job.

paususe: @meaksh just to clarify: we are considering the case where the Uyuni Server would be the Ansible controller, right? Somehow we need to make this case easy to set up (especially considering the Uyuni Server cannot be managed by itself, so entitlements or formulas are not an option).

meaksh: @paususe yes, I've added some notes here.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,304 @@
...
+Of course, we need some new tables in the database, at least something like:
+
+```
+"suseAnsibleController" (label, type, org_1)
+"suseAnsibleControllerConfig" (controller_id, parameter, value)
+"suseAnsibleControllerSystem" (system_id, controller_id)
+"suseAnsibleControllerPlaybook" (controller_id, local_path, content?)
+```
+
+And also create the new "ANSIBLE" entitlement, which should be an addon-entitlement compatible with Salt, Foreign and maybe also for a potential "Management/ANSIBLE" (traditional client) entitlement. Here is the [SQL script](https://github.com/meaksh/uyuni-hacks/blob/master/scripts/ansible/add_ansible_entitleme.sql) used for PoC demo. It's also needed to adapt the rhn databaste functions implemented in the DB.
```
moio, line 114: Is there any particular reason to contemplate the combination of Ansible with the traditional stack?

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,304 @@
...
+
+The option 1 is refers to executing a playbook which is already stored and maintained in the Ansible controller. On the other hand, option 2 would work in an scenario where the maintenance of the Playbooks are done in Uyuni, so we need to make the Ansible controller to get the generated playbook by Uyuni from the Salt file roots.
+
+The SSH credentials to use in both cases to reach the Ansible controller would be the same that "ansible-gatherer" is using.
+
+In case we allow "ansible-gatherer" to add an Ansible controller from a registered system, then playbooks execution can be triggered to the controller either using "salt-ssh" or via normal Salt minion job, depending on how the system is registered in Uyuni.
+
+
+### Execute Salt commands & states to Ansible systems
+The "ansiblegate" module of Salt also allows "salt-ssh" to reuse an Ansible inventory to reach those systems that are being managed by Ansible and execute Salt commands on it.
```
moio, line 210: Is ansiblegate limited to a plain-file inventory or can it consume any inventory from a plugin?

https://docs.ansible.com/ansible/latest/user_guide/intro_dynamic_inventory.html

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,304 @@
...
+  * will it impact performance?
+  * what other parts of the product will be affected?
+  * will the solution be hard to maintain in the future?
+
+# Alternatives
+[alternatives]: #alternatives
+
+* Interfacing AWX.
+
+An alternative would be to use the API of AWX. While this would mean that all options of AWX would be available, it would also mean that we would need to make sure to always stay compatible with all versions used by users, breaking changes could be introduced by AWX and some features might even need to be implemented in AWX first before being able to use them in Uyuni.
```
moio, line 297: Do we have any data about the prevalence of AWX on the total number of Ansible users? If any significant portion uses bare Ansible, then we would also lose that part of the audience.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,304 @@
...
+As you can see, besides of reporting the hosts from the Ansible inventory, the idea is that we also stored it locally in the Uyuni server in order to being able to reuse it later to target those hosts with "salt-ssh" directly from the Uyuni server.
+
+In short, we would gather, for instance:
+- Ansible inventory (hosts)
+- Ansible SSH keys referenced in the inventory
+- Playbooks under specified "remote playbook path" (to be displayed in the UI - readonly mode - not editable)
+
+This would be stored under `/var/lib/spacewalk/ansible/inventory-label/` path in the Uyuni server. At the time of processing the inventory file by "ansible-gatherer", it needs to be tailored to adapt the path of the SSH keys to the local path in the Uyuni server. As done in script from Ansible Integration PoC [here](https://github.com/meaksh/uyuni-hacks/blob/master/scripts/ansible/import_systems_from_ansible_controller.py).
+
+With this information, Java is able to proceed creating these systems in Uyuni, as "Foreign / ANSIBLE" entitled systems.
```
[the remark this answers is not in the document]
meaksh: The idea if caching the inventory in the Uyuni server, in a file (or DB), was proposed in order to be able to reuse it if we want to trigger `salt-ssh --roster=ansible --roster-file=inventory.yml` to reach Ansible managed systems and apply Salt states (like bootstrap state for transition to salt minion).

Consuming a Java API from the `ansible-gatherer`, IIUC, XMLRPC API, would also be an option as well. 

Would this mean those endpoints will be available in the public XMLRPC API ?

accepted/00000-ansible-integration.md
meaksh: Well, I was thinking on the fact that a customer/user may have a traditional client which is also managed by an Ansible control node. 

After running the "ansible-gatherer" and importing the systems, I would like too notice that this traditional client is also being managed by Ansible.

accepted/00000-ansible-integration.md
meaksh: You can pass the inventory you want to use, so you have control of the `-i` option of Ansible. IIUC it should work with dynamic inventories if you pass the module as argument of `ansible -i`

accepted/00000-ansible-integration.md
meaksh: I don't have a clear idea at the moment. Will try to figure out

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,304 @@
...
+- If any of those systems is already existing with "ANSIBLE" entitlement, then nothing to do
+- If the system is already registered but not with "ANSIBLE" entitlement, then add "ANSIBLE" entitlement.
+
+Of course, we need some new tables in the database, at least something like:
+
+```
+"suseAnsibleController" (label, type, org_1)
+"suseAnsibleControllerConfig" (controller_id, parameter, value)
+"suseAnsibleControllerSystem" (system_id, controller_id)
+"suseAnsibleControllerPlaybook" (controller_id, local_path, content?)
```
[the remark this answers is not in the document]
meaksh: Playbooks are located in the Ansible control node. When running `ansible-gatherer` we bring those playbooks to the Uyuni server, in order to display them in the UI and allow triggering them.

The idea here is that those playbooks are read-only, we're not providing a way to edit them in the UI. Just want to shown them as part of the gathered data from the Ansible control node.

At the time of running those playbooks from Uyuni, we don't really require them at the Uyuni server, since we're calling the Ansible control node to apply the playbook that are already stored (and maintained) the control node.

Having the content stored in the DB (in the end, a yaml content), might be better to prevent having orphaned entries in the DB if any of the local stored files has been removed from the filesystem.

Since we're already storing such content in the DB in some other Uyuni features, I would maybe follow the same approach here if there is nothing else to consider.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,304 @@
...
+
+This RFC is more focused on allowing users to import their existing Ansible environments into Uyuni, allow some basic operations, coexistence of Salt and Ansible managed infrastructure in Uyuni, eventually apply playbooks and also to provide an easy way to transition from Ansible to a fully-featured Salt minion managed system in Uyuni, rather than making Uyuni a top-featured UI to build your Ansible infrastructure from scratch.
+
+
+# Detailed design
+[design]: #detailed-design
+
+There are three main parts/goals here, conceptually:
+
+1) Collect data from an Ansible controller (Inventory / SSH Keys / Playbooks, etc) and import the hosts in the inventory as ANSIBLE systems in Uyuni.
```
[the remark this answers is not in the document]
meaksh: Done and pushed.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,320 @@
...
+
+This section elaborates how Uyuni would do for gathering information for an Ansible control node. There are different scenarios that we can consider here:
+
+1. Adding an external host Ansible control node:
+
+The premise here is that there is already an existing and configured Ansible infrastructure somewhere and we want to import it into Uyuni. This Ansible control node is not even a registered system in Uyuni. The hosts defined on the inventory can be imported Uyuni and will be displayed as "Foreign/ANSIBLE" or "Salt/ANSIBLE", etc. The "ANSIBLE" entitlement means that host is being managed by an Ansible control node.
+
+Here we're **not** primarely focused on building a new Ansible managed infrastructure from scratch using Uyuni.
+
+NOTE: Currently Ansible is not shipped in SLE15 any SP. In order to allow an SLE Ansible control node, we would need to ship Ansible in probably in SLE15 client tools.
```
moio, line 43: IIUC in this hypothesis the Uyuni Server would display Ansible-managed systems obtained from an inventory.

What mechanism/protocol would be used to access such inventory?

Would there be any feature other than listing of Ansible-managed clients?

accepted/00000-ansible-integration.md
moio: I can see such a corner case, but I would recommend against any investment on the traditional stack at this point. The idea is to minimize user dependence on that stack to the extent possible, and try as far as possible to avoid creating new dependence.

So any new feature that just happens to work out of the box "by accident" then sure, why not, but otherwise let's stick to the strategy of making any new functionality Salt exclusives to give more incentive to migrate away from the traditional stack.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,320 @@
...
+```
+"suseAnsibleController" (label, type, org_1)
+"suseAnsibleControllerConfig" (control_node_id, parameter, value)
+"suseAnsibleControllerSystem" (system_id, control_node_id)
+"suseAnsibleControllerPlaybook" (control_node_id, local_path, content?)
+```
+
+And also create the new "ANSIBLE" entitlement, which should be an addon-entitlement compatible with Salt, Foreign and maybe also for a potential "Management/ANSIBLE" (traditional client) entitlement. Here is the [SQL script](https://github.com/meaksh/uyuni-hacks/blob/master/scripts/ansible/add_ansible_entitleme.sql) used for PoC demo. It's also needed to adapt the rhn databaste functions implemented in the DB.
+
+NOTE: The profile for a system with registered in Uyuni as "Foreign/ANSIBLE" does not look like as the one Salt/Traditional system, since most of the Uyuni features (channels, software profile, etc) are not implemented for neither "Foreign" nor "ANSIBLE" entitlements. In order to get the entire portfolio of Uyuni features for Salt minion, this "Foreign/ANSIBLE" systems would need to be transitioned to "Salt/ANSIBLE" minion.
```
moio, line 125: Observation: "entitlements" have been renamed to "system types" in all user-facing material (including the UI).

References to "entitlements" still exist in the codebase, but I would recommend to stick to "system types" here, as "entitlements" can be a bit confusing as a concept.

moio, line 125: System types come in two "variants": base and add-on. Each system can have either zero or one base system type, and any number of add-on types. We also have mechanisms to specify which types are compatible (or not compatible) with which types.

Can you break down your proposal of system types by their "variant", and which would be compatible to which (including existing ones)?

https://github.com/uyuni-project/uyuni/blob/master/schema/spacewalk/common/data/rhnServerGroupType.sql
https://github.com/uyuni-project/uyuni/blob/master/schema/spacewalk/common/data/rhnSGTypeBaseAddonCompat.sql

(PS: I know, jargon is very confusing! server group type == entitlement == system type)

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,320 @@
...
+
+Notice also that the Ansible control node is not listed as systems list page, since there is not system entry for it, it would be shown in the respective new UI page for managing Ansible control nodes.
+
+#### New UI tabs & pages
+
+* Ansible
+* Ansible / Ansible control nodes list
```
moio, line 132: What would be the pros and cons of a specialized System list?

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,320 @@
...
+2) Operate Ansible: Apply playbooks via control node node / Apply Salt commands & states directly to Ansible managed systems / Make Ansible system become a fully-featured Salt Minion.
+3) Maintain Ansible infrastructure: Own Uyuni Playbook catalog, revisions, formulas.
+
+## Collecting data from an Ansible control node
+
+This section elaborates how Uyuni would do for gathering information for an Ansible control node. There are different scenarios that we can consider here:
+
+1. Adding an external host Ansible control node:
+
+The premise here is that there is already an existing and configured Ansible infrastructure somewhere and we want to import it into Uyuni. This Ansible control node is not even a registered system in Uyuni. The hosts defined on the inventory can be imported Uyuni and will be displayed as "Foreign/ANSIBLE" or "Salt/ANSIBLE", etc. The "ANSIBLE" entitlement means that host is being managed by an Ansible control node.
```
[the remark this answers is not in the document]
meaksh: > IIUC in this hypothesis the Uyuni Server would display Ansible-managed systems obtained from an inventory.
> 

Correct.

> What mechanism/protocol would be used to access such inventory?
> 

As detailed later, the idea is that `ansible-gatherer` would take care of it. For an external host, the idea is to run a "salt-ssh" using a given "user/pass" or SSH key, to access the inventory on that host. This "salt-ssh" execution would be executed by `ansible-gatherer` directly accessing "salt-api" (or salt python client), in other words, without going over Java in this case.   

> Is this RFC concerning any feature other than mere listing of Ansible-managed clients in this scenario?

At this point in the RFC, we're still only talking about collecting from Ansible control nodes. If we continue further here, and step into "operating" Ansible, this means we could also trigger the execution of the playbooks from this Ansible control node using the UI. Like mentioned here: https://github.com/uyuni-project/uyuni-rfc/pull/53/files#diff-e8e05c6db2b37eadd62b4c5260480df6331de275bd0b84c8376b3c3b21ce67ccR192

accepted/00000-ansible-integration.md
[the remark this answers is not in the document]
meaksh: I'm clarifying this notes and moving them to a separated section :+1:

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,320 @@
...
+3. Adding an Ansible control node from a registered minion:
+
+This scenario refers to the posibility of adding an Ansible control node from a registered system. Similarly to the previous scenarios but in this case, the Ansible control node is a registered system in Uyuni.
+
+### ansible-gatherer
+
+This is the component that takes care of collecting the inventory, host and playbooks from a given Ansible control node.
+
+Therefore, there might be more than one Ansible control node host, even the Uyuni server, proxy or a registered system, could act at some point as an Ansible control node. Since data sources, and type of sources (e.g. single host, AWX API) might be multiple, this RFC proposes an approach similarly to what Uyuni does for handling "Virtual Host Managers (VHM)". This means, using a Python tool, in this case called something like "ansible-gatherer", which is plugin-based (so easily allows implementing different sources of Ansible inventories). This tool would be called via an Uyuni Java schedule, like the "virtual-host-gatherer", passing the necessary information (parameters, type of host, etc) to reach the Ansible control node and collect the necessary information.
```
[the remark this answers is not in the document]
meaksh: I see the idea of rebranding the VHM to something more generic as interesting and worth to consider.

Pros:
- Less components in Uyuni. Probably less code lines in general.
- Reduce a bit the Python implementation (just adding new plugins to "virtual-host-gatherer"). Most of the implementation effors goes into Java / UI.
- Reuse of Taskomatic tasks, some parts of the UI and XMLRPC API.

Cons:
- Adds more complexity on the Java backend that receives the JSON output from "virtual-host-gatherer":  The current Java backend for VHM is basically meant for gathering Virtual Hosts instances. The JSON output produced by "virtual-host-gatherer" is completely different, and with a different purpose, depending if you're, for example gathering virtual instances (where you get names + uuids) or gathering Ansible nodes from an Ansible inventory (not uuid, playbooks, etc). The Java backend for VHM would need to be adapted to check which "plugin" has been triggered and then parse the output accordingly.
- Having the same schedule for gathering from Ansible that for any other VHM might not be necessarely what a user wants.
- While some of the UI can be reused, also new UI would be anywa y required to visualize an Ansible control node (which is not the same UI that we expect when visualizing a VHM node).
- Completely isolate Ansible gathering from VHM sounds simpler.

accepted/00000-ansible-integration.md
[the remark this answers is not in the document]
meaksh: > I was thinking about a public XMLRPC API, yes. Could an Uyuni Server-backed roster make sense in general?

If the XMLRPC API would allow the `ansible-gatherer` to store all necessary things in the DB, then we can consume those endpoints directly from inside the `ansible-gatherer`. This would require effors for building new XMLRPC API endpoints to allow storing different things (playbooks, inventories, etc), and at the same time it makes a bit harder to trigger "salt-ssh" executions to reach Ansible managed nodes, since we need to ask Java to generate the Ansible roster file that would be consumed by the "salt-ssh" execution.

accepted/00000-ansible-integration.md
meaksh: Thanks, I also agree on this. I'll clarify this better on the RFC text

accepted/00000-ansible-integration.md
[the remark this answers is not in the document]
moio: Pros: I agree.

Cons: 
 - I agree about the same schedule argument
 - I do not see the Java JSON parsing part to be significantly different, in terms of complexity, in the two cases. No matter what we do we will always need a new parser of a completely different JSON structure
 - Similarly, Ansible-specific UI parts will have to be new in any case
 - I agree this adds a bit of complexity, but...

Additonal elements I see are:
 - we will need to rebrand the VHM in any case, as it's long past the point of being an accurate name. Doing so in this occasion is "taking two birds with one shot". We will later save that effort
 - we would establish a standard way to add systems to SUSE Manager in addition to onboarding. Anything else we would add in future would have a clear place to fit in, and clear example to copypaste from. It will be closer to a Salt-style plugin infrastructure (for this particular aspect) - a more modular design
 - I also see it could it be easier on user experience in general to have a unifying concept for any "external things". Whatever you want to add to SUMa, you go to the one page to configure details on how to access such an "external thing", and know there is one task to keep it current

How does that sound?

accepted/00000-ansible-integration.md
meaksh: Thanks for the note. Fixed.

meaksh: FYI these are the schema changes initially done during the PoC: https://github.com/meaksh/uyuni-hacks/blob/master/scripts/ansible/add_ansible_entitleme.sql

Basically:

- Ansible (not base / not permanent)
- Salt compatible with Ansible
- Foreign compatible with Ansible

accepted/00000-ansible-integration.md
meaksh: We may probably want to have that at some point.

Do you mean something in the same way we currently do in "Systems -> Virtual System List" where all virtual systems, including the ones gathered by VHM that are not even registered in Uyuni?

moio: My thinking is: all Systems (no matter their entitlement, and even if they have no entitlement at all) show up in Systems lists. We have several, all with different filtering criteria and rendering options (eg. columns, extra buttons in some pages, and so on).

To me a list of Ansible systems (no matter if we are talking controllers or managed systems) could potentially be just another such list, so from a UX perspective it could also belong there - rather than an isolated Ansible section.

accepted/00000-ansible-integration.md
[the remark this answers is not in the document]
meaksh: I see your point and I like the idea of rebrading the VHM to something more generic, but honestly I think that would require more time and will deserve its own RFC.

Rebrading VHM means, not only code/UI changes (probably also package name changes) but also lot of changes in the documentation.

At this point in time, I would prefer to start the Ansible integration without unifying it with the VHM. Of course, we can later work on a "Rebrading VHM RFC", where I think most of the effors invested in the backend of Java for parsing the output from `ansible-gatherer` can be also reused in a future generic importer. Same for the UI.

What do you think?

accepted/00000-ansible-integration.md
meaksh: I agree the ansible systems should be listed in the "all systems" view, as we do for the virtual instances that are not even registered in SUMA.

But here, this "Ansible / Ansible control nodes list" page is simply what we have in "Systems -> Virtual Host Managers", a list with the registered Ansible control nodes so you can add more.

This is perfectly compatible with having a such a new list in "Systems ->  Systems List -> Ansible". I'm adding a note to the RFC :+1:

accepted/00000-ansible-integration.md
mcalmer: I would just remove traditional stack from the RFC. We decided already - no new feature for traditional stack.
We only add something to the traditional stack when a new feature would break an existing functionallity in the traditional stack.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,327 @@
...
+Of course, we need some new tables in the database, at least something like:
+
+```
+"suseAnsibleController" (label, type, org_1)
+"suseAnsibleControllerConfig" (control_node_id, parameter, value)
+"suseAnsibleControllerSystem" (system_id, control_node_id)
+"suseAnsibleControllerPlaybook" (control_node_id, local_path, content?)
+```
+
+And also create the new "ANSIBLE" system type, which should be an addon-system type compatible with Salt, Foreign and, in case this doesn't need any extra effors, maybe also for a potential "Management/ANSIBLE" (traditional client) system type. Here is the [SQL script](https://github.com/meaksh/uyuni-hacks/blob/master/scripts/ansible/add_ansible_entitleme.sql) used for PoC demo. It's also needed to adapt the rhn databaste functions implemented in the DB.
```
mcalmer, line 121: Try to remove "may" and "maybe". The RFC should be clear what is wanted and what not.
In this case I would remove "Management" completly from the RFC.

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,327 @@
...
+      state: started
+```
+
+As you can see, this playbook is defining different tasks to execute on different "hosts" groups. Those groups are of course defined in the Ansible inventory (default at /etc/ansible/hosts).
+
+The execution of a given playbooks is done in the corresponding Ansible control node, which contains the inventory and the hosts definition.
+
+The playbooks available to apply on each Ansible control node, are the ones exposed by "ansible-gatherer" and ultimately, in the future, playbooks created and maintained in Uyuni (via future Ansible playbook catalog) that would be then pushed to the control node node.
+
+At the time of running a playbook from the Ansible control node, Uyuni is able to do it by reaching the control node and operating it using "salt-ssh" when it's an external host. If the control node is the Uyuni server or a registered minion we don't necessarily need to use "salt-ssh".
```
mcalmer, line 187: Should our main goal just be to make all these systems regular minions and avoid using salt-ssh when we can?
I think it is nice that it would work also with salt-ssh, but as it has massive performance problems we better design everything to become real minions and leave salt-ssh as a corner case for special situations.
What do you think?

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,327 @@
...
+* Ansible / Ansible control nodes list
+* Ansible / Add new Ansible control node
+* Ansible / View Ansible control node information & and visualize playbooks (would also allow triggering playbooks via "salt-ssh" in the control node)
+* Systems / Systems list / Ansible (show only Ansible systems and its control node)
+
+or even create a new entry level on the menu as "Automation", something like:
+
+* Automation / Ansible / ...
+* Automation / Salt / ...
+* Automation / Remote commands
```
mcalmer, line 139: This is new and surprising. What should it do? Is this related to Ansible somehow? Or would it be better put in a separate RFC?

accepted/00000-ansible-integration.md
```diff
@@ -0,0 +1,327 @@
...
+
+### Step 0: The basic: gathering from an Ansible control node
+- Implement "ansible-gatherer" to gather from external host (SSH key based).
+- UI to add Ansible control nodes (reused from "Virtual Host Manager").
+- Push Ansible package to SLE15 clients tools to allow having SLE control nodes.
+- Inventory can be synced with Uyuni (add "Foreign/ANSIBLE" and "Salt/ANSIBLE", ...)
+
+### Step 1: Operating an Ansible control node
+- Trigger the playbooks from Ansible control node using "salt-ssh".
+- Transition from "Foreign/ANSIBLE" to "Salt/ANSIBLE".
```
mcalmer, line 275: When I look at this, I have the feeling that only these 2 steps together are the MVP. But we could also remove some topics if it would help to make the implementation easier (not sure about it)

- Ansible for SLE15 is more a business decission then an implementation requirement. At the first place we want to get "existing" Ansible landscapes integrated. And this means there is a control node somewhere.
- "FOREIGN" should be discussed. If it would make things easier I would only go with "Salt/Ansible".

accepted/00000-ansible-integration.md
paususe: I would keep the mention to the traditional stack in the alternatives, just for reference, with an indication we decided against it. The end result is the same but if we need to revisit the RFC in the future, it will be clear we considered and rejected traditional.

accepted/00000-ansible-integration.md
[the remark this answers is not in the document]
paususe: @moio we already discussed the idea of making VHM more powerful and integrated and I am all for it but that's a separate RFC, one for the future. Let's focus on the problem at hand: Ansible.

accepted/00000-ansible-integration.md
paususe: > Should our main goal just be to make all these systems regular minions and avoid using salt-ssh when we can?

Yes, that's what we agreed on when we discussed the RFC a few weeks ago.

Also, using minions would allow us to avoid user+password. 

> I think it is nice that it would work also with salt-ssh, but as it has massive performance problems we better design everything to become real minions and leave salt-ssh as a corner case for special situations.
> What do you think?

I don't think performance is a factor here, though.

It's the Ansible control nodes that will be registered to Uyuni. That's a handful of nodes, tops. Then each of those Ansible control nodes manages (always via SSH) the clients (third level of depth: 
Uyuni Server (salt-master) => Ansible control node (salt-minion) => Client (ansible)

In fact, as a curiosity (it does not affect this RFC for anything), if you think about it the final design of this Ansible integration is not that different from the Hub, where we also have three levels:
Uyuni Hub Server (salt-master) => Uyuni Server (salt-master and salt-minion) => Client (salt-minion)

accepted/00000-ansible-integration.md
paususe: No, it's not. It's a UI thing that I mentioned a month ago, on how to display this:

https://github.com/uyuni-project/uyuni-rfc/pull/53#discussion_r583784531

image

But we'll need @ncounter / @cbbayburt to provide insight

accepted/00000-ansible-integration.md
paususe: > When I look at this, I have the feeling that only these 2 steps together are the MVP. 

Agreed

> But we could also remove some topics if it would help to make the implementation easier (not sure about it)
> 
>     * Ansible for SLE15 is more a business decission then an implementation requirement. At the first place we want to get "existing" Ansible landscapes integrated. And this means there is a control node somewhere.

There are reasons to provide the package, we can discuss on Monday.

>     * "FOREIGN" should be discussed. If it would make things easier I would only go with "Salt/Ansible".

We agreed no foreign and not traditional, only Salt+Ansible (but remember that's for the Ansible controllers!)

accepted/00000-ansible-integration.md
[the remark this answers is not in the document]
moio: I am not suggesting to make the VHM any more powerful than it is now, I am just suggesting to take this occasion to rename it, and put the Ansible integration in that place (UI-wise and code-wise) instead of someplace new.

So far I did not yet see a convincing argument why this would turn out to be substantially more complex. Even docs-wise we are really talking about ~10 files and a few `sed` commands.

I could even agree Uyuni-wise and RFC-wise it would make sense to postpone this, but downstream-wise we do not want such a UI change (even if it's just mere renaming) before a major version, so now it is a good time to do it.

That would make the overall architecture easier to understand and maintain, instead of adding yet another ad-hoc mechanism.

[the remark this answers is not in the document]
ncounter: While I can understand all the point of view explained so far, I'd like to put my thought on this, and I will do it by "cleaning up the table" and start thinking from scratch.

From a general perspective Uyuni is capable to add systems and keep track of them: the main purpose of doing this is to manage them on the configuration and many other parts, and these systems go through the onboarding process. Let's call these systems the **managed systems**.
That said, Uyuni becomes a core application to manage all the systems in a data center, so "core" that it comes by itself that Uyuni is also an _inventory_ tool of everything available in the data center. Now, Uyuni has also a nice feature to **just keep tracking** of other systems in the data center, the ones Uyuni is not there to **manage** them. Let's call these systems the **self systems**. All the latter are the ones gathered by the VHM.

Here it chimes in our Ansible integration: what are we going to implement? IIUC we will let Uyuni be aware of the existing Ansible systems, but we are definitely not going to manage them their configuration through the onboarding process etc. Therefore they will just take part of the **self systems** bunch mentioned before.

Where all of this is going?
I am trying to abstract the thinking and reaching the user POV, and I want to make sure it is clear where things are and what they are meant for. I (as a user) will fork my searching biased on either **registered systems** because I want to manage their configuration **or** I will look for **self systems** because I need to do whatever I can do with them, but not configuration for sure. The clearest way of representing this is to append the Ansible bunch of systems under the VHM umbrella, but definitely renaming the umbrella to a generic one. Unless I am missing something huge, an Ansible system will just be a **type** of the self systems, like Virtual Host Manager are, and the approach to them is the same.

Why should we not distinguish Ansible as a top level menu entry?
Because we are not trying to distinguish a system by its nature, we want to keep the Uyuni perspective, we are here "to manage or not to manage" first. Once we fork the initial step, we can move the next one in the workflow. Otherwise spaghetti connections are to come inside the user mind, therefore in the code logic.

Looking for a first small initial step to add the integration but not refactoring the whole VHM?
I believe an Ansible system is just a sub-type of the gathered-not-managed systems, but I don't really know the difference from refactoring and including the Ansible gathering withing the current VHM code, or just implementing a separate code path and append it, so I will let this up to who knows this part better. If the assumption of adding a separate code path to gather Ansible systems is fine, I can propose a short-term solution (which is meant to be temporary, the final one will be the whole refactoring): we could wrap the two types under an abstract concept level but keeping them distinguished, in order to let untouched the VHM part. But again, the assumption must be right in the light of the final scope of the whole refactoring.

Before:
`Systems > Virtual Host Managers`

After:
`Systems > Self Systems > Virtual Host Managers`
`Systems > Self Systems > Ansible`

PS: "Self Systems" is just for consistency of my comment, it can be "Autonomous Systems" or "Not Manage Systems" or something else.

accepted/00000-ansible-integration.md
ncounter: I gave an overview of my thinking in https://github.com/uyuni-project/uyuni-rfc/pull/53#discussion_r602851071 . Actually I'd stay attached to the approach of splitting first the "managed VS not managed" criteria, then I'd wrap whatever is meant to wrap.
If for some reason this concept cannot be respected, I could be in favor of this "Automation" wrapping idea. But the main aspect I am not comfortable with is to show and present "Ansible" at the very same level of "Salt", which we know it is definitely not from the Uyuni perspective. Salt is the engine and the core of the configuration management Uyuni is built on, Ansible is just an external type we let it in but we don't touch at all instead. And I also wonder what if yet-another-type will come in in the future.

accepted/00000-ansible-integration.md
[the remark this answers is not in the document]
paususe: @ncounter we are going to manage those systems. They are not foreign systems.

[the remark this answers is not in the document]
ncounter: > @ncounter we are going to manage those systems. They are not foreign systems.

This is not what the initial motivation describes:

>This RFC is more focused on allowing users to import their existing Ansible environments into Uyuni, allow some basic operations, coexistence of Salt and Ansible managed infrastructure in Uyuni, eventually apply playbooks and also to provide an easy way to transition from Ansible to a fully-featured Salt minion managed system in Uyuni, rather than making Uyuni a top-featured UI to build your Ansible infrastructure from scratch.

We will not manage them "by Ansible". We will import them as they are (`allowing users to import their existing Ansible environments into Uyuni`). If the user what to manage them instead, the Salt path will occur (`to transition from Ansible to a fully-featured Salt minion managed system in Uyuni`). Until it is just an Ansible system with no Salt configuration, it should be treated as a foreign/not-managed system. Once the transition is completed, it should end up in the System List as other clients, as a Salt client. Of course the Ansible property should be visible, but it's just a property of the system, we are not handling it by Ansible.

And as mentioned in this section as well `this RFC proposes an approach similarly to what Uyuni does for handling "Virtual Host Managers (VHM)"` because they are "gathered" systems. The question is whether we want to add yet-another-gatherer for a different type, or refactor the VHM and gather multiple types of systems with the same logic, which sounds much more inline with the current picture.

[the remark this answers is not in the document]
paususe: @ncounter Step by step, we are going to:

1. Onboard Ansible controllers and gather inventory (Ansible client machines, final leaf in the hierarchy)
2. Gather what playbooks the Ansible controllers can apply on those Ansible client machines
3. Allow running those playbooks from the Uyuni Server instead of the Ansible controller (the Uyuni Server will drive the Ansible controller, not execute directly)
4. Allow adding Ansible playbooks to configuration channels. Mayb we'll call that "Ansible configuration channels", maybe we'll just add them to the current configuration channels. To be determined.
5. Allow adding playbooks directly from the Uyuni Server WebUI so that users do not need to go to the Ansible controller at all
6. Allow editing playbooks directly in the Uyuni Server WebUI, the same way we can edit Salt states

If that is not management, then I don't what it is.

The fact we will use Salt+AnsibleGate to drive Ansible is just a technical detail.

accepted/00000-ansible-integration.md
mcalmer: Seems here is a big missunderstanding. I thought we Salt/Ansible is for every client the goal. It is a salt client in Uyuni with the additional capability to execute also Ansible Playbooks. In this case I would say salt-ssh is a performance problem.

paususe: I am now confused myself and need to re-read the RFC %)

accepted/00000-ansible-integration.md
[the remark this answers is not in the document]
moio: @ncounter let me offer a possible clarification/way to look at this - as of today, there's two ways to have a system show up in the Systems list: either via registration, or via an "import" from the VHM. Those are just the two ways users have at their disposal to add a system to Uyuni.

The fact that the _current implementation_ of VHM exclusively imports systems which have very limited functionality compared to the registered ones I'd consider more about a historical consideration than a design one. Abstracting from the Ansible discussion here, it's not set in stone that systems coming somewhere other than a registration have to be "underpowered" in any way. In principle, we could very well add new imported system types which end up being well featured, or fully featured, or (stretching the concept) even have more features than anything we have been registering to this day. It's just a matter of having some API endpoint to discover them instead of a bootstrap script.

What I am trying to say is that the crucial difference from a UX perspective could be how the system "is originated in Uyuni" - to offer the two entrance gates in a consistent way. Once a system is born, it's a system like all others, with a set of available features determined by its System Type (formerly "entitlement"), which we can use as filtering conditions just like all others.

In that light, the level of feature-completeness for Ansible clients becomes less relevant. Today it can be basic, tomorrow it might become super-advanced, but this is pretty much transparent to the onboarding procedure.

Does that sound any clearer?

accepted/00000-ansible-integration.md
moio: @mcalmer the Uyuni Server will contact, via SSH, any configured **Ansible controllers**.

Each of those controllers will have inventories (lists) of end clients, which we will import to the DB to show and operate. The Uyuni Server will not contact those clients directly (at least as of the first implementation step) – only through their controllers.

We do not expect users to have more than let's say ~100 **controllers**, which would be a number that makes me start wondering about performance.

If ever we get close to that number, then requiring to register controllers as minions (and contacting them over the Salt bus instead of SSH) will make sense, and it is a step we are ready to make eventually.


Another way to look at this: the main initial intent here is to integrate with existing infrastructure, and the minimalistic/most universal way to do so is via SSH access. It's not equally nice to ask for controllers to be registered as minions just to get started. So we will start from SSH, and potentially offer Salt as an additional option at a later point, when it makes sense. In the meantime, nothing stops a user from registering a controller and manage it as usual.

meaksh:
I am closing this now since we finally agreed on this other RFC: https://github.com/uyuni-project/uyuni-rfc/pull/55

Thanks everybody that contributed!

PR #55, opened 2021-04-08
Enable Ansible control node management for registered systems

meaksh:
Read it [here](https://github.com/uyuni-project/uyuni-rfc/blob/master-enable-ansible-control-node-in-registered-system/accepted/00000-enable-ansible-control-node-in-registered-system.md)


accepted/00000-enable-ansible-control-node-in-registered-system.md
```diff
@@ -0,0 +1,136 @@
...
+#### Use "salt-ssh" from Uyuni directly to the "fqdns" of the Ansible managed clients:
+
+* We previously need to fetch the corresponding SSH key that is used to connect to this particular host by Ansible.
+* Then we would reuse this key as SSH credentials to trigger the bootstrap on that host.
+* Downside: Uyuni needs to have direct access to the Ansible managed clients to bootstrap
+
+#### Use "salt-ssh" from Uyuni using the Ansible control node as SSH proxy:
+
+* We do not necessary need to fetch the corresponding SSH key, but we need to have the Uyuni SaltSSH deployed on the Ansible control node to access it
+* The SSH key to connect the Ansible managed client referenced in the Ansible inventory is used to build the ProxyCommand to use in "salt-ssh".
```
paususe, line 74: I think this case can be assimilated to having a SUSE Manager Proxy, where the clients can only reach the Server through the Proxy but not directly. Deploying the SaltSSH seems OK to me.

accepted/00000-enable-ansible-control-node-in-registered-system.md
```diff
@@ -0,0 +1,136 @@
...
+
+At this point, Uyuni doesn't know how to reach those systems, we only know their "fqdns" according to the Ansible inventory, but the Ansible inventory in the control node should contain information about how to reach them via ssh. The idea here is the following:
+
+In order to bootstrap and trigger the registration of these systems, we have two alternatives:
+
+#### Use "salt-ssh" from Uyuni directly to the "fqdns" of the Ansible managed clients:
+
+* We previously need to fetch the corresponding SSH key that is used to connect to this particular host by Ansible.
+* Then we would reuse this key as SSH credentials to trigger the bootstrap on that host.
+* Downside: Uyuni needs to have direct access to the Ansible managed clients to bootstrap
```
paususe, line 69: I don't necessarily see this is as a problem, at least initially

accepted/00000-enable-ansible-control-node-in-registered-system.md
```diff
@@ -0,0 +1,136 @@
...
+
+#### Use "salt-ssh" from Uyuni using the Ansible control node as SSH proxy:
+
+* We do not necessary need to fetch the corresponding SSH key, but we need to have the Uyuni SaltSSH deployed on the Ansible control node to access it
+* The SSH key to connect the Ansible managed client referenced in the Ansible inventory is used to build the ProxyCommand to use in "salt-ssh".
+
+NOTE: It needs to be clarified in the documentation that `ansible_ssh_private_key_file` needs to be defined in the Ansible inventory of your Ansible control node.
```
paususe, line 76: Is this something that's usually defined? Or is this something very special we would be requiring? That's for me the main factor to decide for one or the other alternative.

accepted/00000-enable-ansible-control-node-in-registered-system.md
```diff
@@ -0,0 +1,136 @@
...
+
+* Once we have this, we can trigger the bootstrap of those systems by calling "salt-ssh" like this:
```
paususe, line 78: We will be using salt-ssh but clients may be salt-minions or salt-ssh clients, right?

accepted/00000-enable-ansible-control-node-in-registered-system.md
```diff
@@ -0,0 +1,136 @@
...
+* Another example of usage (without using the ansible roster/targetting):
+
+```console
+# salt-ssh --ssh-option='ProxyCommand="/usr/bin/ssh -i PATH_TO_SALT_SSH_KEY_IN_THE_UYUNI_SERVER -o StrictHostKeyChecking=no -W %h:%p uyuni-stable-ansible-controller.tf.local /usr/bin/ssh -i PATH_TO_THE_SSH_KEY_IN_ANSIBLE_CONTROL_NODE uyuni-stable-ansible-opensuse152-2.tf.local" -o StrictHostKeyChecking=no' uyuni-stable-ansible-opensuse152-2.tf.local test.ping
+```
+
+
+### Triggering playbook executions in an Ansible control node
+
+After exploring the Ansible control node, inventory and playbooks, we can allow Uyuni to trigger the execution of one of those playbooks via a simple Salt state to execute in that control node:
```
paususe, line 94: Should we also provide an easy way to trigger this from the UI? E. g. by adding Ansible playbooks directly to configuration channels (without the Salt state in the middle, we can autogenerate that state)?

accepted/00000-enable-ansible-control-node-in-registered-system.md
```diff
@@ -0,0 +1,136 @@
...
+Considerations and other discussions about shipping Ansible in SLE, the support level, etc, can be discussed outside this RFC since it's out of the main focus of the general Ansible integration in Uyuni.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we **not** do this?
+
+  * obscure corner cases
+  * will it impact performance? Bootstrapping Ansible managed clients might take some time since "salt-ssh" is involved.
```
paususe, line 120: Say I have 1,000 Ansible clients and I want to bootstrap them with salt-ssh. Will it take a lot more time than if I bootstrap them with the normal bootstrap? To me, it's the same, the only additional step is getting the SSH key from Ansible. Or am I missing something?

paususe (commented):
Left a few comments and questions

accepted/00000-enable-ansible-control-node-in-registered-system.md
mcalmer: Maybe not in the first step. The problem is, that playbooks have the targets where it gets applied included. This cannot be changed without editing playbooks which is not planned yet. This topic along would need an own RFC from my point of view.

accepted/00000-enable-ansible-control-node-in-registered-system.md
agraul: That's my understanding, yes. Just like regular boostrapping which is executed over salt-ssh for both salt-ssh and salt-minion clients

accepted/00000-enable-ansible-control-node-in-registered-system.md
paususe: > Maybe not in the first step

OK

> The problem is, that playbooks have the targets where it gets applied included. This cannot be changed without editing playbooks which is not planned yet. This topic along would need an own RFC from my point of view.

OK. Maybe an opportunity to move the editor to VS Codium or alike :-)

accepted/00000-enable-ansible-control-node-in-registered-system.md
mcalmer: Using our UI you can only register 1 and even with API you register one after another.

Here I would expect some kind of parallel onboarding and this requires batches and other things.
Performance is a possible problem. It depends on the implementation.

accepted/00000-enable-ansible-control-node-in-registered-system.md
meaksh: This NOTE actually applies for both approaches. The SSH keys or credentials to connect with the Ansible managed clients should be defined in the Ansible inventory.

If an user do not specify any user/pass/sshkey in the Ansible inventory to connect with different Ansible managed clients, then Ansible will use the default SSH identity of the user that executes `ansible` CLI to try to connect to the Ansible managed clients.

accepted/00000-enable-ansible-control-node-in-registered-system.md
meaksh: Yes, that's correct.

accepted/00000-enable-ansible-control-node-in-registered-system.md
meaksh: It should take the same time in both cases. In the end, the "Boostrapping UI" page is basically also doing "salt-ssh" execution under the hood.

As you mentioned, the only additional step would be to fetch the SSH keys from the Ansible control node first, and not necessary for each client to bootstrap but just once at the time of getting the Ansible inventory.

accepted/00000-enable-ansible-control-node-in-registered-system.md
```diff
@@ -0,0 +1,136 @@
...
+
+After exploring the Ansible control node, inventory and playbooks, we can allow Uyuni to trigger the execution of one of those playbooks via a simple Salt state to execute in that control node:
+
+```yaml
+execute_ansible_playbook:
+  ansible.playbooks:
+    - name: /some/path/in/my/ansible/control/node/playbook_1.yaml
+```
+
+We would need to create a new type of Action in Uyuni for "running a playbook". We want this action to appear in the system event history for the Ansible control node.
```
moio, line 102: It is my understanding that we do not intend to store the list of playbooks in Postgres at least for now. The list is retrieved live at page load, displayed as a list with a button to run each, and once the button is click the action is scheduled.

Correct?

accepted/00000-enable-ansible-control-node-in-registered-system.md
```diff
@@ -0,0 +1,136 @@
...
+  * obscure corner cases
+  * will it impact performance? Bootstrapping Ansible managed clients might take some time since "salt-ssh" is involved.
+  * what other parts of the product will be affected?
+  * will the solution be hard to maintain in the future?
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+- What other designs/options have been considered? An alternative approach based on VHM/"ansible-gatherer" is discussed [in this other RFC](https://github.com/uyuni-project/uyuni-rfc/pull/53)
```
moio, line 128: I'd recommend trying to put down pros and cons to the extent possible. At least I would recommend making it clear that the basic premise of this design (controller is registered) could potentially be limiting compared to the other design.

moio (commented):
@meaksh thanks for the quick write up!

My main remark with the current text comes from the following considerations:
 - we do have mass registration as an open general problem since a while, [in Public Cloud contexts](https://github.com/SUSE/spacewalk/issues/10884) and otherwise. [We kind of suggest working around it in the documentation](https://github.com/uyuni-project/uyuni-docs/blob/master/modules/large-deployments/pages/operation-reqs.adoc#salt-client-onboarding-rate). This could be the right occasion to solve it once and for all, by providing one good mechanism either leveraging the existing 1-system-at-a-time API or introducing a new one, and solve a general day-1 problem many users face. Fact that systems to onboard come from an Ansible controller rather than someplace else does not change the problem's essence - I would not want to have different mechanisms for different instances of the same problem
 - onboarding Ansible end clients is not necessarily desirable in the first place. For starters, users might only really want to just running playbooks. Second, clients might be on platforms we do not support. Third, onboarding can always happen via running a playbook - simply to run a bootstrap script! Fourth, users might not want to have salt-minion co-installed for non-technical reason. Fifth and most important to me, they might not want to adapt their playbooks to make that happen. In fact, when a decision is made to have a system managed both with Salt and Ansible, extra care in writing both playbooks and states, otherwise they will conflict, [particularly in the area of repo management as our own internal sysadmins experimented a while ago](http://mailman.suse.de/mlarch/SuSE/suse-manager/2018/suse-manager.2018.10/msg00015.html)
- we are a bit short on time and want to deliver a minimal increment fast

Keeping those factors in mind, I recommend moving the whole onboarding problem to a separate RFC, to be implemented once we added the ability to run playbooks, which is the core value add at the moment. Text you have created here should be kept (as it's valuable!) in a "next steps" section.

paususe:
> @meaksh thanks for the quick write up!
> 
> My main remark with the current text comes from the following considerations:
> 
>     * we do have mass registration as an open general problem since a while, [in Public Cloud contexts](https://github.com/SUSE/spacewalk/issues/10884) and otherwise. [We kind of suggest working around it in the documentation](https://github.com/uyuni-project/uyuni-docs/blob/master/modules/large-deployments/pages/operation-reqs.adoc#salt-client-onboarding-rate). This could be the right occasion to solve it once and for all, by providing one good mechanism either leveraging the existing 1-system-at-a-time API or introducing a new one, and solve a general day-1 problem many users face. Fact that systems to onboard come from an Ansible controller rather than someplace else does not change the problem's essence - I would not want to have different mechanisms for different instances of the same problem

Mass registration is a nice enhancement but not a must have for the MVP

> 
>     * onboarding Ansible end clients is not necessarily desirable in the first place. For starters, users might only really want to just running playbooks. 

We are not interested in this case. Uyuni is a systems management solution. Users who do not want to manage do not need Uyuni, they can stay with Ansible.

> Second, clients might be on platforms we do not support. 

That's a can of worms. We are not interested in that case. If Ansible controller X has 1,000 clients in inventory and 300 of those clients are on platforms we do not support, then only the other 700 clients should be registered with Uyuni.

> Third, onboarding can always happen via running a playbook - simply to run a bootstrap script! 

OK

> Fourth, users might not want to have salt-minion co-installed for non-technical reason. 

This is an opinionated Ansible integration. If they do not watn salt-minion, then they can use salt-ssh. If they do not want salt-minion nor salt-ssh, then they should not use Uyuni but stay with Ansible.

> Fifth and most important to me, they might not want to adapt their playbooks to make that happen. In fact, when a decision is made to have a system managed both with Salt and Ansible, extra care in writing both playbooks and states, otherwise they will conflict, [particularly in the area of repo management as our own internal sysadmins experimented a while ago](http://mailman.suse.de/mlarch/SuSE/suse-manager/2018/suse-manager.2018.10/msg00015.html)

They will need to take care of that. Conflicts and uncertainly can happen today too, e. g. if you subscribe to configuration channels with directly conflicting orders.

Again, this is an opinionated Ansible integration. Trying to support every possible case for everybody will be a nightmare to support and will lead to surprises, e. g. 
"User: Why is this system not being updated"
"Uyuni developers: Because you are not managing it, just running playbooks on it"
"User: oh, right, I had forgot I have a mix of managend and non-managed systems"

I do not want that to happen.


> 
>     * we are a bit short on time and want to deliver a minimal increment fast
> 
> 
> Keeping those factors in mind, I recommend moving the whole onboarding problem to a separate RFC, to be implemented once we added the ability to run playbooks, which is the core value add at the moment. Text you have created here should be kept (as it's valuable!) in a "next steps" section.

OK

accepted/00000-enable-ansible-control-node-in-registered-system.md
paususe: Why would it be limiting? On the contrary, I see this proposal providing a much better integration than the other one. The cases the registered Ansible controller makes more difficult are cases are do not really want to get into since they would provide a worse user experience or a more difficult to support integration.

accepted/00000-enable-ansible-control-node-in-registered-system.md
meaksh: That's correct, at least for now.

Beside of listing the playbooks, we probably also want the user can display the content of the playbook (as readonly) before triggering it clicking the button, but this can be done in live without storing any content in the DB:

```console
# salt ANSIBLE_CONTROL_NODE cp.get_file_str /path/to/the/playbook
```

accepted/00000-enable-ansible-control-node-in-registered-system.md
moio: Perhaps that was a wrong choice of words - what I mean is "less general" - it just poses an additional requirement (registration) that the other approach does not have.

ncounter: I'd not define it a "worse or better" user experience though, it would be just different because different is the approach and the resulting number of things one can do with an onboarded/imported system.

First bunch of differences I see (@meaksh up to you to organize them in pros and cons) :
- A registered system can be managed, an imported one can just show up in the list
- A registered system needs to be of a supported flavor by us, an imported one can be of any kind
- Replacing a system would means deleting the profile and re-registering it, an imported one simply need to keep the same IP and can be swapped

On the other side I see the "limiting" word as another way of saying "stricter integration", which is fine as long as it is intended to be. I'd say the other RFC was born on a different perspective, now we explored the two paths deeply, and we see differences. Probably it was worth for the awareness sake and for the future. We can always add the "pre-step" of import systems only. I agree it would be good to keep it as a note in the current RFC here to have a full picture in one view.

meaksh: Thanks for the feedback here!

I've just added some notes in the RFC text to highlight some of the relevant different between the two approaches without really evaluating them in Pros/Cons. Just for the record, so we can even take it into consideration for future enhacements of the current integration proposed in the present RFC.

paususe:
@meaksh is there any reason not to merge this RFC already?

meaksh:
@paususe I was waiting for a formal approval here. Ping to PR reviewers to get the approval before merging. 

@moio @mcalmer @ncounter @agraul

mcalmer (approved):

agraul (approved):

ncounter (approved):
:+1:

meaksh:
I'm merging this now. If there are more consideration we can tackle them in followup PRs.

Thanks everybody for participating here and for all the feedback you provided!

PR #56, opened 2021-05-17
PTF repositories in Uyuni

mcalmer:
request for comment

https://github.com/uyuni-project/uyuni-rfc/blob/f6963756ce63b3e38e97c5168839712784616b22/accepted/00081-ptf-repos-in-uyuni.md


mcalmer:
CC @digitaltom

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,101 @@
...
+- mandatory: false
+
+As we need to generate entries for all existing root products we can iterate over all we found.
+For every root product we take the first channel we can find which does have a `parent channel label` set.
+We can take this as parent channel label for our new entry.
+
+As the channel label and name must be unique.
+We generate the label and name from the URL components:
+
+    <Customer Account ID> <Product Identifier> <Product Version> "PTFs" ["Debuginfo"] <Product Architecture>
```
digitaltom, line 72: How about we use that name already in SCC for PTF repos, and you just take it from the API response? Then those repositories would have the same name in SCC, RMT and SUSE Manager.

HaraldMuellerNey: I think this is handled in the alternative below - with downside of parsing the full tree per organization (mirror/credentials)

mcalmer: This would help a lot. In this case we only need to append the suffix when we duplicate it for different root products.

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,104 @@
...
+[alternatives]: #alternatives
+
+Stop using product unscoped SCC endpoint and use the product scope of the requesting organization. In this endpoint we see the PTF repositories directly with its ids.
+SCC needs to be adapted to provide the channel labels and names directly in its API.
+
+Downside:
+
+- when multiple organization credentials are used, we need to request and parse multiple product scopes which cost much more time.
+- we might need to do bigger changes in the backend as we need to adapt to getting only parts of the data.
+- dependency to other product development teams which slow down the development
```
moio, line 99:
```suggestion
- in Public Clouds, PAYG products might be used - which means products the user has no SCC subscription for. Uyuni might still need product data about those
```

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,101 @@
...
+- the product id
+- the product id of the root product
+- the repository id
+
+The following values are set by the product tree:
+
+- channel label: it must be unique on the whole Uyuni Server
+- parent channel label: what is the parent of this channel
+- channel name: similar to channel label but human readable
+- update tag: a special value to modify the errata id during import
```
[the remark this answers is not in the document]
mcalmer: When SUSE starts an incident which requires a patch, the developer submit the affected package sources to the buildservice. They get build under an (internal) icident id. After QA, when the fix should be released it could happen that multiple update channels are affected. Example "salt":

- salt, salt-minion, ... belongs to sle-module-basesystem15-SP2-Updates
- salt-master, salt-api, ... belongs to sle-module-server-application15-SP2-Updates.

But the `id` in the updateinfo (patch) must be unique. The buildservice has for every release channel an id_template configured which take care for making it unique. The id_templates for this example are defined as:

- `SUSE-SLE-Module-Basesystem-15-SP2-%Y-%C`
- `SUSE-SLE-Module-Server-Applications-15-SP2-%Y-%C`

where `%Y` will be exchanged with the current Year and `%C` is a counter. It is guarantied that the counter is the same for one incident.

When Uyuni now import the  patches from different channels we could import them as single patch. This would result in a lot more patches in the DB as really exists and it could happen that users clone/use only "parts" of a fix if they select only 1 and not all for the action they want to do.

The update tag identify the part of the ID Template which make it unique and try to "rollback" the split made by the buildservice when releasing the patch. The Update Tag in the example case are:

- SLE-Module-Basesystem
- SLE-Module-Server-Applications

During reposync this part is cutted away from the ID. Both patches would end up with the ID `SUSE-15-SP2-%Y-%C`.
reposync search in the DB if an errata with this ID already exists. If yes, it merges the details together. In the DB we have now only 1 errata.

When we re-generate the metadata, we add the update tag again to the resulting updateinfo metadata.

This explains also why the "update tag" must not change. If it would change we cannot find already imported errata and would duplicate everything.

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,101 @@
...
+The defined path is:
+
+    PTF/Release/<Customer Account ID>/<Product Identifier>/<Product Version>/<Product Architecture>/ptf[_debug]/
+
+When Uyuni is listing the repositories available for an organizations, it may find repositories which `id` is not known in the complete product tree.
+To identify them as PTF repositories we can check if the path start with `PTF/Release`.
+
+If they are PTF repositories, we need to generate the missing values for the Channel mapping like this:
+
+- repository id: take it from the repository definition
```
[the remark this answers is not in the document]
mcalmer: It is the SCC repository id.

accepted/00000-ptf-repos-in-uyuni.md
mcalmer: Not sure if PAYG customers can get PTFs at all. They do not have a contract with SUSE.

accepted/00000-ptf-repos-in-uyuni.md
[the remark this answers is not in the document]
moio: Thanks, this clarifies a whole lot.

Can you please copypaste the text above in an appendix section at the end of this file? I would really love not to lose this bit of knowledge, and have a place to refer to when the inevitable questions will come up again!

("Appendix" here means it's not really under discussion as part of the RFC, but brings more context)

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,104 @@
+- Feature Name: PTF repositories in Uyuni
+- Start Date: 2021-05-17
+
+# Summary
+[summary]: #summary
+
+SUSE want to provide customers PTFs (Product Temporary Fixes) in repositories.
```
moio, line 7: Do I get correctly that this will only be for PTFs and not TEST packages, right?

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,104 @@
...
+If they are PTF repositories, we need to generate the missing values for the Channel mapping like this:
+
+- scc repository id: take it from the repository definition
+- scc product id: search for the product using `Product Identifier`, `Product Version`, and `Product Architecture` like specified in the URL
+- scc root product id: find all existing root products for the product id. We need to generate an entry for all of them
+
+The values for update tag and mandatory can be hard coded as they are the same 
+for all PTF repositories.
+
+- update tag: empty
```
moio, line 62: Are we assuming the same patch ID delivered as PTF will be later on delivered by the officially released update for all customers? Will the errata entry in Uyuni be maintained in this case, and packages "re-linked" from the PTF channel to the corresponding Update channel?

accepted/00000-ptf-repos-in-uyuni.md
moio: FTR: I checked and right, at the moment this is not possible.

accepted/00000-ptf-repos-in-uyuni.md
mcalmer: I asked back last week. The answer is "no". Also TEST patches should be delivered in these repos.
For SUSE there is a mechanism which should prevent to install it automatically.
But that requires a change in the solver which we get only for libzypp.
This means we need to talk about other OSes in this regards.

I will start a new section about handling PTFs. This is part 2 of Project Lotus, but may need to be included in the reposync discussion.

accepted/00000-ptf-repos-in-uyuni.md
mcalmer: PTFs will not get "Patches". They work package based only. Another reason why we do not need to care about the update tag. I will elaborate more about it in a new section added to this RFC

accepted/00000-ptf-repos-in-uyuni.md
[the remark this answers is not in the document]
mcalmer: done

accepted/00000-ptf-repos-in-uyuni.md
paususe: @mcalmer what do we need to talk to other OSs? to what OSs? AFAIK the ptf() is only used by SUSE operating systems :-?

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,265 @@
...
+- scc product id: search for the product using `Product Identifier`, `Product Version`, and `Product Architecture` like specified in the URL
+- scc root product id: find all existing root products for the product id. We need to generate an entry for all of them
+
+The values for update tag and mandatory can be hard coded as they are the same 
+for all PTF repositories.
+
+- update tag: empty [1]
+- mandatory: false
+
+As PTF repositories do not get patches, an update tag is not needed.
```
paususe, line 67: One could also see the PTF repository as "this is all updates for existing packages/patches". What's the impact of having or not the update tag?

mcalmer: As long as this channel does not get patches, there is no impact. Update Tags are only applied on patches.

accepted/00000-ptf-repos-in-uyuni.md
paususe: For Debian and Ubuntu, we can use `Dpkg::Pre-Invoke` to check whether the package is a PTF and ignore it in that case

https://wiki.debian.org/AptConfiguration#apt.conf

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,265 @@
...
+
+The general PTF design has 3 key points.
+
+1. A new ptf package to install the PTF.
+   
+   As this will be possible for SUSE OSes we need to check if such a package
+   can also be created for non SUSE OSes.
+   
+   - RHEL like OSes. As they also use rpm, it should be possible. We should check if RHEL 7 support the special kind of `Requires` dependencies and if yum and dnf behave like libzypp.
+   - Ubuntu and Debian. This needs some extra research if this is possible
```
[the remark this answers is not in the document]
mcalmer: yes - but I have already an idea for a solution. I think in most cases we would talk about PTFs for salt and here we have a fix list of packages which must be installed. No need for checking "if" one of the subpackages is installed or not. We can just set hard requirements which works for every package manager.

But this is not really a question for this RFC as our maintenance team need to generate these packages.

We only care about the usability in Uyuni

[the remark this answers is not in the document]
paususe: There will be also the Prometheus exporters, spacecmd and who knows what in the future. Maintaining a hardcoded list will be a burden. IMHO this can be better solved at the package level with `Dpkg::Pre-Invoke`. While the maintenance team will generate the packages, we should help them with non-SUSE OSs.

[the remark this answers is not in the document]
mcalmer: I think there is a missunderstanding. The question is, how to do the following dependency in debian?

    Requires: (pkg1 = pkg1EVR if pkg1)

It means: if `pkg1` is installed, the ptf package has a depenency `Requires: pkg1 = pkg1EVR`. If pkg1 is not installed, this dependency does not exist. It is usefull when we create a PTF with multiple packages where some of them might not be installed (they are optional).
I think we do not have this situation with our limited number of packages. It could be sufficient to just specify the direct dependency without using this boolean dependency thingy.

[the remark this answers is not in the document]
mcalmer: But that would be wrong. The goal is:

    Update all packages which are installed to a specific version,
    but ignore dependencies for packages which are not installed.

This is similar to how "patches" works.

mcalmer:
Redulced the scope of this RFC to "How to mirror PTF and TEST repositories in Uyuni".
There is a new RFC about ["How to use PTFs in Uyuni"](https://github.com/uyuni-project/uyuni-rfc/pull/59)

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,162 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+PTFs are Product Temporary Fixes for Service Requests opened by SUSE customers.
+The resulting packages are the last released maintenance update plus a patch which fixes the actual issue.
+
+PTFs are made for dedicated customers. This means every customer has different PTF repositories, with different content and fixes for different service requests.
+
+With Uyuni we should provide the possibility to mirror the content of these repositories and let the customer use it for there clients.
```
moio, line 18:
```suggestion
With Uyuni we should provide the possibility to mirror the content of these repositories and let the customer use it for their clients.
```

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,162 @@
...
+
+
+# Detailed design
+[design]: #detailed-design
+
+## How to generate the Channels from the SCC Data
+
+The mapping of SCC repositories to Uyuni Channels is done via a prepared product tree file which contains the definitions for all products and its repositories.
```
moio, line 26:
```suggestion
The mapping of SCC repositories to Uyuni Channels is done via a prepared product tree file which contains definitions for all products and their repositories.
```

accepted/00000-ptf-repos-in-uyuni.md
moio: @digitaltom I would like this RFC to be concluded whenever possible. Can we assume the generation of this label is on SCC and it will be delivered in the API at this point?

Otherwise I'd suggest adapting the text below to make clear the label generation is on Uyuni, and proceed under that assumption for now. Then, if needed, handle a change at a later point.

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Test Package Channels
+
+Before generating PTFs, SUSE will first create so called TEST packages.
+The customer need to test them and report positive result before SUSE make an official supported PTF out of it.
+
+TEST packages should only be installed on dedicated test systems.
+Therefor it is important that they get not installed on any system where just the version of the TEST package is higher than the installed one.
+
+Current idea is, that TEST packages will go into a different channel. Probably the last directory will be names "test" and "test_debug".
```
moio, line 97: @mcalmer has anything changed on this front lately?

I would ask to remove the "Current idea" language from the RFC, as it should define what is happening/decided/set at this point in time (if at all possible).

accepted/00000-ptf-repos-in-uyuni.md
```diff
@@ -0,0 +1,162 @@
...
+We would use the same rules as for PTF Channels and adjust the name and label.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- we are guessing the values. Some, like the channel label, are sensitive and must not be changed. We have to be sure that the rules and the structure never changes.
+- ptf handling for non SUSE OSes will not be as nice as for SUSE OSes.
```
moio, line 105: Can you elaborate on what will not be nice?

moio (approved):
I am by and large fine with this text, I added a few points/questions for clarity. None are dealbreakers, but would be nice to have them in if possible.

accepted/00000-ptf-repos-in-uyuni.md
digitaltom: @moio I can remember there was a talk about specific naming standard of PTF repos to help SUSE Manager. 
@ngetahun @mssola do you know if we have this tracked as feature request?

mssola: Last time I discussed this with L3 people, we agreed that for now repositories on IBS would be exactly named `ptf`. Other repositories from subprojects within `PTF:Release` would be simply ignored. Thus, for now, they will never be available as a PTF repository or any of that.

When discussing this, it was said that how to handle this scenario was not clear yet from the zypper side of things, and how to adapt all of this with APIs was also to be discussed.

bhdn:
We currently provide no Debuginfo channel. All binaries are shipped in one single repository. This also means there's no channel filtering.

accepted/00000-ptf-repos-in-uyuni.md
mcalmer: For now SUSE Manager will generate the names. We can change it later if wanted.

accepted/00000-ptf-repos-in-uyuni.md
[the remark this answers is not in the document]
mcalmer: this topic was moved to a different rfc

accepted/00000-ptf-repos-in-uyuni.md
mcalmer: I removed the "debug" channels from the RFC. Test repos might come later as this text say now. For now it is just an option with some clarifications who we would handle it.

accepted/00000-ptf-repos-in-uyuni.md
mcalmer: done

lucidd (approved):

PR #57, opened 2021-05-20
Improve the 3rd party usage of XMLRPC API

hustodemon:
At the beginning, this RFC should have been only proposing the new scheme for numbering, but when diving deeper in the topic, I've identified more problems (for example the missing guidelines in the web ui or documentation).


accepted/00080-api-versioning.md
```diff
@@ -0,0 +1,295 @@
...
+Meier
+[1](https://github.com/stevemeier/cefs/blob/master/errata-import.pl). The
+script contains a list of supported versions and checks on runtime,
+whether the API version of the Uyuni/SUMA server is contained in this
+list.
+
+This works well, until the user needs to target both Uyuni and SUSE
+Manager in their scripts - the API version numbers are not consistent
+(API version `25` in Uyuni is currently not the same as version `25`
+in SUMA).
```
mcalmer, line 78: Currently 25 only exists in Uyuni. SUMA 4.2 will get 25 as soon as it is released. But Uyuni is changing (breaking and non breaking) but keep 25. This is one problem.
SUMA 4.1 is 24, but everytime when we backport a function to 4.1 we change the API but do not change the version.
Checking if a new funtion exists in 4.1 is not possible using the version

accepted/00080-api-versioning.md
```diff
@@ -0,0 +1,295 @@
...
+
+
+#### Pros
+- Consumers able to track non-breaking changes easily (same as
+  [Solution 3](#solution-3-bump-api-version-on-any-change-in-the-api))
+
+#### Cons
+- Frequent changes in minor version could mean more burden on
+  developers and release engineers. A [CI job](#ci-job) needs to be
+  present to ease this.
```
mcalmer, line 213: Usage on the client side is also not easy. As a special feature/function could be available in 26.13 and 27.03.
A simple >= 26.13 will also match versions 27.01 and 27.02 which do not have this feature.

accepted/00080-api-versioning.md
```diff
@@ -0,0 +1,295 @@
...
+   <Fault -1: 'redstone.xmlrpc.XmlRpcFault: Could not find method:
+   test in class: com.redhat.rhn.frontend.xmlrpc.ansible.AnsibleHandler with params: []'>
+
+   # Non-existing handler
+   <Fault -1: 'The specified handler cannot be found'>
+   ```
+   Note: This fault code is
+   not reserved solely to report the API calling mismatch cases, but
+   is used in other cases (see the`CustomInfoHandler.java` file in
+   Uyuni).
```
moio, line 56: I am inclined to call this a bug. XMLRPC fault codes are meant to be unique.

Of course, this would be an API-compatibility-breaking bug :-(

accepted/00080-api-versioning.md
```diff
@@ -0,0 +1,295 @@
...
+
+### Solution 1
+
+- Treat the API versions independent in Uyuni and SUSE Manager. These
+  are 2 different projects and version `X` has a different contents
+  and compatibility in Uyuni and in SUMA.
+- Introduce an API "flavor" under a new `api.getFlavor` method, which
+  returns the project name (e.g. `uyuni`/`suse-manager`) read from a
+  config file.
+- Only "bump" the API versions on breaking changes.
```
moio, line 122: I think this RFC should propose a solution to the general problem of client scripts which want to differentiate their behavior based on any change, breaking or non-breaking (especially: the existence of a certain method that was added at some point in time).

I think we can already discard solutions that do not address this somehow.

accepted/00080-api-versioning.md
moio: @hustodemon how would you return such a version? Non exhaustive list of possibilities: as a string, as a floating point number, as a struct, via two `getMinor()`/`getMajor()` methods...

accepted/00080-api-versioning.md
```diff
@@ -0,0 +1,295 @@
...
+  present to ease this.
+
+
+### (Hypothetical) Solution 5: Enhance the introspection calls
+
+They would provide complete info about the parameters and the
+structure. For each Map or Serializer in params and we would need to
+do our best to keep the documentation consintent and the client should
+be able to verify if the structure is the expected one prior to the
+call.
```
moio, line 222: That would not cover semantic changes,right?

moio (commented):
Hey @hustodemon, that's a good start for a hairy subject like this, thanks!

I'd recommend to expand and really be specific about which problems you are looking to solve first of all, then evaluating all solutions about the completeness/UX of each solution against those problems.

Feel free to ping in case I can help you :wink:

paususe:
I do not have a good solution for the problem of having potentially 3 versions (Uyuni, SUSE Manager 4.2 and SUSE Manager 4.1) of the API in parallel, and with backports.

I feel like a capabilities-based approach is the only sane way. Instead of extending introspection to a huge level of detail, in order to deal with backwards-compatible changes (e. g. adding API calls) and to avoid the problem with opaque parameters changing the insides in an incompatible way, maybe we can version *each* API call? E. g:
* When `getFoo(param1, param2)` changes from `param2 = [int, int]` to `param2 = [int, float]`, we increase `getFoo`'s version from v7 to v8.
* When adding a new function `getBar()` to Uyuni 2021.08, we introduce it as getBar v1 (for that specific API call).
* When backporting that `getBar()` to SUSE Manager 4.2.3, we introduce it as getBar v1

That way, clients check if an API call exists, and if it exists, what version of it. Not nice but at least it seems to work in all cases.

accepted/00080-api-versioning.md
hustodemon: Same opinion here.

> Of course, this would be an API-compatibility-breaking bug :-(

We can fix it and bump the api version :wink:

accepted/00080-api-versioning.md
hustodemon: Right. With this paragraph I only wanted to say `UYUNI API 25 != SUMA API 25`. I'll fix at least the fact that SUMA does not have `25` yet.

accepted/00080-api-versioning.md
hustodemon: I suggested the definitive solution in the RFC comment, but that shouldn't be the way of doing this. Let's do it this way: I'll update the RFC and present the definitive solution as solution # 1 and we can iterate on that.

accepted/00080-api-versioning.md
hustodemon: Correct. I'll add that.

hustodemon:
@paususe I suppose we would need to implement an API method for getting versions for an API call (`int getCallVersion(methodName)`). But I also see some complications with this. An API method can't be identified uniquely by the name: we can have coexisting methods `getFoo(int)`, and `getFoo(boolean)`. We would need to bind the version to a method signature, which means the version checker would look like `getCallVersion(methodName, methodSignature)`. 

Implementing the per-method versioning shouldn't be a big problem. I've (successfully) tried out implementing a java annotation, so that in the end the calls would be annotated like this:
```java
@Version(value = 42)
public List<ShortSystemInfo> listUserSystems(User loggedInUser, String login) { ...
```
Then the `getCallVersion(method, signature)` implementation would retrieve the version using java reflection.

I see the main benefit of this approach in tracking semantic changes in the methods, but I'm not sure, if the effort of implementing this outbalances the approach of just checking the signature and catching exception.

accepted/00080-api-versioning.md
moio: Sounds good, please copypaste those observations in the text for the posterity!

accepted/00080-api-versioning.md
moio: Right - RFCs should propose one solution - the one the author thinks is best and wants to propose to the community.

Known alternatives should be listed (and, depending on context, explained at the appropriate level of detail) to inform readers that other possibilities have been explored and their pros/cons evaluated.

hustodemon:
2nd round is here. I've simplified the RFC so that it only contains one (very opinionated) solution. Please take a look, when you get a minute. I'm also available for discussion almost any time, so feel free to ping me in case you want to discuss anything further.

accepted/00080-api-versioning.md
hustodemon: In the end I decided not to go with the major/minor scheme. We can revive it when necessary.

accepted/00080-api-versioning.md
```diff
@@ -0,0 +1,214 @@
...
+- removing or adding a field in a structure accepted by a method / returned by a
+  method
+- changing the fault thrown by a method (changing a fault code, throwing new
+  faults or removing a fault)
+
+All other changes shall be tracked by the introspection methods.
+
+A CI automation job shall be written to ease tracking of such changes.
+
+
```
moio, line 188: Would you start the bumping from the current version number (25), and just increase it in whatever branch a certain change is committed to?

accepted/00080-api-versioning.md
```diff
@@ -0,0 +1,214 @@
...
+### Bumping the API version
+The version of the API is bumped when a breaking change is introduced that
+cannot tracked by introspection, which involves the following:
+- changing a behavior of an existing method
+- removing or adding a field in a structure accepted by a method / returned by a
+  method
+- changing the fault thrown by a method (changing a fault code, throwing new
+  faults or removing a fault)
+
```
mcalmer, line 183: I would add "removing" functions (deprecated) would also increase the version.

accepted/00080-api-versioning.md
```diff
@@ -0,0 +1,214 @@
...
+#### Introduce the API flavor endpoint
+Introduce a new method under the `api` namespace returning the product flavor
+(Uyuni/SUMA).
+
+
+### Bumping the API version
+The version of the API is bumped when a breaking change is introduced that
+cannot tracked by introspection, which involves the following:
+- changing a behavior of an existing method
```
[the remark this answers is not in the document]
hustodemon: I'd say not every bugfix. I would differentiate between 2 types of bugfixes:
1. the bugfix fixes some scenario that wasn't working before (say, subscribing a system to a channel that contains `-` in the label was broken and threw an exception and the fix makes it work). That means that before the fix, the user could not use the feature (or some part of it) at all.
2. the bugfix changes the behavior that could be used before (say, the API method returned fault with error code `-1`, but error code `42` would be more appropriate).

As for 1, I'm saying that  the bump is not necessary. As for 2, it is. Of course, this puts more burden on the programmer.

accepted/00080-api-versioning.md
```diff
@@ -0,0 +1,214 @@
...
+- removing or adding a field in a structure accepted by a method / returned by a
+  method
```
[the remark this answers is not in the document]
hustodemon: Agreed.

accepted/00080-api-versioning.md
hustodemon: I think this is not necessary, since the consumer  needs to check the method existence via introspection anyway (since adding new methods does not bump the API version).

accepted/00080-api-versioning.md
hustodemon: Yes. A sligthly safer alternative would be to bump the version before introducing the policies (just to be on the safe side, in case there have been some breaking changes introduced already), but I don't think it would make such a big difference, especially in the long term.

accepted/00080-api-versioning.md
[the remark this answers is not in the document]
hustodemon: I'm having a second thought about that (sorry, it's been some time I wrote the rfc).

Bumping the version only on "removing": we prevent breakage. This is necessary.

Bumping also on "adding": we allow tracking semantic changes (the consumer can be sure, that their extra parameter gets recognized by Uyuni). I can think of situations, where this would be a good thing to have.

hustodemon:
This has been stuck for a long time a I don't see myself working on this in a foreseeable future (or ever). Closing.

PR #60, opened 2021-08-11
Payg cloud

rjmateus:
This RFC specifies the implementation for suse PAYG instances on cloud.
It's the follow up RFC of https://github.com/SUSE/spacewalk/issues/14053


accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,209 @@
...
+
+we can call such a zypper plugin ("impersonating zypper") to compute URLs and headers in a trivial and cross-Cloud way.
+https://gist.github.com/moio/b064c1d8cb91a00fd4545f3625ee3911
+
+#### HTTP auth credentials
+Plain HTTP authentication credentials are required to access the repositories.
+Those credentials are stored in plain text `/etc/zypp/credentials.d/<repo_name>` and can simply be read by root.
+
+#### RMT Server IP
+RMR server domain name are not registered in DNS server, and the IP is specified on the `/etc/hosts` file of the PAYG instance.
```
mcalmer, line 87:
```suggestion
RMT server domain name are not registered in DNS server, and the IP is specified on the `/etc/hosts` file of the PAYG instance.
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,209 @@
...
+We can call the command `getent hosts <SMT_HOST_NAME>` to retrieve the IP address, which will cope with future changes to this mechanism.
```
mcalmer, line 88:
```suggestion
We can call the command `getent hosts <RMT_HOST_NAME>` to retrieve the IP address, which will cope with future changes to this mechanism.
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,209 @@
...
+### Implementation steps - Register new PAYG repositories
+
+After we get all repositories information from the PAYG instance, for each repository we will need to:
+
+#### Find internal suse Scc repository ID
+
+Repositories are identified by its URL. We can extract repository endpoint from PAYG machine and find the corresponding repository in Uyuni database. Note that the cloud RMT server have a url prefix `/repo/` which we need to remove before compare.
+
+Example:
+- URL on PAYG instance: `https:\\host/repo/SUSE/Products/SLE-Module-Basesystem/15-SP2/x86_64/product/`
```
mcalmer, line 132:
```suggestion
- URL on PAYG instance: `https://host/repo/SUSE/Products/SLE-Module-Basesystem/15-SP2/x86_64/product/`
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,209 @@
...
+- If not exists, add a new entry to table `susecredentials`, with the new authentication type `cloudrmt`, containing the repository basic authentication credentials.
+- Add a new entry to table `susesccrepositoryauth` for each repository, where:
+  - repo_id: the one obtained in previous step
+  - credentials_id: the one obtain from `susecredentials`
+  - source_id: null
+  - auth_type: `cloudrmt`
+  - auth: authentication header obtained from the PAYG instance for the repository
+
+We also need to save the repository URL, since we will not be able to use the existing SCC repository url.
+If this information will be saved in a new table field of `susesccrepositoryauth` table or in a new table is a implementation detail.
```
mcalmer, line 150: Check also for suseCredentials table. It has already a "URL" field which could be reused as well. But it also has another meaning. Maybe a good time to clean this up as well.

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,209 @@
...
+### Implementation steps - Add product with setup wizard
+
+When a product is added in the setup wizard table `rhncontentsource` is populated according to the authentication mechanism defined in table `susesccrepositoryauth` using the repository url defined in table `susesccrepository`.
+
+We need to adapt this feature to consider the new `cloudrmt` authentication mechanism and copy the basic authentication and headers:
+  - basic authentication can be defined on repo URL (`https://user:pass@host/repo/url/`)
+  - authentication header location is an implementation detail
+    - One idea is concatenate in URL as query string parameter and adapt reposync extract and use it
+
+Url to be used should be the one pointing to the cloud RMT server, saved in previous step, instead the one defined in table `susesccrepository`.
```
mcalmer, line 162: In case of multiple options, what should be prefered? I think it also depends on where Uyuni Server is running. If it is running in the cloud, the local RMT servers should be prefered, while if the Server is running outside of the Cloud, SCC should be prefered. This detail could be seen as 2nd iteration step.

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,209 @@
...
+PAYG client could download data from cloud RMT server directly.
+This will be a huge change, not sure if it is even possible to all providers.
+Possible security impact of having repository metadata publicly accessible.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- How to prevent cross product usage. For example, user with one PAYG SLES-SAP. It can start several PAYG SLES and migrate them to SLES-SAP, paying a smaller fee.
+
+- What happens and how to deal if the PAYG instance we loaded the data from get terminated or stopped?
```
mcalmer, line 207: In this case the "certificate" become invalid and RMT reject download requests. We should double check this and "test" for it and remove the affected `cloudrmt` auth configs from the Server. Channels will be kept with all the existing content and assigned server will still see it. But no new updates will added to the channels.

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,209 @@
...
+
+- Should we have a UI showing all the PAYG authentications registered and allow user to remove existing ones?
```
mcalmer, line 209: I think this makes totally sense.

moio:
One additional aspect which is missing: currently RMT instances in Public Clouds do not mirror SUSE Manager Client Tools repos, and pay-as-you-go instances do not have them among their enabled repos.

The Public Cloud Team will need to mirror those channels, and enable them to be downloaded by pay-as-you-go instances (even if it will only be Uyuni actually downloading them). Uyuni will need to enable those channels to be reposynced somehow.

accepted/0000-payg-cloud-integration.md
rjmateus: NOw I'm saving the base URL on susecredentials table, and concatenate that with suse scc repository url path, to create the repository URL to be used by reposync.

accepted/0000-payg-cloud-integration.md
rjmateus: Added it to the unresolved questions

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,198 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Uyuni/SUSE manager PAYG
+
+We could define a uyuni/suse manager PAYG image with access to all cloud RMT repositories.
```
moio, line 182: Problem: in most cases out there the Server is on-prem, managing Cloud clients.

accepted/0000-payg-cloud-integration.md
rjmateus: Since we are using the existing mechanism I think this is already handled by repo sync. More tests need to be done

rjmateus: I part of your comment as a note. Thanks

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,198 @@
...
+At the moment Uyuni can syncronize repositores from:
+  - the SCC CDN
+  - a plain RMT Server ("from-mirror setup", still requires connection to SCC)
+  - a plain directory exported from an RMT Server ("disconnected setup")
+  - a custom repository
+
+Presently, it is not possible to sync content from a SUSE-operated Cloud RMT Server, because:
+  - a Uyuni Server does not come with "tokens" (unlike PAYG instances, see above)
+  - Uyuni does not know how to pass them to RMT Servers
+  - we do not know which repositories are served via such RMT Servers
```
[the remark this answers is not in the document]
rjmateus: Should be  "Uyuni Server". I have updated it.

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,198 @@
...
+
+Our solution will be based on the following steps:
+  - Connect to PAYG instance, extract authentication data to uyuni server
+  - Teach Uyuni how to use PAYG authentication data to reposync product repositories
+
+With this solution the expected user flow would be:
+  - Provide to Uyuni ssh information to connect to the PAYG instance
+    - uyuni will extract data from the PAYG instance
+    - product of the PAYG instance will then be displayed on the Products Setup Wizard
+  - Import product from the products setup wizard
```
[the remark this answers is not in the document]
rjmateus: I'm not sure about this, since we are reusing the existing mechanism to add channels for products. Implement it via API to this PAYG solution will also allow to add product through API

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,198 @@
...
+Uyuni always gets all products meta information from SCC: which products exists and all repositories assigned.
+These products are only showed in the products setup wizard page if an authentication mechanism for the repositories is available, meaning user have access to it.
+
+The proposed solution will implement a new authentication mechanism on uyuni (named `cloudrmt`, for example) to deal with cloud RMT server authentication. With this approach we will be able to reuse the existing product/channel management features.
+
+### Implementation steps - Cloud RMT server access
+
+To access the cloud RMT server uyuni server needs to know is IP address (which is not registered in DNS) and trust the server certificate.
+
+Cloud RMT IP needs to be added to `/ets/hosts` on uyuni server. IP and hostname are returned by the data extraction script.
```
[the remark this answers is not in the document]
rjmateus: I can try to find a different workaround, but RMT server doesn't look to be registered in any DNS, so I'm not sure if we have alternatives in here.

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,198 @@
...
+
+
+### Implementation steps - Add product with setup wizard
+
+When a product is added in the setup wizard table `rhncontentsource` is populated according to the authentication mechanism defined in table `susesccrepositoryauth` using the repository url defined in table `susesccrepository`.
+
+We need to adapt this feature to consider the new `cloudrmt` authentication mechanism and copy the basic authentication and headers:
+  - basic authentication can be defined on repo URL (`https://user:pass@host/repo/url/`)
+  - authentication header location is an implementation detail
+    - One idea is concatenate in URL as query string parameter and adapt reposync extract and use it
```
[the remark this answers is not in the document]
rjmateus: I will update the RFC with this approach, by adding a new column to `susecredentials` table.

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,198 @@
...
+    - another possibility is using the existing mechanism, which reads from a configuration file: https://github.com/uyuni-project/uyuni/pull/2136
```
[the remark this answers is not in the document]
rjmateus: Some configuration is already being written to the file system for repo sync to call zypper. We have a zypper plugin in reposync to process the URL and the authentication headers, and this plugin doesn't have database access.

accepted/0000-payg-cloud-integration.md
rjmateus: For it to work, I'm expecting the on-prem server needs to be on VPN to the cloud where RMT server is located, otherwise will not be able to connect it.

accepted/0000-payg-cloud-integration.md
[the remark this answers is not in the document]
moio: Hmm, maybe we are actually talking about the same thing.

What I mean is that the Uyuni Server has a feature to "add products", which can be used either via the UI, and in that case it's called "the Setup Wizard pages", or via API, and in that case it's called "sync.content" - and now that I think about it, we also have a third way which is via the CLI, `mgr-sync`.

The way the current text reads it sounds like we are looking for a UI-only solution, while actually we aim for a "core" solution that will extend to the API as well - actually possibly API-first, IIUC.

How does that sound?

accepted/0000-payg-cloud-integration.md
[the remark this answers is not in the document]
moio: OK, in that case, I guess we are fine piggybacking the existing mechanism. Moving that to the DB, if it makes sense at all, is a separate effort not really implied by the current RFC. Let's move on.

accepted/0000-payg-cloud-integration.md
moio: Sure, there will be some kind of tunneling, I agree with that.

I am just pointing out that creating a dedicated Cloud image will not address many cases, in which the Server is on-prem.

accepted/0000-payg-cloud-integration.md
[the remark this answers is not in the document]
rjmateus: Makes totally sense. I have updated the text for clarity. Thanks

accepted/0000-payg-cloud-integration.md
rjmateus: I have updated the text to refer to this drawback also. Thanks

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,229 @@
...
+At the moment Uyuni can syncronize repositores from:
+  - the SCC CDN
+  - a plain RMT Server ("from-mirror setup", still requires connection to SCC)
+  - a plain directory exported from an RMT Server ("disconnected setup")
+  - a custom repository
+
+Presently, it is not possible to sync content from a SUSE-operated Cloud RMT Server, because:
+  - a Uyuni Server does not come with "tokens" (unlike pay-as-you-go instances, see above)
+  - Uyuni does not know how to pass them to RMT Servers
+  - Uyuni Server do not know which repositories are served via such RMT Servers
```
mcalmer, line 30: - Uyuni does not know for which repos/products the user is entitled for - which pay-as-you-go instances of which products he run

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,229 @@
...
+  - Enhance reposync [zypper plugin](https://github.com/uyuni-project/uyuni/blob/master/backend/satellite_tools/spacewalk-extra-http-headers) to receive http header. We have two implementation options:
+    - Add http headers to zypper `.repo` file as extra field. Verify if the implementation is possible and doesn't impact zypper command.
+    - Write the headers to a temporary file in the same location of zypper `.repo` file
+
+Another mechanism for [authentication headers](https://github.com/uyuni-project/uyuni/pull/2136) is already in place which loads headers from a configuration file. We should analyze if it can be merged with the new implementation.
+
+
+### Implementation steps - Refresh repository authentication data
+
+RMT authentication data have a time to live (TTL). For this reason, we need to load new authentication data from the pay-as-you-go instance periodically. For that, a new taskomatic job needs to be defined to:
```
mcalmer, line 183: Why a `new` job? Why not doing it together with the nightly refresh job. This would guarantee that we check one time per day and exactly before we refresh the data from SCC. That job also start the reposync of all vendor channels which would guarantee that the data are up to date.

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,229 @@
...
+#### Insert repository authentication data
+
+- If not exists, add a new entry to table `susecredentials`, with the new authentication type `cloudrmt`, containing the repository basic authentication credentials.
+  - URL: RMT server base URL
+  - New column to save authentication header
+- Add a new entry to table `susesccrepositoryauth` for each repository, where:
+  - repo_id: the one obtained in previous step
+  - credentials_id: the one obtain from `susecredentials`
+  - source_id: null
+  - auth_type: `cloudrmt`
```
mcalmer, line 150: Why adding a new column to `suseCredentials` instead of using `auth` column of `suseSCCRepositoryAuth` ?
Are the authentication header equal for all repos on the RMT server?
What happens if you add a SLES and SLES for SAP instance to Uyuni? The credential should be the same. Are the authentication headers the same as well?

rjmateus: > Why adding a new column to suseCredentials instead of using auth column of suseSCCRepositoryAuth`?

I decided to re-use the same mechanism we have on the basic authentication where we pass the suseCredentials ID.
If we have all the data in there it will be much simple to implement changes in repo sync and we avoid pass too much data on the URL on the `rhncontentsource` table.
It also gives the flexibility to add another authentication mechanisms that can show up in the future.

> authentication header equal for all repos? 

At the moment yes, but it can change in future, since cloud team is working on support for product addons, and with that, addon channels can have different data.

> "What happens if you add a SLES and SLES for SAP instance to Uyuni? The credential should be the same. Are the authentication headers the same as well?" 

I have tested it with two different machines with SLES15 and both basic authentication credentials and authentication headers on AWS where different.

accepted/0000-payg-cloud-integration.md
rjmateus: Great hint Micheal. I will have a look on it, and see how that would be implemented and then upate this RFC according.

accepted/0000-payg-cloud-integration.md
rjmateus: Thanks. I added this topic to the list

moio:
Another hint from @dvosburg: as a next step, in the product list, we could add an icon to distinguish products that came from Public Cloud RMT Servers instead of SCC

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+
+Uyuni needs to load repository data from external sources to provide it to the registered machines.
+At the moment Uyuni can synchronize repositories from:
+  - the SCC CDN
+  - a plain RMT Server ("from-mirror setup", still requires connection to SCC)
+  - a plain directory exported from an RMT Server ("disconnected setup")
+  - a custom repository
+
+Presently, it is not possible to sync content from a SUSE-operated Cloud RMT Server, because:
+  - a Uyuni Server does not come with "tokens" (unlike pay-as-you-go instances, see above)
```
moio, line 28: nitpick
```suggestion
  - an Uyuni Server does not come with "tokens" (unlike pay-as-you-go instances, see above)
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+  - Uyuni does not know how to pass them to RMT Servers
```
moio, line 29: nitpick
```suggestion
  - an Uyuni Server does not know how to pass them to RMT Servers
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+  - Uyuni Server do not know which repositories are served via such RMT Servers
```
moio, line 30: nitpick
```suggestion
  - an Uyuni Server does not know which repositories are served via such RMT Servers
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+  - Uyuni does not know for which repos/products the user is entitled for - which pay-as-you-go instances of which products he run
```
moio, line 31: nitpick
```suggestion
  - an Uyuni Server does not know for which repos/products users are entitled for - which pay-as-you-go instances of which products they are running
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+
+Since Uyuni cannot contact the cloud RMT servers the only option for users is to contact SUSE and get one SCC account which allows to synchronize content from the SCC CDN.
```
moio, line 33:
```suggestion
Since Uyuni cannot contact the Cloud RMT servers the only option for users is to contact SUSE and get one SCC subscription which allows them to synchronize content from the SCC CDN.
```

Note for your information: creating an SCC **account** per se is free, self-serviceable and considered not to be a problem. 
What hurts users in this scenario is the fact they have to contact SUSE to get a **subscription**, which requires calling a human (and a chain of human-based back office operations inside of SUSE).

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+The process for this is not straightforward and adds excessive complexity to users and SUSE teams.
+
+The goal for this RFC is to propose a solution to simplify this process and allow Uyuni to synchronize content directly from RMT cloud servers in a user friendly way.
```
moio, line 36: Consider as an alternative wording:

```suggestion
The goal for this RFC is to propose a solution to simplify this process and allow Uyuni to synchronize content directly from RMT Cloud servers in a self-serviceable way.
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+
+
+With this solution the expected user flow would be:
+  - Provide to Uyuni ssh information to connect to the pay-as-you-go instance
+    - uyuni will save this information on the database
+    - start a single job execution to synchronize repositories and cloud RMT authentication data
+    - product loaded from the pay-as-you-go instance will be available to import after the task finishes
+  - Import product using existing "add products" feature (available at UI, API, and cmd)
+    - reposync will be able to download all needed data from cloud RMT server
+  - Bootstrap instances using the existing methods
```
moio, line 60: Content is fine here, but I feel the explanation could be improved by imagining a reader unfamiliar with the topic.

Eg. reading the first few lines, first question that comes to mind is: why are SSH connection data necessary to begin with?

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+
+## Manage pay-as-you-go ssh connection data
+
+User should be able to register and manage ssh connection data to pay-as-you-go instances.
+Connecting to those instances, uyuni should extract repository information and authentication data.
```
moio, line 65: Try to always qualify what do we mean with Uyuni - the word itself is ambiguous (the Server? the Proxy? the whole infrastructure? the project?)

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+Users should be able to manage ssh connection data using the web UI or the XML-RPC API.
+
+Supported operations:
+  - Add new ssh connection data for pay-as-you-go
+    - should trigger an execution of the taskomatic task to extract data from the payg-as-you-go instance
+  - Remove ssh connection data
+    - Associated data in `suseCredentials` and `susesccrepositoryauth` should also be removed
+  - Update ssh connection data
```
moio, line 73: Should the task be triggered automatically also in this case?

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+  - bastion hostname
+  - bastion ssh port
+  - bastion ssh username
+  - bastion ssh password
+  - bastion ssh client certificate
+  - bastion ssh client certificate password
+
+## Taskomatic task
+
+RMT authentication data have a time to live (TTL). For this reason, we need refresh authentication data from the pay-as-you-go instance periodically.
```
moio, line 92: Is "RMT authentication data" the same as "token" above?

Please choose one jargon and use it consistently throughout the document. If needed, add a "definitions" section somewhere to list all "new jargon that gets created here" :-)

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+For that, a new taskomatic job and schedule need to be developed to:
+  - Connect to pay-as-you-go instance and retrieve the authentication data
```
moio, line 94: Will the task be fired per-instance or will it iterate through all the known instances?

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+Uyuni always gets all products meta information from SCC: which products exists and all repositories assigned.
+These products are only showed in the products setup wizard page if an authentication mechanism for the repositories is available, meaning users have access to it.
+
+The proposed solution will implement a new authentication mechanism on uyuni (named `cloudrmt`, for example) to deal with cloud RMT server authentication. With this approach we will be able to reuse the existing product/channel management features.
+
+#### Implementation step - Cloud RMT server access
+
+To access the cloud RMT server uyuni server needs to know is IP address (which is not registered in DNS) and trust the server certificate.
+
+The current Public Cloud setup requires changing `/etc/hosts` to reach the correct RMT server. This might not be possible if ever Uyuni is delivered as containers. In the constext of this RFC implementation we will follow the same approach a update `/etc/hosts` on Uyuni server.
```
moio, line 155: Please cite this concern in "drawbacks" below

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+
+Cloud RMT https certificate is also returned by the data extraction tool and on uyuni server we need to:
+  - add the certificate to folder `/etc/pki/trust/anchors/<label>`
+  - run command `update-ca-certificates`
+
+#### Implementation step - Register new pay-as-you-go repositories
+
+After we get all repositories information from the pay-as-you-go instance, for each repository we will need to:
```
moio, line 163:
```suggestion
After we get all repository information from the pay-as-you-go instance, for each repository we will need to:
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+[alternatives]: #alternatives
+
+## Uyuni/SUSE manager pay-as-you-go
+
+We could define a uyuni/suse manager pay-as-you-go image with access to all cloud RMT repositories.
+It would be possible to synchronize any product directly from cloud in a more simple and straightforward way.
+
+**Drawbacks:**
+- User will have access to repository and product he is not paying for.
+- Will not be suitable for scenarios were Uyuni is on-perm but managing cloud instances
```
moio, line 225:
```suggestion
- Will not be suitable for scenarios were Uyuni is on-premise but managing cloud instances
```

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- What happens and how to deal if the pay-as-you-go instance we loaded the data from get terminated or stopped?
+  - Possible approach is remove the affected cloudrmt auth configs from the Server. Channels will be kept with all the existing content and assigned server will still see it. But no new updates will added to the channels.
+
+- Should we have a UI showing all the pay-as-you-go authentications registered and allow user to remove existing ones?
+
+- In case of multiple options, what should be preferred? I think it also depends on where Uyuni Server is running. If it is running in the cloud, the local RMT servers should be preferred, while if the Server is running outside of the Cloud, SCC should be preferred.
```
moio, line 242: I'd suggest to always prefer the SCC CDN if available out of reliability concerns.

accepted/0000-payg-cloud-integration.md
```diff
@@ -0,0 +1,242 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+# Alternatives
+[alternatives]: #alternatives
+
+## Uyuni/SUSE manager pay-as-you-go
+
+We could define a uyuni/suse manager pay-as-you-go image with access to all cloud RMT repositories.
+It would be possible to synchronize any product directly from cloud in a more simple and straightforward way.
```
moio, line 221: "cloud", "Cloud" and "Public Cloud" are used somewhat interchangeably throughout the document. Please settle on one piece of jargon, if necessary define it  explicitly, and use it consistently.

moio (approved):
OK @rjmateus! Great job. Content-wise, I have no objections to the text.

Now I went into full-throttle nitpicking mode especially about the form - treat my comments more as an exercise in clear written communication than setting an unreasonably high barrier of entry. I know I am being a little bit unfair with you!

accepted/0000-payg-cloud-integration.md
rjmateus: We will have a taskomatic task to collect the data from each payg instance

accepted/0000-payg-cloud-integration.md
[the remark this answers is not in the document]
rjmateus: As discussed in an e-mail thread it's not possible at the moment. So we will stick to the `/etc/hots`.

accepted/0000-payg-cloud-integration.md
rjmateus: We moved to the solution of a new job to retrieve the ssh data from the payg instance periodically. With this approach we expect that even a call to reposybc on the command line will also work.

accepted/0000-payg-cloud-integration.md
rjmateus: yes, it should trigger it

accepted/0000-payg-cloud-integration.md
rjmateus: In this case I referring to: user, password, and header.
I'm not sure what of this have or not a TTL.
Even in future the logic can change, so I didn't want to get tied to referring to the token header only.

accepted/0000-payg-cloud-integration.md
rjmateus: I was thinking of firing the task and go through all the known instances.
However, the task can also be called with a list of instances as parameters to update data from. This can be then used when ssh connection data is added or updated.

accepted/0000-payg-cloud-integration.md
rjmateus: Done

accepted/0000-payg-cloud-integration.md
rjmateus: I agree with you. I remove it from unresolved questions and added it implementation requirement when adding a product.

accepted/0000-payg-cloud-integration.md
rjmateus: Done

accepted/0000-payg-cloud-integration.md
rjmateus: I have change the text to be consistent and always refer "public cloud"

PR #61, opened 2021-08-27
Image Management RFC

nadvornik:
Read it here https://github.com/nadvornik/uyuni-rfc/blob/images/accepted/00083-image_management.md

### DIFF
--- a/accepted/00083-image_management.md
+++ b/accepted/00083-image_management.md
@@ -1,0 +1,438 @@
+- Feature Name: image_management
+- Start Date: 2021-08-26
+
+# Summary
+[summary]: #summary
+
+Improve Image Management in SUMA
+
+# Motivation
+[motivation]: #motivation
+
+Current Image management is broken and incomplete. The image files are not
+referenced in the database so it is difficult to manage them.
+
+For consistent handling of images, these features are needed:
+- create / delete
+- import / export
+- consistent versioning
+
+This RFC proposes changes mainly for OS images, however Docker images share
+the code so they are affected too.
+
+## Current state
+
+This section describes details of current implementation with it's problems.
+
+### Image versioning
+
+In Kiwi, the image name and version is part of the source xml file. It can be
+affected by Kiwi options (--profile) so it can't be reliably queried before
+building.
+
+SUMA stores the built images in `suseImageInfo table`. It currently ignores
+the Kiwi name and uses SUMA profile name and `latest` version.
+
+`revision` field is a number that increases each time given `name-version`
+is built. It is not used in pillars.
+
+### `build_id`
+
+This is a string generated from Build Action Id  and is unique in each build. It is used
+as build directory on Build Host and also appended to image and pillar file names on SUMA,
+to prevent conflicts.
+
+### Image Pillars
+
+Image pillars provide information about built PXE images to branch servers
+and terminals. They are generated for the end of Inspect action.
+Pillars are organized by the Kiwi name and version.
+When the same image version is built multiple times, it generates
+multiple pillar files (different `build_id`). It is assumed that the result
+of pillar merging is the newest build, but this can be buggy.
+The names of pillar files are not stored in the database, so they are
+never deleted.
+
+`/srv/susemanager/pillar_data/images/org1/image-POS_Image_Graphical7.x86_64-7.0.0-build33.sls`
+```
+
+images:
+    POS_Image_Graphical7:
+        7.0.0:
+            arch: x86_64
+            basename: POS_Image_Graphical7.x86_64-7.0.0
+            boot_image: POS_Image_Graphical7-7.0.0
+    ...
+            sync:
+    ...
+                bundle_url: https://server.tf.local/os-images/1//POS_Image_Graphical7.x86_64-7.0.0-build33.tar.xz
+
+```
+
+### Image files
+
+Built Kiwi images are downloaded from Build host to SUMA server and moved to the public http
+directory.
+The file names are referenced in pillars (bundle_url) in the case of PXE images, they
+are not stored in the database so they are never deleted.
+
+### Delta images
+
+Delta images are created by `retail_create_delta` tool. It gets source and target
+image as a parameter and creates delta file and pillar. Delta file is stored
+along with image files. Pillar is stored along with image pillars. Deltas are not
+referenced in SUMA database, so they are invisible in UI or API.
+
+### The update workflow
+
+The current workflow assumes that the customer increases the image version
+in Kiwi sources each time the image is build with new updates. This is
+rather impractical, because this would be the only change in sources.
+Package updates are handled in SUMA Channels.
+
+### Build log files
+
+In case of failed image build, the log files are left on the build host.
+For successful builds the logs are deleted.
+There is also Kiwi stdout availableas part of Salt result from build action,
+but the formatting is lost so it is unreadable.
+
+
+### Docker images
+
+Built Docker images are uploaded to configured docker registry. Registry identifies
+the image by name and version. When a Docker image is rebuilt in SUMA, it re-uses
+the existing entry in suseImageInfo table and updates revision number.
+
+For the older docker image revisions it stores repository digests
+(suseImageHistory and suseImageRepoDigest tables).
+This allows to identify older revisions which are still in use.
+
+This implementation has the following problems:
+
+- The results can be inconsistent if the building of the new revision fails - the info about the
+last successfully built image is lost.
+
+- Package lists and other information about older revisions are not stored.
+
+
+# Detailed design
+[design]: #detailed-design
+
+
+## Proposed changes
+
+1. Use the Kiwi name and version instead of profile name in `suseImageInfo`
+   The Kiwi name is determined later in the build workflow, see Image building workflow
+   below.
+   During the build, a temporary name will be used. The final name will be set
+   at the end of build action.
+   If the build action fails, the temporary name stays.
+
+2. `revision` will be incremented each time an image with the same
+   name and version is built.
+   The image will be identified by `name-version-revision` as in
+   Build Service.
+   Add the revision to image pillar and adjust saltboot and image-sync formulas
+   to handle it.
+   The actual image file will be created before `revision`,
+   so the file name will still contain the build_id in order to make it unique.
+   For details see the Image building workflow below.
+
+
+```
+
+images:
+    POS_Image_Graphical7:  # <name>
+        7.0.0-1:           # <version>-<revision>
+            arch: x86_64
+            basename: POS_Image_Graphical7.x86_64-7.0.0
+            boot_image: POS_Image_Graphical7-7.0.0
+    ...
+            sync:
+    ...
+                bundle_url: https://server.tf.local/os-images/1//POS_Image_Graphical7.x86_64-7.0.0-build33.tar.xz
+
+```
+
+
+3. Do not re-use `suseImageInfo` entries, create new entry for each build.
+
+   Add yes_no column `obsolete` to `suseImageInfo`.
+
+   Drop table `suseImageBuildHistory`.
+
+   Change foreign key of `suseImageRepoDigest` table from `suseImageBuildHistory.id` to `suseImageInfo.id`.
+
+   For Docker the store can hold only one `name-version` image. Pushing new revision
+   to the store replaces the old one. So the old `suseImageInfo` entries can be set to `obsolete` when the
+   new image is built successfully.
+   On failed build the store still contains the previous revision and the corresponding entry
+   in `suseImageInfo` stays unchanged.
+
+   For details see the Image building workflow below.
+
+   For Kiwi it is possible to have multiple revisions of the same `name-version`. So the entries
+   can be deleted individually. Typically an old image can be deleted when the new image is
+   synced to all Branch servers.
+
+   To maintain previous functionality, add functions to 'delete all revisions'
+   and 'delete all obsolete revisions' of an image.
+
+   `suseImageBuildHistory` will need special attention during migration - it will be necessary
+   to create new row in `suseImageInfo` for each row from `suseImageBuildHistory`
+
+4. Add `pillar_id` to `suseImageInfo`.
+
+   Add table `suseImageFile` with columns
+   - `id`
+   - `image_id`
+   - `file`
+   - `type`
+   - `external`
+
+   `pillar_id` points to pillar ID in `suseSaltPillar` table.
+
+   Note: According to the RFC https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00080-jsonb-pillar-db-storage.md
+   there is a constraint `UNIQUE (org_id, category)` in `suseSaltPillar`.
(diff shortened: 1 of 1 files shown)


accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+    Make sure that API allows moving the image from SUMA to external store,
+    with adjusting pillars etc.
+
+    Implement a tool that moves the image from SUMA to external server or CDN.
+
+7. API
+
+   Add the new fields to Image.listImages and Image.getDetails.
+
+   Add methods to get/set image pillars.
```
aaannz, line 173: Why are image pillar get/set methods needed?

I can see usage for getter, but setter?

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+
+   Add a method to change `file` in SUMA store to external url.
+
+   Add a method to import Kiwi image, all the needed info will be passed 
+   as parameters, image file must exist in SUMA store before this call
+   or it can be an URL to external store.
```
aaannz, line 179: How can image already exist in SUMA store when we are trying to import it to the store?

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+
+   Eventually rename Image.importImage to Image.importDockerImage
```
aaannz, line 181: I don't know how much it this used so the actual impact may be minimal, but still sounds like API breakage.

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+   Add UI to edit image pillars.
+
+
+## Image building workflow:
+
+1. Start building
+   - create new `suseImageInfo` entry
+   - for Kiwi: set `name`, `version`, `revision` to temporary values
+   - for Docker: set `name`, `version` according to profile, generate revision
+   - schedule build action
```
aaannz, line 200: Again, why is the docker here. Isn't this only about OS images?

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+9. Pillar editor (optional)
+
+   Add UI to edit image pillars.
+
+
+## Image building workflow:
+
+1. Start building
+   - create new `suseImageInfo` entry
+   - for Kiwi: set `name`, `version`, `revision` to temporary values
```
aaannz, line 198: Please reword, this confused me wrt step 2. e.g.

```suggestion
   - for Kiwi: use temporary values for `name`, `version`, `revision`
```

Still not clear why `build_id` cannot be used.

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
+- Feature Name: image_management
+- Start Date: 2021-08-26
+
+# Summary
+[summary]: #summary
+
+Improve Image Management in SUMA
```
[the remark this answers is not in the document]
nadvornik: It proposes some changes also for container images.

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+   The Kiwi name is determined later in the build workflow, see below.
+
+2. `revision` will be incremented each time an image with the same 
+   name and version is built.
+   The image will be identified by `name-version-revision` as in 
+   Build Service.
+   Add the revision to image pillar and adjust saltboot and image-sync formulas
+   to handle it.
+   The actual image file will be created before `revision`,
+   so the file name will still contain the build_id in order to make it unique.
```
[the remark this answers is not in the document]
nadvornik: In production, build_id is a large random number.
For for user-visible part of `name-version-revision` we need a sequential revision that is increased only if name-version stays the same.

`revision` already exists in `suseImageInfo` table but it is not used in pillar.

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+    ...
+                bundle_url: https://server.tf.local/os-images/1//POS_Image_Graphical7.x86_64-7.0.0-build33.tar.xz
+
+```
+
+
+3. Do not re-use `suseImageInfo` entries, create new entry for each build.
+
+   For Docker the old entry can be deleted when the new image is built 
+   successfully.
```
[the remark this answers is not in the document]
nadvornik: It is the same code and docker needs changes too.

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+   
+   For Kiwi the entries can be deleted only manually. Typically it can
+   be deleted when the new image is synced to all Branch servers.
```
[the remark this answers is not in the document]
nadvornik: I mean remove old image when new image is synced.

accepted/00081-image_management.md
nadvornik: Main reason is adjusting URLs.

accepted/00081-image_management.md
nadvornik: SUMA store is just a directory, so I think that the file transfer does not have to go through API.

accepted/00081-image_management.md
nadvornik: no :)

rjmateus (commented):
some comment form my side

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+1. Use the Kiwi name and version instead of profile name in `suseImageInfo`
+   The Kiwi name is determined later in the build workflow, see below.
+
+2. `revision` will be incremented each time an image with the same 
+   name and version is built.
+   The image will be identified by `name-version-revision` as in 
+   Build Service.
+   Add the revision to image pillar and adjust saltboot and image-sync formulas
+   to handle it.
+   The actual image file will be created before `revision`,
```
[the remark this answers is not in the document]
nadvornik: The build_id must be passed to Kiwi before building. `name-version` is determined after building.
So we have to either keep build_id in file name or rename the file.
Renaming the file would be IMHO more error-prone, but doable too.

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+   For Docker the old entry can be deleted when the new image is built 
+   successfully.
+   
+   For Kiwi the entries can be deleted only manually. Typically it can
+   be deleted when the new image is synced to all Branch servers.
+
+4. Add `pillar` and `file` to `suseImageInfo`.
+
+   `pillar` points to pillar file or to the `category` column in `suseSaltPillars` table
+   if https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00080-jsonb-pillar-db-storage.md is implemented.
```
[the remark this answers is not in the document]
nadvornik: How can this work together with `UNIQUE (org_id, category),` ?

accepted/00083-image_management.md
```diff
@@ -0,0 +1,291 @@
...
+   `file` is a local path for images in SUMA Store or URL for
+   external images (determined by existing `external` flag).
+
+   Pillars and local image files will be deleted when the image is deleted
+   in UI or API.
+
+   `file` will be used to show image URL in UI.
+
+   Pillar entries will be deleted when the corresponding `suseImageInfo` entry is deleted.
+   This will be implemented with an sql trigger.
```
aaannz, line 196: 1) I am going a little off topic here but in my workshop draft I needed to have uncompressed bundle on the SUMA server so I could register initrd and kernel to cobbler distro. For that I got rid of bundle upload and instead uploaded KIS image individually. For this having single `file` column would not be enough. What do you think about creating new linking table image id <-> file?
2) How about not using `file`, but `URI` instead? As it could be either `file://` or URL in the future?

accepted/00083-image_management.md
```diff
@@ -0,0 +1,291 @@
...
+
+   Display deltas in Image details UI.
+
+   Deleting source or target image would delete also the delta image.
+
+6. Store build logs in the database
+
+   - add `log` column to `suseImageInfo` table
+   - collect the logs at the end of build action, use `cp.push` module
+   - add UI to display it
```
aaannz, line 214: Optional?

accepted/00083-image_management.md
```diff
@@ -0,0 +1,291 @@
...
+
+
+7. External stores
+
+    Make sure that API allows moving the image from SUMA to external store,
+    with adjusting pillars etc.
+
+    Implement a tool that moves the image from SUMA to external server or CDN.
```
aaannz, line 222: Optional, or do you envision to be present in first implementation?

aaannz (commented):
Looks very good. I have couple questions, particularly about possibility of dropping bundle download/upload.

I know that we had this discussion years ago, but with relation to possible cobbler integration having separate kernel and initrd would be needed.

accepted/00083-image_management.md
nadvornik: I think the API should be in the first implementation.
The tool is needed at least for testing the API.

accepted/00083-image_management.md
nadvornik: 1. good idea, added.
2. I think that using `file://` for local files just complicates things, so `file` is probably better name.

accepted/00083-image_management.md
nadvornik: I think it is simple enough to add it to the first implementation. Currents state regarding build logs is quite bad.

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+
+# Summary
+[summary]: #summary
+
+Improve Image Management in SUMA
+
+# Motivation
+[motivation]: #motivation
+
+Current Image management is broken and incomplete.
```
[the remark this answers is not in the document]
nadvornik: done

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+9. Pillar editor (optional)
+
+   Add UI to edit image pillars.
+
+
+## Image building workflow:
+
+1. Start building
+   - create new `suseImageInfo` entry
+   - for Kiwi: use temporary values for `name`, `version`, `revision`
```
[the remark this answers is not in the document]
nadvornik: Temporary values are used during build.
Successful end of build action indicates that the names are final.

accepted/00081-image_management.md
```diff
@@ -0,0 +1,233 @@
...
+   
+   `file` will be used to show image URL in UI.
+
+5. Add `suseDeltaImageInfo` table with columns
+   - `sourceImageID`
+   - `targetImageID`
+   - `pillar`
+   - `file`
+   
+   Add API to create/delete entries, use it in `retail_create_delta`.
```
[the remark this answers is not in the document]
nadvornik: delta images can't be generated automatically, there is too many combinations.

accepted/00083-image_management.md
```diff
@@ -0,0 +1,311 @@
...
+   external images (determined by `external` flag).
+
+   Pillars and local image files will be deleted when the image is deleted
+   in UI or API.
+
+   `file` will be used to show image URL in UI.
+
+   Pillar entries will be deleted when the corresponding `suseImageInfo` entry is deleted.
+   This will be implemented with an sql trigger.
+
```
aaannz, line 215: Same for `suseImageFiles` entries. But in that case `ON DELETE CASCADE` can be used.

accepted/00083-image_management.md
```diff
@@ -0,0 +1,311 @@
...
+   - `sourceImageID`
+   - `targetImageID`
+   - `pillar_id`
+   - `file`
+
+   Add API to create/delete entries, use it in `retail_create_delta`.
+
+   Display deltas in Image details UI.
+
+   Deleting source or target image would delete also the delta image.
```
aaannz, line 226: Sorry to make it more difficult. In case of having KIS image instead of its bundle, delta will be done only for system image, right?

So delta procedure will look for either `bundle` or `image` file type in `suseImageFiles` table for given source and target image id and create new entry just for this file. Is that correct?

Can you clarify this bit a little?

aaannz (commented):
One more thing to clarify and the rest LGTM.

accepted/00083-image_management.md
nadvornik: This is common practice, no need to mention it explicitly, IMHO.

accepted/00083-image_management.md
nadvornik: Currently we support deltas only on bundles.
Even if we extend it to KIS images, this would be an implementation detail. IMHO it is out of scope here.

accepted/00083-image_management.md
aaannz: I disagree, you mentioned pillars and deltas, why not mention that `suseImageFiles` entries would be deleted as well when related `suseImageInfo` is removed. (The part with cascade sure can be omitted)

accepted/00083-image_management.md
aaannz: I do not think it is implementation detail, but design decision. Here you are limiting delta to one file and we are already talking about future possibility to use plain KIS image which consists of 3 files.
I agree that implementing delta for plain KIS is out of scope for implementing this RFC, but in this design phase we need to address obvious development directions.

accepted/00083-image_management.md
```diff
@@ -0,0 +1,311 @@
...
+
+6. Store build logs in the database
+
+   - add `log` TEXT column to `suseImageInfo` table
```
rjmateus, line 230: How big will be this log information? Should we keep in on database or should be a file in the filesystem?

accepted/00083-image_management.md
```diff
@@ -0,0 +1,311 @@
...
+
+
+7. External stores
+
+   Make sure that API allows moving the image from SUMA to external store,
+   with adjusting pillars etc.
+
+   Implement a tool that moves the image from SUMA to external server or CDN.
+
+8. API
```
rjmateus, line 242: Limited to API you should we have UI for it also?

rjmateus (commented):
Just some comment, but overall looks good to me

accepted/00083-image_management.md
```diff
@@ -0,0 +1,311 @@
...
+            sync:
+    ...
+                bundle_url: https://server.tf.local/os-images/1//POS_Image_Graphical7.x86_64-7.0.0-build33.tar.xz
+
+```
+
+
+3. Do not re-use `suseImageInfo` entries, create new entry for each build.
+
+   Add yes_no column `obsolete` to `suseImageInfo`.
```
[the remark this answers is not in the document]
nadvornik: I think this could complicate and slow down queries.
OTOH, keeping the flag up to date should not be a problem.

accepted/00083-image_management.md
nadvornik: The log is around 1MB.
I think database is better, at least for containers.

accepted/00083-image_management.md
nadvornik: I think that these actions mainly make sense as part of more complex workflow, which should be scripted.

rjmateus (approved):
LGTM

aaannz (approved):

PR #62, opened 2021-10-13
Containerized Proxy

moio:
First in a series on containerization.

Comments welcome.

[**RENDERED VERSION**](https://github.com/uyuni-project/uyuni-rfc/blob/containerization_proxy/accepted/00083-containerized-proxy.md)


accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,82 @@
...
+
+It is believed implementing Proxy containerization will surface a subset of common problems and solutions also useful to implement a containerized Server in future.
+
+# Detailed design
+[design]: #detailed-design
+
+## Delivery
+  - Proxy is delivered as one container
+    - Updates are delivered via new images on a public Registry
+  - Documentation is produced to describe how this container can be run on popular conainer runtimes: [docker](https://docs.docker.com/engine/reference/commandline/cli/) and [podman](https://podman.io/)
```
aaannz, line 24:
```suggestion
  - Documentation is produced to describe how this container can be run on popular container runtimes: [docker](https://docs.docker.com/engine/reference/commandline/cli/) and [podman](https://podman.io/)
```

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,82 @@
...
+Having the option of deploying Proxy functionality as a container can be easier to operate (eg. deployment, independence from underlying OS, keeping up to date) and more flexible (eg. implementing active/passive HA by leveraging an existing Kubernetes environment).
+
+It is believed implementing Proxy containerization will surface a subset of common problems and solutions also useful to implement a containerized Server in future.
+
+# Detailed design
+[design]: #detailed-design
+
+## Delivery
+  - Proxy is delivered as one container
+    - Updates are delivered via new images on a public Registry
```
aaannz, line 23: So updating and management is left on customer's tools, right? How to know when to update container, what is the status (package versions, CVEs, ..) of container?

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,82 @@
...
+  - Documentation is produced to describe how this container can be run on popular conainer runtimes: [docker](https://docs.docker.com/engine/reference/commandline/cli/) and [podman](https://podman.io/)
+  - A minimal [Helm](https://helm.sh/) chart is delivered and tested against [k3s](https://k3s.io/)
+
+## Onboarding
+  - Containerized Proxy is not a managed client, does not run `salt-minion`, is not registered in the usual way
+  - A new UI page (and API endpoint) allows to add containerized Proxies by specifying their FQDN
+    - After confirmation a System entry (in the System list) is created 
+      - Normal client actions (package management, configuration file management...) will not be available on that System
+      - Proxy-specific Server UI/API functionality, like showing the list of attached clients, will be available like on regular Proxies
+    - for each such System, the Server generates a tarball ("configuration directory") that can be downloaded
```
aaannz, line 33: This would be good to do also for regular proxy, so update configuration scripts to take advantage of such tarball.
So IMO that API should work with these proxy entries and server system entries.

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,82 @@
...
+    - Proxy container needs the configuration directory to be mounted as a volume. After start, it is fully connected and operational
+
+## Operating
+  - Container will expose Salt (4505/4506) and https (443) ports
```
aaannz, line 37: I guess specifying ports will be part of the config in tarball, but IMO we should default for containers to listen on some higher ports and then do mapping in docker/podman/some ingress service.

I think it is reasonable to assume that there will be multiple services/containers running on one machine and they may want to use https as well.

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,82 @@
...
+  - Container will mount volumes:
+    - one with the configuration directory
+    - one with cached content (`/var/cache/squid/`, `/var/cache/rhn`)
+    - one with log files (`/var/log/apache2`, `/var/log/squid`, `salt-broker` log)
+  - Proxy Containers can be started/stopped/exchanged - as long as the configuration directory stays the same the Proxy identity, history and functionality will be preserved. Optionally, the cached content directory can also be preserved for optimal performance
+  - Creating bootstrap scripts from containerized Proxies will not be possible. `mgr-bootstrap --hostname=$PROXY_NAME` from the Server will still be possible
+  - SSL certificates can be handled in the two ways which are currently supported:
+    1. using the same CA as the Server, as generated at Server installation time. A new SSL certificate for the Proxy is generated at "configuration tarball" creation time. This will be the default option
+    2. using the same CA as the Server, as generated by the user ("bring your own SSL certs"). A new SSL certificate for the Proxy has to be placed in the configuration directory by the user
```
aaannz, line 46: We may want to plan with some other SSL termination than just our in container apache. At least to provide documentation and option to listen on unencrypted localhost port.

moio:
> I forgot one more question or rather clarification. What will this container image be based on? SLE15 BCI and Leap15 base or something else?

We will need Leap for Uyuni for sure. I wonder whether there is a possibility to exploit the Leap/SLE binary equality in order not to have a second image with a different base.

@mcalmer, do you foresee pitfalls in using the Leap base image + packages from `repo-sle-update` exclusively as the basis for a supported Proxy images?

accepted/00083-containerized-proxy.md
aaannz: Yes, sorry. I always mix up expose and port publishing.

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,87 @@
...
+    - After confirmation a System entry (in the System list) is created 
+      - Normal client actions (package management, configuration file management...) will not be available on that System
+      - Proxy-specific Server UI/API functionality, like showing the list of attached clients, will be available like on regular Proxies
+    - for each such System, the Server generates a tarball ("configuration directory") that can be downloaded
+    - Proxy container needs the configuration directory to be mounted as a volume. After start, it is fully connected and operational
+
+## Operating
+  - Container will expose Salt (4505/4506) and https (443) ports
+  - Container will mount volumes:
+    - one with the configuration directory
```
aaannz, line 39: Cobbler PXE files are considered as a configuration directory? (`/srv/tftpboot`)

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,87 @@
...
+  - Containerized Proxy is not a managed client, does not run `salt-minion`, is not registered in the usual way
+  - A new UI page (and API endpoint) allows to add containerized Proxies by specifying their FQDN
+    - After confirmation a System entry (in the System list) is created 
+      - Normal client actions (package management, configuration file management...) will not be available on that System
+      - Proxy-specific Server UI/API functionality, like showing the list of attached clients, will be available like on regular Proxies
+    - for each such System, the Server generates a tarball ("configuration directory") that can be downloaded
+    - Proxy container needs the configuration directory to be mounted as a volume. After start, it is fully connected and operational
+
+## Operating
+  - Container will expose Salt (4505/4506) and https (443) ports
```
aaannz, line 37: What about cobbler and network boot? I think that is expected to be supported as well for autoyast - so tftp (69/udp) exposed as well.

UEFI HTTP support https as well, but I am not sure how often will customer want to or is will to properly configure clients with CAs (I don't know how that even can be done, but it should be possible). So exposing http (80/tcp) should be considered as well. At least for specific URL path.

aaannz: Or we can have separate cobbler container? Can be viable for customers which want minimal proxy and do not network boot.
This cobbler container then can be orchestrated together with proxy container. Can be covered in separate RFC? (WDYT @SchoolGuy )

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,82 @@
...
+## Delivery
+  - Proxy is delivered as one container
+    - Updates are delivered via new images on a public Registry
+  - Documentation is produced to describe how this container can be run on popular conainer runtimes: [docker](https://docs.docker.com/engine/reference/commandline/cli/) and [podman](https://podman.io/)
+  - A minimal [Helm](https://helm.sh/) chart is delivered and tested against [k3s](https://k3s.io/)
+
+## Onboarding
+  - Containerized Proxy is not a managed client, does not run `salt-minion`, is not registered in the usual way
+  - A new UI page (and API endpoint) allows to add containerized Proxies by specifying their FQDN
+    - After confirmation a System entry (in the System list) is created
```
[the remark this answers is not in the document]
moio: Resolving per chat

accepted/00083-containerized-proxy.md
SchoolGuy: Technically they shouldn't be stored on the host as the Squid Proxy currently serves that. Thus we would need an ephemeral volume for such a thing in my eyes.

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,87 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+  - Long term this is a separate deployment mechanism to maintain and support. In future it should be decided whether keeping both is worth it
+
+# Alternatives
+[alternatives]: #alternatives
+
+ - Proxy could be split into more containers, according to the microservices pattern. This would make deployment and operations more complex, and no significant benefits have been highlighted so far
```
SchoolGuy, line 81: In my eyes the benefit is that we don't have to have process management inside the container. Which makes them smaller and easier to move around. This is especially in edge scenarios indeed an advantage I believe. However I don't want to be the judge if this advantage is big enough. This would also mean that we should optimize the layers of our containers if we are thinking in such directions.

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,87 @@
...
+  - Proxy Containers can be started/stopped/exchanged - as long as the configuration directory stays the same the Proxy identity, history and functionality will be preserved. Optionally, the cached content directory can also be preserved for optimal performance
+  - Creating bootstrap scripts from containerized Proxies will not be possible. `mgr-bootstrap --hostname=$PROXY_NAME` from the Server will still be possible
+  - SSL certificates can be handled in three ways:
+    1. using the same CA as the Server, as generated at Server installation time. A new SSL certificate for the Proxy is generated at "configuration tarball" creation time. This will be the default option
+    2. using the same CA as the Server, as generated by the user ("bring your own SSL certs"). A new SSL certificate for the Proxy has to be placed in the configuration directory by the user
+    3. leave SSL termination to the user's infrastructure altogether
+
+## Limitations
+  - Only ZeroMQ Salt clients are supported
+  - No monitoring functionality
```
SchoolGuy, line 51: This limitation really bugs me. Monitoring is a weakness in SUSE imho, so disabling this would even make it worse. If this is planed to be fixed in a later RFC I am more then happy to resolve this point without further discussion.

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,87 @@
...
+  - add Salt SSH support
+  - enable the use of "configuration directories" for VM Proxies as well
+  - Retail Branch Server: to be defined in a separate RFC
+
+## Other details
+  - Container image includes Apache, Proxy-specific Python WSGI applications and `salt-broker`
+  - base image will be Leap 15.3
+  - API/UI will need the following data to add a Proxy:
+    - FQDN
+    - http proxy hostname, username and password (optional, default from [ProxySettingsManager.getProxySettings()](https://github.com/uyuni-project/uyuni/blob/b7cf9270cc1c5edd9c15fa1217e280d645bfcef9/java/code/src/com/redhat/rhn/manager/setup/ProxySettingsManager.java#L42))
```
SchoolGuy, line 66: This is a bad idea according to my knowledge. The reason for this is that the hostname of a proxy is not stable.

- Docker: https://docs.docker.com/config/containers/container-networking/#ip-address-and-hostname
- K8s: https://kubernetes.io/docs/concepts/containers/container-environment/#container-information
- Podman: https://docs.podman.io/en/latest/markdown/podman-run.1.html#hostname-name-h

At least in the Docker usecase this means that the hostname is subject to change on image updates (and thus in a fat image package updates which happen regularly I hope).

In Kubernetes it could be that because of resource constrains or node drains the container is moved and then gets a new deployment id and thus new name.

We would need a proxy in front of our proxy who represents a stable entry point so that traffic (thus also the registration and communication of the proxy <-> server) is done in a reproducible manner.

SchoolGuy (commented):
I don't want to block this but I hope I did understand your your points correctly and my answers do make sense in such a way.

accepted/00083-containerized-proxy.md
aaannz: Then each time cobbler container or proxy container is started, cobbler sync would need to be called on SUMA, right?

Also squid proxy does not serve grub and pxe configs and kernel, initrds as they are using tftp (extept UEFI http, or course). Or did you have something else in mind?

SchoolGuy: 1. `cobbler sync` needs to be called each time we start the container - Yes and No the problem is that I hadn't have time to improve this. The sync is currently slow. Thus we have https://github.com/uyuni-project/uyuni-rfc/pull/44 . With the code of today: Yes - which is stupid because the content doesn't change and we sync to ALL proxies then AND Cobbler regenerates everything.
2. Yes TFTP is another matter and to be viewed separately from my comment.

accepted/00083-containerized-proxy.md
aaannz: Do we have second full cobbler installation? I was thinking that only proxy part of cobbler would be spun out to separate cobbler. Basically `/srv/tftpboot`, cgi scripts needed for tftpsync and tftp and http server.

SchoolGuy: Well if it is just this then we already have this. And no currently we don't have a full Cobbler installation. The proxy part of Cobbler is split out since we only use a single custom wsgi script for this at the Proxy.

moio: Point taken about exposing ports 80 and 69.

I will open a separate discussion over the possibility of using multiple containers when the rest of this RFC is set, as I am now learning about `podman` features that might change the cost/benefit balance a bit.

accepted/00083-containerized-proxy.md
moio: Process management and smaller images: points taken, will expand them better.

moio:
@aaannz, @mcalmer, @SchoolGuy: the remaining thing IMO is about this choice of one or more than one containers, and I will further expand on alternatives ASAP. In the meantime please let me know if there is anything else that does not look convincing.

SchoolGuy (approved):
I find this document much better now. I would vote for multiple containers but I can understand why this would not be desired. As a process manager inside the container in case we need one, I can recommend [supervisor](https://pypi.org/project/supervisor/).

accepted/00083-containerized-proxy.md
```diff
@@ -0,0 +1,103 @@
...
+  - A minimal [Helm](https://helm.sh/) chart is delivered and tested against [k3s](https://k3s.io/)
+
+## Onboarding
+  - Containerized Proxy is not a managed client, does not run `salt-minion`, is not registered in the usual way
+  - A new UI page (and API endpoint) allows to add containerized Proxies by specifying their FQDN
+    - After confirmation a System entry (in the System list) is created 
+      - Normal client actions (package management, configuration file management...) will not be available on that System
+      - Proxy-specific Server UI/API functionality, like showing the list of attached clients, will be available like on regular Proxies
+    - for each such System, the Server generates a tarball ("configuration directory") that can be downloaded
+    - Proxy container needs the configuration directory to be mounted as a volume. After start, it is fully connected and operational
```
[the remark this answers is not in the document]
SchoolGuy: In theory a cobbler sync should never be required. It is a "rescue" command if you "just want to start from scratch". The problem is our tftp plugin in my eyes. If we can get rid of it we should never have to call it. Normally (without SUMA) Cobbler touches the files required after a change of data automatically. #44 might be a way to get there.

aaannz:
> About using config files. Have you considered using environment variables? If you use config files, you need to make sure the container runs on a host that has those files in sync. Using environment variables, you can start the container without it. Imagine you have a big k8s cluster where you roll out the configuration via config maps or secrets. I would at least consider being able to rewrite the configuration with environment variables.

I have no prior experience with ConfigMaps, but reading the documentation it should have no problem with using and producing config files as well. ( https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#populate-a-volume-with-data-stored-in-a-configmap )

SchoolGuy:
> > About using config files. Have you considered using environment variables? If you use config files, you need to make sure the container runs on a host that has those files in sync. Using environment variables, you can start the container without it. Imagine you have a big k8s cluster where you roll out the configuration via config maps or secrets. I would at least consider being able to rewrite the configuration with environment variables.
> 
> I have no prior experience with ConfigMaps, but reading the documentation it should have no problem with using and producing config files as well. ( https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#populate-a-volume-with-data-stored-in-a-configmap )

Yeah that is true but that would not work with a basic docker-compose setup which we do want to support as written here: https://github.com/uyuni-project/uyuni-rfc/pull/62/files#diff-29675c50857585c858728a89459af6661c8e7e94dc897a41b7921d6abec668b3R26

moio:
@SchoolGuy we have several configuration files, which in this proposal would be Server-generated. True, in principle we could generate very many configuration variables instead, to be written into the right files by a helper script before the main process starts. But how do you envision "transferring" them from Server to the Proxy container host/cluster? We can certainly not output a huge list and expect users to copy and paste them.

(the idea of the "configuration tarball" comes mainly out of convenience)

SchoolGuy:
@moio This is where the fun really starts, we are getting closer to Cloud Native applications here. For Prometheus there is a good example for such a situation. They have a "Prometheus-Operator" which basically takes care of managing the application state. Thus the user is not interfacing for configuration with the application but the Operator. See here: https://prometheus-operator.dev/

This operator would then take care of the problem you just described (and more) and thus managing the transfer of said configuration as far as I interpret this pattern. I am learning and discovering all of these things in the last weeks on my private k3s cluster.

More details on the k8s operator pattern can be found here: https://kubernetes.io/docs/concepts/extend-kubernetes/operator/

The problem now is the following: There is no way this is compatible with Docker anymore. We are then deep inside Kubernetes only environments, also podman is very hard to support then. I don't see this as a problem because with k3s we have a super lightweight solution. Also there is the k3d project which runs k3s inside Docker, so docker would be back on the table via this "roundtrip".

aaannz:
> Yeah that is true but that would not work with a basic docker-compose setup which we do want to support as written here: https://github.com/uyuni-project/uyuni-rfc/pull/62/files#diff-29675c50857585c858728a89459af6661c8e7e94dc897a41b7921d6abec668b3R26

Good point, but does this limit us somehow? I mean we will prepare containers and documentation about those containers, not whole infra how to run it. We say that we require configuration files, but how users will achieve this is on their decisions. We prepare examples for docker-compose, helm, etc. where we mount volumes with configuration files. If users would want to use ConfigMaps, they are free to use k3s or full k8s. If ConfigMaps can work with files, then I think we are done here.

SchoolGuy:
> > Yeah that is true but that would not work with a basic docker-compose setup which we do want to support as written here: https://github.com/uyuni-project/uyuni-rfc/pull/62/files#diff-29675c50857585c858728a89459af6661c8e7e94dc897a41b7921d6abec668b3R26
> 
> Good point, but does this limit us somehow? I mean we will prepare containers and documentation about those containers, not whole infra how to run it. We say that we require configuration files, but how users will achieve this is on them. We prepare examples for docker-compose, helm, etc. where we mount volumes with configuration files. If users want to use ConfigMaps, they are free to use k3s or full k8s. If ConfigMaps can work with files, then I think we are done here.

This is an assumption which is not right in my mind. A container is a piece of infrastructure and we do want to tell the customer how to use it. This also is valid for container networks and other things related to that. In the past we did limit the supported situations on a bare metal installation and in the container world there is just too many ways to get to the same result in the end. The problem with this for us is that this then becomes a blackbox which is highly relevant for problem solving of the support and the developers.

Take for example the customer drops on host level the capability to do certain syscalls. We don't do them in Java but the jvm does them at the end. We have no way of knowing this if we don't tell the customer that he is not supposed to drop them. The customers & community have then a good understanding of how to use our cattle of containers if we give them the reference setup of the infrastructure.

Our colleges at Rancher also say for example that you can't use [Longhorn with spinning disks](https://longhorn.io/docs/1.2.2/best-practices/#minimum-recommended-hardware) because it wouldn't work. Technically this is not of the concern of the application because it just manages the storage which is shown to it but due to different argumentation they did say that the people using the piece of software have to meet certain requirements in the infrastructure to run the application successfully.

accepted/00083-containerized-proxy.md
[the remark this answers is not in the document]
moio: From a UX perspective I can agree with that. Problem is that all existing Proxy functionality is bolted on the assumption there is a system profile (the Proxy page, Proxy related DB tables and so on).

It is not clear to me at this point how much bigger of a change it would be not to have a system profile. We should take a look and at the very least mention your approach as an alternative in this RFC - or adopting it.

moio:
About configuration:

> Have you considered using environment variables?

Yes. Two observations:

1. the number of such variables is significant
2. some of them are security-sensitive (eg. SSL keys)

So in principle they could be exposed to the containers via environment variables + secrets. But still they need to be generated on the Server and transported somehow, and given they are many and some are confidential that would mean, either a set of files, or a YAML file with blobs in it.

Generating configuration files directly in the Server and mounting them in the container seems to me the simplest option. 

Alternatively, we would need a mechanism to take the YAML file and put it in the right config files on the container end (in K8s this can be done via `ConfigMaps` or init containers, but we would still need something different for `podman` and/or `docker`). I am unsure whether the extra decoupling is worth the effort.


About Operators, it is my understanding that this pattern does a whole lot more than just exposing configuration via native Kubernetes API (actually that part is kind of a side effect) - the main purpose is to automate management at the cluster level. From that angle, it seems way over-heavy approach for the use case we are looking at here (1-4 containers) - and we can always add it later if a stronger need appears.

moio:
About logs: addressed now.

moio:
> Alternatively, we would need a mechanism to take the YAML file [...] I am unsure whether the extra decoupling is worth the effort.

Ultimately after some discussion with @ncounter we decided to settle on the configuration via YAML file. This should make the overall approach more resilient to version upgrades, and it will be simpler to pass config via `ConfigMap`s/`Secret`s - despite being a little bit more work to create the format.

moio:
I also pushed a change to use BCI images as the starting point - they seem to be even better supported than Leap, still fully opensource, and will remove one moving piece when we will look at the commercial version for SUSE Manager.

PR #64, opened 2021-12-01
Hub reporting RFC

mcalmer:
https://github.com/uyuni-project/uyuni-rfc/blob/hub-reporting-rfc/accepted/00085-hub-reporting.md


mbussolotto:
Couple of things that are not clear for me:
- So we will have a package that it will be installed in the hub and server, right? The package will create the db and support setup, migration etc.. (both in the server and in the hub). Then another configurations created based on the choice we will have:
  * if the hub pulls the server for information (so the hub will know all the server and servers will be "dummy"). If I understand correctly, we are going in this direction.
  * if the server sends autonomously the info to the hub (in this case the server will have info about the hub and the hub will be "dummy")
- Related to the previous one: are we sure the hub should collect the data from the servers, or just store them and send them to an external DB (so the server will send info to the hub autonomously?)
- Timing is something that scares me a little bit, because it will affect the type of report available. E.g. let's supposed to have a report like "number of lock packages in all the system in the last hour": In this case the results would be almost random, because it depends on when the server sends the data (The goal is to get the data from 1000 Servers in maximal 3 hours.).

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,177 @@
...
+In summary:
+- DB in Uyuni Hub stores, collects and eventually aggregates data coming from all the DB in Uyuni Server,
+- DB in Uyuni Server stores it own data,
+- The reporting tool can be connected either to the Hub or to any Uyuni Server,
+- all connections are secure.
+
+### The Database Schema
+
+The schema should export the most important tables from the main Uyuni Database as a slightly de-normalized
+variant containing only data which are relevant for a report.
```
moio, line 46: I think it would be fair to say that main Uyuni DB schema is _very_ normalized - which is great from a DB design perspective, but generally difficult to write query against.

Thus, given the goals of this effort, I would have expected a statement like "heavily de-normalized" instead of "slightly" so - what trade-offs are you projecting to balance here?

What pros and cons do you see in denormalizing more or less?

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,177 @@
...
+
+Every row should get a report timestamp column (`exported`), which is set to the time when the data
+were exported from the main Uyuni Server database.
+
+Indexes should be set on the typical columns which might be searched for.
+
+Data which belong to the traditional stack only (like osad status) should not be made available in the
+reporting database as the traditional stack will be deprecated soon.
+
+A detailed Database schema is not part of this RFC.
```
moio, line 64: I wholeheartedly agree producing a full schema for the RFC is not appropriate, but would it help to go already into the exercise of listing what are the main tables that we expect (and roughly how many they are), or at least what objects/entities/concepts we intend to cover.

Could a shortlist be added here?

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,177 @@
...
+All the other information required (e.g. Uyuni Server info) will be retrieved by already existing configuration.
+
+## Reporting Database
+A reporting DB is setup in all Uyuni Servers and Uyuni Hub nodes to store reporting data.
+The default setup will create a Read/Write user, to manage the database and write the data from it into the Reporting DB.
+This user and the DB connection parameters are written into `/etc/rhn/rhn.conf` similar to the default DB options.
+
+## Initial Setup
+Uyuni Hub should create its own read-only user on the Reporting Databases of the single Servers.
+As the Servers are managed with salt, we will write a state to create an account.
```
moio, line 84: What pros and cons do you see compared to a solution in which the read-only user is automatically created on each Server, and then those credentials are collected via Salt by the Hub?

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,177 @@
...
+The username and password are generated on the Hub and provided as pillar data.
+On the server the state take care of the existence of the account and the Hub
+can store the parameters in its database under the system entry.
+
+# Workflow
+
+## Scheduled job: collect from Uyuni DB
+A taskomatic job is responsible to fetch and prepare the data from the main Uyuni
+Database and insert them into the Reporting Database.
```
moio, line 93: What pros and cons do you see in using [Postgres' materialized views](https://www.postgresql.org/docs/13/rules-materializedviews.html) instead of Java code to do the selectiong/processing/inserting?

The Taskomatic job would then have to call `REFRESH MATERIALIZED VIEW reportingschema.reportingtable` for each table in the reporting schema.

accepted/00000-hub-reporting.md
mcalmer: Example: de-normalizing "rhnServer". 

org_id => org name
creator_id => creator name
contact_method_id => method name

This is the kind of normalization we want to do. instead of using IDs to point to single values, we can write the values directly. We also want to remove unimportant values from the report.

But multi value fields like "network interfaces" or maybe virtualization functionality we want to provide via extra tables which still needs to be joined.

With this we let the customer the choise on "how" he want to make use of it in his report.

Not sure if we name it "slightly" or "heavy" de-normalized. Maybe we just say "de-normalized"

accepted/00000-hub-reporting.md
mcalmer: @mackdk maybe you can add what you prepared as an example.

accepted/00000-hub-reporting.md
mcalmer: If we create the password on the hub, we send it only via secured pillar data to the server and it does not need to be stored in plain text on the single server. Only the R/W credentials used on the server itself are stored in the config.
Also credentials the admin creates for his reporting tool do not need to be stored in any kind of config on the server.

I will add this to the text.

accepted/00000-hub-reporting.md
mcalmer: We design already for using an external DB for the report. Materialized Views do not work with a different database and not over network when it is also external.

accepted/00000-hub-reporting.md
mackdk: I added the example in the RFC. I don't think however that this fully answers @moio questions, as the example was build by examining only the inventory report and trying to generalise it. In order to have a better overview on what are the tables/entites/concepts we want to cover we need to scope a bit more the data and what pieces of information are valuable for the user.

accepted/00000-hub-reporting.md
moio: OK, I think I agree in principle - the goal should be to try to reduce the table count significantly somehow. Let's iterate on the detailed schema when the time comes. Thanks!

accepted/00000-hub-reporting.md
moio: OTOH, all denormalizing code would be in Java, which among other things means that columns have to be mapped and when the schema changing mapping code will also need to be adapted accordingly. Moreover Java mapping/unmapping could become a funnel performance-wise.

With views, we will know when the view breaks because of an underlying schema change, as it will not compile. Also, I see that that some trivial changes like adding a column that was not there before could be easier than changing the appropriate mapping on the appropriate classes with the appropriate types.

I see the objection of contemplating the external DB as a deployment possibility from the start and I agree it is a valid one. Still, I feel like the approaches are not mutually exclusive - we could for example still keep a (non-materialized) view on the peripheral server's schema, then copy over data from it to a (potentially external DB) plain table. That copy could be implemented "blindly" ie. just assuming that the schemas match, `SELECT` from one side and `INSERT` into the other - with no object mapping at all.

How does that sound?

mcalmer: We never had the idea to write a real hibernate object mapping. We suggested to use mode queries with select and insert.
Using a view and do a blind "select * from" and "insert into VALUES()" is a good idea in my opinion.

moio: JFYI - very very similar automated-schema-inspection and generic (safe) SELECT/INSERT code exists in the ISSv2 codebase - in Golang of course.

With this comment I am not necessarily advocating to have a self-contained Golang tool based on that code to perform the data copy - I am just mentioning it as an alternative option for your consideration.

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,199 @@
...
+
+Implementing a taskomatic simple java job should be sufficient as we need only one task which run at a
+certain point in time.
+
+As we design this feature to be ready to use an external database for reporting, we do not use materialized views.
+
+## Scheduled job: collect remote data
+We also have an additional taskomatic job which collect the data from all the managed
+Uyuni servers and insert them into the Hub Reporting Database which could be
+again an external DB.
```
admd, line 127: This will be a delta between the last sync time and the current time, a few words about "how" part in this regard, in RFC will be nice.

admd: Moreover, it will not always be the case that data will be inserted, it could be also deletion. We have solved a similar issue in ISSv2 so might be helpful to take a look at that approach.

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,199 @@
...
+
+- Q: where and how to configure the connection parameters for the reporting database
+  * Q1: on a single Server ?
+    - A1: The R/W account should be in `/etc/rhn/rhn.conf` like the other DB account
+    - A1: additional accounts should be stored where they are needed. The single server only needs the R/W account. Users which want to connect a reporting tool to it, should remember it or store it in the account management of the reporting tool.
+  * Q2: on the Hub for every managed single Server?
+    - A2: As the Hub "manages" the single Servers, it should create an own Read-only account and store it in its DB under the system item.
+
+- Q: how to identify a Uyuni Server in the Hub database and distinguish it from a normal managed client?
+  - (?): current assumption in Hub is, that all systems are Uyuni Servers.
```
admd, line 198: This is a correct assumption. Although, I can't remember if we already have some rules in place which enforce it.

accepted/00000-hub-reporting.md
mcalmer: The idea was to do a "delete all and re-insert all". I will write this.

accepted/00000-hub-reporting.md
mcalmer: Richardo told me that. We are trying to find an easy test to officially mark a system as a Uyuni/SUSE Manager Server.
Currenly I think about:
- checking for a special file
- connecting to the XMLRPC API on "localhost"
More ideas are welcome.

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,199 @@
...
+### The Database
+The main database is a PostgresDB in the Uyuni Hub system (but in the future it will be possible to 
+use also an external db): it stores all the information collected from all the server, and eventually
+aggregates them. Other databases with the same schema are also present in all the server, to collect 
+information for that system.
+All the databases (in the Uyuni Hub and in all Uyuni Server) needs to be made available on the network to either connect in a secure way (using the SSL
+certificates provided by Uyuni Server) with the reporting tool or (for Uyuni Server) with the Hub to gather the data. 
+If possible the connection to the main Uyuni DB from the outside should be forbidden.
+In summary:
+- DB in Uyuni Hub stores, collects and eventually aggregates data coming from all the DB in Uyuni Server,
```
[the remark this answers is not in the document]
mcalmer: I added another point which say this

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,208 @@
...
+### The Database
+The main database is a PostgresDB in the Uyuni Hub system (but in the future it will be possible to 
+use also an external db): it stores all the information collected from all the server, and eventually
+aggregates them. Other databases with the same schema are also present in all the server, to collect 
+information for that system.
+All the databases (in the Uyuni Hub and in all Uyuni Server) needs to be made available on the network to either connect in a secure way (using the SSL
+certificates provided by Uyuni Server) with the reporting tool or (for Uyuni Server) with the Hub to gather the data. 
+If possible the connection to the main Uyuni DB from the outside should be forbidden.
+In summary:
+- the DB in Uyuni Hub stores, collects and eventually aggregates data coming from all the DBs of the managed Uyuni Server,
```
moio, line 38:
```suggestion
- the DB in Uyuni Hub stores, collects and eventually aggregates data coming from all the DBs of the peripheral Servers,
```

For consistency's sake with existing documents, can you refer to Servers registered to a Hub as "peripheral Servers" throughout this RFC please?

accepted/00000-hub-reporting.md
moio: Can installed products data be used to this end?

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,208 @@
...
+again an external DB.
+We must parallelize the jobs to be able to gather the data from all Servers even in a large environment.
+The goal is to get the data from 1000 Servers in maximal 3 hours.
+
+The job get the list of candidates from the Hub Database including the user and connection paramaters to the
+reporting databases of the managed Uyuni Servers.
+
+The reporting DB does not use own ID columns with own sequences. To keep things simple we will
+remove all rows for the Uyuni Server identified by its `mgm_id` and insert all which we get new from the it.
+In case we see performance problems we need to re-think this approach and need to implement a more inteligent
```
rjmateus, line 140: I'm not sure if it should be part of this RFC, but with this amount of delete/insert, statistics may need to be regenerated for the target tables.

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,208 @@
...
+mechanism.
+
+## Consistency during upgrade
+The database schema on the Hub and the Uyuni Servers might differ as not all server might be updated
+at the same point in time. To support schema differences we should:
+
+- query tables on the Uyuni Server and compare them with the tables available on the hub.
```
rjmateus, line 147: This looks really close to what we will be doing in ISS. Should the synchronization part be implemented in ISS v2?
Or at least the export and import data.

accepted/00000-hub-reporting.md
rjmateus: Maybe we can see if the pattern is installed?

rjmateus (commented):
Great work. Some small questions.

accepted/00000-hub-reporting.md
mcalmer: I was thinking about future container work. What is a SUSE Manager or Uyuni Server when it comes to containers.
I think the central part is the Java Public API (XMLRPC or HTTP). So I was thinking about a connection check to localhost.
But this is not that easy.

accepted/00000-hub-reporting.md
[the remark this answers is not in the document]
mcalmer: I think here is the question what is meant with "aggregate"? We say before "collect". This already means we copy all the data from the peripheral Reporting Databases into the Hub.
I think aggregate should mean doing more with the data and currently we do not have an example where this makes sense.

All the data will be in the Hub DB. No need to connect to multiple DBs if you want to see all data.
But if we need to reduce them even more than what was done already for every single Server is unknown, but in principle possible.

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,208 @@
...
+A ready-to-use report can be provided with views doing joins over multiple tables.
+No foreign keys should be used to make data update and refreshing easier and independent from the order
+of the tables.
+
+Every table gets an extra column for the Uyuni server id of server which provide the data (`mgm_id`).
+On a single Uyuni Server this column has the standard value `1` which represent "localhost".
+On the Hub it will be replaced with the real server id the managed server has in the hub database.
+
+Every row should get a report timestamp column (`exported`), which is set to the time when the data
+were exported from the main Uyuni Server database.
```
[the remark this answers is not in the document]
mcalmer: yes, it is the time when this data was valid and correct, independent of when they were collected into the Hub DB.

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,208 @@
...
+
+Indexes should be set on the typical columns which might be searched for.
```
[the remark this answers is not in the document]
mcalmer: No idea yet. We will provide views which combine some tables for reports like we have in spacewalk-report.
Here we can already check if some indexes are usefull to speedup the view itself.

More indexed means: import is slower. But not having indexes means reports are slower.

accepted/00000-hub-reporting.md
```diff
@@ -0,0 +1,208 @@
...
+
+Data which belong to the traditional stack only (like osad status) should not be made available in the
+reporting database as the traditional stack will be deprecated soon.
+
+#### Schema example
+A detailed database schema is not part of this RFC, but as an example, the data related to servers could be 
+denormalized and presented in the reporting database in the following way:
+
+![Server schema example](images/hub-reporting/server-example.png)
```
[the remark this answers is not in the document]
mcalmer: Having 1 IP (primary) directly in Server table might be possible. But if the server has multiple we need another table to store them and this is the normal case. Then we would have the primary address 2 times or the ip address table has not all addresses.
I think this makes it harder to explain to the customers.

accepted/00000-hub-reporting.md
mcalmer: Well, you are writing files and we want to connect to the DB directly. I am not sure if we can use what you have.
We can have a look if this is usable for us.

accepted/00000-hub-reporting.md
moio: Differently from Proxy's case, I do not see Servers being something other than managed minions in the immediate future. That will mean, one of the containers will have a salt-minion instance running.

Long term we might want to remove it as we are doing with the Proxy, but that is not something I see coming for at least the first version.

accepted/00000-hub-reporting.md
moio: I tend to consider that an implementation detail, if and when we detect actual performance problems directly tied to statistics.

moio (approved):

PR #65, opened 2021-12-09
Salt-SSH with Salt Bundle

vzhestkov:
[Rendered text](https://github.com/vzhestkov/uyuni-rfc/blob/Salt-SSH-with-Salt-Bundle/accepted/00000-salt-ssh-with-salt-bundle.md)


agraul (commented):
What you wrote is a good start! I think we could improve it a bit by being more explicit about the roles that the different components play and how they interact. 

Just as an example, for the "Pre flight script" section I would start by saying something like:
> Salt can execute a shell script on the salt-ssh client before the "real salt-ssh execution" starts. This script downloads the salt-bundle package from the appropriate bootstrap repository and extracts the virtual environment.
>
> The "Pre flight script" calculates the bootstrap repository URL in a similar manner to the existing boostrap.sh script, based on the operating system and CPU architecture of the system it is executed on. It requires one parameter from the Uyuni server: `(host, port)` (entry point) for SSH connections through tunnels.
>
> ...
>
> \### Salt Changes
> - Add ability to pass parameter to pref_flight_script 

If I may suggest one stylistic change it is: prefer active verbs. Passive verbs turn the subject-verb-object structure around, it is then easy to forget the subject. Personally I like to use http://www.plainenglish.co.uk/how-to-write-in-plain-english.html as a guidance. I don't yet follow it as much as I could, though.

vzhestkov:
@SchoolGuy I pushed some changes on base of your suggestions

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,59 @@
+- Feature Name: Salt-SSH with Salt Bundle
+- Start Date: 2021-12-08
+
+# Summary
+[summary]: #summary
+
+Use Salt Bundle to handle Salt-SSH sessions on managed system side.
```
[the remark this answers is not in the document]
SchoolGuy: Your addendum improves the context a lot. Although no external documentation is linked, I am satisfied.

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,59 @@
...
+Additionally we are extending the list of working modules as Salt Bundle contains binary python modules required by some of Salt modules.
+
+This approach will help us to limit the number of supported versions of Salt codebase and drop `py26-compat-salt`, `py27-compat-salt` and all Python 2 dependencies maintenance.
+
+# Detailed design
+[design]: #detailed-design
+
+## Uyuni roster
+
+We can use Uyuni roster module to provide roster data instead of creating flat roster files with Java. PR moving the logic to Uyuni roster on Java side: https://github.com/uyuni-project/uyuni/pull/4457
```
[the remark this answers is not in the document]
SchoolGuy: Thanks for the clarification!

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Pre flight script
+
+Salt can execute a shell script on the salt-ssh client before the "real salt-ssh execution" starts. This script downloads the salt-bundle package from the appropriate bootstrap repository and extracts the virtual environment.
+
+The "Pre flight script" calculates the bootstrap repository URL in a similar manner to the existing boostrap.sh script, based on the operating system and OS architecture of the system it is executed on. It requires one parameter from the Uyuni server: (host, port) (entry point) for SSH connections through tunnels.
+
+By-default there is no way to pass additional parameters to pre fligt script running on the managed system. We need to pass the entry point (host and port) to get bootstrap repository from. The salt codebase need to be modified to pass additional parameters, otherwise we have to generate pre flight script individually for each managed system.
```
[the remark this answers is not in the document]
SchoolGuy: In the new version of this RFC I understand this part better a lot. Thanks for addressing it!

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,59 @@
...
+
+# Motivation
+[motivation]: #motivation
+
+Due to the deprecation of Python 2 and dropping support of it, while some of supported Linux systems still has no proper Python 3 version supported by Salt, we need to have a way to process Salt-SSH events on managed system side.
+
+We can deploy Salt Bundle with bundled Python inside the virtual environment and handle all the Salt-SSH events using Salt Bundle with making Salt-SSH independent from Python on the managed system.
+Additionally we are extending the list of working modules as Salt Bundle contains binary python modules required by some of Salt modules.
+
+This approach will help us to limit the number of supported versions of Salt codebase and drop `py26-compat-salt`, `py27-compat-salt` and all Python 2 dependencies maintenance.
```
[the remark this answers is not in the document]
SchoolGuy: This is also cleared up. Thanks again!

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,59 @@
...
+
+By-default there is no way to pass additional parameters to pre fligt script running on the managed system. We need to pass the entry point (host and port) to get bootstrap repository from. The salt codebase need to be modified to pass additional parameters, otherwise we have to generate pre flight script individually for each managed system.
+
+## Deploying Salt Bundle to the managed system
+
+We don't need to install Salt Bundle package on the managed system. We need to extract the virtual environment of the bundle from the package. Pre flight script is deploying it as secribed above.
+
+## Using proper salt codebase
+
+On deploying Salt Bundle with pre flight script, without modification salt-thin deployment, we will get two different codebases of salt deployed on the managed system. We can prevent deploying salt-thin in case if Salt Bundle has been deployed already and use it as preferred salt codebase to handle salt-ssh session with on managed system side.
```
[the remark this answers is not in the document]
SchoolGuy: Okay the first part is done in my eyes. But can we link what Salt-Thin is in this case!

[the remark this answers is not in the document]
vzhestkov: I think this link is outdated. `salt-thin` is a part of main salt codebase now.
Maybe just to clarify what `salt-thin` is? Like:
```suggestion
`salt-thin` is a tarball with salt codebase created with the master to deploy salt on the managed system.
```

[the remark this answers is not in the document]
SchoolGuy: Yes that is a step. I would even be happier if we have a link to the Saltstack Docs for that mechnism. Is it maybe this here? - https://docs.saltproject.io/en/latest/ref/runners/all/salt.runners.thin.html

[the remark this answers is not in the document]
vzhestkov: Bit explained in https://github.com/uyuni-project/uyuni-rfc/pull/65/commits/9519b521d6d65989aedd9895e7c6bb796f1ca2a3 and link is also there

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,59 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+- Prepare Salt Bundle for each OS and architecture on the Server, but it just make pre flight script bit smaller, the overall complexity is higher due to the bunch of additional changes on server side.
+- With not doing this the systems having not having sufficient Python version can't be managed with Salt-SSH and bootstrapped with web UI
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- The way to update Salt Bundle used for Salt-SSH on the managed system and validate it (cases when we need to redeploy it).
```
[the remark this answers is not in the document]
vzhestkov: Put one suggested solution to check for the updates in https://github.com/uyuni-project/uyuni-rfc/pull/65/commits/9519b521d6d65989aedd9895e7c6bb796f1ca2a3
But the validation of the deployed salt codebase is still unresolved.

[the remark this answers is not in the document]
SchoolGuy: For me this is a possible approach and thus my concern is resolved.

SchoolGuy (approved):
With my knowledge can understand what this RFC desires to do and thus I find this LGTM. Since we have the Bundle it is only logical to use it.

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,77 @@
...
+[summary]: #summary
+
+Use Salt Bundle to handle Salt-SSH sessions on managed system side.
+
+Salt Bundle is a single package containing salt codebase, Python and python modules required for salt.
+
+# Motivation
+[motivation]: #motivation
+
+Due to the deprecation of Python 2 and dropping support of it, while some of supported Linux systems still has no proper Python 3 version supported by Salt, we need to have a way to process Salt-SSH events on managed system side.
```
moio, line 14: Can you qualify what those systems/versions are exactly?

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,77 @@
+- Feature Name: Salt-SSH with Salt Bundle
+- Start Date: 2021-12-08
+
+# Summary
+[summary]: #summary
+
+Use Salt Bundle to handle Salt-SSH sessions on managed system side.
```
moio, line 7:
```suggestion
Use Salt Bundle to handle Salt-SSH sessions on the minion side.
```

Unless there is a reason to, I suggest to stick to "minion" everywhere in this document.

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Pre flight script
+
+Salt can execute a shell script on the salt-ssh client before the "real salt-ssh execution" starts.
+This script downloads the salt-bundle package from the appropriate bootstrap repository and extracts the virtual environment.
+By-default there is no way to pass additional parameters to pre fligt script running on the managed system.
+
+The "Pre flight script" calculates the bootstrap repository URL in a similar manner to the existing boostrap.sh script, based on the operating system and OS architecture of the system it is executed on.
+It requires one parameter from the Uyuni server: (host, port) (entry point) for SSH connections through tunnels.
+The salt codebase need to be modified to pass additional parameters, otherwise we have to generate pre flight script individually for each managed system.
```
moio, line 39: What are the pros and cons of each of the two alternatives presented here? Which one is preferred and thus proposed as part of this RFC?

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,77 @@
...
+Pre flight script is deploying it as described above.
+
+## Using proper salt codebase
+
+On deploying Salt Bundle with pre flight script, without modification salt-thin (`salt-thin` is a tarball with salt codebase created by the server to deploy salt on salt-ssh managed system: https://docs.saltproject.io/en/latest/ref/runners/all/salt.runners.thin.html) deployment, we will get two different codebases of salt deployed on the managed system (one codebase is from salt bundle package for the OS+arch and the other one is from salt-master delivered with salt-thin tarball).
+As the sources of these two codebases are different (update channel of the Server and client tools relevant for the system) these codebases are not in sync.
+We can prevent deploying salt-thin in case if Salt Bundle has been deployed already and use it as preferred salt codebase to handle salt-ssh session with on managed system side.
+In most cases the client tools channels are updating more frequently than the server and for the minions (not salt-ssh systems) we are using the salt codebase from client tools, it's better to use Salt Bundle codebase for salt-ssh systems.
```
moio, line 52: IIUC this means we do not expect to ever use the salt-thin any longer. Is that correct, or are there exceptions?

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,77 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+  * Salt Bundle consumes more space than salt-thin on managed system
```
moio, line 57: Is it possible to have a quantification, even approximate? Eg. a 20% increase vs. a 200% increase.

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,77 @@
...
+  * Salt is creating additional SSH connection to run the pre flight script 
+  * Pre flight script requires an access to the bootstrap repository (salt-thin is deployed with ssh, no http(s) connection needed for salt-ssh)
```
moio, line 59: What else is needed from the bootstrap repository other than the Salt Bundle package?

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,77 @@
...
+[alternatives]: #alternatives
+
+- Prepare Salt Bundle for each OS and architecture on the Server, but it just make pre flight script bit smaller, the overall complexity is higher due to the bunch of additional changes on server side.
+- With not doing this the systems having not having sufficient Python version can't be managed with Salt-SSH and bootstrapped with web UI
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- The way to update Salt Bundle used for Salt-SSH on the managed system and validate it (cases when we need to redeploy it).
+  It can be easily solved with including sha256 to `venv-enabled-ARCH.txt` in the root of bootstrap repo.
```
moio, line 72: It looks like this unresolved question is now resolved, or at least we have a tentative answer. Please move this information in the RFC body, leaving only really-still-open questions here.

moio (commented):
Hey @vzhestkov, nice job here! The general direction sounds good to me. I added some questions inline, mostly for refinements.

Let me know what you think

accepted/00000-salt-ssh-with-salt-bundle.md
vzhestkov: With the described approach and in my testing environment `salt-thin` is deployed only in case if salt bundle was not deployed for some reason (bootstrap repo is unavalaible or doesn't contain the bundle, or http(s) from the server is not accessible) and in case of using `salt-thin` python 3 must be installed on the system

accepted/00000-salt-ssh-with-salt-bundle.md
vzhestkov: current extracted `salt-thin` with 3 salt codebases (`3002.2` from the master, `py26-compat-salt` and `py27-compat-salt`) is about 90Mb (~30+ Mb per each salt codebase), extracted salt bundle slightly larger than 100 Mb. On removing py26- and py27- `salt-thin` will be ~40Mb.

accepted/00000-salt-ssh-with-salt-bundle.md
vzhestkov: `venv-enabled-ARCH.txt` to define the path to the Salt Bundle package. Pre flight script gets just these 2 files.

accepted/00000-salt-ssh-with-salt-bundle.md
vzhestkov: Yup, I realized it on discussing with @SchoolGuy and PR appending the hash to `venv-enabled-ARCH.txt` was already merged. Will move it.

vzhestkov: But the validation of the code deployed (is it complete and unchanged) is still unknown.

vzhestkov:
@moio I appended some details on your comments with https://github.com/uyuni-project/uyuni-rfc/pull/65/commits/e5f87d4a3b6632e6f82f8aed698c7c2b673d879b

accepted/00000-salt-ssh-with-salt-bundle.md
moio: Would it be possible / could it make simpler to transport those two files over the already-existing SSH channel to the minion instead of opening an http channel back to the Server?

vzhestkov: It requires quite significant changes in salt codebase related to salt-ssh sessions handling and impossible to do with pre flight script only. The problem is that pre flight script is used to calculate which salt bundle package should be used on base of system type and architecture. In case of using ssh do deploy these files we need few more ssh sessions to be open to the minion before, so it's more heavy from point of view of amount of open ssh connection prior the actual connection used to process the request on minion side. For this case it's better to implement something totally different than just pre flight script.

What we need:
1. Define the system type and architecture. In case of not having salt deployed yet, it's easier to do with a logic we get to the pre flight script from the bootstrap script.
2. Pass the system type and architecture to the server (salt-api in this case) and let it select the proper package to deploy on the system.
3. Extract the package on the system.

So yes, it's possible, but in this case it's better to avoid using the pre flight script feature from upstream and implement it different way with salt codebase on the server.

moio: I am fine with the currently described alternative, feel free to copypaste information from the response you gave above as an alternative in the "Alternatives" section.

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,95 @@
...
+
+We don't need to install Salt Bundle package on the minion.
+We need to extract the virtual environment of the bundle from the package.
+Pre flight script is deploying it as described above.
+
+## Using proper salt codebase
+
+On deploying Salt Bundle with pre flight script, without modification salt-thin (`salt-thin` is a tarball with salt codebase created by the server to deploy salt on the salt-ssh minion: https://docs.saltproject.io/en/latest/ref/runners/all/salt.runners.thin.html) deployment, we will get two different codebases of salt deployed on the minion (one codebase is from salt bundle package for the OS+arch and the other one is from salt-master delivered with salt-thin tarball).
+As the sources of these two codebases are different (update channel of the Server and client tools relevant for the system) these codebases are not in sync.
+We can prevent deploying salt-thin in case if Salt Bundle has been deployed already and use it as preferred salt codebase to handle salt-ssh session with on the minion side.
```
moio, line 59: As this is an RFC, you should be assertive in your proposal - the intent here is to describe what the problem is, how you would solve it and why that way is better than others rather than just proposing alternatives.

"We can" is not an assertion - your sentences should sound like "X will happen", "X will not happen" and so on (with justifications).

In this case: assuming the RFC is completely and correctly implemented, as a user, will I see salt-thin being used or not? If so, at what conditions?

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,95 @@
...
+This change can help to avoid creating temporary roster files (https://docs.saltproject.io/en/latest/topics/ssh/roster.html) and implements caching additionally.
+
+## Pre flight script
+
+Salt can execute a shell script on the salt-ssh client before the "real salt-ssh execution" starts.
+This script downloads the salt-bundle package from the appropriate bootstrap repository and extracts the virtual environment.
+By-default there is no way to pass additional parameters to pre fligt script running on the minion.
+In this case we have to generate individual pre flight script to each of the minion. The main drawback here is maintaining these pre flight script files.
+
+The other solution is to pass additional parameter to the static pre flight script.
```
moio, line 44: As this is an RFC, please select one of the solutions and present it as the chosen one, listing the other in alternatives with its pros/cons as a justification of the one you picked.

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,95 @@
...
+The pre flight script calculates the bootstrap repository URL in a similar manner to the existing boostrap.sh script, based on the operating system and OS architecture of the system it is executed on.
+It requires one parameter from the Uyuni server: host, port (entry point) for different contact methods and pathes (SSH from server/proxy, direct access to the repos or with the tunnel).
```
moio, line 46: Does this match the way normal repos are accessed by onboarded Salt-SSH minions?

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,95 @@
...
+The salt codebase need to be modified to pass additional parameters, therwise we have to generate pre flight script individually for each minion.
+
+## Deploying Salt Bundle to the minion
+
+We don't need to install Salt Bundle package on the minion.
+We need to extract the virtual environment of the bundle from the package.
```
moio, line 52: What are the pros and cons of actually installing the package?

accepted/00000-salt-ssh-with-salt-bundle.md
vzhestkov: Yes, it works the same way, the only difference that the repo is not assigned to system's package manager.

accepted/00000-salt-ssh-with-salt-bundle.md
vzhestkov: For some of the usecases users may use the minion connected to the server with venv-salt-minion, but use it with the other system as salt-ssh minion with handling salt-ssh with salt bundle. I don't think we should really support such cases. The main reason is that we are not installing any package on salt-ssh minions, so why to change this with salt bundle with salt-ssh approach? Most of the users expects salt-ssh to be agentless management method, in case of installing the package the agent will be installed, but not used. And additionally: salt bundle package also deploys pkgset beacon notification plugins, configs etc. which are not required for salt-ssh.

moio (commented):
OK @vzhestkov, great job so far. In this round of review, I am focusing much more on legibility than substance - which is a good sign, as I have less to wonder about the core concepts and more about polishing.

I think the RFC is in a good shape apart from the one remaining open question, rest is copy editing basically.

accepted/00000-salt-ssh-with-salt-bundle.md
```diff
@@ -0,0 +1,96 @@
...
+## Deploying Salt Bundle to the minion
+
+We don't need to install Salt Bundle package on the minion as users concider salt-ssh as an agentless contact method and it's better not to install any service on the system even if it will never been started.
+It can also help for the cases when venv-salt-minion is used on the system for any other purpose.
+Pre flight script extracts only virtual environment of the bundle from the package.
+Pre flight script is deploying it as described above.
+
+## Using proper salt codebase
+
+On deploying Salt Bundle with pre flight script, without modification salt-thin (`salt-thin` is a tarball with salt codebase created by the server to deploy salt on the salt-ssh minion: https://docs.saltproject.io/en/latest/ref/runners/all/salt.runners.thin.html) deployment, we will get two different codebases of salt deployed on the minion (one codebase is from salt bundle package for the OS+arch and the other one is from salt-master delivered with salt-thin tarball).
```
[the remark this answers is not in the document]
vzhestkov: We excpect Salt Bundle to be updated more frequently in the repo than the salt codebase installed on the master. Mostly due to the customer usecases when client tools are updating automatically and bootrap repos are regenerating after reposync and most of the customers have some delay in updating the server. In this case we can expect salt bundle codebase more fresh. We can't be 100% due to the different usecases and update policies on customer side. But in case if there is no Salt Bundle in the bootstrap repo for some reason, or pre flight script unable to reach it due to some network issues or limitations, we will add extra logging to salt-api process to show that pre flight script returns some extra data. (Currently salt is not checking the output of the script, but only fails and show error message if pre flight script returns something other than 0). But finally we will use `salt-thin` as a fallback. The main drawback here that `salt-thin` in general is limited in number of modules shipped, and in case of getting rid of `py2X-compat-salt` it will work only with Python 3. On fails of handling salt-ssh sessions with salt-thin, salt is producing error on master side showing the root cause of the issue.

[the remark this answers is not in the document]
vzhestkov: Let's discuss some additional details regarding this point.
Now we have two options:
1. Deploy Salt Bundle on the minion and fail if it's not available for some reason and do not use `salt-thin` at all. (we can just show an error message from the pre flight script)
2. Deploy Salt Bundle on the minion if possible (do not fail if not) and use `salt-thin` as a fall back method and get the error only if `salt-thin` fails.

With option 1 we can cause the issues for some of the customers. Here are the possible cases:
- The customer disabled auto bootsrap repo creation for some reason (it's mostly possible on migrating from previous version) and bootstrap repo was not updated with Salt Bundle at all. Bootstrapping new clients and handling of the sessions for already registered salt-ssh system for the OS will not be possible until bootstrap repo recreated manually.
- The customer is using salt-ssh with not supported OS and have no bootstrap repo for it at all.

The main possible drawback of the 2nd option: it's possible that some event can be handled different ways with Salt Bundle and `salt-thin` as the last one is limited in the functionality and is not shipped with all required modules inside. Additionally it builds with the salt codebase of the master, but not from the client tools (we expect client tools should be updated more frequently). As @moio pointed it could add some headache on debugging salt-ssh issues the customers have.

@meaksh @agraul @SchoolGuy not sure if I listed all of the cases we mentioned on the call.

The possible solution from me: we can use 1st option as default, but make it configurable to allow to switch to 2nd if the customer really need it for some reason.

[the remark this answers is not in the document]
agraul: I think you cover the gist of what we discussed. I would like to point out that "customer using salt-ssh with not supported OS and have no bootstrap repo for it at all" is not only a potential problem for SUMA customers but also for uyuni contributors who want to enable new operating systems. While the former group is not supposed to use unsupported operating systems, we encourage the latter group to do exactly that. 

To me it makes sense to not have an automatic fallback but a manual "use salt-thin" configuration option, which we declare unsupported in SUMA.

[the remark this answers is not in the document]
moio: > To me it makes sense to not have an automatic fallback but a manual "use salt-thin" configuration option, which we declare unsupported in SUMA.

To me this sounds like a fair balancing act to start with. We can always change the policy and make that fully supported later, in case we observe a high volume of requests because of conditions we cannot predict today.

meaksh (approved):
After the latest discussions and refinements on the text, the current version looks good to me. Thanks!

admd:
@vzhestkov @meaksh  If RFC has been finalized and one has anything against it, can we merge it.

PR #66, opened 2021-12-10
Hub data reporting changes

mbussolotto:
To see how the changes looks: https://github.com/mbussolotto/uyuni-rfc/blob/hub-reporting-rfc/accepted/00000-hub-reporting.md


accepted/00000-hub-reporting.md
```diff
@@ -70,20 +56,50 @@ from the main Uyuni Server database.
...
+- to setup and manage the reporting database,
+- to setup and manage DBs in all Uyuni Server.
+
+The management tools should support:
+- Database setup and schema initialization
+- Database schema migration
+- Simple User management. The default setup will create two users:
+  * a Read/Write user, to manage the database and write the data from the main Uyuni Database into the Reporting DB. This user and the DB connection parameters are written into `/etc/rhn/rhn.conf` similar to the default DB options.
+  * a Read-Only user, used by the reporting tools for gathering information from the hub database.
+The package will also setup a taskomatic job to retrieve informations from all the other SUSE Manager Server
```
mackdk, line 74: Formatting stuff: I think we need an empty line here as this two paragraphs are not separated from the list in the preview I see at the moment.

accepted/00000-hub-reporting.md
```diff
@@ -70,20 +56,50 @@ from the main Uyuni Server database.
...
-## Adding Data to the Reporting DB
+We provide a new package installed in the hub system:
+- to setup the reporting DB schema,
+- to setup and manage the reporting database,
+- to setup and manage DBs in all Uyuni Server.
+
+The management tools should support:
+- Database setup and schema initialization
+- Database schema migration
+- Simple User management. The default setup will create two users:
```
mackdk, line 71: Do we want this management feature to allow creating both read-only and read-write users or only the former?

accepted/00000-hub-reporting.md
```diff
@@ -70,20 +56,50 @@ from the main Uyuni Server database.
...
+  * a Read/Write user, to manage the database and write the data from the main Uyuni Database into the Reporting DB. This user and the DB connection parameters are written into `/etc/rhn/rhn.conf` similar to the default DB options.
+  * a Read-Only user, used by the reporting tools for gathering information from the hub database.
+The package will also setup a taskomatic job to retrieve informations from all the other SUSE Manager Server
+database.
+All the other information required (e.g. Uyuni Server info) will be retrieved by already existing configuration.
+
+### Account management on the Hub for the Reporting Databases of the managed Servers
+
+During installtion, hub should create its own user on the Reporting Databases of the single Servers.
```
mackdk, line 80: Just to verify that I have understood this correctly, this is an additional users, created by the hub server on each servers in addition to the two default one mentioned in the paragraph above. And this is read only as well, I guess, since the hub does not need to change the data. I also would guess that the "Simple User management" tool mentioned above should not alter this hub account in any way, otherwise the sync will break.

accepted/00000-hub-reporting.md
mbussolotto: I haven't a strong opinion on that, but if I need to choose I would say both read-only and read-write, in order to provide all the users required by the reporting tool

accepted/00000-hub-reporting.md
mbussolotto: I tried to add more info but yes, probably we should rephrase the user part at all and make it more clear

accepted/00000-hub-reporting.md
mcalmer: I think we should automatically create only the read/write user as we need it directly.
The hub should create his own user when connecting to the Server DB. We described it below.
We should document how to create a user for a reporting tool and the admin decide if he do this on the single server or the hub or both/all.
What do you think?

mackdk: Yes indeed makes sense to leave the final user full control on their accounts, so +1 also for me.

mbussolotto: Changed this section and add more information below regarding how to use data collected in a reporting tool and documentation

accepted/00000-hub-reporting.md
```diff
@@ -26,30 +26,16 @@ let us stay with the good well known PostgreSQL database.
...
-and login data.
-The database needs to be made available on the network to either connect with the reporting tool
-or with the Hub to gather the data.
-The connection should be secured with SSL using the certificates we configure anyway for Uyuni Server.
+### The Database
+The main database is a PostgresDB in the hub system (but in the future it will be possible to 
+use also an external db): it stores all the information collected from all the server, and eventually
+aggregates them. Other databases with the same schema are also present in all the server, to collect 
+information for that system.
+The hub database needs to be made available on the network to either connect in a secure way (using the SSL
```
mcalmer, line 34: I think not only the hub DB needs to be made accessible from the outside. I also every reporting DBs for the single server need that. At least if we want to syncronize the data using JDBC connections.

accepted/00000-hub-reporting.md
```diff
@@ -70,26 +56,58 @@ from the main Uyuni Server database.
...
 
+# Installation
+The reporting tools will be installed by default; all the upgrade to a newer version, will install the 
+reporting tools as well.
 
-## Adding Data to the Reporting DB
+We provide a new package installed in the hub system:
+- to setup the reporting DB schema,
+- to setup and manage the reporting database,
+- to setup and manage DBs in all Uyuni Server.
```
mcalmer, line 66: Is this only for the Hub ? I think this is generic as we try to keep the reportig DB on a single server compatible with the reporting DB on the Hub.

accepted/00000-hub-reporting.md
```diff
@@ -70,26 +56,58 @@ from the main Uyuni Server database.
...
+- Database schema migration
+- Simple User management. The default setup will create a Read/Write user, to manage the database and write the data from into the Reporting DB. This user and the DB connection parameters are written into `/etc/rhn/rhn.conf` similar to the default DB options.
+
+### Uyuni Server
+On the server side, during installation, hub should create its own read-only user on the Reporting Databases of the single Servers. 
+This user is used by the hub to read data from the Uyuni Server DB.
+As the Servers are managed with salt, we will write a state to create an account.
+The username and password are generated on the Hub and provided as pillar data.
+On the server the state take care of the existance of the account and the Hub can store the parameters
+in its database under the system entry.
```
mcalmer, line 86: Maybe we should better state, that Hub and Single Server is the same and only name real differences?

accepted/00000-hub-reporting.md
```diff
@@ -70,26 +56,58 @@ from the main Uyuni Server database.
...
+# Workflow
+
+## Uyuni Server: retrieve and organize Data
 
 On an Uyuni Server a taskomatic job is responsible to fetch and prepare the data from the main Uyuni
-Database and insert them into the Reporting Database.
-The job should be written in a way, that Uyuni and Reporting Database could be on different Hosts.
+Database and insert them into the Reporting Database. The infomation will be stored in a local DB and
+collected by Uyuni Hub.
+The job should be written in a way, that Uyuni Hub and Reporting Database could be on different Hosts.
```
mcalmer, line 95: The Bug is always on a different server. 
```suggestion
The job should be written in a way, that the Reporting Database could be on different Host.
```

accepted/00000-hub-reporting.md
mbussolotto: right, tried to rephrase it

accepted/00000-hub-reporting.md
mbussolotto: Same as above, I ll try to rephrase it to made more clear the single-server scenario

PR #67, opened 2022-02-08
RFC: HTTP API

cbbayburt:
Provide an HTTP API exposing the existing methods via plain JSON, in addition to the XMLRPC protocol.

See the [rendered version](https://github.com/uyuni-project/uyuni-rfc/blob/http-api/accepted/00000-http-api.md).


accepted/00000-http-api.md
```diff
@@ -0,0 +1,214 @@
...
+
+The actual handler methods are designed to use POJOs as parameters. However, Spark expects a specific signature that takes `request` and `response` as parameters for route handlers. A simple wrapper shall unpack the parameters from the request and call the appropriate handler method for each request.
+
+A typical resource-oriented REST API relies on various HTTP verbs to provide contextual information on what type of operation an endpoint is used for. However, the XMLRPC protocol is designed around methods, rather than resources. Therefore, HTTP verbs will be restricted to GET for read-only, and POST for read-write operations regardless of whether they require a payload or not.
+
+Spark provides 3 common ways to attach data to a request. Named route parameters that are part of the base URL, querystring parameters, or the payload itself.
+
+For primitive types, either route parameters or querystring parameters may be used. The desired method shall be specified using a parameter annotation in the handler method signature (explained below).
```
Etheryte, line 54: I think simply don't understand the rationale here: why do we want to support both, not only one? My initial gut feeling is that picking one and sticking to it would make both documentation and examples easier to read, as well as code more consistent (both for our use and for external parties).  

Similarly for 

> For complex types, the request payload of type JSON must be used.  

I don't know what complex types mean in this context, are there any specific things for which we don't/won't/can't support query strings for example? File uploads and the like come to mind, but I think either some context or an example or two would help a lot for understanding here.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,214 @@
...
+
+
+## 3. I/O
+
+On the HTTP API side, I/O serialization is mostly handled by Spark, and the existing `GSON` library. XMLRPC uses specialized "serializer" classes to write structured output on XML. These serializer classes shall be used exclusively for the XMLRPC output.
+
+
+## 4. Exception Handling
+
+The current implementation of the business logic uses exceptions that are designed specifically for XMLRPC (see: `FaultException` and subclasses). To be able to use the business logic with both interfaces, a new, internal exception structure shall be designed to offer enough flexibility in exception handling. These exceptions are then translated to the specific interface, and processed accordingly. For HTTP endpoints, the internal exceptions can be represented by different HTTP status codes where appropriate.
```
Etheryte, line 190: No specific comment here, just a note that would be good to keep in mind if we can standardize/centralize this in the current network layer on the frontend. Currently, error and error message handling is a bit of a bazaar where everything is mix and match.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,214 @@
...
+
+
+## 5. Security
+
+The HTTP interface reuses the existing Spark framework. Therefore, we're typically not exposed to any new security issues. However, with the introduction of the new API, existing interface will be more critical for external use, so it would be beneficial to propose an approach to security testing.
+
+
+## 6. Documentation
+
+APIDOC is an extensive tool that automatically generates documentation for the API endpoints written in many languages. It can be used to generate the documentation for the HTTP API as well.
```
Etheryte, line 202: Out of curiosity and a complete tangent: could this possibly also output Typescript types in the future?

accepted/00000-http-api.md
```diff
@@ -0,0 +1,214 @@
...
+More research is required to provide the framework for creating documentation for both interfaces with minimum duplication.
+
+
+# Unresolved Questions
```
Etheryte, line 206: One topic I would like to see added is versioning. If we don't plan to do anything specific for versioning, it would also be good to write that down along with a rationale.

Etheryte:
In general, I didn't see anything I don't like, at most some ambiguities that could be cleared up. Thanks a lot for putting the time and effort into this, looks very promising. I have no input on the Java annotation logic etc.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,214 @@
...
+
+
+# Detailed Design
+
+The design aims to provide a plain HTTP endpoint to the existing methods by reusing the available business logic with minimum redundancy. The bulk of the effort is to decouple the existing layers from protocol-dependent parts as much as possible, so that the most of the workflow can be utilized for both interfaces in an agnostic manner.
+
+![Current workflow](images/http-api-before.png)
+*Current workflow*
+
+The components marked with red are tightly coupled to the XMLRPC-specific functionality, and cannot be reused with HTTP as-is. In contrast, the green parts are agnostic, and can be used for both interfaces.
```
mcalmer, line 23: Nitpick

On my screen i do not see anything in "green". It looks just black

accepted/00000-http-api.md
```diff
@@ -0,0 +1,214 @@
...
+
+On the HTTP side, Spark allows registration of an endpoint as a "route" per each endpoint method, optionally including a number of named route parameters that construct the route path.
+
+The actual handler methods are designed to use POJOs as parameters. However, Spark expects a specific signature that takes `request` and `response` as parameters for route handlers. A simple wrapper shall unpack the parameters from the request and call the appropriate handler method for each request.
+
+A typical resource-oriented REST API relies on various HTTP verbs to provide contextual information on what type of operation an endpoint is used for. However, the XMLRPC protocol is designed around methods, rather than resources. Therefore, HTTP verbs will be restricted to GET for read-only, and POST for read-write operations regardless of whether they require a payload or not.
```
mcalmer, line 50: Do we not use DELETE requests for removing things?

accepted/00000-http-api.md
```diff
@@ -0,0 +1,214 @@
...
+## 3. I/O
+
+On the HTTP API side, I/O serialization is mostly handled by Spark, and the existing `GSON` library. XMLRPC uses specialized "serializer" classes to write structured output on XML. These serializer classes shall be used exclusively for the XMLRPC output.
+
+
+## 4. Exception Handling
+
+The current implementation of the business logic uses exceptions that are designed specifically for XMLRPC (see: `FaultException` and subclasses). To be able to use the business logic with both interfaces, a new, internal exception structure shall be designed to offer enough flexibility in exception handling. These exceptions are then translated to the specific interface, and processed accordingly. For HTTP endpoints, the internal exceptions can be represented by different HTTP status codes where appropriate.
+
+This is the main point that couples the business logic with a specific interface, and will require the most work to implement.
```
mcalmer, line 192: The changes in Exceptions is a bit unspecific. I wonder if it is not possible to use the existing exceptions and just enhance them with a HTTP status code. Depending on what layer is used to return it, it takes the fault or the http status.

accepted/00000-http-api.md
parlt91: One example that comes to my mind here would be:
```
List<RescheduleResult> updateCalendar(User loggedInUser, String label, Map<String, String> details,
            List<String> rescheduleStrategy)
```
in `MaintenanceHandler.java`

The `details` parameter can contain a potentially very large string (stringified ical file) as well as other additional parameters. In this case I guess using querystrings might not be feasible anymore. (But I don't know about limitations when it comes to URL length, but where talking multiple thousand characters here)

But I agree in with you in general at least personally I don't see any reason to further complexify this by using params in the base URL as well and we should IMHO stick with `?query=params` wherever possible.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,214 @@
...
+Consider the following XMLRPC handler method:
+```java
+public ContentProject updateProject(User user, String label, Map<String, Object> props)
+```
+
+With the proposed annotations, it may be written as:
+```java
+@XmlRpcApiEndpoint
+@HttpApiEndpoint
+public ContentProject updateProject(User user, @RouteParam String label, Map<String, Object> props)
```
parlt91, line 141: I really wonder if there is some way to automate this. I think adapting every single handler method by hand would be a major pain this also applies to the API-docs (it may be as simple as to just add both in all the cases and only care about potential exceptions).
Just as an idea: Do you think it would be possible to drop these and just add the annotation in case the method only applies to one of them?
Since we already have the doc annotations for xmlrpc I wonder if there is a way to make them work for the HTTP-API as well (or at least some mechanism to translate them).

admd (commented):
We will also need to take care of users with read-only roles just like we are doing in case of xmlrpc API, which means no access to any other method other than the ones annotated with @Readonly annotation according to the design mentioned in this RFC.

accepted/00000-http-api.md
rjmateus: Or could the spark API map them directly to an HTTP status code automaticly? 
In what will a new structure differ from the current one?
Could we add a payload (as json) to the error response HTTP server would send with the data `FaultException` already have?

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+APIDOC is an extensive tool that automatically generates documentation for the API endpoints written in many languages. It can be used to generate the documentation for the HTTP API as well.
+More research is required to provide the framework for creating documentation for both interfaces with minimum duplication.
+
+
+# Unresolved Questions
+
+### Naming standard
+
+We use the term REST causally when describing the HTTP API, but as covered in this document, we don't intend to implement "REST" in a traditional way. Therefore, the term REST might be misleading for the user. On the other hand, "HTTP API" does not reflect the whole picture as well, since the existing XMLRPC API technically also utilizes HTTP. For the most clarity, we must determine a standard name for the API and use it consistently on all user-facing media (web UI, documentation, etc.).
```
rjmateus, line 216: Json over HTTP?

rjmateus (commented):
Thanks for the get RFC.
Here is my review for it.

accepted/00000-http-api.md
cbbayburt: Ok, in that case, let's stick with querystring parameters, plus JSON body for complex types (maps, objects, etc.)

accepted/00000-http-api.md
cbbayburt: So far we only use it to generate docs, but not code. It may be technically possible, but I'm sure there would be better alternatives for typescript generation reading the code directly, rather than the Javadoc.

accepted/00000-http-api.md
cbbayburt: Fixed :+1:

accepted/00000-http-api.md
cbbayburt: The difference between a POST and a DELETE is only semantics. If we had a resource-oriented API, it would make sense to implement multiple verbs on the same endpoint like:
```
PUT /api/contentmanagement/project (create a project)
POST /api/contentmanagement/project (update a project)
DELETE /api/contentmanagement/project (delete a project)
```
But in our case, the action is already in the method name: `updateProject`, `deleteProject`, etc. so there's no real benefit in using alternative verbs.

accepted/00000-http-api.md
cbbayburt: I will export and study the whole exception tree in the API context, then I'll update this section with more specific info.

accepted/00000-http-api.md
cbbayburt: IMO, registering all the public methods in a handler without any explicit marks is potentially a bad practice. For example, currently, we cannot have any public helper methods in the handlers (it is an obstacle for unit testing).

I don't worry much about the one-time work for adding the annotations. It might be possible to do it with an automated script.

However, let's discuss designing the annotations in a way that is easiest to use for us when adding new endpoints.

Some alternatives that I can think of:
 - Have an annotation to exclude the method from being a handler, rather than including it
 - Have just one annotation, and specify the protocol optionally in an annotation parameter (as suggested by @rjmateus)

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+To provide a uniform approach to handler registration, Java annotations can be used to specify a method as an endpoint handler. These annotations can be interpreted during initialization to register all the handlers to both frameworks. As a side benefit, all the handler classes in a predetermined package tree can be automatically identified, without the need to add them in a list manually.
+
+Java annotations shall also be used to determine attributes of an endpoint, such as whether it is a read-only method or not.
+
+Below is an example of such an annotation structure:
+
+#### @ApiHandler("namespace") Target=Class
+
+A class marked with this annotation will be registered as the handler of the specified namespace in the XMLRPC framework. The class will further be inspected on initialization to discover any annotated handler methods. Found methods will be registered to Spark as route handlers.
```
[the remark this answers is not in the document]
cbbayburt: Unfortunately not. The XMLRPC framework is designed in a way that it recognizes a whole class per endpoint (namespace). The additional level of multiple methods per namespace is our own implementation.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+#### @XmlRpcApiEndpoint Target=Method
```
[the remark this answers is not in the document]
cbbayburt: > Instead of using these annotations why not have one with a parameter for the list of targets to where we want to expose, with default to all.

That sounds like a good alternative :+1:

> Besides that, I think we should follow the same approach Spring is using: have annotations for each method type: @apiGet, @apiPost, @apiDelete, @apiHead, etc.

Since the handler classes will have the common business logic shared between two protocols, I wanted to keep it independent of a specific protocol as much as possible. Moreover, as mentioned above, we don't plan to use anything other than POST and GET unless we plan a whole new proper REST API, which targets resources rather than methods, because method names already include the semantics, so we don't need different verbs for that.

So considering this, and all the other comments, I guess the best way would be to have just one annotation, with the target protocols as a parameter, something like:
```
@ApiEndpoint(target=[ XMLRPC | HTTP | BOTH(default) ])
```

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+Each of these annotations shall mark a method to act as a handler for one of the APIs.
+
+**HTTP:** These methods will be registered to Spark as route handlers during initialization.
+
+**XMLRPC:** `BaseHandler` will search for a requested method in a list of methods annotated with `@XmlRpcApiEndpoint` in the class during a request.
+
+In most cases, both annotations shall be used together. Exclusive use of any of them shall be reserved for edge cases where the endpoint is not applicable to the specific protocol.
+
+#### @ReadOnly Target=Method
```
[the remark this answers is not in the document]
cbbayburt: Other than determining the method type, this would be also used to enforce the access rules for RO users, so it's still beneficial to have.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+An endpoint handled by this method will be set as a read-only endpoint.
+
+**HTTP:** Spark will create `GET` endpoints for the methods marked with this annotation. If not used, Spark's method wrapper will check the requesting user and block access if restricted.
+
+**XMLRPC:** `BaseHandler` will check this annotation against the requesting user and block access if restricted.
+
+#### @Unauthorized Target=Method
```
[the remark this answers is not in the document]
cbbayburt: The role check is already implemented inside of the business logic in all the existing methods so at the moment we don't have any benefit in changing this.

The only exception is a few methods where we don't enforce authentication (e.g. the `login` endpoint itself).

> If we follow the approach I'm proposing in the previous comment, then this can also be a parameter to the annotation.

I think this is a perfectly valid alternative to having a dedicated annotation. It's just a matter of taste, and it is a trivial distinction, so let's decide on this after we see the whole thing demonstrated.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+```java
+public ActivationKey getDetails(User user, String key)
+```
+
+With the proposed annotations, it may be written as:
+```java
+@XmlRpcApiEndpoint
+@HttpApiEndpoint
+@ReadOnly
+public ActivationKey getDetails(User user, String key)
```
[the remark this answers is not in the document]
cbbayburt: The plan is to interpret the registered methods with the following algorithm:
1. ~Get all the `@RouteParam` annotated parameters from the base URL sections~
2. Get all the rest of the parameters of **primitive** type from the querystring
3. Get all the object parameters from the request body in JSON

I can already say from the comments that we can eliminate No.1 and use querystring parameters only.

The registration code will take care of the differentiation between the primitive and complex types and look for the values in the right place.

Consider this example:
```java
public myApiEndpoint(User user, int id, MyComplexObject myObject, Map<..> myMap)
```
A request to this endpoint will be registered in a way that it must be called in the following form:
```
POST .../myApiEndpoint?id=<ID>
{
  myObject: {/*MyComplexObject serialized in JSON*/},
  myMap: {/*Map serialized in JSON*/}
}
```

I've chosen this approach because it'll require a minimal amount of work with the existing API. However, there are some limitations with this:
1. You need to know how the above algorithm works to know how exactly your request should look like. Hopefully, we'll be able to reflect this in the doc generation
2. Primitive parameters will always go to querystring, complex types will always go to method body. I don't see any problems with this since all the existing methods can follow this just fine.
3. Since we also have the concrete distinction of GET=read-only, POST=write and complex types are put into the POST body, it won't be possible to add complex type params (maps, objects) to a readonly method. This also shouldn't be a problem since it doesn't make much sense to have complex parameters in any read-only methods (mostly list/lookup methods that need just an ID). I will double-check this with all the existing RO endpoints.

Another, most simplistic alternative would be to use `POST` and only `POST`, embedding all the parameters into the JSON body. This requires the least amount of work for the implementation, but in the end, we must decide from the users' perspective, and how the API looks from outside above all.

I'll try to rephrase all this info into the actual RFC as well.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+**HTTP:** Spark will create `GET` endpoints for the methods marked with this annotation. If not used, Spark's method wrapper will check the requesting user and block access if restricted.
+
+**XMLRPC:** `BaseHandler` will check this annotation against the requesting user and block access if restricted.
+
+#### @Unauthorized Target=Method
+
+An endpoint annotated with `@Unauthorized` may be called with or without a logged in user. The handler method may or may not have a `User` parameter in the signature. If it has, the parameter may be `null`, in case the user is not logged in.
+
+#### @RouteParam Target=Parameter
```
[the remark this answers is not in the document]
cbbayburt: The point is, in this design, the Spark routes are generated automatically reading from the method signature like the following:
```
// Spark registration auto-generated
post("myMethod/:<route_param1>/:<route_param2>", (req, res) -> {
  // Translate the arguments and call the actual method
})
```

However, looks like we're moving away from using named route params anyway.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+  //   and the JSON object named "props" from the request body
+  // 2.Call to the `updateProject` method, passing the unwrapped arguments
+  // 3.Serialize and return the JSON result
+});
+```
+
+
+## 2. Authentication
+
+The web UI performs token-based authentication via a session cookie called `pxt-session-cookie`. Currently, the web UI login page exposes an API endpoint to support this process. This endpoint shall also be used for API authentication.
```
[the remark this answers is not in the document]
cbbayburt: > Authentication should have two different implementations, one for XML-RPC and another for HTTP?
Correct, it's the way it is now. I don't think we need to change anything on that front in the initial implementation.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+HTTP/1.1 200 200
+...
+<JSON data follows>
+```
+
+
+## 3. I/O
+
+On the HTTP API side, I/O serialization is mostly handled by Spark, and the existing `GSON` library. XMLRPC uses specialized "serializer" classes to write structured output on XML. These serializer classes shall be used exclusively for the XMLRPC output.
```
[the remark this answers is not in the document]
cbbayburt: Yes. However, we might have to register all the types to `gson` in advance. We'll see that with a PoC soon.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+With the proposed annotations, it may be written as:
+```java
+@XmlRpcApiEndpoint
+@HttpApiEndpoint
+@ReadOnly
+public ActivationKey getDetails(User user, String key)
+```
+
+At registration, Spark route wrapper will create the following route definition for this method:
```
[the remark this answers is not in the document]
cbbayburt: Correct, at the same time with the existing XMLRPC handler registration

accepted/00000-http-api.md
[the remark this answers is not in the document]
rjmateus: I get your point, and we will not use any other than Get and Post. But my feeling is that we are creating new name to give to something that already has a name.
If we define and use only two verbs, that is a different discussion, but we are basically giving new names, like `ReadOnly`.
If we use the verbs, in future we could easily add support to more if we decide in that way.

accepted/00000-http-api.md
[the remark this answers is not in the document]
rjmateus: I was thinking more in the long term for the project. Decoupling the business logic from the user role. This would simplify the day we decided to refactor the authorization logic in the project.
However, if is already implemented at the method level, we don't have much benefit now for the amount of work it would bring.

accepted/00000-http-api.md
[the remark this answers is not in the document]
cbbayburt: We need `ReadOnly` in any case, also for the XMLRPC API.

Currently, we use a regex `^(list|get|is|find).*$` against method names to determine if a method is readonly, and restrict access accordingly. IMO this is kinda flaky. I wanted to use this opportunity to change this behavior to be more explicit.

accepted/00000-http-api.md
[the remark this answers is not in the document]
cbbayburt: I agree. If we decide to go over the authorization logic in the future, that would be the best way to implement it.

accepted/00000-http-api.md
cbbayburt: The exception structure is simpler than I thought. We have only one level of inheritance (all from `FaultException`) and there's no XMLRPC-specific attribute in any of the exceptions other than the error number.

> I wonder if it is not possible to use the existing exceptions and just enhance them with a HTTP status code. Depending on what layer is used to return it, it takes the fault or the http status.

Yes, I think this will be possible.

So the question is if we should allow people to arbitrarily assign any HTTP code to any exception they create in the future? Or should we implement some OOP structure to put some ground rules? We probably don't want to end up in an inconsistent mess as we have with the fault codes.

cbbayburt: Here's a summary of what we have in XMLRPC so far:
https://github.com/uyuni-project/uyuni-rfc/blob/http-api/accepted/attachments/00000-exception-layout.md

Etheryte: I think it would make sense to have a-la an enum of status codes we support, that way we can also standardize on the frontend, and then we can provide additional information in a separate field if applicable. I wouldn't allow [the full range of status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) since it's too large and very subjective. Looking through the XMLRPC exception list, I would say _most_ exceptions can be mapped to one of the following: 

 - 200 OK for all successes, I wouldn't really bother differentiating between 200 and 201 for example etc.
 - 400 Bad Request for all cases where the input values make no sense (if I'm not misunderstanding, then a-la InvalidParameterException etc would fall in this camp?)
 - 401 Unauthorized for missing/incorrect authentication
 - 403 Forbidden for permission issues (a-la NotPermittedByOrgException if I'm not misunderstanding?)
 - 404 Not Found
 - 500 Internal Server Error for other problems, again I don't think it makes sense to differentiate between a-la 500 and 503 etc.

admd:
Folks, in case your concerns are addressed and you are fine with the design, please approve the PR.

@moio @lucidd can you please also take a look so we can proceed further.

rjmateus (approved):

cbbayburt:
I will schedule a final meeting after we've got all the reviews to make final decisions about naming conventions, annotation styles, etc.
I guess so far we have agreed on the core design. The rest should not block the implementation right away.

accepted/00000-http-api.md
mcalmer: ok

mcalmer:
Overall I think this is ok. Not sure how much work it is. Especially on the Exception side. It is still a bit unclear if your want to rewrite all Exceptions or just add some additional data to the existing.

cbbayburt:
> Overall I think this is ok. Not sure how much work it is. Especially on the Exception side. It is still a bit unclear if your want to rewrite all Exceptions or just add some additional data to the existing.

We'll use them as-is, and just map each of the exceptions to some HTTP status code. Most of the exceptions should be mapped to just a bunch of common status codes, so here we don't have much work to do.

I still have to update parts of the RFC with the latest decisions, but I'm currently working on the PoC to be able to answer more questions.

accepted/00000-http-api.md
cbosdo: Our doclet is a Javadoc addon that visits the Javadoc comments AST to get the `@xmlrpc*` tags as well as the `@deprecated` and `@since`. Those are then passed to velocity to be used as data for templates.

Templates are generating what we build them for... so far we generate docs in several formats (some unused) but we could generate Typescript definitions from them too. We already have to maintain Velocity only for that, I would avoid using yet another thing to generate typescript if possible... unless that can also help getting rid of Velocity.

accepted/00000-http-api.md
moio: > Have an annotation to exclude the method from being a handler, rather than including it

I like this direction and I think it is important to keep it in mind for this whole RFC.

As an API is a UI for programmers, in general, any design should take into account human factors, such as the power of laziness and the power of defaults. One important long-term success factor is to minimize "programmer bad feelings"!

To me that implies picking defaults that offer the best possible experience with least possible resistance, and then provide additional options for any corner/unforeseen cases.

A design that adds work for every method or even every parameter is probably going to be hated by future generations of programmers because of the bookkeping. One that prevents legitimate use cases will be perceived as limiting.

Instead, having reasonable defaults - such as for example exporting all public methods of a class that has been annotated as read-write on both protocols and all parameters via querystring/body without additional work is from this perspective a good thing. Having additional per method and/or per parameter annotations to override any of those aspects will help with "making easy things easy, hard things possible".

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+On the HTTP side, Spark allows registration of an endpoint as a "route" per each endpoint method, optionally including a number of named route parameters that construct the route path.
+
+The actual handler methods are designed to use POJOs as parameters. However, Spark expects a specific signature that takes `request` and `response` as parameters for route handlers. A simple wrapper shall unpack the parameters from the request and call the appropriate handler method for each request.
+
+A typical resource-oriented REST API relies on various HTTP verbs to provide contextual information on what type of operation an endpoint is used for. However, the XMLRPC protocol is designed around methods, rather than resources. Therefore, HTTP verbs will be restricted to GET for read-only, and POST for read-write operations regardless of whether they require a payload or not.
```
cbosdo, line 50: Do you plan to also use DELETE for resources removals? That would be more RESTy

accepted/00000-http-api.md
```diff
@@ -0,0 +1,220 @@
...
+
+The exposed HTTP routes shall follow the same namespace-method structure with a few extensions:
+
+```
+/rhn/manager/api/namespace/method[/:route/:params][?query=params]
+```
+
+On each request, the Spark route wrapper shall unpack both types of arguments in the request, deserialize the payload, and call the appropriate handler method with unpacked arguments.
+
+To provide a uniform approach to handler registration, Java annotations can be used to specify a method as an endpoint handler. These annotations can be interpreted during initialization to register all the handlers to both frameworks. As a side benefit, all the handler classes in a predetermined package tree can be automatically identified, without the need to add them in a list manually.
```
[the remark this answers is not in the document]
cbosdo: Not that big a deal: it's a finite amount of work 😜

accepted/00000-http-api.md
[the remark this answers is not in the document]
cbosdo: Even if the XML-RPC recognizes a whole class per endpoint, couldn't annotations make it easier to match the methods within the namespace?

accepted/00000-http-api.md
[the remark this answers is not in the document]
cbosdo: Also keep in mind that those annotations could also be used to generate the docs

cbbayburt:
Sorry for the long wait. Instead of giving half-answers, I wanted to go through with a proof of concept implementation first.

Please check it out: https://github.com/uyuni-project/uyuni/pull/4910

If you want to try it out, the `contentmanagement` namespace is fully functional as an HTTP endpoint! (Instructions are in the PR description)

Noticeably, there are a few fundamental differences between the proposed design and the RFC.
I'll try to answer the comments and update the RFC accordingly soon.

accepted/00000-http-api.md
cbbayburt: > Would it be an option to always be able to receive any parameters in the body, plus any "simple-typed" parameters via querystring optionally?

I think this is a good approach. I updated the text accordingly.

I tried this in the PoC without much effort. We look for both query params and the body for any parameters, with the body having precedence. Moreover, I let GSON parse all kinds of params so, in theory, it accepts even complex JSON in the query string, but I'll restrict this to work only with primitives, otherwise, we'll be supporting some crazy URL strings.

> What pros and cons do you see?

The only drawback is to somehow reflect this in the documentation without confusing people. We can offer both alternatives as examples, or we can document the preferred way depending on the type of the request (GET vs POST).

accepted/00000-http-api.md
cbbayburt: Updated the text with more specific info

accepted/00000-http-api.md
cbbayburt: > Have an annotation to exclude the method from being a handler, rather than including it

Updated the text to reflect this. Implemented in the PoC as well.

cbbayburt: Meanwhile I realized we have a JavaDoc tag `@xmlrpc.ignore` for this purpose. This will be replaced with the `@ApiIgnore` annotation, and used in both interfaces.

accepted/00000-http-api.md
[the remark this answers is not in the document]
cbbayburt: A huge blocker for this is the limitation that we cannot do a simple annotation scan in plain Java. There's a library called `reflections` for this purpose.
But honestly, since we're not adding any new handlers at this point and will use them as-is, implementing this is a bit overkill, so I removed this from the text.

accepted/00000-http-api.md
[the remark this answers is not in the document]
cbbayburt: So it turns out we need to implement the `JsonSerializer` interface for `gson` to recognize our return types. I'll try to design this in a generic way so we can write one serialization code and use it with both frameworks.

accepted/00000-http-api.md
cbbayburt: One big difference from traditional REST is that we're calling method names, not resources. So all our delete methods are named "deleteThis()" and "deleteThat()". Other than the semantics, DELETE does not bring any other benefits, so it seems redundant in our case.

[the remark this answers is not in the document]
cbbayburt: I should've checked this out at the very beginning :see_no_evil: but here's what I've found:

We have 6 endpoints that have simple lists in input (mostly for system IDs).

We have only one that has a map:
```java
SnapshotHandler.listSnapshots(User user, Integer sid, Map dateDetails)
```
`dateDetails` contains only 2 constant optional keys: `startDate` and `endDate`

We're not doomed, however, since query strings (also Spark) actually support these data types. Here's what it would look like:

List example:
```
GET https://uyuni.server/rhn/manager/api/system/listLatestAvailablePackage?name=mypackage&systemIds[]=1001&systemIds[]=1002 (and so on)
```

Map example:
```
GET https://uyuni.server/rhn/manager/api/system/provisioning/snapshot/listSnapshots?sid=1001&dateDetails[startDate]=2022-02-27&dateDetails[endDate]=2022-02-28
```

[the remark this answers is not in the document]
cbbayburt: Solved for lists: https://github.com/uyuni-project/uyuni/pull/4910/commits/ac9d9947866b2c08239c320359c77b9741eac80d

Maps will be a bit more complex, though.
Considering the super-specific single edge case of `SnapshotHandler.listSnapshots`: I wonder if we're allowed to break its signature to adapt it.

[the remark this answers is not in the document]
moio: How about:
 - deprecating this method
 - keeping it exposed via XMLRPC, but not exposing it to HTTP
 - creating a new method with a different signature (or just an overload, if possible) and expose it on both protocols

From a downstream perspective this all will happen in a major version, so it should not be a problem.

accepted/00000-http-api.md
```diff
@@ -0,0 +1,219 @@
...
+## 4. Exception Handling
+
+In XMLRPC, all the exceptions used inherit `FaultException` directly. Every exception has a unique integer code that describes an XMLRPC error. For HTTP, the `FaultException` tree must be enhanched with an HTTP status code that will be used in the response in any case of exception. Unlike the unique XMLRPC fault codes, HTTP codes will be chosen from a small subset of the standard HTTP status codes.
+
+Apart from the ehancement above, the current exception structure can be used as-is with the HTTP interface.
+
+
+## 5. Security
+
+The HTTP interface reuses the existing Spark framework. Therefore, we're typically not exposed to any new security issues. However, with the introduction of the new API, existing interface will be more critical for external use, so it would be beneficial to propose an approach to security testing.
```
thesp0nge, line 186: This sentence is so **strong**. Why using Spark keeps us away of new security issues? This is a dangerous assumption from a security point of view

accepted/00000-http-api.md
thesp0nge: I suggest using versioning in the URL (https://restfulapi.net/versioning/)

accepted/00000-http-api.md
```diff
@@ -0,0 +1,219 @@
...
+
+### Automatic handler discovery
+
+Currently, all the handler classes must be registered manually in the `HandlerFactory` class. This can be improved by the use of class annotations and package scanning. However, this would require use of a 3rd-party library since automatic scanning for annotations in packages is not possible with plain Java. Since the HTTP API implementation will not introduce any new handlers initially, this improvement will only provide convenience for the future additions in expense of additional work and an extra dependency in the initial implementation.
+
+### Parameter name matching
+
+XMLRPC relies strictly on the order of the sent arguments to match a method signature to a request. In contrast, the arguments sent via HTTP or JSON does not have any explicit ordering. For that reason, we must rely on parameter names and types to be able to match a signature to a request.
```
thesp0nge, line 207: I'm confused this. Why not passing parameter in a POST request? This will allow parameter names to be used. Take a look at this document: https://stackoverflow.blog/2020/03/02/best-practices-for-rest-api-design/

accepted/00000-http-api.md
```diff
@@ -0,0 +1,219 @@
...
+
+By default, parameter names are not preserved in the compiled Java classes, and this makes parameter name matching impossible. To solve this problem,  `-parameters` `javac` argument must be used when compiling to preserve parameter names in the compiled bytecode.
+
+As demonstrated in the [proof of concept](https://github.com/uyuni-project/uyuni/pull/4910), this option has no major side effects (e.g. ignorable file size difference). However, the CI flow has to be considered when applying this change. 
+
+### CSRF tokens
+
+To be able to provide API POST routes, the CSRF token must be disabled for the POST endpoints in the new HTTP API. It is possible to define exceptions to CSRF token validation via a path root. However, in the current proposal, the API path root is shared with the internal React API (`/rhn/manager/api`). We must decide on an approach to disable CSRF exclusively for the HTTP API endpoints:
```
thesp0nge, line 215: This is not correct. CSRF protection is not useful if there is no auth cookie or basic auth token. However, if there is, it must be in place a way to discriminate if an authenticated request is legitimate or fake.

Please consider the following answer on this topic: https://security.stackexchange.com/a/166798

thesp0nge: @renner told me that since Sparkle doesn't have a native anti CSRF mechanism, you implemented it on your own. Is it correct? Can you share the code to investigate further?

accepted/00000-http-api.md
Etheryte: I might be mistaken, but I read the sentence to mean that we won't be adding any additional (as in new) attack surface compared to right now since it uses a framework that's already exposed. But feel free to correct me if I misunderstood @cbbayburt.

thesp0nge: Ok, I read the sentence as "since we are using Spark, we won't suffer by any new security issue in the future". I would suggest rephrasing in "The HTTP interface reuses the existing Spark framework. Therefore, the attack scenario is unchanged"

accepted/00000-http-api.md
cbbayburt: @thesp0nge: the API described in this RFC is meant to be consumed by users directly via automated scripts (python, JS, etc.). It's not meant to be used in the web application, so there's no conventional form submission via a browser.

Considering this point, we have 2 unresolved questions:
1. Are we still exposed to CSRF, if so, how?
2. What should be the proper workflow for protection? Since we don't send any forms or any initial data before a POST request, the only way I can see is to not use any tokens, but send credentials with every request.

accepted/00000-http-api.md
cbbayburt: The restriction is on the Java side, not HTTP.

We try to avoid using the term REST, because in nature, this API is more of an RPC over JSON than REST.

The RPC methods in question are part of the existing XMLRPC API, and they are implemented as plain Java methods in the backend. Consider this example:
```java
public String myMethod(String foo, String bar)
```
with this design, we must match the following request to this method:
```
https://uyuni.server/rhn/manager/api/myMethod?foo=value1&bar=value2
```
or this:
```
POST https://uyuni.server/rhn/manager/api/myMethod
{
  'foo': 'value1',
  'bar': 'value2'
}
```

In either form, we have the parameter names and values on the request, and the order of the parameters is arbitrary. But by default, the Java compiler does not preserve parameter names in method signatures, so in compiled code, the method becomes this:
```java
public String myMethod(String arg0, String arg1)
```
in this case, it's not possible to know which of these parameters is `foo` and which is `bar`.

accepted/00000-http-api.md
Etheryte: If I understand correctly, the problem here would be if we share the credentials for the HTTP API (as in this new API) and the website API (the stuff we have right now). In that case, an attack scenario would be a fake page, such as:

```html
<form action="https://suma.server/http-api/deleteAllSystems" method="POST">
  <input type="submit" value="Claim your prize!" />
</form>
```

If the user is logged in on `suma.server` and clicks the buttons, without further checks, we have a problem.  

If we don't allow the regular user session to be used for the new HTTP API, this should not be an attack vector, correct? Please correct me if I'm wrong @thesp0nge.

moio: My naive understanding is we could drop CSRF checks altogether if we added `SameSite` headers (although this is irrespective of this RFC). Can you confirm @thesp0nge?

https://scotthelme.co.uk/csrf-is-dead/
https://scotthelme.co.uk/csrf-is-really-dead/

moio:
@cbbayburt as implementation is already complete, I suggest you want to merge this.

PR #69, opened 2022-07-12
A plan to remove DWR

wweellddeerr:
Remove the DWR library from SUMA.

See the [rendered version](https://github.com/uyuni-project/uyuni-rfc/blob/6b8deb6118d29b67b49f76dfdf7c664a329d7d88/accepted/00092-replace-dwr.md).

### DIFF
--- a/accepted/0000-replace-dwr.md
+++ b/accepted/0000-replace-dwr.md
@@ -1,0 +1,74 @@
+- Feature Name: Remove DWR
+- Start Date: 2022-07-10
+
+# Summary
+
+Remove the DWR framework from SUMA.
+
+# Motivation
+
+## Why are we doing this?
+DWR project has not an active community and does not appear to be under active development:
+- its last version was released 5 years ago;
+- the last change on its codebase was applied more than a year ago;
+- [its official website and documentation](https://directwebremoting.org/) is not functional.
+
+Additionally:
+- As described [here](https://github.com/SUSE/spacewalk/issues/5967) there is a suspicion that the javascript files related to DWR could potentially unburden tomcat on situations of high overload; and
+- The attributes of the `DWRSESSIONID` cookie, part of DWR library, are causing the following warning to appear in the browser logs (see [this issue](https://github.com/SUSE/spacewalk/issues/13995)):
+```
+Cookie “DWRSESSIONID” will be soon rejected because it has the “SameSite” attribute set to “None” or an invalid value, without the “secure” attribute. To know more about the “SameSite“ attribute, read https://developer.mozilla.org/docs/Web/HTTP/Headers/Set-Cookie/SameSite
+```
+
+## What use cases does it support?
+DWR is a library that hide the details of request/response handling during AJAX interactions, it allows JavaScript code to use Java methods running on the backend just as if they were pure JavaScript code. Currently, it supports some features in Suma and the points where it is used are mapped as follows.
+
+* `engine.js` is the core js file from DWR, it is imported in [`layout_head.jsp`](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/decorators/layout_head.jsp#L67) and referenced in [`template.js`](https://github.com/uyuni-project/uyuni/blob/master/web/html/src/styleguide/template.js#L31), so loaded in all pages.
+* `util.js` is a js library that comes with DWR and contains utility functions that can be used to update HTML using js data. It is imported in [`layout_head.jsp`](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/decorators/layout_head.jsp#L68) and referenced in [`template.js`](https://github.com/uyuni-project/uyuni/blob/master/web/html/src/styleguide/template.js#L32), so also loaded in all pages. Its functions seem to be used only [here](https://github.com/uyuni-project/uyuni/blob/master/web/html/javascript/spacewalk-checkall.js#L193) and [here](https://github.com/uyuni-project/uyuni/blob/master/web/html/src/manager/audit/cveaudit/cveaudit.tsx#L119).
+* **Dynamically generated JS files** - the following js files are generated dynamically by DWR, exposing some methods of Java classes as pure js functions that call the real code using Ajax requests. The classes and methods are defined in [`dwr.xml`](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/dwr.xml) file: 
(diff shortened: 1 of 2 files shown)


accepted/0000-replace-dwr.md
```diff
@@ -0,0 +1,74 @@
...
+  * <sup>*</sup> `SystemGroupsRenderer`: exposes the `renderAsync` method of the [`SystemGroupsRenderer`](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/frontend/action/renderers/SystemGroupsRenderer.java) class. It is [loaded](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/pages/yourrhn.jsp#L30) and [used](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/pages/yourrhn.jsp#L85) only in the `yourrhn.jsp` file.
+  * <sup>*</sup> `TasksRenderer`: exposes the `renderAsync` method of the [`TasksRenderer`](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/frontend/action/renderers/TasksRenderer.java) class. It is [loaded](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/pages/yourrhn.jsp#L15) and [used](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/pages/yourrhn.jsp#L48) only in the `yourrhn.jsp` file.
+
+  <sup>*</sup> All of these are only used in the [Your RHN](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/pages/yourrhn.jsp) page.
```
moio, line 42: Historical note FYI: Your RHN was envisioned to be a customizable homepage with "widgets". What actually came to be is a limited set of such "widgets" and, to my knowledge, basically nobody ever changes them.

I would not object a redesign of the whole homepage, although that goes above and beyond the intent of this PR.

accepted/0000-replace-dwr.md
```diff
@@ -0,0 +1,74 @@
...
+
+For the frontend side, the capabilities provided by DWR (both the utility functions from `util.js` file and the ajax calls) are replaceable using jQuery.
+
+There is a PoC in [this commit](https://github.com/wweellddeerr/uyuni/commit/9ddb68030dc49fa3272a062ceb7015c35b68d457) exploring how it is possible to use jQuery to make the ajax request and a raw Servlet as backend replacing the `DWRItemSelector` component of DWR. The SSM (one of the features that uses it) seems to work well after applying this patch.
+
+Using a raw Servlet as backend seems to be the best approach since (a) the methods currently exposed by DWR are only available through the `DwrServlet`, which is a raw servlet; and (b) these methods typically manipulate the HttpSession, making it difficult to use the HTTP API, for example.
```
moio, line 56: Can you elaborate on this comment a bit, pointing to what exactly we would need to manipulate and why?

accepted/0000-replace-dwr.md
```diff
@@ -0,0 +1,74 @@
...
+
+# Alternatives
+
+The first alternative is to just keep using DWR and try to solve the warning mentioned above, maybe trying to create a Pull Request on the DWR repository or even opening an issue on it.
+
+The second alternative is to gradually replace the pages using DWR to use React. The main drawbacks of this alternative are: (1) some issues to migrate pages using DWR were aborted in the past, probably due to non technical decisions; and (2) the `rl:selectablecolumn` custom tag mentioned before is used in several JSP pages, it would be necessary to find an alternative to this component in order to avoid the need to migrate so many pages.
+
+A hybrid approach could also be used, migrating some pages (specially the `Your RHN`) to React and applying the suggested solution (or even the first alternative) only in the remaining cases, reducing the risks.
```
moio, line 70: Ultimately we will want to port all pages to React, but I think the subgoal of getting rid of DWR is an important one as well.

Please use your best judgement to select the cheapest approach - if migrating a page to React is easier than adapting the corresponting DWR methods, please go down that path.

moio (commented):
Apart from a question inline I like this proposal and I think it goes in the right direction.

This is a long-time tech debt item I'd love to see tackled.

rjmateus (approved):
This looks like a great approach to me, to be used as a transitional step before we get rid of JSP pages (someday in the future).
In terms of risk, the most critical part is the selector one. But from a server point of view, the call will be the same, only the communication style will change. That said, it should work or fail catastroficly.

admd:
@cbbayburt @Etheryte @cbosdo can you please gives a thumbs up to Welder if you have nothing against this approach so @wweellddeerr can proceed further?

Etheryte (approved):
I'm not all that familiar with the current ins and outs of DWR, but the proposed approach to replace it looks good to me. 👍

cbosdo:
@wweellddeerr There have been a successful attempt at replacing a page using DWR feature: the virtual systems list. The new systems list also gets rid of quite a number of places using the `DWRItemSelector`. We now have an API endpoint to add items in the SSM: https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/suse/manager/webui/controllers/SetController.java#L76.

I like your approach, but I fear the temporary socket would stay for ever... We would still need to remember to progressively convert to proper API endpoints.

accepted/0000-replace-dwr.md
```diff
@@ -0,0 +1,74 @@
...
+* **Dynamically generated JS files** - the following js files are generated dynamically by DWR, exposing some methods of Java classes as pure js functions that call the real code using Ajax requests. The classes and methods are defined in [`dwr.xml`](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/dwr.xml) file: 
+  * `DWRItemSelector`: exposes the method [`select`](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/frontend/taglibs/DWRItemSelector.java#L52) of the DWRItemSelector java class. It is loaded in [`layout_head.jsp`](https://github.com/uyuni-project/uyuni/blob/master/java/code/webapp/WEB-INF/decorators/layout_head.jsp#L69) and referenced in [template.js](https://github.com/uyuni-project/uyuni/blob/master/web/html/src/styleguide/template.js#L33), so loaded in all pages. The function is currently used in all the pages using the `rl:selectablecolumn` custom tag defined [here](https://github.com/uyuni-project/uyuni/blob/master/java/code/src/com/redhat/rhn/frontend/taglibs/rhn-list.tld#L242), including SSM. **This seems to be the most risky component, since it is used across several pages**.
```
cbosdo, line 29: This one is used for the SSM Count. For sure it is used in a lot of pages, but as long as we find a proper way to implement it using API endpoints it should be easy to propagate to React pages at least.

accepted/0000-replace-dwr.md
cbosdo: reactifying YourRHN would be really cool and kick a number of DWR uses: two birds killed with one stone.

PR #70, opened 2022-08-01
Upgrading Bootstrap

Etheryte:
A proposition on why and how to upgrade Bootstrap from 3.x to 5.x.

See the [rendered version](https://github.com/Etheryte/uyuni-rfc/blob/bootstrap/accepted/00000-upgrading-bootstrap.md).


cbbayburt (approved):
Thanks, @Etheryte, very well written.

"Simplifying existing styles" part could be a bit more detailed with concrete pointers, but I guess that can be left to the design phase as well.

I'd like to hear some input from the QE as well, as to me it sounds like the testing would be critically important with this, and I believe it will need more effort than what our integration tests cover.

accepted/00000-upgrading-bootstrap.md
```diff
@@ -0,0 +1,70 @@
...
+
+Coming out successfully from the other side, we would hope to have simpler stylesheets with less to maintain, and fewer security issues.  
+
+
+# Detailed design
+[design]: #detailed-design
+
+## Theme & styles
+
+Bootstrap 3 uses Less while Bootstrap 5 only offers Sass styles. This switch by itself is not very problematic, however there are additional issues that make the upgrade more difficult. Our current styles make heave use of numerous mixins from Bootstrap 3 which no longer exist in Bootstrap 5. Additionally, there are numerous breaking styling changes going from 3 to 4 and even more so moving from 4 to 5.
```
moio, line 31: > This switch by itself is not very problematic

Do automated translation tools exists, or at lease references/guidelines?

accepted/00000-upgrading-bootstrap.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Theme & styles
+
+Bootstrap 3 uses Less while Bootstrap 5 only offers Sass styles. This switch by itself is not very problematic, however there are additional issues that make the upgrade more difficult. Our current styles make heave use of numerous mixins from Bootstrap 3 which no longer exist in Bootstrap 5. Additionally, there are numerous breaking styling changes going from 3 to 4 and even more so moving from 4 to 5.  
+
+Instead of trying to manually migrate our existing very large theme source, I propose we instead try to first reduce our existing theme footprint as much as possible. We already have component scoped styles working and in use, so we can try move component or view specific styles (e.g. table styles, setup wizard etc) into component scoped styles and remove them from the global stylesheet.  
+
+This allows us to cut up the effort required and do at least some of the work gradually. I suspect the last core theme change will still be painful, though. It might be easier to rebuild the existing theme from the base Bootstrap style rather than migrating the existing one, but this is something we can consider after we've simplified the theme.
```
moio, line 35: Would the footprint reduction step be wasted if we ultimately decide to rewrite from scratch?

accepted/00000-upgrading-bootstrap.md
```diff
@@ -0,0 +1,70 @@
...
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Suma is huge and the UI likewise. Last time we touched the theme to match corporate styling for 4.3.0 we broke Many Things and the tail end of layout issues is long. Realistically I would expect this upgrade to require considerable time with QA and testing. In an ideal case, the upgrade would be merged as early in a new version cycle as possible, as to give everyone the maximum amount of time possible to ensure the end result is up to par. 
+
+
+# Alternatives
+[alternatives]: #alternatives
+
+The easiest alternative is to do nothing. We've been shipping Bootstrap 3 along with its known issues for a long time and so far it hasn't blown up in our face. Likewise, our current themes are hard to maintain, but so long as corporate branding doesn't change, there isn't too much to change day-to-day. It is possible to go ahead with the stylesheet simplification etc by itself without upgrading Bootstrap.
```
moio, line 65: Would it be possible to patch the XSS vulnerabilities downstream instead? Could it be Red Hat, Debian or some other open source project has done that already?

accepted/00000-upgrading-bootstrap.md
Etheryte: Automated tools exist to translate most of Less into Sass. There are edge cases which aren't translatable since the languages function slightly differently, but I wouldn't be too worried about those as I would expect the manual effort to be limited. The main problem I would say is the mixins, since some of them no longer exist or have changed in the newer version.

accepted/00000-upgrading-bootstrap.md
Etheryte: Partially, there are still components for which this makes sense. Some examples would be our table component, the calendar widget we use, etc. But some of the effort would be wasted, yes. It's hard to guesstimate which would be less effort, as both detangling existing styles and styling Suma from scratch are both considerable challenges.

accepted/00000-upgrading-bootstrap.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Component and class name changes
+
+This is tied to the previous section, but worth discussing on its own. Numerous component, layout and utility classes have been dropped or changed in both upgrades, e.g. `input-group-addon` was changed in Bootstrap 4 and dropped in 5, so we need to use `input-group-text` and check if it renders correctly etc. The list is pretty long. As discussed in [this ticket](https://github.com/SUSE/spacewalk/issues/18346), it's unlikely that we'd find a tool that can handle all different ways we use Bootstrap (in JSP, JPSF, React, etc), however writing a migration script similar to the Typescript migration might be feasible. For this, we can assemble a list that covers all the breaking changes that affect us by going through the list of changes for 4 and 5, checking whether we use the affected classes, and if so, write down what the upgrade path is.
```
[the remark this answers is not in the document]
Etheryte: Mostly I would hope the same, that we can grep and replace after verifying we have the right set of matches. My main worry is dynamically created class names etc which I know we have in some places, but hopefully not much. Translations are a good point I did not consider, thanks for the heads up, I will keep it in mind.

accepted/00000-upgrading-bootstrap.md
Etheryte: Possible yes, reasonable, I'm inclined to say no. Bootstrap is not a trivial piece of code, as evidenced by these vulnerabilities, and since we don't have anyone who's familiar with their codebase I wouldn't be sure we would be fixing these issues correctly without introducing any new issues. It's possible someone has looked into it, but I doubt they would still be maintaining it at this point — I did a cursory search and could not find anything though.

accepted/00000-upgrading-bootstrap.md
srbarrios: I would not see it as a waste time, if we dive and learn better our code, by working in this reduction.

accepted/00000-upgrading-bootstrap.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Summary
+
+In short, the suggested upgrade path can be summarized as follows:  
+
+ - Simplify the existing styles, trying to use standard Bootstrap tooling wherever possible
+ - Gather the list of breaking changes
+ - Write a migration script and/or manually migrate the above list
+ - Migrate existing styles from Less to Sass
+ - Verify the results with Cucumber
```
srbarrios, line 53: At that point, it will not be only verifying but also adapting the test framework code to use the new WebElement locators.

accepted/00000-upgrading-bootstrap.md
```diff
@@ -0,0 +1,70 @@
...
+ - Clean up any dangling ends
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Suma is huge and the UI likewise. Last time we touched the theme to match corporate styling for 4.3.0 we broke Many Things and the tail end of layout issues is long. Realistically I would expect this upgrade to require considerable time with QA and testing. In an ideal case, the upgrade would be merged as early in a new version cycle as possible, as to give everyone the maximum amount of time possible to ensure the end result is up to par.
```
srbarrios, line 59: Fully agree, we can't have that close to 4.4 Alpha 1 :D

accepted/00000-upgrading-bootstrap.md
```diff
@@ -0,0 +1,70 @@
...
+ - Clean up any dangling ends
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Suma is huge and the UI likewise. Last time we touched the theme to match corporate styling for 4.3.0 we broke Many Things and the tail end of layout issues is long. Realistically I would expect this upgrade to require considerable time with QA and testing. In an ideal case, the upgrade would be merged as early in a new version cycle as possible, as to give everyone the maximum amount of time possible to ensure the end result is up to par. 
+
+
+# Alternatives
+[alternatives]: #alternatives
```
srbarrios, line 63: How coupled we are with backend? Can we support at same time two frontends?
We might want to start developing from scratch a new frontend, without any kind of restriction not even in web frameworks, while keeping the old one as alternative.

The new frontend will not have test coverage for a while, but once we can feel all is developed, then we can switch on the test framework.

accepted/00000-upgrading-bootstrap.md
moio: Thanks for the information. Please add a note about usage of translating tools if that is your current best plan going forward.

accepted/00000-upgrading-bootstrap.md
moio: > I would not see it as a waste time, if we dive and learn better our code, by working in this reduction.

Well, it would be a waste if all we learnt goes to the trash bin because we ultimately rewrite it all from scratch.

@Etheryte: the spirit of an RFC is to try to come up with the best possible alternative _given available evidence at the time of writing_. We can always course-correct later - just in principle (take it as an exercise) we strive to describe only one course of action as the suggested one by the RFC author, and all other alternatives in their own section.

In this case, I would encourage you to spend a bit of time and try to decide what path you see as most promising, so that only one proposal ends up in the main RFC text, and that will be our starting bet. As always we remain open to change plans in face of contrary evidence, and ready to switch bets!

accepted/00000-upgrading-bootstrap.md
moio: I would definitely encourage to follow best practices when writing new code or changing existing code in a significant way, but I do not think adapting the whole testsuite to a new style of locators really belongs to this RFC (or effort).

In principle it could be a separate one, correct?

accepted/00000-upgrading-bootstrap.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+# Detailed design
+[design]: #detailed-design
+
+## Theme & styles
+
+Bootstrap 3 uses Less while Bootstrap 5 only offers Sass styles. This switch by itself is not very problematic, however there are additional issues that make the upgrade more difficult. Our current styles make heave use of numerous mixins from Bootstrap 3 which no longer exist in Bootstrap 5. Additionally, there are numerous breaking styling changes going from 3 to 4 and even more so moving from 4 to 5.  
+
+Instead of trying to manually migrate our existing very large theme source, I propose we instead try to first reduce our existing theme footprint as much as possible. We already have component scoped styles working and in use, so we can try move component or view specific styles (e.g. table styles, setup wizard etc) into component scoped styles and remove them from the global stylesheet.
```
ncounter, line 33: What I would like to point out here in reducing the footprint itself is making sure to reduce also the layers that make the footprint what it is like. It's not only about the width of it but also (and more complicated) about the depth. What I've found the most frustrating part to maintain was the multiple overriding pieces they are just partially overlapping again and again through the layers.
It is by design that the history made it complicated: overriding spacewalk with a full copy of SUSE Manager and/or Uyuni. But this is not the only complexity aspect: the core theme itself is made of overrides and fixes on top to adjust the behavior as desired.
So carving out specific components would help a lot reducing the **width**, but I'd really suggest to squash the **depth** of  components, and I expect this last point to be the most complicated one actually.
For instance, for a table component style, at the moment we have:
```
base theme style +
base table style +
Uyuni theme style fully overriding base theme style +
Uyuni table style fully overriding the base table style
```
and if you apply a single change in the middle of this chain, the result can be different.

Ideally we should have one and only one `theme style` (maybe more a _neutral-branded_ one), then single components would run over it and that's it.

@Etheryte : technically speaking, from your point of view, what would be a good approach to squash the duplicated theme styles into only one?

accepted/00000-upgrading-bootstrap.md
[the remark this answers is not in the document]
ncounter: IIUC the most of the time we dynamically create class names is more often for `font-awesome` icons, but it might also be the case here and there for bootstrap ones. I don't expect that many but I'm afraid that yes, those will have to be caught and tackled manually.

accepted/00000-upgrading-bootstrap.md
ncounter: Let's rephrase it this way: reducing and simplifying the stylesheet is a short-mid term goal, upgrading Bootstrap is the mid-long term goal that can come only after the short-mid one. Both are worth to be chased and both would deliver value for developers and customers at the same time. I don't see this as an alternative, but just a split of the Epic task into more specific and defined steps to run over. Estimating the whole RFC to be done would be something for _only the braves_, starting with the simplification described above instead sounds more reasonable :wink:

accepted/00000-upgrading-bootstrap.md
Etheryte: This is a larger can of beans, but technically it would be at least somewhat possible, depending on what you mean exactly. Given the work Can has done on the HTTP API, building any kind of new UIs to match the endpoints available there would be possible. Routing-wise there wouldn't be much issue in sending certain URLs to one and and the other, we already figuratively do this with Struts and React. I think the main question to ask here would be what we would hope to get out of this idea. At some point in the past, the React based UI was this new frontend, and in that sense we are already doing it. The React parts share a lot of the baggage with the old UI such as Bootstrap and themes outlined in this RFC though.

moio: > [...] Can we support at same time two frontends?

FMPOV it is not really in the scope of this RFC to support two _frontend stacks_, so let's table that discussion for the time being.

A different consideration _could_ be made about supporting two _css frameworks_ - the current one and the new-and-streamlined Bootstrap 5 one. So instead of having a big-bang migration, we could ship it behind a feature flag - especially for the Uyuni community and for example our test systems - and have it more human tested before deleting the old framework completely.

Question to @Etheryte is: do you see a lot of changes necessary in the HTML structure and tag attributes, so that such an approach is not feasible/causes too much overhead?

Etheryte: I think that would increase the complexity considerably, since then for every change you'd need to check both old stylesheets and new stuff that everything lines up correctly. What would be a viable alternative would be to instead have different stylesheets per page. E.g. on `/rhn/foo` you'd get the old Uyuni/Suma theme, but on `/rhn/bar` you'd get the new theme. That way we could move forward page-by-page and gradually put in the work to move things over and test them.

Etheryte: I feel I don't understand what you mean, or there's some misunderstanding. I'll try to rephrase what I'm proposing and perhaps you can explain where the issue lies @moio?

What I have in mind is to have e.g. a configuration file that includes a list of paths for pages that we have updated and verified to work. For those pages, in the page header template where styles are loaded, we could load new stylesheets. For pages that are not found in that list, we can serve the old stylesheets instead.  

So for example, if we have `list-of-updated-pages.conf` such as:  

```
/rhn/foo-1
/rhn/foo-2
...
```

and then in `layout_head.jsp` a-la

```
<c:choose>
  <c:when test="${currentPage exists in list-of-updated-pages.conf}">
    <link rel="stylesheet" href="/css/new/${webTheme}.css" />
  </c:when>    
  <c:otherwise>
    <link rel="stylesheet" href="/css/old/${webTheme}.css" />
  </c:otherwise>
</c:choose>
```

This way we can create a minimal functioning stylesheet to cover one page at first, verify it works, and then that style is only served for that one page, every other page keeps working as is. Do you think this idea could work?

moio: That idea can work. Having half pages on the new stylesheet and the other half on the old one is, in fact, a lot better than having all pages on the new style - but with half of them broken. Anyway this is implemented, I would surely encourage to proceed page-by-page: be it with a centralized configuration file as you described or just by adjusting individual page headers to point to new stylesheets when each is ready. Let's informally call such a mechanism _developer switchable_.

The idea I was entertaining with was to make themes _user switchable_ instead, so that enterprising Uyuni community members or fearless developers could try the new theme on all pages and gauge the degree of brokenness and report bugs. What I got from your last message is that, given the coupling between Bootstrap and the markup which is expected to change from 3 to 5, in order to attain _user switchability_ we would have to maintain two copies of every page, with and without Bootstrap 5 compatibility fixes - and that does not seem like a sensible idea given the amount of code that would substantially need duplication.

Does that make sense to you, is it clearer now?

Etheryte: Yes, the user switchable part was what I was missing from your idea. I think that would not be feasible because of the duplication and extra overhead. I think we're on the same page now, thanks a lot for clarifying.

accepted/00000-upgrading-bootstrap.md
SchoolGuy: Yes it could be in theory, however there is absolutely no benefit to it except tech-debt. Since the locators then need to be rewritten again after the PR with the work this RFC outlines (because of the breaking bootstrap changes), this would be a duplicated effort imho.

rjmateus:
@Etheryte if the decision goes in the direction @moio and you were discussing, of having pages migrate one by one, and use some kind of list of all pages that are in the new style, how will this cope with the single page mechanics we have at the moment?
We may be in a page with bootstrap 3 layout and load one which is already moved to version 5. In this case should we do a full reload of the page?
In that solution, we will need a duplicated page template, one for bootstrap 3 (the current one) and a new one in bootstrap 5?

The plan looks good to me:
- reduce the existing footprint
- migrate one page and make sure all works (including testsuite)
- prepare a cookbook on how to migrate pages
- apply that cookbook to a set of pages and adapt anything if needed 
- mass apply it all (if possible with some scripts/tools)

PR #71, opened 2022-08-04
Add "Replace Salt Toaster" RFC

agraul:


SchoolGuy (commented):
So obviously the TODO needs to be filled and the open questions need to be discussed. However I had some talk in person with @agraul about this already and I like his approach. Especially the part about nudging people to fix tests properly, rather then adding things to a skip-list is for me a very important thing that I am happy it got into this RFC.

For me this RFC reflects my opinion about what we should do in the future.

A question that we maybe should answer is how we want to integrate this approach into `openSUSE/salt`, as this is the current place where salt-toaster is currently used. The content of this RFC is very focused on Uyuni atm.

accepted/00000-replace-salt-toaster.md
```diff
@@ -0,0 +1,138 @@
...
+a skip list, one for each test group.
+
+Example:
+```sh
+$ cat test_config.yml
+groups:
+  unit:
+    tests_dirs:
+      - tests/pytests/unit
+      - tests/unit
```
vzhestkov, line 94: Does it mean we are going to use it for `unit` tests only and avoid running `functional` and `integration` tests?

agraul: No, we will run all four test groups (unit, functional, integration, scenario). This is just a short example of an incomplete config to show the desired input and output of the (to-be-developed) test execution script.

vzhestkov: Thanks for explanation :+1:

agraul:
> A question that we maybe should answer is how we want to integrate this approach into openSUSE/salt, as this is the current place where salt-toaster is currently used. The content of this RFC is very focused on Uyuni atm.

I just pushed a section about setting up GH runners using the same sources and workflow that we use in sumaform. I'll also add something that explains where which tests (& groups) are run, once I thought more about that.

SchoolGuy:
@agraul That is a good paragraph. I like the idea of using Cache to speed things up. From my eyes we should fill in the TODOs and then we should already be in a state where we can do a first review with requested changes vs approvals.

agraul:
I added a few unresolved questions. I'll continue with them once I'm back from vacations. In the meantime, everyone is welcome to leave their comments.

accepted/00000-replace-salt-toaster.md
```diff
@@ -0,0 +1,192 @@
...
+different operating systems we support. These VMs are set up with sumaform and
+salt-jenkins. Terracumber triggers terraform to create the VMs from within a
+Jenkins job and launches the test suite, connecting to each VM over SSH.
+
+## Provisioning Test VMs with Sumaform + Salt-Jenkins
+
+A new module "salt\_testenv" is added to sumaform. This module defines a new
+role, "salt\_testenv", and the default values needed to run all test suites, e.g.
+4GiB of RAM. Sumaform uses this role to apply the `golden-images-provision.sls`
+state file from the Salt-Jenkins repository.
```
moio, line 52: I guess it would be great to have a generic sumaform option to run arbitrary Salt files at the end of provisioning.

accepted/00000-replace-salt-toaster.md
```diff
@@ -0,0 +1,192 @@
...
+
+Up to now, all state files used by sumaform are maintained in the sumaform Git
+repository. For provisioning the salt\_testenv VMs, we don't want to maintain
+our own states that define which dependencies need to be installed on what
+operating system. Salt upstream already maintains such states for their own test
+infrastructure. We re-use their states by adding `salt-jenkins` as a Git
+submodule to sumaform and merge the salt state trees.
```
moio, line 59: What advantages and disadvantages you see with using submodules vs. consuming states from another repo (=defined path, for example)?

accepted/00000-replace-salt-toaster.md
```diff
@@ -0,0 +1,192 @@
...
+provider, that sumaform uses to upload the `salt/` directory to each VM for
+provisioning can't handle symbolic links well. A
+[workaround](https://github.com/hashicorp/terraform/issues/13202#issuecomment-321578202)
+is possible, but it will touch a core sumaform mechanism.
+
+#### Alternatives
+
+Alternatively, we can copy the files instead. This loses the information where a
+state file is maintained and requires manually updating the copies whenever the
+used submodule revision is updated. Hard links would solve the last issue, I
```
moio, line 82: How about having Jenkins copy trees over to the sumaform directory, and then allowing sumaform to copy them over to the VM in turn? Is that what you are proposing here?

accepted/00000-replace-salt-toaster.md
```diff
@@ -0,0 +1,192 @@
...
+haven't yet confirmed that they work with the `file` provider.
+
+## Provisioning Test VMs for GitHub Actions
+
+GitHub Actions use so-called runners, which are VMs running Windows, macOS or
+Linux. It is possible to host your own runners, but we currently neither do nor
+plan to do that. While we can't use sumaform to provision GitHub-hosted runners,
+we can prepare the runner before executing a test suite run. With Salt Toaster,
```
moio, line 90: One option that may or may not be applicable in this case is to use sumaform in SSH mode - that does not do any VM creation, it only runs the SCP+salt part.

accepted/00000-replace-salt-toaster.md
```diff
@@ -0,0 +1,192 @@
...
+to red is a strong signal that either the product code or the test code has not
+reached an acceptable quality level yet. At the same time, we inherit flaky
+tests from upstream and don't have the time to fix them. These tests can be
+skipped to keep the overall test results green. But we need to be careful: when
+skipping tests is more convenient than fixing them, we risk skipping too
+many -- which directly lowers the value we get from running tests in the first
+place. Adding a test to the "skip list" is as big of a bad smell as is turning
+the result red.
+
+The skip list is maintained in a GitHub repository. All changes must be
```
moio, line 153: Does that mean a _different_ GitHub repository (than the main salt one)?

SchoolGuy: Yes that is correct (don't have the link at hand sadly).

accepted/00000-replace-salt-toaster.md
SchoolGuy: I am against this. Sumaform has a defined scope and if we generify everything we run into a kraken of features again. The slightest hickup in the wrong step causes a complete destroy and apply cycle. Let's not do more then needed. But that is only my view on things.

agraul: A generic sumaform option could be useful if we wanted to define states to apply to a module in `main.tf`, without having to touch `salt/top.sls`. This would allow us to extend an existing "role" with arbitrary state files (that are already present in `salt/` for a single "module".

If we had that option, we could define this salt_testenv module completely in a `main.tf`. I prefer not to go this route because it would make setting up a VM harder. The proposed module defined in "modules/salt\_testenv" makes it easy to deploy such a VM without having to know the name of the state file, or the required memory.

Generally this is could be a nice feature. I believe it's orthogonal to adding this new module though.

accepted/00000-replace-salt-toaster.md
agraul: A Git submodule is another Git repository, it's just a way of specifying the relationship. It's  possible to just `git clone` another Git repository instead, that way we would need to manually check out and we couldn't take advantage of a unified `git diff`. I don't really see a reason not to have the submodule relationship set up at this point.

The second question is more interesting to me, at least if I interpret it correctly: Is there a way to apply the states from another directory without merging them into `salt/`?
I don't know yet if it works and am curios to try a different approach: Could we inject this other directory as a second `file-root` into the `salt-call` execution that runs the provisioning and let Salt merge the state trees? 

I'm trying this out and see what it would take (if possible), I'll let you know when I got something.

accepted/00000-replace-salt-toaster.md
agraul: Currently a similar list in defined in [Salt Toaster](https://github.com/openSUSE/salt-toaster/blob/master/conftest_source_nox.py#L1343).

I would put the skip list in the same repository as the above mentioned test Script, mainly to reduce the number of repositories used. Re-using `openSUSE/salt` only works if we disable direct pushes to the branches that contain the skip list and always remember to exclude all changes touching the test script and skip list from our packaging automation. IMO it makes sense to keep the script and the definition outside of the `openSUSE/salt` repo.

accepted/00000-replace-salt-toaster.md
agraul: Both the symbolic link approach and the copy approach to the state tree merger must be executed before running `terraform apply`. I haven't thought too hard about where to define the merging. I wouldn't hard-code it into a Jenkins job, at the very least it should be in a shell script to maintain the code properly.
 My preferred way, that should work based on my (limited) experience with it, is to define the merging in terracumber (symbolic links, hard links or copying). This area needs a bit more research.

accepted/00000-replace-salt-toaster.md
agraul: It's good to know about the SSH mode, I wasn't aware of that!

For this use case, I don't think it's a good alternative. As far as I know, we would need to add a main.tf just for that to openSUSE/salt. The main.tf would need a module that targets localhost, just to execute a Salt state (which we first need to merge as well). That's a bit complicated for our use case compared to my (ugly) PoC GH Actions jobs definition:
```yaml
jobs:
  build-then-test:
    runs-on: ubuntu-22.04
    steps:
      - name: Check out repository code
        uses: actions/checkout@v3
      - name: Install salt-call
        run: |
          sudo apt-get update
          sudo apt-get install salt-common
      - name: Check out Salt Jenkins
        uses: actions/checkout@v3
        with:
          repository: saltstack/salt-jenkins
          path: salt-jenkins
      - name: Install test dependencies
        working-directory: ./salt-jenkins
        run: sudo salt-call --local --file-root=state-tree state.apply golden-images-provision
      - name: Run tests
        run: nox --session 'pytest-3(coverage=False)' -- tests/pytests/unit/

File: accepted/00000-replace-salt-toaster.md [hunk 7]
```diff
@@ -0,0 +1,192 @@
...
+
+A new module "salt\_testenv" is added to sumaform. This module defines a new
+role, "salt\_testenv", and the default values needed to run all test suites, e.g.
+4GiB of RAM. Sumaform uses this role to apply the `golden-images-provision.sls`
+state file from the Salt-Jenkins repository.
+
+Up to now, all state files used by sumaform are maintained in the sumaform Git
+repository. For provisioning the salt\_testenv VMs, we don't want to maintain
+our own states that define which dependencies need to be installed on what
+operating system. Salt upstream already maintains such states for their own test
```
meaksh, line 57: Hmmm, I'm not sure here, as one of the thing the current Salt toaster was taking care of is to test the installation of the Salt package (RPM or DEB) that we built.

Also run the tests using the Salt dependencies that are installed also from the OS + custom Salt repos we built.

This way, we ensure we are testing the Salt codebase with the actual dependencies and versions that customer are going to get when installing this Salt package in real life.

Do you have any idea how to solve this?

If we rely on Salt upstream and install the Salt dependencies via "pip", with a predefined version specified by upstream, there might be differences with the versions the customers will get when install Salt coming from our repos, or that some of the dependencies are not even available in these repos.

So, ideally during provisioning, we would like to test the package installation of our generated Salt package, and then while testing, using the dependencies that are pulled as requires when installing the Salt package.

This way we could cover the cases of: 
- Missing installation dependencies for Salt in our repositories. 
- The version of an installed RPM/DEB dependency of Salt is different that what "pip" installs. (due available packages in repo or due spec/control speficiation)

meaksh (commented):
Thank you Alex for pushing this RFC forward!

I added a question about "package installation" test and dependencies, but beside of that, I'm also wondering how we would test old OSes, like SLE12SP5, CentOS7 or Ubuntu 18.04.

BTW, any idea about making the Salt Toaster to test "venv-salt-minion" on different OSes? So, for example, to really run the tests using "venv-salt-minion" (Py3.10 + 3004 based) on a, for instance, SLE12SP5 system?

I think we should have some thoughts also on this and make the Salt Toaster to help us on covering all these situations.

accepted/00000-replace-salt-toaster.md
moio: > I am against this. Sumaform has a defined scope and if we generify everything we run into a kraken of features again. The slightest hickup in the wrong step causes a complete destroy and apply cycle. Let's not do more then needed. But that is only my view on things.

I respect that.

> I believe it's orthogonal to adding this new module though.

Completely agreed.

FMPOV, in general, having clear mechanisms to extend functionality of a project helps in use cases in which it does not make sense to add overspecific code - although it's of course delicate to set the interface well. I will let sumaform maintainers make the final call.

accepted/00000-replace-salt-toaster.md
agraul: That's quite a big comment, let me try to clarify this a bit.

> Hmmm, I'm not sure here, as one of the things the current Salt toaster was taking care of is to test the installation of the Salt package (RPM or DEB) that we built.
> 

We still need to put the Salt source code on the test VM, I didn't cover that yet. Salt should come from our packages to ensure we test the program we want to test. That this will also act as an "installation test" is a nice side effect, but IMO not really important within the scope of running Salt's testsuite. 


> Also runs the tests using the Salt dependencies that are installed also from the OS + custom Salt repos we built.
> 

That stays the same.

> This way, we ensure we are testing the Salt codebase with the actual dependencies and versions that customers are going to get when installing this Salt package in real life from our repos.
> 
> Do you have any idea how to solve this?
> 

I haven't looked deeply into this, right now I expect we can install Salt from the correct repository with `terracumber` just ahead of executing the test suite. 

> If we rely on Salt upstream and install the Salt dependencies via "pip", 

The Salt state tree that provisions the VMs does not install Salt dependencies with `pip`, with the exception if two state files. We need to decide what to do about these two.

>with a predefined version specified by upstream, there might be differences with the versions the customers will get when installing Salt coming from our repos, or that some of the dependencies are not even available in these repos.
> 
> So, ideally during provisioning, we would like to test the package installation of our generated Salt package, and then while testing, using the dependencies that are pulled as requires when installing that Salt package.
>

I agree that this would be ideal. At this point, we can't enforce that the dependencies used during the test execution come from packages because `nox` uses `pip` to fill the virtual environment used for testing. I don't know how much work it would be to override that behavior. 

> This way we could cover the cases of:
> 
>     * Missing installation dependencies for Salt in our repositories.
> 
>     * The version of an installed RPM/DEB dependency of Salt is different that what "pip" installs. (due available packages in repo or due spec/control speficiation)
> 
> 
> What do you think?

I think we need to consider the trade-off between how closely the test environment resembles the production environment and how much work it is to maintain the test infrastructure. If it turns out it's not possible to use the very same dependencies in the test environment but we see that as a requirement, we would need to rethink our approach to testing without `nox` or using `pip`.

accepted/00000-replace-salt-toaster.md
```diff
@@ -0,0 +1,192 @@
...
+infrastructure. We re-use their states by adding `salt-jenkins` as a Git
+submodule to sumaform and merge the salt state trees.
+
+In order to avoid future state id clashes, all existing state ids will get a
+"sumaform-" prefix.
+
+### Merging Sumaform's and Salt-Jenkins' Salt Trees
+
+The state trees are merged by creating symbolic links inside sumaform's `salt`
```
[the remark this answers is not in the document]
agraul: Yes, that's a good idea as well. So far I looked at this being just another sumaform module that is treated the same as all others.

I see two elements in your comment that I want to address separately:

1) Decoupling the salt state trees
2) Running a third `salt-call` invocation for this particular module

I looked into 1) last week, and I like this more than merging the trees on the sumaform host system. In my [experiment](https://github.com/agraul/sumaform/commit/0187986b326dd2177eb4f7795a151e3b87a05565) I added another `provisioner "file"` module to the libvirt host module that uploads another salt state tree based on a variable that can be pre-filled in "salt\_testenv". What didn't work was using it as another `--file-root` parameter for `salt-call` (only the last `--file-root` is used), but it's possible to write an `/etc/salt/minion` config with both `salt/` and this arbitrary salt state tree. It was just a PoC and is rough around the edges, but this could be a better alternative to handle 2 distinct salt state trees.

I still included the role->golden-images-provision mapping in `salt/top.sls`. This could be moved out of `salt/top.sls` and into an explicit `salt-call state.apply golden-images-provision` `remote-exec` call in the "salt\_testenv" module. Or even a generic "run all states specified by a variable" feature that moio brought up in a different comment above.

I don't have a strong preference for how we apply the state, what would you prefer from a sumaform maintenance POV?

accepted/00000-replace-salt-toaster.md
agraul: `--file-root` can only be used once, but it's possible to define multiple file roots in `/etc/salt/minion`. `sumaform` could generate such a config file (like it does for `/etc/salt/grains`), and we won't need to merge the trees ourselves.

accepted/00000-replace-salt-toaster.md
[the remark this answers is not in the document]
rjmateus: > What didn't work was using it as another `--file-root` parameter for `salt-call` (only the last `--file-root` is used), but it's possible to write an `/etc/salt/minion` config with both `salt/` and this arbitrary salt state tree. It was just a PoC and is rough around the edges, but this could be a better alternative to handle 2 distinct salt state trees.

Instead of adding a second `--file-root` would be possible to run a second salt-call using the additional_salt_tree as file-root?
This way we could just run the current existing workflow, and as a secondary step, run the state apply (or high-state) for the additional salt-tree.

Do you foresee any dependency between the two salt-trees that would force us to run them at the same time?

> 
> I still included the role->golden-images-provision mapping in `salt/top.sls`. This could be moved out of `salt/top.sls` and into an explicit `salt-call state.apply golden-images-provision` `remote-exec` call in the "salt_testenv" module. Or even a generic "run all states specified by a variable" feature that moio brought up in a different comment above.

I would keep this in the separated salt-tree.


> I don't have a strong preference for how we apply the state, what would you prefer from a sumaform maintenance POV?

I like our approach of passing an additional salt-tree, but I would run the high state as a separate additional step during the provisioning.
This way we can debug what is sumaform specific and what is being provided outside sumaform, and if any problem pop-up we would be able to point the right direction.

accepted/00000-replace-salt-toaster.md
agraul: We discussed this topic a bit on a call. The outcome is that we evaluate if `nox` actually makes running the tests easier (against our packaged dependencies). The alternative to compare this against is to run a plain `pytest`. In that case we need to install all dependencies used in the tests manually. A `python3X-salt-tests` subpackage could require all the dependencies (that are for tests only) and ship the tests themselves.

accepted/00000-replace-salt-toaster.md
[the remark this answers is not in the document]
agraul: > Instead of adding a second --file-root would be possible to run a second salt-call using the additional_salt_tree as file-root?

I'm not sure how easy it is to design this in sumaform with the right amount of abstraction. I tried adding a `provisioner "remote-exec"` block to my PoC salt-test module's main.tf but got the error `Blocks of type "provisioner" are not expected here.`. There likely is a way to do this in a nice way, I haven't found it yet.

> Do you foresee any dependency between the two salt-trees that would force us to run them at the same time?

No. I thought of provisioning as a single step, that's why I kept them together.

> This way we can debug what is sumaform specific and what is being provided outside sumaform, and if any problem pop-up we would be able to point the right direction.

That's a fair point.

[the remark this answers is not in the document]
rjmateus: > I'm not sure how easy it is to design this in sumaform with the right amount of abstraction. I tried adding a `provisioner "remote-exec"` block to my PoC salt-test module's main.tf but got the error `Blocks of type "provisioner" are not expected here.`. There likely is a way to do this in a nice way, I haven't found it yet.
> 
I can give you a hand on this, I have some ideas for it. It would be easier to explain in a call. If you can we can arranje one, otherwise I can try to explain in the issue.

accepted/00000-replace-salt-toaster.md
```diff
@@ -0,0 +1,220 @@
...
+when running the tests, but adds the overhead of maintaining ".deb" packages of
+all test dependencies.
+
+Using the Salt Bundle flavor that we install in the Sumaform-provisioned test
+VMs could be an option, but then we would need to move the application and test
+code into the bundle.
+
+## Terracumber
+
+Terracumber is a tool used by our QE squad to operate terraform and cucumber for
```
rjmateus, line 97: terracumber downloads sumform and run a terraform file with the definition we want, and passes a set of variables. IT can also be done without terracumber. 
It works as a help tool to reduce the load on configuring and managing the pipelines on jenkins.

rjmateus (approved):
LGTM. Just one small clarification

accepted/00000-replace-salt-toaster.md
```diff
@@ -0,0 +1,220 @@
...
+GitHub Actions use so-called runners, which are VMs running Windows, macOS or
+Linux. It is possible to host your own runners, but we currently neither do not
+do that. We can't use sumaform to provision GitHub-hosted runners, so GH Actions
+will use a different approach.
+
+Since the GitHub-hosted Linux runners only run Ubuntu, we will use a Docker
+container that is based on openSUSE Leap (latest version). This container's
+Dockerfile installs a Salt sub-package that pulls in all required test
+dependencies. The container image can be build in OBS, which will automatically
+rebuilt it when included packages change.
```
SchoolGuy, line 82: Open question: How will we wait for the built container image with the new code? If everything comes from OBS (code + test-code). Or do we plan on not building this as an RPM on the GH pipeline?

SchoolGuy (approved):
Overall this is a better solution imho. The projects we depend upon are required by the project as a whole, thus we don't introduce new dependencies. Of course they are new for the Salt distribution itself but I believe this is an acceptable risk.

accepted/00000-replace-salt-toaster.md
agraul: For Pull Requests, I think we only need to install the dependencies (`pytest`, `pytest` plugins, 3rd-party libs used in testing etc.) and both the application and test code is "installed" with `actions/checkout`. I don't expect this container to change often and we don't run such tests a lot either. Therefore I didn't consider that we might run into "tests fail due to outdated container".

accepted/00000-replace-salt-toaster.md
agraul: That's true. Do you think there is a better way to introduce terracumber?

accepted/00000-replace-salt-toaster.md
SchoolGuy: Okay that would be fine by me to remove that additional step of package the test code and the application code on GH.

accepted/00000-replace-salt-toaster.md
rjmateus: I think is fine as it is

meaksh (approved):
I like this new approach, it feels consistent and I also like we are integrating Salt testing into sumaform/terracumber tools.

FMPOV this is good to go! Thanks :+1:

PR #72, opened 2022-08-25
Hub Retail Extension

nadvornik:
https://github.com/SUSE/spacewalk/issues/9594
This is a rewrite of  https://github.com/uyuni-project/uyuni-rfc/pull/45.

Render available [here](https://github.com/nadvornik/uyuni-rfc/blob/hub/accepted/00095-hub_retail.md)

### DIFF
--- a/accepted/0000-hub_retail.md
+++ b/accepted/0000-hub_retail.md
@@ -19,70 +19,132 @@ Reasons for use of Hub with Retail environment
 - disconnected operation
   - Peripheral server must stay fully functional when the Hub connection is down
 
+See [Hub requirements](https://github.com/SUSE/spacewalk/wiki/Hub-requirements) for details.
 
 Possible configurations
 -----------------------
 
-1. Full
+1. Standard
+```
+    Terminal -> Branch Server -> SUMA Server
+```
+  This is the currently supported configuration.
+  It is suitable for one large department store.
+  For larger deployments there may be problems with the number of terminals or with network outages.
+  The SUMA Server must be reachable in order to perform any change of Terminal configuration.
+
+  Terminal
+    - deployed from OS Image during boot (by applying Saltboot state)
+    - then works as standard SUMA minion
+
+  Branch Server
+    - SUMA Proxy
+    - downloads Images from SUMA (over HTTP) and distributes them to Terminals (over HTTP, TFTP, etc.)
+    - runs services needed by Terminals (DHCP, TFTP, HTTP, ...). The services are configured via Formulas with Forms.
+
+  SUMA Server
+    - standard SUMA operations
+    - holds the configuration for Branch server services in Formulas
+    - controls boot process of each Terminal by applying Saltboot state
+    - triggers update of PXE data on Branch Server
+
+
+2. Full
 ```
     Terminal -> Branch Server -> Peripheral Server -> Hub
 ```
+  Adding Hub on top of previous configuration solves the problems with scalability and offline operations.
+  It is suitable for for large organization with many department stores. It can look like this:
 
-2. Combined SUMA and Branch server 
+  - Branch server on each floor
+  - Peripheral server in each building
+  - Hub in datacenter
+
+
+3. Combined Peripheral and Branch server
 ```
     Terminal -> Branch + Peripheral Server Combined -> Hub
 ```
 
-3. Standard
-```
-    Terminal -> Branch Server -> SUMA Server
-```
+  Customers with many small shops do need Hub for offline operations but do not need
+  separate Branch server and Peripheral Server because the number of connected terminals is low.
+  Running Branch + Peripheral Server on one machine could save resources.
+  The Proxy part of Branch server is not needed then.
+
 
 4. Single server
 ```
     Terminal -> Branch + SUMA Server Combined
 ```
+  Suitable for one small shop.
+  This scenario is not a hard requirement but there were such requests for SLEPOS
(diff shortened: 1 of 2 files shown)


nadvornik:
The remaining open question is how to creates global groups with associated formula data.
https://github.com/nadvornik/uyuni-rfc/blob/hub/accepted/0000-hub_retail.md#hwtype-configurations

@rjmateus any opinion?

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,211 @@
...
+
+  - Images - distributed as large tarballs via http
+  - Image metadata - pillars, activation keys, list of repos and packages (in DB)
+  - [HWTYPE](https://www.uyuni-project.org/uyuni-docs/en/uyuni/retail/retail-deploy-terminals.html#_create_a_hardware_type_group) configurations
+
+## Local data
+
+The following items do not need central management and can stay on Peripheral Server (like in the current implementation).
+
+  - Branch groups -  groups containing Branch Server and connected terminals, includes Saltboot groups for Containerized Branch Server
```
rjmateus, line 111: This Banch groups are normal systems groups, right? This can be managed from HUB server with the configuration formula.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,211 @@
...
+
+Combined Peripheral and Branch server:
+
+TBD
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- is nested Hub needed?
+  The use case with one Peripheral Server per shop might exceed the maximum number of Peripheral Servers.
```
rjmateus, line 209: I'm not expecting that one peripheral per shop is main use case.
Even in that scenario, one hub could manage up to 10k suse manager servers, so It should be enough. If not, that is a problem I think everyone would be glad to solve :)

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,211 @@
...
+[YAML](https://github.com/uyuni-project/retail/blob/master/python-susemanager-retail/example.yml) file.
+[Hub Users RFC](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00061-hub-users.md) introduced similar concept, based on Salt.
+Maybe these parts could be merged.
+
+
+## Combined Peripheral and Branch server
+
+The components from Containerized Branch Server can be re-used also in this scenario.
+
+Required components:
```
rjmateus, line 159: With the containerized proxy this would work out of the box?

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,211 @@
...
+- Cobbler / TFTP   
+- Mapping of branch-local URLs
+- DHCP server
+
+
+### Mapping of branch-local URLs
+
+https://github.com/SUSE/spacewalk/issues/18455
+
+Non-containerized retail branch server uses image pillars to map branch-local image URLs to SUMA server URLs.
```
rjmateus, line 171: I'm not sure if we should plan to support non-containerizes branch proxy with suse manager in the same VM.
If they run on the same site, but in different VM it should work as is.

accepted/0000-hub_retail.md
```diff
@@ -0,0 +1,211 @@
...
+
+Containerized branch server does not have access to pillars, so it currently implements this 
+functionality by apache mod_rewrite. This is quite error-prone, because it expect URL in certain 
+format. It can't work with images on external server.
+https://github.com/SUSE/spacewalk/blob/Manager-4.3/containers/proxy-httpd-image/uyuni-configure.py#L166
```
rjmateus, line 176: This shouldn't be extracted into self-contained issue? It looks like a general problem and not a hub-specific one.

accepted/0000-hub_retail.md
nadvornik: They are normal groups with attached group formulas. AFAIK the HUB configuration formula can't handle group formulas yet.

accepted/0000-hub_retail.md
nadvornik: It turned out that it does not need any extra services on SUMA server, besides what is already needed for cobbler (tftp, dhcp) and pillar mapping https://github.com/SUSE/spacewalk/issues/18455.

accepted/0000-hub_retail.md
nadvornik: Containerized proxy basically forwards the requests to SUMA server. That means that the required services are already available on SUMA server and can be used also by directly connected terminals.

accepted/0000-hub_retail.md
nadvornik: yes, it is self-contained issue, 
https://github.com/SUSE/spacewalk/issues/18455

accepted/0000-hub_retail.md
rjmateus: correct, it can only create system groups.

nadvornik: @rjmateus so is it OK to extend the HUB configuration formula to create also group formulas? It is the alternative 2 here https://github.com/nadvornik/uyuni-rfc/blob/hub/accepted/0000-hub_retail.md#hwtype-configurations

rjmateus: from my point of view yes. In this way, we can have a centralized place to configure the peripheral servers.

nadvornik: I have updated the RFC accordingly.

PR #73, opened 2022-09-23
SLE Micro - report when a reboot is needed

wweellddeerr:
[Updated version](https://github.com/wweellddeerr/uyuni-rfc/blob/d245a14da0e27e7ddebd00d7ab7255c54f230e12/accepted/00090-sle-micro-reboot-needed.md)


aaannz:
Instead of execution module, can we use salt beacon and then watch for beacon events in our reactor java handler? This way we wouldn't need to adapt sls files. For transactional system we would just need to ensure our beacon is installed.

accepted/0000-sle-micro-reboot-needed.md
```diff
@@ -0,0 +1,61 @@
...
+# Detailed design
+[design]: #detailed-design
+
+The SLE Micro uses transactional updates. Transactional updates are atomic (all updates are applied only if all updates succeed) and support rollbacks. They don't affect a running system as no changes are activated until after the system is rebooted.
+To be able to have up-to-date information regarding reboot needed, we need to update this information after every state apply.
+For that, this proposed solution is based on the following steps:
+ - Create an execution module to check if a reboot is required
+ - After every state apply run this execution module
+ - Store the result in a postgresql column
+ - Report reboot needed in the UI based on this column
```
aplanas, line 26: There is another module that can help (I am not sure if the service is present in SLE Micro) when doing a reboot.

[reboot_manager](https://docs.saltproject.io/en/3004/ref/modules/all/salt.modules.rebootmgr.html) is also supported in salt.  With this we can create "reboot windows" parameterized by the user.

We can trigger a reboot (using the configured option by the user) using the [reboot](https://docs.saltproject.io/en/3004/ref/modules/all/salt.modules.transactional_update.html#salt.modules.transactional_update.reboot) command from the `transactional_update` module.

accepted/0000-sle-micro-reboot-needed.md
```diff
@@ -0,0 +1,61 @@
...
+
+[This draft Pull Request](https://github.com/SUSE/spacewalk/pull/19036) can help to understand the general idea, and the details are explained below.
+
+### The execution module
+
+To check if a reboot is needed we can use `snapper` to determine if the default snapshot is not active. If so, a reboot is pending. One can find more details about this strategy in [Jira](https://jira.suse.com/browse/CSD-95) discussion. The execution module based on this strategy can be seen [here](https://github.com/SUSE/spacewalk/blob/3c2031bac04fbe88cd5b6692a3e511a457cdf5e7/susemanager-utils/susemanager-sls/src/modules/rebootutil.py).
```
aplanas, line 32: I think that `pending_transaction` is the correct call here ... not sure why it is not mentioned. `t_u` and `r_m` modules are complex and cover all the use cases.

rjmateus:
> Instead of execution module, can we use salt beacon and then watch for beacon events in our reactor java handler? This way we wouldn't need to adapt sls files. For transactional system we would just need to ensure our beacon is installed.

I agree with @aaannz If possible, the best approach could be to develop a beacon. Is it enough to what the existence of the file " /var/run/reboot-required" and trigger the beacon if the file changes?
If not possible, then we can develop a custom beacon that checks the snapshots, as @wweellddeerr did in the salt module.
With this approach, the only change needed in the sls would be to install the beacon if the system is using transactional updates.

rjmateus:
One more topic, with the beacon we would also cope with any changes the user could do directly to the system. It's more or less the same approach we do when a user directly installs a package and a package refresh needs to be triggered.

mcalmer:
@rjmateus SLE-Micro support currently 4 different reboot manager. All indicate a required reboot different. So we need to decide if we want to rely on them, or if we introduce a 5th one called "SUMA" and we put out own indicator.
Which raise the next question: should we support the other reboot manager or not?

accepted/0000-sle-micro-reboot-needed.md
```diff
@@ -0,0 +1,61 @@
...
+This is not ideal when it comes to usability. The user is responsible for knowing when a reboot is needed, and scheduling those reboots.
+
+
+# Detailed design
+[design]: #detailed-design
+
+The SLE Micro uses transactional updates. Transactional updates are atomic (all updates are applied only if all updates succeed) and support rollbacks. They don't affect a running system as no changes are activated until after the system is rebooted.
+To be able to have up-to-date information regarding reboot needed, we need to update this information after every state apply.
+For that, this proposed solution is based on the following steps:
+ - Create an execution module to check if a reboot is required
```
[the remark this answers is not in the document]
aplanas: WRT grains: sadly I did not added the grain in the [extra](https://github.com/saltstack/salt/commit/94b6ab1cdbe78513e066fc95c70039ff727c8674) grains, but luckily there is a [place](https://github.com/saltstack/salt/blob/master/salt/grains/pending_reboot.py) to add one upstream!!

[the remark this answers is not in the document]
aplanas: > No, grains do not update automatically. Grains are made for values which do not change often. It will not work for this usecase.

Grains can be refreshed, and this is the approach the upstream is using for Windows (see last comment)

aplanas:
> SLE-Micro support currently 4 different reboot manager.

`pending_transaction` will select the correct one (unless bugs): https://github.com/saltstack/salt/blob/master/salt/modules/transactional_update.py#L151

vzhestkov:
I like the solution suggested by @aaannz . I think it fits better for the usecase.

accepted/0000-sle-micro-reboot-needed.md
agraul: That's true, and it's a good way if we can assume that `rebootmgr` is the only reboot manager used. As far as I am aware, `transactional-update` can schedule reboots for 4 reboot managers (https://github.com/openSUSE/transactional-update/blob/2e15fee623b24e9ab8e382253d4930e91a912bc8/lib/Reboot.cpp#L30) which we probably all need to support in Uyuni. We could get around this by defining that SLE Micro managed by Uyuni can only use reboot manager X, which could also be Uyuni itself (i.e. transactional-update would need to be expanded to trigger a reboot for a 5th reboot manager)

aplanas: Right, and the method is set in [transactional-update.conf](https://github.com/openSUSE/transactional-update/blob/master/man/transactional-update.conf.5.xml#L76), so can be configured by Salt.

That is also with the [reboot](https://docs.saltproject.io/en/3004/ref/modules/all/salt.modules.transactional_update.html#salt.modules.transactional_update.reboot) command, together with [pending_transaction](https://docs.saltproject.io/en/3004/ref/modules/all/salt.modules.transactional_update.html#salt.modules.transactional_update.pending_transaction) command pair should be used.

accepted/0000-sle-micro-reboot-needed.md
wweellddeerr: @aplanas I think you are right, it was not mentioned because I was not aware of this possibility. I will update the RFC to use `pending_transaction` to check if a reboot is needed, seems to be the best way to do it.

mbussolotto:
Hi @wweellddeerr , JFYI I'm working on this issue https://github.com/SUSE/spacewalk/issues/18381: the description might be misleading, shortly reboot does not work in SLE Micro on an action chain (no issue if reboot is triggered outside an action chain). I think there's no confict between my changes and your RFC, but if you want to have a look the PR is here ( https://github.com/uyuni-project/uyuni/pull/5772 )

rjmateus:
I think we could develop a beacon, as ondrej suggested, and implement it in the same way Alberto mentioned in the salt module [1].
To perform the update, I think we could leverage the existing salt module [2] and delegate the reboot to whatever is configured in the system.
One more improvement could be to allow changing the configured reboot mechanism on the transactional system being in use.
@aplanas @aaannz @mcalmer any though on this?

@mcalmer I'm not sure if we should develop a SUMA reboot mechanism. What would be the benefits? Do we need some special behavior? Or does SUMA even care how the reboot is done, or only care that it happens?


[1] https://github.com/saltstack/salt/blob/master/salt/modules/transactional_update.py#L889-L909
[2] https://docs.saltproject.io/en/3004/ref/modules/all/salt.modules.transactional_update.html#salt.modules.transactional_update.reboot

aplanas:
> I think we could develop a beacon, as ondrej suggested, and implement it in the same way Alberto mentioned in the salt module

Makes sense. For the beacon, should be noted that there is a state machine implicit:

```
[!Pending Transaction] --{update}--> [Pending Transaction] --{t_u.reboot}--> [Reboot Planned]
                                                          <--{rmgr.cancel}--
         ^                                                                          |
         \ -----------------------------{system reboot}-----------------------------/
```
With `reboormgr` there is a reboot window, so the system can be in `Reboot Planned` state for long. Also there is a missing cycle from `Reboot Planned` to itself, when a second `t_u_reboot` is done.

mcalmer:
> I think we could develop a beacon, as ondrej suggested, and implement it in the same way Alberto mentioned in the salt module [1]. To perform the update, I think we could leverage the existing salt module [2] and delegate the reboot to whatever is configured in the system. One more improvement could be to allow changing the configured reboot mechanism on the transactional system being in use. @aplanas @aaannz @mcalmer any though on this?

Beacon sounds interresting. I think we can give it a try.

> @mcalmer I'm not sure if we should develop a SUMA reboot mechanism. What would be the benefits? Do we need some special behavior? Or does SUMA even care how the reboot is done, or only care that it happens?

I think SUMA must not ignore the configured rebootmanager. If the customer wants "kued" or "rebootmgr" than he has a reason and in SUMA we should not allow to reboot using our action.

But what is a reboot manager? You can define rules when and how a reboot happens. This can be some with SUMA as we have maintenance windows and things like this. So SUMA "is" a reboot manager. Just transaction-update tool does not now it (yet).

mcalmer:
> > I think SUMA must not ignore the configured rebootmanager. If the customer wants "kued" or "rebootmgr" than he has a reason and in SUMA we should not allow to reboot using our action.
> 
> Agree. But it should be safe to call the transactional update reboot action, wich should be able to deal with this scenario.

Yes, but in this case we cannot know "when" the reboot will happen. It may take hours or days.
SUMA normally schedule a reboot and keep the action in state "Picked Up". Only on a minion start event SUMA set the reboot action to "completed". But there is a timeout or 6 hours. If the reboot does not happen within 6 hours, the action is set to failed.

 
> > But what is a reboot manager? You can define rules when and how a reboot happens. This can be some with SUMA as we have maintenance windows and things like this. So SUMA "is" a reboot manager. Just transaction-update tool does not now it (yet).
> 
> Sure, we can also be a reboot manager. But should we be "introduced" to the transaction-update tool or should we follow the path of re-using the existing mechanisms?

Well, I think we should provide a method where a customer has 100% control about every aspect of management of a system. And for this we require that SUMA has the control over the reboot.

> 
> For example, maintenance windows. Suma has it, and I think transaction-update tool also has a manager for that.

No, this is not the case. The transactional-update tool can just request a reboot from the configured rebootmanager.
transactional update does not know about how they work.

> We can leverage that manager our we can develop our own to reboot. If we follow the path of developing our own mechanism, then we can re-use the existing action chains. If we decided to only re-use existing mechanisms, then we can focus on the easy configuration of them from suse manager side. (I'm just dropping some possible paths in here, but it's not the focus of the RFC, since this one is for report need reboot)

I think we should get deeper into "how the reboot managers work". I think there are some assumption in the room which are not true.

There is one possibility I see and this would be to set the rebootmanager to "systemd". This would always reboot immediately when transactional-update request it. With this we could be sure that the reboot happens when we request it or when it is needed. But configured maintenance windows on SUMA side would maybe ignored.

Btw: keeping SUMA maintenance windows in sync with  the maintenance windows configured in the reboot manager on the system is maybe not an option. For this both would need to provide the same level of configuration options. In SUMA you just add a calendar and with this you can very free define when your maintenance window is. Not sure if this is possible with rebootmgr and/or kured.

wweellddeerr:
> Instead of execution module, can we use salt beacon and then watch for beacon events in our reactor java handler? This way we wouldn't need to adapt sls files. For transactional system we would just need to ensure our beacon is installed.

@aaannz @rjmateus @vzhestkov do you have any suggestion about how to know whether the status changed from the last execution of the beacon? Would be storing the status in a file? Otherwise the only thing that is possible to know is the current status.

aaannz:
> @aaannz @rjmateus @vzhestkov do you have any suggestion about how to know whether the status changed from the last execution of the beacon? Would be storing the status in a file? Otherwise the only thing that is possible to know is the current status.

Maybe I'm missing something, but why do you want to track changes? Is the knowledge about current status not enough?

Beacon event fires -> store reboot is needed flag in the database.

wweellddeerr:
> > @aaannz @rjmateus @vzhestkov do you have any suggestion about how to know whether the status changed from the last execution of the beacon? Would be storing the status in a file? Otherwise the only thing that is possible to know is the current status.
> 
> Maybe I'm missing something, but why do you want to track changes? Is the knowledge about current status not enough?
> 
> Beacon event fires -> store reboot is needed flag in the database.

It is enough in order to have up-to-date information about the status. But, is it a good approach to keep sending the information even if nothing changed in the system? Maybe it will not be a problem, but from my perspective, the ideal would be to update the information only when the status has changed. If we have a large installation, can't the frequent sending of this data from thousands of systems flood the master?

vzhestkov:
@wweellddeerr try to check `__context__` dict first. It could be used to store any variables between execution the modules for the minion service. In breif the module can be reloaded few times with LazyLoader, but this variable could be used to store the data betwen the shots of the beacon. The other possible way to store or if you need more persistent store (will be available even if the service is restarted) is the salt cache.

admd:
> Well, I think we should provide a method where a customer has 100% control about every aspect of management of a system. And for this we require that SUMA has the control over the reboot.

Correct. 

To keep things in control and in separation, uyuni should have its own reboot manager in this case and once that is enabled, all other reboot managers should be disabled(ideally! unless I am overlooking a critical use case). This reboot manager could be as dumb as just stating if a reboot is needed and rebooting the machine using systemd or whatever when instructed to do so. I don't think there are hard specifications that a reboot manager needs to fulfill.
@mcalmer @rjmateus

aaannz:
> To keep things in control and in separation, uyuni should have its own reboot manager in this case and once that is enabled, all other reboot managers should be disabled(ideally! unless I am overlooking a critical use case). This reboot manager could be as dumb as just stating if a reboot is needed and rebooting the machine using systemd or whatever when instructed to do so. I don't think there are hard specifications that a reboot manager needs to fulfill. 

If SLE Micro instance is used as k3s node or just any container host, in case of our custom reboot manager we would need to implement `kured` functionality as well. So we can drain/cordon node before reboot.

mcalmer:
> > To keep things in control and in separation, uyuni should have its own reboot manager in this case and once that is enabled, all other reboot managers should be disabled(ideally! unless I am overlooking a critical use case). This reboot manager could be as dumb as just stating if a reboot is needed and rebooting the machine using systemd or whatever when instructed to do so. I don't think there are hard specifications that a reboot manager needs to fulfill.
> 
> If SLE Micro instance is used as k3s node or just any container host, in case of our custom reboot manager we would need to implement `kured` functionality as well. So we can drain/cordon node before reboot.

Well, depends. I am totally in favor of allowing kured to be configured as reboot manager. In this case SUMA should ask kured to reboot. But we need to handle this action a bit different to not fail the action too early.

But if the customer decide in this case to use SUMA, then he should implement such a mechanism  with states or workflows.
Not sure if we need to provide this feature integrated into SUMA. This would be more a nice to have than a requirement.

admd:
> > To keep things in control and in separation, uyuni should have its own reboot manager in this case and once that is enabled, all other reboot managers should be disabled(ideally! unless I am overlooking a critical use case). This reboot manager could be as dumb as just stating if a reboot is needed and rebooting the machine using systemd or whatever when instructed to do so. I don't think there are hard specifications that a reboot manager needs to fulfill.
> 
> If SLE Micro instance is used as k3s node or just any container host, in case of our custom reboot manager we would need to implement `kured` functionality as well. So we can drain/cordon node before reboot.

Thank you @aaannz, I overlooked this. 

Just thinking out loud, if suma should be just a delegator to send reboot request to the actual reboot manager, suma should give this information back in UI that a reboot has been requested/scheduled , then depending on the configured reboot manager this could be an immediate reboot or later at some point and once actual reboot will take place, we will update the SUMA UI. 

Once the reboot is requested, is the user still allowed to make changes in SLE Micro before that previously requested reboot to actually happen?

rjmateus:
> Just thinking out loud, if suma should be just a delegator to send reboot request to the actual reboot manager, suma should show this information in UI that a reboot has been requested/scheduled, then depending on the configured reboot manager this could be an immediate reboot or later at some point and once actual reboot will take place, we will update the SUMA UI.

I think that is the path to follow, on suma be a delegator (and not a reboot manager by itself). As @mcalmer said, suma expects that the reboot will happen in the next 6 hours (IIRC), otherwise it will be set as failed. 

We should also think about what types of reboot we should allow from suma. I would say at least "kued" and systemd (for the immediate reboot)

admd:
> We should also think about what types of reboot we should allow from suma. I would say at least "kured" and systemd (for the immediate reboot)

@rjmateus why should we limit to these methods if all SUSE Manager needs to do is to delegate? Do you think we will need to handle it differently for each rebootmanager that SLE micro supports? 

I would expect SLE Micro to have a single entry point that we need to use, no matter what reboot manager is configured and then the rest is an implementation detail for us. Would that not be the case?

accepted/0000-sle-micro-reboot-needed.md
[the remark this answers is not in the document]
mcalmer: but refreshing grains after every change can be very expensive. You refresh "all" grains, not only the one which you are interested in.
Additionally I see the following happen.
- user  update a kernel package
- user request reboot needed grain
  -> no reboot needed

Why? Forgot to refresh. The grains was taken out of the cache and that a kernel was updated did not refresh the grain.

PR #77, opened 2023-03-04
RFC: Recurring Actions

parlt91:
RFC about how to expand the existing recurring highstate framework with new types of recurring actions


accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,193 @@
...
+In the end I don't have a strong opinion which approach to take.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+## SSM
+
+There are two ways recurring actions could work with SSM. However both come with significant downsides.
+
+1) Creating a recurrent action via SSM will schedule a `MinonRecurringAction` for every selected minion. This would schedule one `Taskomatic` job for every minion that would be scheduled to run at the same time.
```
admd, line 174: The primary advantage of SUSE Manager is its ability to allow users to perform tasks on multiple systems simultaneously. It is rare for users to perform actions on each system individually one by one.

But your concern is valid. If we can achieve the same result with a system group, then we can omit offering it as part of SSM.

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,193 @@
...
+
+On the system level the `Recurring States` tab is currently displayed as a sub to of the `States` tab. It will be moved to the top level tabs and renamed to `Recurring Actions` since it's no longer going to be state specific.
+
+For action type specific frontend implementations see below.
+
+## Specific implementations
+
+### Custom States (Config Channels)
+
+The idea here is to schedule a `Apply states[custom]` action on a recurring basis. The states executed depend on what is configured in `States` -> `Configuration Channels` on System / Group / Org level.
```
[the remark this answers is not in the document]
mcalmer: I would suggest to allow to add items as "free text".
We have the list of config channels which you can select to add the "string" to a list.
Additionally you can add entries using "free text".
All together is put in a listbox which can be sorted using "arrow buttons" moving items up and down.

accepted/00097-recurring-actions.md
mcalmer: Remember: SSM is a temporary "group" and specific to the loged-in user.
What would work is:
SSM => create a group from current SSM selection => define a recuring action on that group. Right?
Recurring requires a permanent collection of systems. SSM is not permanent.

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,193 @@
...
+
+The `RecurringActionApplyJob` `Taskomatic` job would then call the functions to build/promote a CLM project on recurrent basis.
+
+On the frontend side I'd suggest to add a new main menu item called e.g. `Recurring Lifecycle Management` in `Content Lifecycle` (below `Filters`).
+
+Here a list with existing recurring CLM build/promote schedules would be displayed as well as a button to prompt to create new schedules.
+
+The create page would add new input fields to select the project, the type (build/promote) and the environment to promote (if chosen).
+
+### Action Chains
```
rjmateus, line 112: I'm not sure about this one, and it would need further investigation.
If we allow the user to define a list of stalt states to run, and also their order, that can be seen more or less as an action chain.

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,193 @@
...
+
+The `suseRecurringActionType` table would then only need to include an id referencing the openSCAP profile entity.
+
+The rest of the implementation would work similar to other action types.
+
+### Ansible Playbook execution
+
+TODO
+
+## Traditional Clients
```
rjmateus, line 145: out of scope since traditional clients support is deprecated.

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,193 @@
...
+
+I see two possibilities two implement new functions:
+
+1) For each type implement individiual CRUD function that states the action type in its name. E.g. `RecurringActionHandler.createCustomState(...)`. This is essentially just a wrapper that calls the generic `create` function with the action type as a parameter.
+
+2) Have one CRUD function for all the types and pass the action type as an argument.
+
+Option 1) could be more intuitive to use and its function would be clear just by reading its name, but add a lot of extra API endpoints that all call the same underlying base function, so option 2) would be less overhead.
+
+In the end I don't have a strong opinion which approach to take.
```
rjmateus, line 165: I would say that we can follow the approach that would be simpler and less error prom for users, which looks like option 1. But I also don't have a strong opinion on this

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,193 @@
...
+Needless to say this would set us up for disaster and overwhelm `Taskomatic` if done with more than a few minions.
+
+2) Create a new `SSMRecurringAction` entity that would remember the Minions selected for the schedule by creating a db table e.g. called `RecurringSSMMinions` with a one-to-many relationship with the SSM entity. This way entries would be updated properly if a system is deleted (given it's configured properly with a foreign key and delete cascade to the minion).
+This way we could have one `Taskomatic` job that would perform the action for all the minions it was created with.
+
+Each SSM action could be displayed in the minions recurring action list. Updating/Editing the action on minion level would either be forbidden or update it for all the minions it was created for. 
+
+However adding/removing minions from an existing SSM schedule would require implementation of dedicated UI to support.
+
+In the end I'd advise against supporting recurring actions with SSM. We would map something that's designed to be not persistent (SSM selection) to something that is (Recurring Action schedule).
```
rjmateus, line 184: agree. User should/can create a system group to apply the recurrent actions.

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,193 @@
...
+
+In the end I'd advise against supporting recurring actions with SSM. We would map something that's designed to be not persistent (SSM selection) to something that is (Recurring Action schedule). 
+
+# Alternatives
+[alternatives]: #alternatives
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+**Question:** With the ability to schedule workflows on recurrent basis, the load on Taskomatic is expected to increase. Especially if there is a loot of schedules with frequent recurrance. Some actions the customers used to schedule manually are going to be perfomed automatically and thus most likely more often.  Is taskomatic up for the task?
```
rjmateus, line 192: this needs to be tested. Specially the ones that will bring load to salt-master and the CLM promote/build, which are heavy.

rjmateus (commented):
overall looks good, just a few topics that need to be discussed.
Thanks for the great contribution.

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Backend
+
+See [here](https://github.com/uyuni-project/uyuni/pull/6678/commits/fc365ebce46a9c59e540acf9e12473e504d1ccd8) for an outline on how the use of a specific entity created above would look like for the existing recurring highstate.
+Recurring action type entities would be created/managed based on the action type received from the frontend.
+
+The workflow to add new recurring action types would look like:
+1) Think about which action specific parameters need to be stored to schedule the action from the `RecurringActionApplyJob` `Taskomatic` job.
+2) Add a new element to the `ActionType` enum in `RecurringActionType`.
+3) Add new column(s) to the `suseRecurringActionType` table that store the parameters from  1).
```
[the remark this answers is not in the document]
parlt91: I thought about saving the type specific data as json initially, but not being able to use foreign keys deterred me. But now that you mention it I think we can still use the `action_type` as `discriminator column` to determine the type and just (de)serialize the json data based on the type.
This would give us more flexibility. We'd need to run another query to fetch the type related object from the id that we store in the taskomatic job, but I think this is acceptable.

I'll  play around with it a bit and check if it works the way intend it to and update the RFC (with this approach as a 2nd (and likely preferred) option)

accepted/00097-recurring-actions.md
parlt91: As MC explained it to me the justification for action chains is that salt states don't work with a reboot in between. Also I guess there are customers that want to strictly stick to the web UI for their workflows. For them it would be nice if they'd be able to 'click' their own reusable/automated workflow together on the UI.

But I agree this needs further investigation

accepted/00097-recurring-actions.md
parlt91: Okay, I think we have an agreement here. Customers can just create a system group from their SSM selection and use it to create the recurring action. No need to go through the trouble to make it work with SSM

accepted/00097-recurring-actions.md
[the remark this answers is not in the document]
parlt91: @rjmateus I just realized the main reason I didn't consider json is what would happen to the recurring schedule if the object it references is deleted.
Let's say we store the `project_id` of a CLM project as json in the `suseRecurringActionType` table and that project gets deleted, then the recurring schedule will still be present.
Is there an elegant way to handle this?

Other then trying to query the project in the taskomatic job and then assume the project no longer exists if we don't find it and consequently delete the schedule.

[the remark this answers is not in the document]
rjmateus: I would say it can fail with a comprehensive error message and then can be up to the customer to delete the task (or we can consider delete it automatically)
Regarding the json, you can keep use the `action_type` as `discriminator column` and use the subclasses approach with a "getdata" method method that would return the de-serialized data specific for each type.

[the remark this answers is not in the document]
parlt91: As an alternative solution:
What do you think about handling it with a trigger?

It could look like
```
CREATE OR REPLACE FUNCTION suse_contentproject_del_trig() RETURNS TRIGGER AS $$
DECLARE project_id int;
BEGIN
  project_id = OLD.id;
  DELETE FROM suseRecurringActionType
     WHERE (SELECT (type_data->>'project_id')::int FROM suseRecurringActionType) = project_id;

  RETURN NULL;
END;
$$ LANGUAGE plpgsql;

CREATE OR REPLACE TRIGGER suse_contentproject_del_trig
AFTER DELETE ON suseContentProject
FOR EACH ROW EXECUTE PROCEDURE suse_contentproject_del_trig();

```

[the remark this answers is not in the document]
rjmateus: That can also work. In the trigger, you should also look for the action type.
Another option is to update the delete project java code to also delete from this table.
btw, I really hate database triggers and functions. They are harder to debug and hidden inside the database. But hey, they are efficient when done properly :)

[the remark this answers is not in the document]
parlt91: I updated the rfc with your suggestions

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,193 @@
...
+
+For action type specific frontend implementations see below.
+
+## Specific implementations
+
+### Custom States (Config Channels)
+
+The idea here is to schedule a `Apply states[custom]` action on a recurring basis. The states executed depend on what is configured in `States` -> `Configuration Channels` on System / Group / Org level.
+
+Since which states to run is preconfigured for each entity type we don't actually need a new parameter in `suseRecurringActionType` table.
```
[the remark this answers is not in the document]
parlt91: Section has been updated to hopefully be clearer

accepted/00097-recurring-actions.md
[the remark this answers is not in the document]
cbbayburt: Jumping in with some counter-arguments here :) sorry in advance for the long comment

We will have many types of actions, but in the end, it's gonna be a fixed number of types in our control. Considering its mentioned drawbacks, I don't think we need that much dynamism that json offers.

Yes, every type has different fields, but we know of all of them in advance. So in this case, maybe a structured way would be better.

IIUC, a recurring action record would store the same data as a single action of its kind (plus a fixed amount of data for the recurring functionality) so maybe we could actually reuse the Action tables themselves, so it would work like this:

- We add a flag column in the base Action table: `recurring (Y/N)`
- We create a view (rhnActionView) with the following query for all the existing functionality around actions: `SELECT FROM rhnAction WHERE recurring = 0` (we need to plug the existing action table mappings to this view instead).
- We create another view (rhnRecurringAction) with the following query for retrieving any recurring actions: `SELECT FROM rhnAction WHERE recurring = 1`

So the `rhnActionView` would be our new Action table and `rhnRecurringAction` would be our Recurring Action table, with the same data structure.

The main reason I'm being skeptical is that I think using triggers/procedures is somewhat problematic mainly because it's impossible to unit test them, and they're very hard to debug, maintain and refactor.

cbbayburt (commented):
Thanks, @parlt91! This already draws a very clear picture for moving forward.

One thing to be clarified for me is how we assign recurring tasks to system groups.

We will have the functionality to assign recurring tasks to system groups, but are they going to be assigned to the individual systems by the time the task is assigned? Or is the task going to be assigned to the system group itself?

The main difference is when the group memberships change in the future. For example, when the user removes a system from a group, do the recurring tasks stay with the system?

IMO, just like the configuration channels implementation where a channel can be either assigned to a system, to a group, or to the whole organization, we should be able to assign recurring tasks to a single system or to a system group itself (maybe even also org-wide). This way, if you add systems to a group when the group already has recurring tasks, the existing tasks will be executed on the newly assigned systems as well.

parlt91:
> One thing to be clarified for me is how we assign recurring tasks to system groups.
> 
> We will have the functionality to assign recurring tasks to system groups, but are they going to be assigned to the individual systems by the time the task is assigned? Or is the task going to be assigned to the system group itself?

The task is assigned to the entity (System / Group / Org) so if you e.g. remove a system from a group it no longer performs the task that was assigned to the group.

We have the `RecurringAction.computeMinions()` function that computes which minions to schedule the action for in the `RecurringStateApplyJob` taskomatic job.

accepted/00097-recurring-actions.md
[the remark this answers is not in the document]
rjmateus: @cbbayburt I get your point, and I don't like triggers either.

I would prefer @parlt91 initial approach and have separated tables: keep the existing `rhnAction` and re-use the table `rhnRecurringAction` for recurrent actions.
They are not exactly the same and if they start differently the expectation is that they will differ even more in the future.

We would only change the `rhnRecurringAction` which defines the target (minions, groups or org), and then add a new table with the target action (like Pascal proposed). In that table, since we know the type of data we can use, the initial approach of having separated columns would be fine.

My problem with each column per data is the lack of flexibility, and we know special requests will arrive for sure.

cbbayburt:
> @cbbayburt The task is assigned to the entity (System / Group / Org) so if you e.g. remove a system from a group it no longer performs the task that was assigned to the group.
> 
> We have the `RecurringAction.computeMinions()` function that computes which minions to schedule the action for in the `RecurringStateApplyJob` taskomatic job.

Ok, that's what I was hoping for. Thanks :)

rjmateus:
I prefer Pascal initial proposal of having separated tables and columns for data. I don't like triggers and would avoid them as much as possible. They are hard to test and to have visibility.

accepted/00097-recurring-actions.md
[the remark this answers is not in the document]
parlt91: IIUC you propose to go back to the initial approach of using columns for the data instead of using json?
I think we don't need to rely on triggers and can make sure the recurring schedules get deleted properly in the java code.
E.g. explicitly look for recurring CLM `build/promote` schedules when deleting a CLM `project` and delete it if applicable.

Not every type of recurring action would have another entity it relates to (the highstate, custom state don't).
So I don't think it will be to hard to implement and maintain this.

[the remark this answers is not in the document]
rjmateus: What I would like to avoid is the trigger.
I think is safe to move forward with what you propose in this last comment.
Json column with the needed data. If one deletes the entity we adapt the java code to look for the recurrent action and delete also from there.

admd:
@parlt91 We also need to make a few improvements that may not be directly related to this RFC design. However, I would like to mention them here to ensure that these usability aspects are taken into consideration.


-   It is currently unclear which recurring tasks are associated with which group/minion/org. To address this issue, it would be helpful to provide a link to the specific entity (minion/org/group) so that users can easily navigate to it and determine the ownership of the recurring tasks.
Screenshot 2023-03-10 at 19 15 13


- At the minion level, it is necessary to display the recurring tasks that have been inherited from the groups and organization level. As shown in the screenshot, this particular minion has three distinct recurring states, with one specifically assigned to it and the other two belonging to the group and organization. All of them should be displayed here. While there is no need to allow editing or deleting of inherited recurring states from this page, a link should be provided to the group or organization where the original recurring task belongs.

Screenshot 2023-03-10 at 19 15 41

parlt91:
@admd

1) Is already mentioned in the RFC as an improvement we need to make
2) I can look into. I imagine it won't be that difficult to add as well. I'll add a section in the RFC about it.

Thanks for bringing it up

accepted/00097-recurring-actions.md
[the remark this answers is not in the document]
parlt91: I've added a section with an alternative solution for the db using the `rhnAction` table as @cbbayburt proposed, including a POC that would make it work with CLM. Just so we have an option to consider.
Having implemented both I don't have a strong preference towards either.

accepted/00097-recurring-actions.md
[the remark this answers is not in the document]
renner: (I missed that discussion and cannot find "free text" in the RFC, so maybe my comment is obsolete)

I am not sure if I like the idea of the "free text" option. How should customers know what to write, wouldn't it be better to at least scan for available states and list their names for selection?

Further I doubt that all (or even any) of the internally used states should be available for selection. IIRC they often serve a very specific purpose that is not really documented anywhere as they are basically implementation details, and often require special pillar data as input, which means that it would probably not even work to just call them. I might suggest that only a very limited list of internal states that make sense in this context can be selected, like e.g. channels, if at all.

accepted/00097-recurring-actions.md
mcalmer: I agree, but please write down that the alternative is to use a group. Using a group instead is missing here.

accepted/00097-recurring-actions.md
mcalmer: Right. I think Action Chains should be the "last" to implemented and it require a re-design of Action Chains.
I would propose to only think about 5.0 or later where "traditional management" is removed.
Then we can think of removing the UI parts of action chains from the single pages like package install/update/delete, etc.
We need to generate states from UI and we would need new UIs for it.
This is a new and independent RFC.

rjmateus: I agree with that. But in that case, we need to be careful with what we implement now in recurrent actions.
If we allow users to define a list of states to be applied and in which order, that will be similar to action chains, with the exception of the restart. This means we should not allow the restart to be added or it can fail to execute.
This lead to other another question: should we allow a list of states (and its order) to be defined or just one single state?

If we allow the list of states, that will be overlapping with action chains and make the implementation hard for this first stage.

We could say that in the phase we only support one single state (can be a configuration channel, some of the internal suma states, or a free text) and then when we have recurrent action channel, the user could define the list of states/actions to be applied.

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,321 @@
...
+# Detailed design
+[design]: #detailed-design
+
+## General design
+
+The existing recurring highstate framework already provides us with the bulk part of what be need for a generalized implementation. Some changes to the `RecurringAction` entity are required to make it work with multiple recurring action types. See details below.
+
+### Database
+
+**Solution #1:**
```
mcalmer, line 31: Is it already clarified which "Solution" you want to use? If yes, it is time to remove the other one which is out of the game. A final RFC should only describe the future implementation and provide no room for interpretations (if possible)

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,321 @@
...
+
+The preferred solution in this case is option 1).
+
+See ERM model below:
+
+![Recurring Action Type table](images/00097-recurring-actions-erm.png)
+
+See [here](https://github.com/uyuni-project/uyuni/pull/6678/commits/ba13bb918c0220a9455ddf1cf2d4373c12742d3b) for a POC of the proposed entity changes.
+
+**Solution #2:**
```
cbbayburt, line 73: We should specify that this was the chosen solution.

accepted/00097-recurring-actions.md
```diff
@@ -0,0 +1,321 @@
...
+See [here](https://github.com/parlt91/uyuni/commit/ac608e7865b0d6e6eee0ff495cc886664489c86d) for a POC of the `ContentManagementAction`.
+
+And [here](https://github.com/parlt91/uyuni/commit/898103d38e194fc1dcf4b72877d566e49f3c799c) for a POC of a working recurring CLM project build backend implementation.
+
+### Backend
+
+See [here](https://github.com/uyuni-project/uyuni/pull/6678/commits/61cadb982fcbdd46546ff219f658aaefa21c1f4b) for an outline on how the use of a specific entity created above would look like for the existing recurring highstate.
+Recurring action type entities would be created/managed based on the action type received from the frontend.
+
+The workflow to add new recurring action types would look like:
```
cbbayburt, line 95: IIUC this workflow is based on Solution 1. Maybe we can rewrite it for solution 2.

cbbayburt (changes requested):
@parlt91 I had another look into the document just now. There are of course parts that haven't been implemented (yet), but in general, it looks in line with what was actually implemented, so I don't think any major changes are necessary.

I just made notes below for some minor updates. After that, I think we can merge it finally.

accepted/00097-recurring-actions.md
parlt91: Done

cbbayburt (approved):
LGTM :+1: Thanks for the update.

PR #78, opened 2023-03-22
Automatic migration to Salt Bundle

ycedres:
Force an automatic migration for Salt clients from Salt 3000 to Salt Bundle.

### DIFF
--- a/accepted/00096-automatic-migration-salt-bundle.md
+++ b/accepted/00096-automatic-migration-salt-bundle.md
@@ -1,0 +1,81 @@
+- Feature Name: Automatic migration from Salt 3000 to Salt Bundle
+- Start Date: 2023-03-22
+
+# Summary
+[summary]: #summary
+
+Force an automatic migration for Salt clients from Salt 3000 to Salt Bundle.
+
+
+# Motivation
+[motivation]: #motivation
+
+
+Salt 3000 is EOL as of Aug 31, 2022. It is advisable for Salt clients using Salt 3000 to migrate to Salt Bundle. Right now this migration can be done by applying the already existent `util.mgr_switch_to_venv_minion` state from the CLI. The purpose of this RFC is to automatically perform this migration for Salt 3000 clients to Salt Bundle so they can keep receiving updates for the Salt package, including security updates.
+
+NOTE: This RFC excludes SSH minions.
+
+
+# Detailed design
+[design]: #detailed-design
+
+An state to perform the migration from Salt 3000 to Salt Bundle is added to the highstate. This state checks the 'saltversion' grain and others in order to determine if the targeted minion needs to be migrated to Salt Bundle.
+
+The migration will be triggered by the state if the following conditions are met:
+- The running minion is not Salt Bundle.
+- The running minion is not an SSH minion.
+- The running minion is Salt 3000.
+- The `mgr_avoid_venv_salt_minion` pillar is not `True` for this minion.
+
+If so, the migration is performed in the following steps:
+
+- Install the `venv-salt-minion` package.
+- Copy the minion id file, minion configuration files, PKI files and static grains from the existing `salt-minion` in `/etc/salt` to the `venv-salt-minion` corresponding directories in `/etc/venv-salt-minion`.
+- Enable the `venv-salt-minion` service.
+- Disable the `salt-minion` service.
+- Remove `/etc/salt/minion.d/susemanager.conf` file, to prevent possible accidental reconnection of classic "salt-minion" after migration.
+
+The whole process is transparent to the user.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- The `salt-minion` package remains installed and the configuration files in the `/etc/salt` are not deleted.
+
+- After migration this check is done in every highstate.
+
+- If a client was bootstrapped avoiding the Salt Bundle following the steps in the documentation, the migration would force this client to use Salt Bundle. The options to avoid installing Salt Bundle and keep using `salt-minion` that currently appear in the documentation are:
+
+  * Execute mgr-bootstrap with --no-bundle option.
+
+  * Set AVOID_VENV_SALT_MINION to 1 in the generated bootstrap script.
+
+  * For bootstrap state set the `mgr_avoid_venv_salt_minion` pillar to `True`.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Force the removal of the `salt-minion` package and related files. This could be done by means of a custom grain module that informs if the minion is running  `venv-salt-minion`. With that in place, the application of the highstate a second time could handle that second stage for purging.
+- The impact of not doing this is that Salt 3000 clients will stop receiving updates for the Salt package.
+
+
+# Resolved questions
+[resolved]: #resolved-questions
+
+1. What is a better solution?
+  - The sls contains the logic for triggering the migration or not.
+  - There is logic for triggering the migration that decides if adding or not the sls to the highstate.
+  
+The option chosen is adding the logic to the state as this stays in line with our usage of SLS files.
+
+2. Could this automatic migration behaviour be opted out? It seems it would be reasonable in case bootstrapping with the Salt Bundle was explicitly avoided.
+
+The automatic migration can be opted out by setting the pillar `mgr_avoid_venv_salt_minion` to `True` for the specific minion.
+
+3. Is it best to take a cleaner approach and remove the `salt-minion` package and related files or to take a more conservative approach and do not do it in case there are unexpected scenarios?
+
+Neither the `salt-minion` package nor the configuration files in `/etc/salt` are going to be removed in case the user had configured another `salt-minion` between stages.
+
+However, the `/etc/salt/minion.d/susemanager.conf` file is going to be removed to prevent an accidental reconnection.
+


accepted/0000-automatic-migration-salt-bundle.md
```diff
@@ -0,0 +1,62 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+An state to perform the migration from Salt 3000 to Salt Bundle is added to the highstate. This state checks that the 'saltversion' grain in the targeted minion and that the minion is not already running Salt Bundle. In that case, the migration is performed in the following steps:
+
+- Install the `venv-salt-minion` package.
+- Copy the minion id file, minion configuration files, PKI files and static grains for the current `salt-minion` in `/etc/salt` to the `venv-salt-minion` corresponding directories in `/etc/venv-salt-minion`.
+- Enable the `venv-salt-minion` service.
+- Disable the `salt-minion` service.
```
admd, line 25: What was the reason behind not removing it and just disable it? What could be the possible problems if we completely remove it?

If we don't see any major problem, ideally I would suggest removing it, just to remove the possibility that user don't go and start salt-minion service manually. As user will still be used to older way to starting/stopping minion so this could be a pretty common scenario.

Besides that, If we keep this state as part of high-state even after the migration is done, we need to make sure that there is no way we could mess a different salt-minion instance connected with a different salt master.

ycedres: Thank you Abid. This is precisely one of the points that was still subject to discussion, that's why I wrote it as one of the drawbacks. 

I was not really sure if there was any multi-master use case that would require to keep specifically the existing `salt-minion` running. In that case, `venv-salt-minion` would need a different minion id. If this is unlikely, removing it would be a better idea, as you say.

Right now I don't see the way that having this as part of the highstate could be a problem if a different master is present. However, we could only include the state as part of the highstate only if we want to trigger the migration and avoid any unexpected behaviour.

accepted/0000-automatic-migration-salt-bundle.md
```diff
@@ -0,0 +1,62 @@
...
+Salt 3000 is EOL as of Aug 31, 2022. It is advisable for Salt clients using Salt 3000 to migrate to Salt Bundle. Right now this migration can be done by applying the already existent `util.mgr_switch_to_venv_minion` state from the CLI. The puprose of this RFC is to automatically perform this migration for Salt 3000 clients to Salt Bundle so they can keep receiving updates for the Salt package, including security updates.
+
+
+# Detailed design
+[design]: #detailed-design
+
+An state to perform the migration from Salt 3000 to Salt Bundle is added to the highstate. This state checks that the 'saltversion' grain in the targeted minion and that the minion is not already running Salt Bundle. In that case, the migration is performed in the following steps:
+
+- Install the `venv-salt-minion` package.
+- Copy the minion id file, minion configuration files, PKI files and static grains for the current `salt-minion` in `/etc/salt` to the `venv-salt-minion` corresponding directories in `/etc/venv-salt-minion`.
```
rjmateus, line 23:
```suggestion
- Copy the minion id file, minion configuration files, PKI files and static grains from the existing `salt-minion` in `/etc/salt` to the `venv-salt-minion` corresponding directories in `/etc/venv-salt-minion`.
```

accepted/0000-automatic-migration-salt-bundle.md
rjmateus: I agree with Abid on this one. But we can put some information in the release notes if the user has multi-master configured in the salt-minion, this will be a breaking change. We should be defined what should happen in that case. But I think documentation only should be available since we shouldn't have many cases.

accepted/0000-automatic-migration-salt-bundle.md
```diff
@@ -0,0 +1,62 @@
...
+
+The whole process is transparent to the user.
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- The `salt-minion` package remains installed and the configuration files in the `/etc/salt` are not deleted.
+
+- After migration this check is done in every highstate.
```
rjmateus, line 35: We may need to add another check to this and see if suse manager salt master or proxy is configured in the salt-minion. 
If a user decides to start using multi-master we will be removing they configurations.

accepted/0000-automatic-migration-salt-bundle.md
```diff
@@ -0,0 +1,62 @@
...
+
+- If a client was bootstrapped avoiding the Salt Bundle following the steps in the documentation, the migration would force this client to use Salt Bundle. Those steps are:
```
rjmateus, line 37: Long term, we should only support bundle, and all this options should go away.

accepted/0000-automatic-migration-salt-bundle.md
ycedres: Agreed. I think this checks are already done in the `util.mgr_switch_to_venv_minion` state.

accepted/0000-automatic-migration-salt-bundle.md
agraul: At the time of the migration from salt-minion to venv-salt-minion, the customer can't have salt-minion configured to a different master. The problem that Yeray ran into was that migration+removal of salt-minion in a single step is technically not possible. Logically, it would be safe to do. 

As a result, the salt-minion needs to be removed at some time in the future, after the migration (i.e. the agent doing the removal has to be venv-salt-minion). And that's where things could be tricky, because this is when a customer could have configured salt-minion for a different Salt Master after the migration has finished. I don't expect this to happen often, but we should acknowledge this technical possibility.

All that said, the current state checks that the specified `master:` in the minion's config file is "uyuni" (looks at the pillar data for it). So, afaik, we should be safe to do the removal.

Last, but not least, let's not call this situation "multi-master". That's term is already used for a [redundant master setup in the Salt world](https://docs.saltproject.io/en/latest/topics/tutorials/multimaster.html).

accepted/0000-automatic-migration-salt-bundle.md
```diff
@@ -0,0 +1,62 @@
...
+- Force the removal of the `salt-minion` package and related files. This could be done by means of a custom grain module that informs if the minion is running  `venv-salt-minion`. With that in place, the application of the highstate a second time could handle that second stage for purging.
+- The impact of not doing this is that Salt 3000 clients will stop receiving updates for the Salt package.
+
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+
+- What is a better solution:
+  * Adding the sls to the highstate. The sls contains the logic for triggering the migration.
```
agraul, line 57: This would be the way that Salt is expected to be used. I know we don't really follow this approach, we just apply states and let them decide if there is something to do. The other option would be in line with our historic use of Salt states, even if it's not the idiomatic way.

accepted/0000-automatic-migration-salt-bundle.md
rjmateus: understood. Thanks for the great explanation Alex

admd: Thank you guys, this was helpful.

> As a result, the salt-minion needs to be removed at some time in the future.

@agraul  and when would be that time in the future?

agraul: > Thank you guys, this was helpful.
> 
> > As a result, the salt-minion needs to be removed at some time in the future.
> 
> @agraul and when would be that time in the future?

Once the venv-salt-minion is used. The current migration start stops the salt-minion.service afaiu, the next highstate should already hit venv-salt-minion.service. I think we currently have no way to detect on the master side if a minion is using salt-minion or venv-salt-minion, so we can't use that information to trigger the removal.

accepted/0000-automatic-migration-salt-bundle.md
ycedres: Should this avoid-venv-salt-minion  feature be removed in the context of this RFC or it is something do in the future?

accepted/0000-automatic-migration-salt-bundle.md
meaksh: As mentioned, removing the old "salt-minion" service at the time of the migration is not really possible, as we cannot remove the old "salt" while it is actually running the migration state. This is problematic, and therefore the current manual migration state that Victor implemented is done in two steps, as our documentation already describes: one for migration and the other one for "purging" the old "salt-minion": https://documentation.suse.com/suma/4.3/en/suse-manager/client-configuration/contact-methods-saltbundle.html#_switching_from_salt_minion_to_salt_bundle

That being said, in the context of triggering this migration as part of the highstate, we are un the same situation, we cannot do this migration/removal in a single shot.

Then the question, should we trigger also the "removal" step also automatically as part of highstate, when somehow we detect that "venv-salt-minion" is running? Well.. we need to be careful here, as this could be problematic: what if a customer already configured the old "salt-minion" to use another Salt solution, we need to take care of really double checking that old "salt-minion" is not really or configured against any other Salt master before removing... I would maybe suggest that "removal" part, is NOT done automatically via highstate, but I can still be triggered manually as of today (triggering the 2nd state manually as documentation already described if customer wants)

accepted/0000-automatic-migration-salt-bundle.md
meaksh: Am I missing something or was your comment actually targetting the other solution?

I'm don't a really have strong opinion here bu I would probably prefer to keep same approach than we currently use in SUMA (the logic is inside the SLS itself and we always add it to the highstate via our custom `mgr_master_tops` module).

accepted/0000-automatic-migration-salt-bundle.md
rjmateus: Pablo statement looks good to me.

accepted/0000-automatic-migration-salt-bundle.md
admd: I get that. My only concern was around old salt-minion trying to connect again with the SUSE Manager while that machine is already being managed with the salt bundle. Would it be feasible during the migration to change or simply remove the /etc/salt/minion.de/susemanager.conf  so old minion at least doesn't point to SUSE Manager anymore unless manually done by the user?

accepted/0000-automatic-migration-salt-bundle.md
ycedres: I think the sls containing the logic is a cleaner solution, when reading the sls file you get the whole picture and I believe is what Alex means when he says that it's hos Salt is expected to be used.

accepted/0000-automatic-migration-salt-bundle.md
ycedres: I guess the only option that would be a problem is if the same minion is registered to a different Uyuni/SUMA server. This is not very common, is it? In that case I'd say is safe to remove `/etc/salt/minion.d/susemanager.conf.

admd: Then I would vote for removing that file. Nothing further from my side.

admd (approved):

meaksh:
I've added some commits here to reflect some of the topics discussed during the review of this RFC.

I'm requesting review so we can get final approval and proceed with merging. Thanks!

accepted/0000-automatic-migration-salt-bundle.md
```diff
@@ -0,0 +1,72 @@
...
+  * For bootstrap state set the mgr_avoid_venv_salt_minion pillar to True.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Force the removal of the `salt-minion` package and related files. This could be done by means of a custom grain module that informs if the minion is running  `venv-salt-minion`. With that in place, the application of the highstate a second time could handle that second stage for purging.
+- The impact of not doing this is that Salt 3000 clients will stop receiving updates for the Salt package.
+
+
+# Unresolved questions
```
rjmateus, line 63: Are these unresolved questions still valid? It sounds strange to merge an RFC without these questions being answered.

rjmateus (commented):
Just one question to review

accepted/0000-automatic-migration-salt-bundle.md
ycedres: I think they all have been addressed during the discussion and now this questions are resolved, so it does not make sense to keep them.

accepted/0000-automatic-migration-salt-bundle.md
agraul: I meant that the idiomatic way would be to have a state that describes the end situation, and then have this state applied to those systems that are targeted in top.sls. 

I'm okay with having the logic in the state to stay in line with our usage of SLS files.

accepted/0000-automatic-migration-salt-bundle.md
agraul: It would be great if the outcomes of the discussion made their way into the final document. I agree with removing them as unresolved questions, let's just make sure we don't "ignore" their outcomes.

rjmateus: Or you can keep it, put them as resolved questions, and add the outcome for each of them, for completeness.

rjmateus (approved):

agraul (approved):

accepted/0000-automatic-migration-salt-bundle.md
```diff
@@ -0,0 +1,81 @@
...
+  
+The option chosen is adding the logic to the state as this stays in line with our usage of SLS files.
+
+2. Could this automatic migration behaviour be opted out? It seems it would be reasonable in case bootstrapping with the Salt Bundle was explicitly avoided.
+
+The automatic migration can be opted out by setting the pillar `mgr_avoid_venv_salt_minion` to `True` for the specific minion.
+
+3. Is it best to take a cleaner approach and remove the `salt-minion` package and related files or to take a more conservative approach and do not do it in case there are unexpected scenarios?
+
+The approach that is going to be followed is removing the `salt-minion` package in the second stage in order to avoid the user from startig the `salt-minion` service manually. Configuration files in `/etc/salt` are not going to be removed in case the user had configured another `salt-minion` between stages.
```
meaksh, line 78: Is this sentence correct? My understanding was that we are NOT removing the "salt-minion" package as part of this migration.

meaksh (approved):

PR #80, opened 2023-06-13
CVE Auditing with OVAL

HoussemNasri:
Use OVAL data in addition to channels to conduct more accurate CVE audits.

See the [rendered version](https://github.com/HoussemNasri/uyuni-rfc/blob/oval/accepted/0000-cve-auditing-with-oval.md).


msmeissn:
Abid asked me to review.

to me it looks good.

Note that we also express "not affectedness" via a version = 0 relation, but this should not matter. (And also have a livepatching logic trick you will probably see in SUSE and Red Hat OVAL.)

Only big concern from my side is the question of scale. 
With "openscap" we see the regular SUSE OVAL file taking multiple GIGABYTE of memory to process.

I think with intelligent parsing and data storage it might work.

rjmateus (commented):
Hey Houssem Nasri, Thank you for the great RFC.
It looks good but I have some notes that can make it more focused and informative.

In the UYUNI Backend is mentioning 3 modules, that I think are to detail in some parts but doesn't cover all the steps we need to archive the CVE analyzes goal (more at the end of this comment).
Also we should just put the API call, not specifying if is JSON or XMLRPC, since both will be available.

I would not put all the methods for the DAO in this RFC. I would consider that as a implementation detail.

For the database I think we need to decompose the metadata into database tables to help in the needed features, like search for CVE number, of list CVE for platform. A database diagram would help.

I'm not sure if I fully agree with the new names for the "PatchStatus".
I would put something like:
  - Add the "AFFECTED_PATCH_UNAVAILABLE" as you made.
  - Keep "AFFECTED_PATCH_INAPPLICABLE", this mean a full or partial patch is available in a different channel. I think keeping the name will help existing users (but I don't have a strong opinion on it)
  - "AFFECTED_FULL_PATCH_APPLICABLE" similar, but name closer to what we have now
  - "AFFECTED_PARTIAL_PATCH_APPLICABLE", meaning the path is applicable but will not solve all the issues.

# Proposed Modules

## Oval data processing Module
Responsible for Download oval files, parse them and insert the structure data in the database base.
For this module we should see the proposed schema and how the fields will map.
The goal for it is to transform the data from XML to postgresql database to consumed by the other modules.

For OVAL data you can also have a look in the Neuvector information, and from where they are getting the oval files: https://open-docs.neuvector.com/scanning/cve_sources#cve-database-sources

We can then discuss if we should have a schedule to periodically update this data.

## System CVE analyzes
Run a CVE analyze for a system (or set of systems).
It will use the information we have for the system and the OVAL data inserted in the database. This will result in which CVE's are affecting the system, and the status of each CVE.
This result should be saved in the database to be consumed later. We expect that this to be an expensive task so we need to avoid run it to often and save the results in the database.

For this you can use the diagram you presented for the verification process and CVE status in the system.

I also would like to see some more details on how we plan to run the criteria scan on a system.

## API access
This would expose API methods (Json and XMLRPC) to access data and perform actions.
Some examples can be:
- Get System CVE
- Schedule CVE scan for system(s)
- Get CVE's for platform (OS)
- etc

parlt91 (commented):
Thank you Houssem for your effort so far. It's really looking good already. One thing I'd like a bit more detail on is the schema part. It's really crucial that we design it with performance in mind. Would you mind explaining a bit more why you designed it the way you did so we can understand a bit better what your thought process was when creating it.

HoussemNasri:
@parlt91 @rjmateus Thank you for the review. Below, is a list of tasks to be done on the RFC based on your comments, and comments from @admd  that were shared in meetings. Please let me know if it's missing something. 
### To be done
- [ ] Show only important methods in DAO
- [ ] Update PatchStatus values
- [ ] Decompose the RFC into higher-level modules
- [ ] Propose a schedule to periodically update OVAL data
- [ ] Add a CVE system analysis section
- [ ] Add an API access section
- [ ] Explain in more detail the evaluation of criteria trees
- [ ] Explain in more detail the database schema design decisions
- [ ] State performance metrics
- [ ] Explain the vulnerable package extraction process and why it's not feasible to cache the result to the database
- [ ] Add links to implementation in RFC
- [ ] Document the unselected alternative cve auditing algorithm that depends on vulnerability and patch definitions

accepted/0000-cve-auditing-with-oval.md
```diff
@@ -0,0 +1,188 @@
...
+  public static List<OVALPlatform> getPlatformsAffectedByCve(String cve) {}
+  public static Optional<OVALDefinition> lookupVulnerabilityDefinitionByCve(String cve) {}
+}
+```
+
+### Database Storage for OVAL Documents
+
+Once the OVAL file is parsed, the next step is to store it in a database. A relational database, such as PostgreSQL, seems like an obvious choice given that it is the default database used by Uyuni.
+
+In Uyuni, multiple database-access patterns are being used. Skimming over the codebase, it appears that the newer parts of the code is using JPA Criteria API and Annotation-based Named queries. For that reason, the same patterns will be used to query OVAL data.
```
mcalmer, line 64: About how many objects stored in DB are we talking? I assume this is a lot.
Hibernate is bad when it comes to big tables with a lot data and combining them.
We did that mistake already too often in the other tables.
So it maybe a good idea to use DTO objects and SQL instead of hibernate objects.

accepted/0000-cve-auditing-with-oval.md
```diff
@@ -0,0 +1,188 @@
...
+
+#### OVAL Extensions Support
+
+The primary objective of the OVAL parser is to parse openSUSE and other Linux-based distributions OVAL files. This requires that we add support to the Linux OVAL extension. We can take it a step further and offer an abstraction that can be used to implement other extensions. This should, for example, facilitate the integration with Android[^5] and Windows[^6] clients when they become supported in Uyuni.
+
+### OVAL Downloader
+
+The OVAL Downloader is responsible for finding OVAL documents online, downloading them, and caching them for easy access. However, since the relevant OVAL data is saved in PostgreSQL, it is possible for the downloader to skip caching the OVAL files or remove them once they are stored in the database. This approach would reduce the amount of storage needed.
+
+OVAL data comes from several sources – OvalRepo, OpenSUSE OVAL repository and other Linux distributions repositories – which means that the OVAL Downloader needs to be robust in the presence of errors. It is essential also that the component is designed with comprehensive error messaging to provide instant alerts in case an OVAL source is moved or the server is down.
```
mcalmer, line 39: People wants things going automatically. We need to run the downloader recurring at least once per day and update all the information in the database.
We also need a place/way to configure which OVAL to download. Best would be to attache this to the products.
When a customer choose a product to mirror, we automatically download the OVAL.

accepted/0000-cve-auditing-with-oval.md
```diff
@@ -0,0 +1,188 @@
...
+
+# Drawbacks
+
+# Alternatives
+
+# Unresolved questions
+
+- How to distinguish QA from released patches in non-openSUSE distributions?
+- Can we depend on OVALRepo to provide us with OVAL definitions?
+
```
mcalmer, line 174: I wonder if we get the querying of data fast enough. First for the existing case, searching for affected systems of 1 CVE number. 
A second case was requested since a long time already. Show all systems with all affected CVE numbers.
Currently this data set it to big to query and present to the user. Maybe this can be done with this style.

During implementation I would try as early as possible to try out the performance of the algorithms.

accepted/0000-cve-auditing-with-oval.md
HoussemNasri: The writing of the data below took around 10 minutes with Hibernate
- Definitions: 13660 (The writing of definitions will result in the writing of affected products, references, and mentioned cves)
- Objects: 9657
- States: 19036
- Tests: 67547

Weirdly enough, using native queries took even more. I think it's because of JPA's auto-flushing strategy which prevents running native queries in bulk. Right now, I'm out of ideas to optimize it further. I'll try not to rely on Hibernate anymore and use plain SQL as you suggested. You can look at https://github.com/HoussemNasri/uyuni/blob/oval-db/java/code/src/com/suse/oval/OVALCachingFactory.java for more details about implementation.

mcalmer: Tip: switching between hibernate and native queries is expensive. On every switch, hibernate need to verify the cache if a flush is required. Alone this check cost most of the time.
I think writing is only one aspect. It will happen during the night somewhen. If it takes some minutes more does not matter.
More imporant is reading. How long do the queries take to display data. This is where people "click" and wait for the result

HoussemNasri: Reading affected products from the database is one aspect of the algorithm and the fastest step I would say. There is also the evaluation of the criteria tree, the extraction of vulnerable packages from the criteria tree, and the reading of patch information from the database/channels. 

Right now, I have an early draft implementation of the algorithm running and it takes between 0.2 and 14 seconds to evaluate a single CVE. It's not optimized though. With some caching, we can get those numbers way below. Unfortunately, showing all systems with all affected CVE numbers might not be feasible (it would take too much time) even with optimization.

HoussemNasri: Update: It turns out that the whole process of auditing one system against one CVE takes less than a second, but it grows over time (probably Hibernate dirty checking) when checking multiple CVEs, hence why the 14 seconds.

HoussemNasri:
Greetings,

I've been exploring ways to support Red Hat OVALs, and I have some thoughts I'd like to share with you. It appears that enabling Red Hat support requires changes to the algorithm implementation, database schema, and test evaluator algorithm. Similar modifications were necessary to support Debian. However, I'm concerned that this pattern of scattered changes to deal with new products may render the design less robust. I've been thinking about the reasons behind this, and I believe I underestimated the structural differences between OVALs provided by various distributions.

For example, in Debian, each OVAL definition describes the vulnerability state of one package only. Consequently, if multiple packages are vulnerable to a CVE (which is very common), there will be multiple definitions technically describing the same CVE. On the other hand, SUSE OVALs map vulnerability definitions by CVE, with each definition describing a CVE along with its patched, unpatched, or not affected packages. Red Hat exhibits different characteristics, and while I haven't thoroughly investigated Ubuntu yet, I'm certain it has its own differences. Aside from the lack of robustness in the current solution, this approach also requires "smart" implementation tricks that may not be evident unless you wrote the code.

I propose an alternative approach: modifying the algorithm to exclusively operate on vulnerable packages (a similar approach to the channels algorithm currently used in Uyuni). This way, the algorithm won't be concerned with whether each OVAL definition describes one or multiple packages or references multiple CVEs, or whether it describes a patch or a vulnerability. It will operate at a more atomic level, with the structural differences hidden in the vulnerable package extractors, where each distribution can have its specific implementation. This approach offers several advantages:

- It's fast (the algorithm = 1 SQL query)
- Supporting more products is easy (we only need a new implementation of the vulnerable package extractor)
- Implementation is easier to reason about

However, this approach does come with some drawbacks. Notably, it creates a stronger dependency on the structure of the criteria tree. Extracting vulnerable packages from the criteria tree requires following specific patterns in the OVAL data. For instance, Red Hat OVAL definitions with an ID containing "unaffected" describe vulnerabilities for which a patch was backported to all distributions, so they can be ignored. Similarly, in SUSE OVALs, if a criterion has a comment matching "'package' is affected," then 'package' is unpatched. The concern here is that if the security team decides to change "is affected" to "is unpatched," the implementation would need to change. Keep in mind that even the current algorithm needs to extract vulnerable packages at some point so the concern applies there too but it is less dependent on it.

I'd greatly appreciate your thoughts and feedback on this approach.

 @mcalmer @rjmateus @parlt91

parlt91:
Hi Houssem,

see some of my thoughts below:

> Greetings,
> 
> I've been exploring ways to support Red Hat OVALs, and I have some thoughts I'd like to share with you. It appears that enabling Red Hat support requires changes to the algorithm implementation, database schema, and test evaluator algorithm. Similar modifications were necessary to support Debian. However, I'm concerned that this pattern of scattered changes to deal with new products may render the design less robust. I've been thinking about the reasons behind this, and I believe I underestimated the structural differences between OVALs provided by various distributions.
>
> For example, in Debian, each OVAL definition describes the vulnerability state of one package only. Consequently, if multiple packages are vulnerable to a CVE (which is very common), there will be multiple definitions technically describing the same CVE. On the other hand, SUSE OVALs map vulnerability definitions by CVE, with each definition describing a CVE along with its patched, unpatched, or not affected packages. Red Hat exhibits different characteristics, and while I haven't thoroughly investigated Ubuntu yet, I'm certain it has its own differences. Aside from the lack of robustness in the current solution, this approach also requires "smart" implementation tricks that may not be evident unless you wrote the code.
>
I agree with your concerns. Each distribution has it's own nuances in their OVALs. The more distributions we support the more  likely we'll run into issues. While it would be nice to have a one fits all kind of solution I am afraid we're not going to get that. 
>
> I propose an alternative approach: modifying the algorithm to exclusively operate on vulnerable packages extracted from the criteria tree (a similar approach to the channels algorithm currently used in Uyuni). This way, the algorithm won't be concerned with whether each OVAL definition describes one or multiple packages or references multiple CVEs, or whether it describes a patch or a vulnerability. It will operate at a more atomic level, with the structural differences hidden in the vulnerable package extractors, where each distribution can have its specific implementation. This approach offers several advantages:
> 
> * It's fast (the algorithm = 1 SQL query)
> * Supporting more products is easy (we only need a new implementation of the vulnerable package extractor)
> * Implementation is easier to reason about
>
I agree that using the most atomic level (packages) should give us the best result here. I am fine with this approach as long as our functionality to query data about a specific CVE on a per system basis remains intact. This means we'll still have to look at the channels and installed packages to decide which patch status we're going to display for a given CVE.
We should also make sure that our test cases for the `CVEAuditManager` remain intact to avoid running into regressions (with some adaptions for the new patch statuses of course). A lot of these tests target a specific bug/issue we ran into in the  past. So we should review them one by one and check if they're still relevant.
> 
> However, this approach does come with some drawbacks. Notably, it creates a stronger dependency on the structure of the criteria tree. Extracting vulnerable packages from the criteria tree requires following specific patterns in the OVAL data. For instance, Red Hat OVAL definitions with an ID containing "unaffected" describe vulnerabilities for which a patch was backported to all distributions, so they can be ignored. Similarly, in SUSE OVALs, if a criterion has a comment matching "'package' is affected," then 'package' is unpatched. The concern here is that if the security team decides to change "is affected" to "is unpatched," the implementation would need to change. Keep in mind that even the current algorithm needs to extract vulnerable packages at some point so the concern applies there too but it is less dependent on it.
> 
I wouldn't worry about the structure changing, that's not something we really have control over (except maybe in SUSEs case). What's important in my opinion is that we have test coverage that would make us aware of such changes.
>
> We can of course not rely on the comments in criterions and look up the associated rpm_test. The tests are more stable since they are part of the OVAL specification. However, this can slow down the extraction time, but only when syncing OVALS.
>
Unless the slowdown is significant, I think it's fine. If we run the extraction once per day during low load hours it shouldn't be an issue unless it takes multiple hours so sync. (On a side note: We should make sure that we have a configurable `taskomatic` schedule available for that sync task so users can decide when best to run it).
>
> This also seems like the more common approach to leveraging OVAL in vulnerability scanning (instead of criteria tree evaluation)
> 
> * https://github.com/quay/claircore/blob/33e39bc1d855b21ee477fa5136c31d6842341ec3/pkg/ovalutil/rpm.go#L43
> * https://github.com/vulsio/goval-dictionary/blob/38d8605fb186a601577319f8eac8ca5230c8bf81/models/suse/suse.go#L233
> 
> I'd greatly appreciate your thoughts and feedback on this approach.
> 
> @mcalmer @rjmateus @parlt91

cheers

HoussemNasri:
Hi @rjmateus, I'm glad we agree on this! 

Your assumptions are absolutely right. During synchronization, we extract the list of vulnerable packages from OVAL definitions and then store them in the database along with the corresponding CVE and vulnerable operating systems. Subsequently, during the auditing process, we can efficiently retrieve the necessary information by querying the database based on the operating system of the client being tested and the specific CVE. 

Storing vulnerable packages in the database is something I thought was impossible before because it would take too long, but recently I did some experiments by using stored procedures and batch processing and the results were acceptable (~35 seconds for a ~60 MB OVAL file) especially since the syncing would run once a day.

HoussemNasri:
Hello @msmeissn,

Currently, I'm working on matching the client's operating systems against the vulnerable OS as per OVAL data. So I was thinking of converting the user-friendly OS names found in comments (.e.g. SUSE Linux Enterprise Desktop 15 SP3 is installed), to a CPE. The simplest approach seems to involve referencing the corresponding OVAL test and converting the comment into a CPE, like this: "sled-release is ==15.3" --> `cpe:/o:suse:sled:15:sp3`. However, when dealing with OS versions like 15.3, there are two potential CPE formats: "...:15:sp3" and "...:15.3". To address this, I'm planning to create both variations and retain the one that matches one of the entries in `<affected_cpe_list>`.

I wanted to check with you to ensure the correctness of this solution and whether there might be any edge cases that I've overlooked.

parlt91 (approved):
From my side this looks fine now. I just left one remark (see below). Thank you for your effort so far :)

accepted/0000-cve-auditing-with-oval.md
```diff
@@ -0,0 +1,366 @@
...
+It's worth noting that the diagram does not explicitly depict the addition of a `cpe` column to the `rhnServer` table.
+This modification is necessary to enable the matching of vulnerable platforms from the `suseOVALPlatform` table
+against actual systems.
+
+Uyuni already contains tables to store package names and EVRs, they're called respectively `rhnPackageName`
+and `rhnPackageEVR`. One possible optimization would be to create foreign keys for `name` and `fix_version`
+inside `suseOVALVulnerablePacakge` instead of hard coding their values. While the optimization results in
+reduced redundancy and increased storage efficiency, it does introduce potential performance implications.
+An experiment should be made to decide whether the optimization is worth it or not.
```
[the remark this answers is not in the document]
HoussemNasri: No, not yet. Right now, it's not clear how I should go about testing the performance impact. I could run the CVE auditing query before and after the optimization and measure the impact in seconds, but I only have 4 clients registered. The impact might be exponentially bigger if 100s of clients were registered.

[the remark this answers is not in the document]
cbosdo: You can easily add thousands of systems by copying one of yours with this script: https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/performance/copy-system.py. To test the performance impact I think you would also need to have many channels and packages in your DB. Again a small script adding the entries in the DB would be enough... but we don't have that one yet 😉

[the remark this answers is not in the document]
HoussemNasri: Thanks, that seems like what I've been looking for.

accepted/0000-cve-auditing-with-oval.md
```diff
@@ -0,0 +1,366 @@
...
+
+Once the OVAL file is parsed, the next step is to store it in a database. A relational database, such as PostgreSQL,
+seems like an obvious choice given that it is the default database used by Uyuni.
+
+#### Database Schema
+
+The database employs a straightforward schema consisting of three new tables. Notably, there is no need to create a CVE
+table, as we will use the existing `rhnCVE` table integrated into Uyuni. Here's the purpose of each of these tables:
+
+1. **suseOVALVulnerablePackage**: This table represents a single vulnerable package along with its corresponding fix
```
rjmateus, line 144: Can be fixversion be null, for example for CVE that where detected but no fix was yet released?

rjmateus (approved):
Looks good to me. Just one small question that should not block the RFC to be merged.

accepted/0000-cve-auditing-with-oval.md
HoussemNasri: Yes, I remember trying to use a non-nullable string and set it to an empty string when fix version is unavailable but there was a database check that prevented it.

rjmateus: ok, thank sounds good. I think we should have it as null if no fix version is available.

mcalmer (approved):

PR #81, opened 2023-07-21
Adding RFC for password limitations removal epic task

thesp0nge:
RFC for https://github.com/SUSE/spacewalk/issues/22066

### DIFF
--- a/accepted/0000-remove-password-limitations.md
+++ b/accepted/0000-remove-password-limitations.md
@@ -1,0 +1,31 @@
+- Feature Name: remove_password_limitations
+- Start Date: 2023-07-21
+
+# Summary
+Remove password limitations when setting up UYUNI with YAST
+
+Today we give users, strict limitations when choosing a password during UYUNI setup in YAST. We ask not to use spaces, $ or ! and we suggest that a 7 characters password is the lower bound for a good choice.
+This suggests to skilled users that we store passwords somewhere in clear text and that we use them in shell scripts. This can also give advice to an attacker to find weaknesses in customer installations.
+
+I suggest moving over from these limitations, allowing passphrases over passwords with no restriction from characters a user would choose.
+
+# Motivation
+Improve security posture of UYUNI / SUSE Manager product and give more confidence to customer that we handle secrets in a secure way. 
+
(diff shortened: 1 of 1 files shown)


rjmateus (commented):
Thank you for your contribution to this topic.
AFAIK, yast is collecting data and saves it to a file named "setup_env.sh". After this a shell script called ("/usr/lib/susemanager/bin/mgr-setup") that loads those properties [1] and setup everything.

I don't think it's possible to save that data in the database, since the creating is done during the setup. Changing this would mean refactoring most of the setup script.

It can be changed, but we would need to investigate what will break in the scripts.
For Suse Manager 5.0 we are developing a new tool to setup the containerized environment [2]. @cbosdo this is something in need to consider in this new code base.

Another option that we may explore in the future is to create some a dummy certificate to be able to start apache, and the have a simple webUI to collect all setupdata and bootstrap from in stages (first database, then we can save the configurations to there).

[1] https://github.com/uyuni-project/uyuni/blob/master/susemanager/bin/mgr-setup#L58
[2] https://github.com/uyuni-project/uyuni-tools

cbosdo:
I don't think we need to move anything to database to properly handle passwords. It smells like a refactoring of the setup scripts which is now a collection of shell, python and perl scripts. Of course this can be done totally independently of the containerization of the server.

thesp0nge:
As you can see from my request, a huge refactor is exactly what I propose. How do you suggest removing password limitations?

mcalmer:
I think one requirement is, that we need some passwords in clear text. Hashing them is not an option.
The one we can handle as hashes, are already stored in that way. 
I think all the limitations are coming from cases where we need the password in clear text as we use it to login to a service, like login to the database.

thesp0nge:
any news on this?

mcalmer:
We have password to authenticate against the DB itself in rhn.conf .
We ask also for a password for a self generated SSL Certificates. But they are not stored permanently.
We need them only to feed the commands at generation time.
We have other passwords stored in the DB to authenticate against SCC, VMware or Cloud (virtual-host-gatherer) container registries or RMT server in the cloud.

PR #82, opened 2023-11-28
hub online synchronization

rjmateus:
This is a RFC to improve content synchronization for HUB scenarios. It also relates to Inter Server Synchronization.

[Rendered version](https://github.com/uyuni-project/uyuni-rfc/pull/82/files?short_path=2cf5174#diff-2cf517457bdce7fb7b50dd31c7c74dcee549a8e64965e96221fac04f05cfe5a0)


accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,199 @@
...
+
+### HUB XML-RPC API
+
+Component responsible for broadcasting API requests to peripheral servers. This component loads the peripheral servers that are registered on the HUB by looking at the minion entitlements.
+
+With the proposed solution we can retrieve the list of the registered peripheral servers and avoid the peripheral server registration as minion.
+
+### Inter-server-sync-v2
+
+Will be replaced by this solution. It will continue to exist but for disconnect set-ups only.
```
cbosdo, line 166: May be this should also state that ISSv2 will also stay for image and config channels sync until this is implemented in the new solution.

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,199 @@
...
+## Future Steps
+
+This implementation sets the foundations to add more online content synchronization through the API, like configuration channels, images and activation keys.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Full synchronize a high number of peripherals at same time can overload the HUB server
+  - After the first full synchronization it should not be a problem
+  - One Peripheral can manage up to 10.000 minions. If we add 10 peripherals we would be able to manage 100.000 minions (more than our biggest customer)
```
cbosdo, line 187: This item is not really a drawback to me and doesn't change over ISSv2 architecture

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,199 @@
...
+
+One of SUSE Manager focus areas is synchronizing and managing content, especially RPM's. To do so, it synchronizes remote channels, apply filters to produce new channels (CLM) and makes these channels available to be consumed.
+
+The overall idea of the Hub Online synchronization is to re-use the existing repo-sync mechanism and synchronize channels in the peripheral from the HUB published repositories (online connection between hub and peripherals servers).
+
+In the first stage the focus will be synchronizing software channels, since it's the most problematic one in terms of performance and automation. However, we should be able to re-use the integration mechanism, since everything will be implemented through API.
+
+We will need a mechanism to create the channels in the peripheral servers (vendor and customs ones) in the desired organization. Those channels need to have as content source the corresponding HUB server repository URL, with the correct authentication token. With this, repo-sync will take care of channel synchronization (proof of concept already tested).
+
+POS:
```
cbosdo, line 71:
```suggestion
Pros:
```

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,199 @@
...
+- Automatically update channel during the night with repo-sync/mgr-sync
+- No need to have specific code adaptations in case of schema changes. It will be handled any way in repo-sync changes
+- Rely on a stable specification: repositories metadata
+- After initial configuration, all synchronization between hub and peripheral should be automatic and transparent
+  - Only differences would be transferred
+- Promote environments in CLM will automatically propagate changes to all connected peripherals
+- Automatic creating of bootstrap repository done by repo-sync tool
+
+Cos:
```
cbosdo, line 80:
```suggestion
Cons:
```

accepted/0000-hub-online-synchronization.md
rjmateus: This is mentioned in the tools scope: https://github.com/uyuni-project/uyuni-rfc/pull/82/files#diff-2cf517457bdce7fb7b50dd31c7c74dcee549a8e64965e96221fac04f05cfe5a0R90

accepted/0000-hub-online-synchronization.md
rjmateus: In ISSv2 if one adds more peripherals it will not reflect in more HUB load since the export is done once and the import is done in parallel in each peripheral. 
With this approach, adding one more peripheral will add load to the HUB server

rjmateus:
@mcalmer Good point Michael.
I will add that to the next steps-section, but with some detail in the solution. 
This HUB online synchronization will not be able to fully replace ISSv2 because of the disconnected environments. However we may think about a solution using RMT to sync data, and then export to disconnected environment (but I'm not sure if I like it)

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,199 @@
...
+
+We can re-use and improve existing ISSv1 database tables to save the needed data.
+
+
+## Hub as a proxy for SCC data
+
+The SUSE Manager server needs a set of metadata to be able to operate. Currently that metadata is provided by SCC directly or, in the case of PAYG, provided by the cloud RMT infrastructure. We should also provide this data in SUSE Manager HUB to be consumed by the peripheral servers.
+The minimal endpoints to be provided are:
+   - "/suma/product_tree.json"
+   - "/connect/organizations/products/unscoped"
```
admd, line 123: To list the correct migration path, we use upgrade_paths.json and some information from SCC in combination. We need a way to have this information on peripheral servers to in order to make product migration work correctly.

@mcalmer am I missing anything or misunderstood around this?

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,199 @@
...
+
+On top of this, we should also provide an endpoint for peripherals to send status data needed by SCC (example of this is the minions registered and hardware information). Peripheral servers should send this data to the HUB instead of SCC, and the HUB server should consolidate it and send it to SCC.
+
+## Peripheral software channels creation
+
+We need a mechanism to create the channels in the peripheral servers (vendor and CLM's) in the desired organization. The peripheral channel creation must be done from the HUB server through an API. Since we are making special channel creation (defined next), those API methods should be available to server-to-server communication only.
+
+For each peripheral we should define the organization mapping between HUB organizations and peripheral ones, following the rules (similiar to what we have on ISSv1):
+  - Each Hub organization can only be mapped to one peripheral organizations
+  - Peripheral organization can receive data from several different HUB organizations
```
admd, line 133: Why should we keep following this design? I mean what's the downside the of peripheral org to receive data from only one Hub org?

accepted/0000-hub-online-synchronization.md
rjmateus: @admd I think that data is static and coming from the file "/usr/share/susemanager/scc/upgrade_paths.json" which is part of package "susemanager-sync-data", but MC can confirm this assumption.

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,199 @@
...
+
+From the webUI we should be able to select which channels should be added to each peripheral (vendor or custom). This needed to be saved in a new database table on the HUB side.
+
+Steps needed to create the channels on peripheral:
+   - In the HUB, generate authentication tokens to access the channel(s). Re-use the existing channel access table.
+   - In the peripheral create all needed channels (vendor and custom) point to the HUB server in the rhncontentsource table. Each channel needs to have its own entry.
+   - Minimal set of tables will be populated on the peripheral: rhnchannel,  rhnchannelcontentsource, rhnchannelproduct, rhnchannelfamilymembers, rhnproductname, rhnchannelcloned, suseproductchannel.
+
+
+When creating the channels in the peripheral server the content source will reference the correct HUB repository that provides that channel. Peripheral servers need to be configured with the flag `java.unify_custom_channel_management`, which will synchronize custom channels during the nightly mgr-sync process.
```
admd, line 143: Can you please provide further details on the methodology for establishing the mapping between channels and Hub repositories. Is it a manual process where users need to define the repository for each newly created channel in the peripheral, or does it operate automatically through predefined heuristics?

admd (commented):
This is lovely, and I appreciate it. I only have a few questions, but aside from that, everything appears satisfactory from my perspective.

accepted/0000-hub-online-synchronization.md
mcalmer: upgrade_paths.json is static and I think we just dropped it for 5.0 as it contains only outdated OSes.
Mainly SLES10 systems.

accepted/0000-hub-online-synchronization.md
rjmateus: I don't see a problem with one peripheral organization receiving data from one HUB organization. I was trying to keep it flexible.
This way, user can have Organizations on the HUB and decide to which peripheral organization they want to map it. If they select the same org to map on a peripheral, that would merge them in there. This is safe to do since channel labels are unique in the system, doesn't matter the organization.

accepted/0000-hub-online-synchronization.md
rjmateus: This should be automatic when creating the channel on the peripheral. It will create an entry in the table rhncontentsource and associate it with the channel. With this association repo-sync will work out of the box without any code change.
I will update the description to make it more clear.

rjmateus: @admd I have updated the description. Can you have a look if is clear enought

accepted/0000-hub-online-synchronization.md
rjmateus: @admd I was now looking at how we can also synchronize configuration channels. For that we cannot have a peripheral receiving data from several different HUB organizations since the channel label is not unique across orgs. I will change this requirement to only allow one org.

accepted/0000-hub-online-synchronization.md
cbosdo: If stating it once and for all is fine, then I'm good with this

rjmateus: In the mean time I added that information to this sentence. It would not hurt and bring claririty.

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,228 @@
...
+
+Configuration should be pushed from the HUB to peripherals, to avoid the need of direct management of the peripherals.
+
+For now this solution is focused on software channels only, but can be extended to synchronize more data types.
+
+
+## Define connection between HUB and Peripherals
+
+We can follow a similar approach to what exists on ISSv1. On the hub side we can define multiple peripheral servers to connect to by providing the FQDN and an authentication token.
+On the peripheral side we also need to define the Hub server FQDN and an Authentication token.
```
aaannz, line 110: Hub will have peripheral and generate associated auth token, right?

Why do we need to provide peripheral FQDN? Wouldn't generic peripheral name (may be FQDN) and generated token be enough? Or do we approach this as username/pass scenario?

I am assuming that connection will be always from peripheral to the hub, right?

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,228 @@
...
+The existing API methods under namespace `configchannel` cannot deal with cross organization mapping. To overcome it we have two options:
+   - Develop a new endpoint to create and update channels with server-to-server authentication and the possibility to defined the target organization. Besides organization identifier, all business logic should be the same as the API namespace `configchannel`.
+   - Collect organization admin credentials (user+password) on the HUB, and use the existing API. The downside is that it's more fields to be collected and managed.
+
+### Images
+
+TODO
+- Is all data saved in the database?
+- Is the identifier unique in the system, or we can have duplicates across organizations?
+- Are the existing endpoint sufficient to update all the data?
```
aaannz, line 203: > Is all data saved in the database?

Metadata + pillar data are in the database, images are separate, http(s) accessible

> Is the identifier unique in the system, or we can have duplicates across organizations?

Image IDs are unique across the installation.

> Are the existing endpoint sufficient to update all the data?

Split over multiple API calls + http for image download is there.
Importing the image is lacking. Only basic information + pillar + image files can be imported, e.g. package list is not possible to import using API.

Example implementation is in [contrip/os-image-tools](https://github.com/uyuni-project/contrib/blob/main/os-image-tools/osimage-import-export.py)

From my POV, once the API authentication using the token is available, the exporting of images is ready. Only thing needed is that peripheral server regularly calls hub server API to ask for new or removed images.
Some performance optimizations can be done, like time based image search and adding more complex API calls to limit needed number of calls can be done.

For import, this is quite lacking:
- image building handling is mangled together with salt event handling -> quite a refactoring to do so
- once refactored, we can reuse same classes and methods to do import of images + add an API endpoints.

accepted/0000-hub-online-synchronization.md
rjmateus: It should be used for authentication from HUB to peripheral. 
Communication will be bi-directional. Some cases will be peripheral calling hub (like synchronize software channels and calling SCC endpoints) other cases will be HUB calling the peripheral API like creating channels, pushing configuration channels, etc)

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,228 @@
...
+
+For now this solution is focused on software channels only, but can be extended to synchronize more data types.
+
+
+## Define connection between HUB and Peripherals
+
+We can follow a similar approach to what exists on ISSv1. On the hub side we can define multiple peripheral servers to connect to by providing the FQDN and an authentication token.
+On the peripheral side we also need to define the Hub server FQDN and an Authentication token.
+
+Each peripheral server can only have 1 Hub server (master). This will avoid dealing with problems like channel label conflicts between multiple masters. Configure a hub server will block access to a set of menu items like: "Admin" -> "Setup wizard" -> "Organization Credentials" and "Products" and "PAYG Connection" (similar to what we already have for ISSv1).
```
mcalmer, line 112: master => main ?

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,228 @@
...
+The SUSE Manager server needs a set of metadata to be able to operate. Currently that metadata is provided by SCC directly or, in the case of PAYG, provided by the cloud RMT infrastructure. We should also provide this data in SUSE Manager HUB to be consumed by the peripheral servers.
+The minimal endpoints to be provided are:
+   - "/suma/product_tree.json"
+   - "/connect/organizations/products/unscoped"
+
+On top of this, we should also provide an endpoint for peripherals to send status data needed by SCC (example of this is the minions registered and hardware information). Peripheral servers should send this data to the HUB instead of SCC, and the HUB server should consolidate it and send it to SCC.
+
+## Peripheral software channels creation
+
+We need a mechanism to create the channels in the peripheral servers (vendor and CLM's) in the desired organization. The peripheral channel creation must be done automatically from the HUB server through an API. Since we are making special channel creation (defined next), those API methods should be available to server-to-server communication only.
```
mcalmer, line 129: I think it would be good to define a little bit more in detail, how this server-to-server API should look like.
This should also say something about the exiting API namespaces `sync.master` and `sync.slave`.

- What namespace should be used for it?
- one namespace or multiple? 
- design it with an outlook to the future and what else needs to be added in future to this API. E.g. activation keys, config channels, images, formulas, etc.
- how should the authentication work?

accepted/0000-hub-online-synchronization.md
rjmateus: Thank you Ondrej
I have added the information about images.

accepted/0000-hub-online-synchronization.md
rjmateus: Hey Michael. I added some clarification about API namespace and use cases. I didn't add any details about the exact API methods to develop because it looks to me like it's an implementation detail.
Could you have a look if it's more clear now? Thank you

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,252 @@
...
+
+### Images
+
+For images metadata and pillar information are saved on the database. Images are store on disk but accessible through http(s). Images are associated to a Organization but it's IS is unique across installation.
+
+To copy data from one SUSE Manager server to another we would need to use several API calls and http(s) calls. Some data cannot be imported using the API, like package list information.
+
+An implementation example is available from the community [link](https://github.com/uyuni-project/contrib/blob/main/os-image-tools/osimage-import-export.py).
+
+Communication can be done from the HUB to the peripheral server to synchronize all the necessary data.
```
aaannz, line 227: I would like to point out not everyone would like to sync from HUB to peripheral.

Case being one of our users where they have different SUMAs for different environments - prod, qa, dev.
They build their images using SUMA in the dev environment, once image pass basic validation it is exported and imported to the qa SUMA using above mentioned script. Here image passes more thorough testing and once pass and maint. window open it is again moved to the prod. This process ensures no further changes to the image is done as the import/export do not modify image in any way.

Centrally managed hub network would help them with ensuring same configuration of the those SUMAs, however they would certainly need an ability to either:
- sync images from the peripheral to another peripheral and to HUB (this can be done outside HUB arch by existing APIs)
- prevent auto-syncing images from peripheral to HUB and/or overwriting peripheral images from HUB

aaannz (commented):
I am missing one important section and that is failure scenarios:

- what happens when peripheral is to be synced but is unavailable?

Seems like particularly in channel creation and CLM, where sync should be done automatically, this scenario can happen.
Since in this case connection direction is expected to be:

> other cases will be HUB calling the peripheral API like creating channels, pushing configuration channels, etc)

So if peripheral is unavailable, do we keep track what was updated and what not? And how?

- what happens when peripheral or hub crashes during the sync?

With ISSv2 everything was just one transaction so inconsistencies should not happen.
We should however check ACIDity of our APIs. And not only individual API calls, but sequences of them which we will use for the sync and define expected failure modes.

rjmateus:
@aaannz I added a section about failing scenarios. Do you think is clear enough or should it be more completed?

accepted/0000-hub-online-synchronization.md
rjmateus: Good point in here Ondrej.
Let's move by parts, and I will start from the end.

The idea is for users to be able to define in the HUB if they were to synchronize all data or only a select set. This way they could control when an image lands on each peripheral server.

HUB server could also have the ability to build images. It will have all the channels, so they create a build host assigned to the HUB server.

Considering those two assumptions would make sense to build the image on HUB server and then transfer that to dev preipheral server and make all the necessary tests. After all dev test where made, we transfer the new image version to other environments (qua, prod) and make it available to all.

If this is not the case, then we can always use the script you mentioned or ISSv2, since it will stay around.

The goal for this solution is scalability only, but other use cases will stay around and may need different implementation and components.

cbosdo:
Remember that a lazy  repo-sync has been started. This may have an impact on the design.

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,267 @@
...
+
+- Full synchronize a high number of peripherals at same time can overload the HUB server
+  - After the first full synchronization it should not be a problem
+  - One Peripheral can manage up to 10.000 minions. If we add 10 peripherals we would be able to manage 100.000 minions (more than our biggest customer)
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Create a new UI and use ISSv2 to synchronize data
+- Solve the existing and known problems of ISSv2
```
srbarrios, line 261: I'd not discard this point, in addition to implement ISSv3.
As ISSv2 it's going to be used for disconnected environments, we can still bring a better user experience to that use case. For example, can we consider some improvement around performing parallel sql queries?

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,267 @@
...
+- After initial configuration, all synchronization between hub and peripheral should be automatic and transparent
+  - Only differences would be transferred
+- Promote environments in CLM will automatically propagate changes to all connected peripherals
+- Automatic creating of bootstrap repository done by repo-sync tool
+
+Cons:
+- All peripheral can start synchronizing at same time
+  - Can be problematic if we have several peripherals performing full synchronization at same time. However we can configure the peripherals to run repo-sync in different hours and spread the load
+  - This is only a problem in first sync, since subsequent sync only transfer differences
+
```
srbarrios, line 84: When synchronizing from SCC service, in theory, we rely on a service with HA.
In that proposal, the peripherals will rely in a unique Hub instance, what happen if that goes down or.. the Hub disk burns? I would also consider how we can recover under these cases.

accepted/0000-hub-online-synchronization.md
rjmateus: In terms of performance, the main issues are in the import process. In there we cannot change it to run in parallel. 
We can however change the way we do the transaction, and have one transaction per channel, instead of one transaction per export as we have now.
Another possibility is to not have a transition at all, but that can be risky if users start to use the channel during the import process, or if an error occurs during the import.

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,359 @@
...
+To define this connection the existing API namespace should be re-use (`sync.master` and `sync.slave`), with some adaptations to cope with the new requirements. Example of this is the need to define an authentication method between servers to the connection more secure.
+
+During development phase we should consider if we can join both namespaces in one name `hub`, for example.
+
+
+### Alternative 2:
+We can follow a similar approach to what we do to activate a proxy.
+On the Hub we create an API which is called authenticated and can take additional data like credentials for the peripheral server.
+This API create a system (kubernetes) or change an existing system (podman host exists) to an Uyuni Peripheral Server.
```
rjmateus, line 127: I would separate it from the current systems list and system profile registration. We have made that for hub/peripheral on 5.0 because we didn't have a better approach.
I would not tie the UB/peripheral configuration on the need to have a system profile on SUSE Manager, that is a detail

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,359 @@
...
+
+On the peripheral side we call the API on the Hub and fill out the required configuration data needed locally on the peripheral server (e.g. auto generated mirror credential password).
+We will also change the `scc_url` configuration on the Peripheral to point to the Hub Server.
+
+On the hub, the activate call will create the peripheral configuration with first data.
+
+Alternatively for the Hub/peripheral communication an new API might be useful to have a clear separation.
```
rjmateus, line 134: I would upvote for this.

accepted/0000-hub-online-synchronization.md
rjmateus: On top of that, I would follow the approach of ISSv1, and have dedicated UI/API call to configure the relationship and not rely on the peripheral being registered as minion or system profile on the HUB

rjmateus (commented):
First review iteration of alternative 2

accepted/0000-hub-online-synchronization.md
mcalmer: I think we need to discuss it how we want to show a Peripheral or Proxy "server" in the UI. 
From my point of view we need to handle both in the same way, but we should use the time now to find a final solution and not a another workaround or short cut.

rjmateus: 100% agree with you. After our conversation in the morning, I'm better understanding what was your goal.
The initial configuration have a circular dependency, and to avoid to many manual steps we must think of a way to improve it. Using a similar mechanism and the proxy activation makes sense to me. As you have said we could call an API on the HUB, passing the credentials for the peripheral server and we setup the ISSv3 basic configurations automatically. 
 
ON the management side, I see a few questions that can be managed differently:
- Manage a proxy and peripheral servers. In my opinion, this should done through the existing Systems UI with a new element(s). The logic for server and proxy should be similar, with different entitlements.
- Configure HUB-perpheral authentication: this can be done with a new tab on the system details, or on a dedicated page on Admin menu. A dedicated page would be better, in my opinion.
- Configure HUB-perpheral content synchronization: We should make it on a dedicated page on the admin menu.

We can make a differentiation about the system management, like upgrading the peripheral server/proxy, and the ISSv3 configuration and management. They can exist independently of each other, and use just one or the other.

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,359 @@
...
+When creating the channels in the peripheral server the content source will reference the correct HUB repository that provides that channel. This should be done by a new special API (server to server communication) which will add a entry to table `rhncontentsource` with the URL to the HUB channel repository. With this information repo-sync will be able to synchronize data from HUB server without any code change.
+
+Peripheral servers need to be configured with the flag `java.unify_custom_channel_management`, which will synchronize custom channels during the nightly mgr-sync process.
+
+One important aspect is to recreate the connection between custom channels and vendor channels, so we can have SP migration and avoid the need to synchronize all channel clone chains (ISSv1 also does this implementation).
+
+Vendor channels are not linked to any organization, and can also be synchronized with this method.
+
+
+### Alternative 2:
```
rjmateus, line 219: looks perfect to me

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,359 @@
...
+
+For use cases 2) and 3), we have a problem with API authentication and authorization.
+Currently, each user is part of an organization, even if it has the role `sat_admin`. Since we need to make cross-organizations call we would need a new authentication/authorization mechanism or change the authorization for the existing API methods.
+
+For this reason, the best approach would be develop a new namespace with a token authentication to be used for server-to-server communication. This namespace will contain all the needed API methods for hub content synchronization. Some methods will be similar to other existing methods in other namespaces, with differences in authentication/authorization and maybe an extra parameter for the organization (since it would be extracted from the authenticated user).
+
+The exact list of methods and parameters are seen as an implementation detail.
+However for data creation (like create software channels, synchronize images, etc) we should have consider one single API method for each data type and be sure of is ACIDity.
+
+#### Alternative 2:
```
rjmateus, line 263: this one looks perfect to me

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,398 @@
...
+   - the FQDN of the Hub
+   - the credentials to setup on the Peripheral Server
+
+The authentication credentials works only for a special server-to-server API and is not connected to any user on the hub or peripheral server.
+The new server-to-server API is required as the existing API is tied to the organization of the calling user. The new API should work organization independent.
+
+The Peripheral Server may already be configured. When restarted with this additional configuration, it configures the Hub configuration and setup the credentials
+that the Hub can connect and finish the configuration.
+
+We may want to provide the possibility to generate a full peripheral configuration which makes it possible to setup the peripheral server from scratch including the Hub configuration (similar to the proxy configuration).
```
rjmateus, line 143: I would put this as a next step if needed. By configuring just the server-to-server API any other subsequent steps should be automated as much as possible.

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,398 @@
...
+
+
+After the peripheral server is setup and the Hub can connect via the server-to-server API, the final configuration can be triggered by the Hub.
+This include:
```
rjmateus, line 147: This can all be done in the same taskomatic process, with several multiple steps. Hub decide to apply configuration from HUB and then all this data is transferred and stored in the right locations.

accepted/0000-hub-online-synchronization.md
```diff
@@ -136,6 +136,29 @@ This is needed when we want to make this API organization independent.
...
+Similar as in the Proxy case, we should have a "system" entry which represent the Peripheral Server.
+This can be used to show special information or relations.
+
+There are 2 different cases we need to consider.
+   1. A host system exists with the same FQDN (podman running the container)
+   2. No host system got registered (kubernetes case or podman where the host should not be registered at the Hub)
+
+To handle these 2 cases, we should create a new base entitlement for "Container Workloads" (similar to Foreign Entitlement) and 2 add-on entitlements for Peripheral and Proxy Server.
+In case we have a host system, we would not use the "Container Workload Entitlement" as base, but the "Salt Entitlement".
+Only in case if no host is available, the "Container Workload Entitlement" would be used.
```
admd, line 166: Every 5.0 suse manger instance currently has same static hostname  `uyuni-server.mgr.internal`, how are we we gonna differentiate between all these peripheral servers.

accepted/0000-hub-online-synchronization.md
```diff
@@ -122,18 +122,34 @@ During development phase we should consider if we can join both namespaces in on
 
 
 ### Alternative 2:
-We can follow a similar approach to what we do to activate a proxy.
-On the Hub we create an API which is called authenticated and can take additional data like credentials for the peripheral server.
-This API create the configuration on the Hub side and also handle the creation of a system which represent an Uyuni Peripheral Server.
 
-On the peripheral side we call the API on the Hub and fill out the required configuration data needed locally on the peripheral server (e.g. auto generated mirror credential password).
-We will also change the `scc_url` configuration on the Peripheral to point to the Hub Server.
+We need to establish a secure connection between Hub and Peripheral Server.
```
admd, line 126: @rjmateus @mcalmer are we going to drop the existing way of Hub and peripheral servers way of work? What would be the hub_xmlrpc_api role, will that be gone too? 

What path are we providing for the users who are working with the current approach.

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,398 @@
...
+This configuration can be exported and needs to be transferred to the Peripheral Server.
+This configuration include:
+   - the Root CA of the Hub
+   - the FQDN of the Hub
+   - the credentials to setup on the Peripheral Server
+
+The authentication credentials works only for a special server-to-server API and is not connected to any user on the hub or peripheral server.
+The new server-to-server API is required as the existing API is tied to the organization of the calling user. The new API should work organization independent.
+
+The Peripheral Server may already be configured. When restarted with this additional configuration, it configures the Hub configuration and setup the credentials
```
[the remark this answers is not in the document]
mcalmer: I rephrase this to "import the configuration".

accepted/0000-hub-online-synchronization.md
mcalmer: I put this in an extra section

accepted/0000-hub-online-synchronization.md
mcalmer: yes, but this happens typically every night. But when you setup a new peripheral, you want to trigger it "now".
So we need the button and a job which does it nightly.

accepted/0000-hub-online-synchronization.md
mcalmer: This is the "internal" hostname in the container network. We need the hostname from external. Otherwise the hub cannot connect as the internal name is not routed.

accepted/0000-hub-online-synchronization.md
mcalmer: hub_xmlrpc_api is for ISSv2. In worst case the user can stay in it.
I cannot say if this is compatible with ISSv3. I do not really know what this is doing.
@rjmateus might be able to answer this

accepted/0000-hub-online-synchronization.md
admd: Got it! I just wanted to highlight that the internal hostname doesn’t become an issue.

accepted/0000-hub-online-synchronization.md
rjmateus: Sound fine to me. however, would be a option to have the nightly job and at the same time trigger a job execution for that peripheral when he is created or when an update to the configurations is made? This way we would avoid the button.

accepted/0000-hub-online-synchronization.md
rjmateus: hub_xmlrpc_api will stay as is, since it's note dependent on the ISS version in use. This means it can work with any version of ISS, even the v3.
However, we may need to make some changes to one API that is providing the hub servers FQDN to the xml-rpc API, to instead of looking at the system entitlement it returns a list of configured peripherals

admd: And what about all different authentication methods https://documentation.suse.com/suma/5.0/en/suse-manager/specialized-guides/large-deployments/hub-auth.html ? 
To support them, we will leave it up to the user to create the appropriate API users, ensuring that those users can be utilized to correctly make the API calls later on?

accepted/0000-hub-online-synchronization.md
mcalmer: Well, you first generate this config. This is a change and you trigger this update, but at that point in time the peripheral is not yet setup and the job would fail.
Than you configure the peripheral with this config. Maybe that's all what you want at that point in time, but you would like to know if the communication works now. You will ask for a button.
I think we need a button. Especially when an admin do not want to deploy all the changes directly to the peripheral server he is configuring on the hub side.

rjmateus: ok, got you point. I forgot about this chicken & egg problem :)
It's better to have the button, worth case if users don't want they are not forced to click it.

accepted/0000-hub-online-synchronization.md
rjmateus: @srbarrios is this stilla question for you?

accepted/0000-hub-online-synchronization.md
```diff
@@ -0,0 +1,402 @@
...
+### Alternative 2:
+
+We need to establish a secure connection between Hub and Peripheral Server.
+For this we create a Peripheral configuration on the Hub server.
+It mainly defines the FQDN of the Peripheral and generate credentials to authenticate from the Hub.
+The Root CA must be configured as well when the peripheral server is already setup.
+
+This configuration can be exported and needs to be transferred to the Peripheral Server.
+This configuration include:
+   - the Root CA of the Hub
```
rjmateus, line 133: We are providing the Root CA from hub to be setup on the peripheral.  Do you think that HUB would need to also get the peripheral CA, in case it's different? I am thinking about the SSL connection to the API, for example.

cbosdo: Any outcome here would probably also help for Kubernetes case where it may be hard to share the same RootCA across clusters if using `cert-manager`. We'll face similar issues for proxies on Kubernetes too.

PR #83, opened 2023-11-30
Reboot Required Flag for non-SUSE Distros

wweellddeerr:
See [rendered version](https://github.com/uyuni-project/uyuni-rfc/blob/2235b410a8809689c5dff8b90f9409915aeb8663/accepted/00000-reboot-required-flag.md).

### DIFF
--- a/accepted/00000-reboot-required-flag.md
+++ b/accepted/00000-reboot-required-flag.md
@@ -9,8 +9,8 @@ The feature of showing in the Web UI when a system managed by Uyuni requires a r
 
 ### Detailed Design
 
-#### Current Approach Used for SUSE Distros
-Currently, except for transactional systems (i.e. SLE Micro), Uyuni lacks a system attribute to indicate directly whether a given system requires a reboot at a given moment. The information is derived from other attributes: the time of the last boot and the packages installed in the system. If a package has the “reboot suggested/needed/required” characteristic, and it was installed in a system after its last boot, Uyuni then shows in the WebUI that a reboot is required for this system. It is interesting to note that Uyuni is sensitive only to package installations. If a package removal or any other change causes a reboot to be required, this will not be detected by Uyuni. The “reboot suggested/needed/required” characteristic of a package is also derived from other data that is only available for SUSE Distros.
+#### Current Approach Used for SUSE distributions
+Currently, except for transactional systems (i.e. SUSE Linux Enterprise Micro), Uyuni lacks a system attribute to indicate directly whether a given system requires a reboot at a given moment. The information is derived from other attributes: the time of the last boot and the packages installed in the system. If a package has the “reboot suggested/needed/required” characteristic, and it was installed in a system after its last boot, Uyuni then shows in the WebUI that a reboot is required for this system. It is interesting to note that Uyuni is sensitive only to package installations. If a package removal or any other change causes a reboot to be required, this will not be detected by Uyuni. The “reboot suggested/needed/required” characteristic of a package is also derived from other data that is only available for SUSE distributions.
 
 
 #### Proposed Solution
@@ -22,8 +22,8 @@ The proposed solution entails extracting information directly from the system to
 The suitable way for checking if a reboot is required in a Linux system varies according to its family and version. Here, the following approach will be used:
 
 - Debian/Ubuntu: check if `/var/run/reboot-required` file exists
-- Suse: check if `/boot/do_purge_kernels` or `/run/reboot-needed` file exists.
-- RedHat:
+- SUSE: check if `/boot/do_purge_kernels` or `/run/reboot-needed` file exists.
+- Red Hat:
   - Major release >=8: check if the exit code of running `dnf -q needs-restarting -r` command is 1.
   - Major release < 8: check if the exit code of running `needs-restarting -r` command is 1.
 
@@ -39,9 +39,9 @@ Additionally, a boolean indicating whether a reboot is required or not will be i
 
 As the `reboot_info` beacon will be adapted to not fire events indicating that the reboot is not required, it is necessary to adapt the logic of the reboot required flag in server side. Currently there is a boolean database column `reboot_needed` at `suseMinionInfo` table for this purpose. This column will be modified to be a timestamp `reboot_required_after` and the server will need to compare this timestamp with the `uptime` of the system to determine if a reboot is required or not in a given moment.
 
-##### Changing Suse distros approach
+##### Changing SUSE distributions approach
 
-As it is possible to handle the reboot required information using the proposed approach, it will be used for all distributions. This implies changing the current approach for Suse systems to not query package/patch metadata for this purpose. Thereby, there will be a standardized implementation for the feature. Furthermore, in large installations, the query to list systems requiring reboot will have a better performance.
+As it is possible to handle the reboot required information using the proposed approach, it will be used for all distributions. This implies changing the current approach for SUSE systems to not query package/patch metadata for this purpose. Thereby, there will be a standardized implementation for the feature. Furthermore, in large installations, the query to list systems requiring reboot will have a better performance.
 
 
 ### Drawbacks


accepted/00000-reboot-required-flag.md
```diff
@@ -0,0 +1,59 @@
...
+
+  
+#### Proposed Solution
+The proposed solution is to introduce an execution module that reads directly from the system whether a reboot is required, following specific approaches according to the OS family and version. This execution module will run together with package profile update and the result will be persisted in the database using the `reboot_needed` column at `suseminioninfo` table. During a minion startup event the flag will be set to `false` assuming that the reboot should not be necessary anymore as the system is starting.
```
rjmateus, line 17: I'm not sure about the start-up event, as you mentioned it cannot be reliable since it can be caused by a salt-minion service restart. 
Would be possible (or even make sense) to save the date when we change the reboot flag, and in the startup even we try to get a grain with the last boot time and if it's after the reboot flag change date then we can change it.
Or maybe have a grain with the reboot flag information and send it in the startup event?

rjmateus (commented):
Looks good to me. I'm not sure only about the way to clean the reboot flag with the startup even.

accepted/00000-reboot-required-flag.md
cbbayburt: Isn't there any indicator that differentiates between an actual reboot and a service restart in the event payload?

accepted/00000-reboot-required-flag.md
```diff
@@ -0,0 +1,59 @@
...
+It's feasible to introduce a configuration property that determines the activation or deactivation of the reboot information beacon. Subsequently, the server would detect and display the 'reboot required' indication based on this property. If the beacon is active, the execution module's outcomes are ignored, and the flag isn't removed during the processing of startup event results. Conversely, when the beacon is inactive, we resort to employing the execution module strategy. Moreover, an automatic disabling of the beacon could be implemented if the server surpasses a certain threshold of registered minions, considering that performance issues would likely arise primarily in large installations.
+
+However, the principal drawback of this approach lies in its inherent complexity, posing challenges in its implementation, comprehension, and maintenance in the future.
+
+
+### Unresolved Questions
+- Should the strategy for identifying reboot required for SUSE distros be replaced for the sake of performance and standardization with other distros? Or is it a better idea to don’t change it considering that it is running for a long time and therefore it is stable?
```
cbbayburt, line 58: IMO it's well worth the risk. Our current implementation is not perfect anyway, and keeping both implementations is an extra maintenance burden.

cbbayburt (commented):
LGTM, too :+1:

accepted/00000-reboot-required-flag.md
```diff
@@ -0,0 +1,59 @@
...
+- Is checking reboot information at the times of package changes (package profile update) sufficient to cover the most relevant use cases? While this strategy is currently employed for SUSE distros, should we be concerned with addressing other potential causes for reboots?
```
cbosdo, line 59: Since you are considering reworking the whole thing I would not restrict the use case of other causes

accepted/00000-reboot-required-flag.md
```diff
@@ -0,0 +1,59 @@
...
+
+Although there are other possible reasons why a reboot may be necessary (a kernel parameter change, for instance), our current approach for SUSE distros is based only on packages/patch installation and seems to be enough for covering the most relevant use cases. Considering that a package profile update already runs after any package/patch change - even if it runs directly on the system - running the module together with package profile update should be enough for capturing when the reboot is required.
+
+The suitable way for checking if a reboot is required in a Linux system varies according to its family and version. The execution module will use the following approach:
+- Debian/Ubuntu: check if `/var/run/reboot-required` file exists
+- Suse:
+  - Major release >= 12: check if the exit code of running `zypper ps -s` command is 102.
```
mcalmer, line 24: Problem: this is not working always. I had here cases where the zypper update/install command exist with 102. But when calling "zypper ps" after this showed not 102.
Please contact @mlandres to find out what is a reliable way to get reboot needed

accepted/00000-reboot-required-flag.md
mcalmer: what about making this a timestamp "require_reboot_after".
Now we can compare the uptime / last reboot date with that date to be sure that it was a reboot and not just a salt-minion restart.

accepted/00000-reboot-required-flag.md
mcalmer: Not sure. Please talk to the libzypp team. Maybe we need both as only special conditions trigger the one **or** the other.

accepted/00000-reboot-required-flag.md
rjmateus: Upvote on MC idea

accepted/00000-reboot-required-flag.md
```diff
@@ -0,0 +1,59 @@
...
+
+  
+#### Proposed Solution
+The proposed solution is to introduce an execution module that reads directly from the system whether a reboot is required, following specific approaches according to the OS family and version. This execution module will run together with package profile update and the result will be persisted in the database using the `reboot_needed` column at `suseminioninfo` table. During a minion startup event the flag will be set to `false` assuming that the reboot should not be necessary anymore as the system is starting.
+
+Although there are other possible reasons why a reboot may be necessary (a kernel parameter change, for instance), our current approach for SUSE distros is based only on packages/patch installation and seems to be enough for covering the most relevant use cases. Considering that a package profile update already runs after any package/patch change - even if it runs directly on the system - running the module together with package profile update should be enough for capturing when the reboot is required.
```
admd, line 19: > Considering that a package profile update already runs after any package/patch change - even if it runs directly on the system


This assertion is inaccurate. We do not execute a package profile update after each package or patch installation. It is only triggered in the event of substantial changes. Typically, we receive and apply the delta, updating the database directly.

admd (commented):
We also need to address scenarios where information regarding the necessity of a restart is unavailable. The key is to present accurate information; if we cannot definitively determine whether a restart is required, it is essential to communicate this uncertainty to the user.

@agraul @meaksh can you please also take a look here.

agraul (commented):
I like the approach of asking the systems about their status instead of computing the status for every system on the server.

accepted/00000-reboot-required-flag.md
mcalmer: First: "`zypper ps -s` command is 102" does not work. The `needs-rebooting` command check for `/run/reboot-needed` file. I think this is what we should do. Check if either  `/run/reboot-needed` or `/boot/do_purge_kernels` exists?
We can ask for backporting needs-rebooting to SLE12. It would help. If it just look for patch data, we would be as good as we currently are when we do the same in SUMA directly.

accepted/00000-reboot-required-flag.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+
+#### Hybrid (Salt Beacon + execution module)
+It's feasible to introduce a configuration property that determines the activation or deactivation of the reboot information beacon. Subsequently, the server would detect and display the 'reboot required' indication based on this property. If the beacon is active, the execution module's outcomes are ignored, and the flag isn't removed during the processing of startup event results. Conversely, when the beacon is inactive, we resort to employing the execution module strategy. Moreover, an automatic disabling of the beacon could be implemented if the server surpasses a certain threshold of registered minions, considering that performance issues would likely arise primarily in large installations.
```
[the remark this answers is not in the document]
agraul: Okay, that would be good. We probably don't have that PR backported, I think we didn't know about that. But our beacon config is currently not aligned with upstream's anyway, we had to roll back some of their changes. (https://github.com/SUSE/spacewalk/issues/15051 tracks adding the upstream changes again)

accepted/00000-reboot-required-flag.md
mcalmer: JFYI - `needs-rebooting` is currently not working. I reported a bug and it should be fixed with the next maintenance update for libzypp/zypper for all SLE15 and openSUSE products.

accepted/00000-reboot-required-flag.md
```diff
@@ -10,50 +10,53 @@ The feature of showing in the Web UI when a system managed by Uyuni requires a r
...
+##### Reading from the systems that a reboot is required
+
+The suitable way for checking if a reboot is required in a Linux system varies according to its family and version. Here, the following approach will be used:
 
-The suitable way for checking if a reboot is required in a Linux system varies according to its family and version. The execution module will use the following approach:
 - Debian/Ubuntu: check if `/var/run/reboot-required` file exists
-- Suse:
-  - Major release >= 12: check if the exit code of running `zypper ps -s` command is 102.
-  - Major release <12: check if `/boot/do_purge_kernels` file exists.
+- Suse: check if `/boot/do_purge_kernels` or `/run/reboot-needed` file exists.
```
deneb-alpha, line 25:
```suggestion
- SUSE: check if `/boot/do_purge_kernels` or `/run/reboot-needed` file exists.
```

accepted/00000-reboot-required-flag.md
```diff
@@ -10,50 +10,53 @@ The feature of showing in the Web UI when a system managed by Uyuni requires a r
 ### Detailed Design
 
 #### Current Approach Used for SUSE Distros
```
deneb-alpha, line 12:
```suggestion
#### Current Approach Used for SUSE distributions
```

accepted/00000-reboot-required-flag.md
```diff
@@ -10,50 +10,53 @@ The feature of showing in the Web UI when a system managed by Uyuni requires a r
...
+Currently, except for transactional systems (i.e. SLE Micro), Uyuni lacks a system attribute to indicate directly whether a given system requires a reboot at a given moment. The information is derived from other attributes: the time of the last boot and the packages installed in the system. If a package has the “reboot suggested/needed/required” characteristic, and it was installed in a system after its last boot, Uyuni then shows in the WebUI that a reboot is required for this system. It is interesting to note that Uyuni is sensitive only to package installations. If a package removal or any other change causes a reboot to be required, this will not be detected by Uyuni. The “reboot suggested/needed/required” characteristic of a package is also derived from other data that is only available for SUSE Distros.
+
```
deneb-alpha, line 14:
```suggestion
Currently, except for transactional systems (i.e. SUSE Linux Enterprise Micro), Uyuni lacks a system attribute to indicate directly whether a given system requires a reboot at a given moment. The information is derived from other attributes: the time of the last boot and the packages installed in the system. If a package has the “reboot suggested/needed/required” characteristic, and it was installed in a system after its last boot, Uyuni then shows in the WebUI that a reboot is required for this system. It is interesting to note that Uyuni is sensitive only to package installations. If a package removal or any other change causes a reboot to be required, this will not be detected by Uyuni. The “reboot suggested/needed/required” characteristic of a package is also derived from other data that is only available for SUSE distributions.

```

accepted/00000-reboot-required-flag.md
```diff
@@ -10,50 +10,53 @@ The feature of showing in the Web UI when a system managed by Uyuni requires a r
...
+The suitable way for checking if a reboot is required in a Linux system varies according to its family and version. Here, the following approach will be used:
 
-The suitable way for checking if a reboot is required in a Linux system varies according to its family and version. The execution module will use the following approach:
 - Debian/Ubuntu: check if `/var/run/reboot-required` file exists
-- Suse:
-  - Major release >= 12: check if the exit code of running `zypper ps -s` command is 102.
-  - Major release <12: check if `/boot/do_purge_kernels` file exists.
+- Suse: check if `/boot/do_purge_kernels` or `/run/reboot-needed` file exists.
 - RedHat:
```
deneb-alpha, line 26:
```suggestion
- Red Hat:
```

accepted/00000-reboot-required-flag.md
```diff
@@ -10,50 +10,53 @@ The feature of showing in the Web UI when a system managed by Uyuni requires a r
...
- - Have a standardized way of checking for reboot required.
+##### Storing reboot information
 
-On the other hand, the main reason for not using this alternative is the risk of overloading the server with beacon events, especially for large installations. For example, applying a patch that makes a reboot to be required in multiple systems will prompt these client systems to transmit the beacon event in addition to package profile update, resulting in a doubling of the events that the server must process. Likewise, upon the client system reboot, the beacon will send the reboot required events just after the startup event.
+As the `reboot_info` beacon will be adapted to not fire events indicating that the reboot is not required, it is necessary to adapt the logic of the reboot required flag in server side. Currently there is a boolean database column `reboot_needed` at `suseMinionInfo` table for this purpose. This column will be modified to be a timestamp `reboot_required_after` and the server will need to compare this timestamp with the `uptime` of the system to determine if a reboot is required or not in a given moment.
 
+##### Changing Suse distros approach
```
deneb-alpha, line 42:
```suggestion
##### Changing SUSE distributions approach
```

accepted/00000-reboot-required-flag.md
```diff
@@ -10,50 +10,53 @@ The feature of showing in the Web UI when a system managed by Uyuni requires a r
...
 
+As it is possible to handle the reboot required information using the proposed approach, it will be used for all distributions. This implies changing the current approach for Suse systems to not query package/patch metadata for this purpose. Thereby, there will be a standardized implementation for the feature. Furthermore, in large installations, the query to list systems requiring reboot will have a better performance.
```
deneb-alpha, line 44:
```suggestion
As it is possible to handle the reboot required information using the proposed approach, it will be used for all distributions. This implies changing the current approach for SUSE systems to not query package/patch metadata for this purpose. Thereby, there will be a standardized implementation for the feature. Furthermore, in large installations, the query to list systems requiring reboot will have a better performance.
```

deneb-alpha (commented):
a few suggestions left

mcalmer:
@wweellddeerr JFYI - https://bugzilla.suse.com/show_bug.cgi?id=1217948
Let's see if we can get this feature

PR #84, opened 2024-01-16
RFC: Confidential Compute Attestation Report

mcalmer:
Confidential Compute Attestation Report in SUMA

[Rendered Version](https://github.com/uyuni-project/uyuni-rfc/blob/coco-attestation/accepted/00099-coco-attestation.md)

Issue: https://github.com/SUSE/spacewalk/issues/23421


accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,80 @@
...
+
+- Apply a salt state on the VM
+  - send a nonce file
+  - request the attestation report
+  - request the vcek certificate
+- Store the data into the database
+  - Status: pending
+  - Store nonce, report and vcek data
+  - assign to a systemid and an execution date
+- Build a Container which does the attestation
```
srbarrios, line 46: What about having an .sls for this task, so we can perform these salt states on each VM, instead of building a container to run the attestation? Also, this approach will facilitate scheduling it as an action every X time, and it will help to run it in parallel for all the minions we want to check it.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,80 @@
...
+  - uses a cron like approach running "every minute"
```
srbarrios, line 47: Why do we need this high cadence? 🤔 Can't we just check it when the user asks for it plus once every 12 hours or so?

mcalmer: This require an API endpoint in the container which accept "pings" from SUMA.
Also when running multiple instances of the container, you need a way to identify a container which has free capacity and a load balancer. 
The "every minute" is similar to what taskomatic is currently doing. Wakeup and check if there is something to do.
If not, go and sleep for another minute.
Currently all just ideas.

accepted/00000-coco-attestation.md
mcalmer: The attestation container is for more security. If you attest things on a hacked VM the result cannot be trusted.
The verification should happen on a different system. A container sounds like a perfect candidate for this job.

mcalmer:
@dvosburg and @BLandorff maybe you want to review as well

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,244 @@
...
+You can start a Confidential VM instance (currently no SNP available)
+Attestation via Cloud Logging. 
+
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Currently limited to KVM guests and only available as technical preview.
+Attestation of Cloud Workloads is questionable as the cloud provider writing there own attestation.
```
vandabarata, line 233:
```suggestion
Attestation of Cloud Workloads is questionable as the cloud provider is writing their own attestation.
```

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,244 @@
...
+Optionally the VCEK could be read from the extended memory if this is supported by the environment.
+In this case the VCEK is send together with the attestation report and stored in the database.
+
+Additionally it should be tested if secure boot is enabled.
+
+### The Attestation Container
+
+The Container should contain the `snpguest` tool and the CA Certificates for the different Processor Models (ARK and ASK for Milan and Genoa).
+In case the VCEK is available, the verification can directly start and the container only need access to the database.
+If the VCEK is not available, the container need to fetch it from the AMD Key Distribution Center.
```
vandabarata, line 197:
```suggestion
If the VCEK is not available, the container needs to fetch it from the AMD Key Distribution Center.
```

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,244 @@
...
+
+Optionally the VCEK could be read from the extended memory if this is supported by the environment.
+In this case the VCEK is send together with the attestation report and stored in the database.
+
+Additionally it should be tested if secure boot is enabled.
+
+### The Attestation Container
+
+The Container should contain the `snpguest` tool and the CA Certificates for the different Processor Models (ARK and ASK for Milan and Genoa).
+In case the VCEK is available, the verification can directly start and the container only need access to the database.
```
vandabarata, line 196:
```suggestion
In case the VCEK is available, the verification can directly start and the container only needs access to the database.
```

mikeletux:
I've got some suggestion regarding tables in the database:

For example:
```
environment VARCHAR(128) NULL
                      CONSTRAINT suse_srvcocoatt_rep_env_ck
                        CHECK(environment IN ('kvm_amd_epyc_milan', 'kvm_amd_epyc_genoa')),
```

Shouldn't be better to create a new table with all the environment instead of having them hardcoded so that, when in the future new environments are added we do not change the schema itself but we add those into that new table?
We could reference it a a FK to that new table. 

Not a database expert but the idea came to my mind that might be helpful.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,245 @@
...
+- succeeded
+
+The initial status is `pending` which indicate that the verification has not yet happen.
+
+### Step two
+The second step is a process which run in a dedicated environment like a container which verify the result.
+The attestation container for that environment search regular for pending attestations of its environment type.
+It read the data from the report table, performs all the needed checks and write the outcome into a result
+table with a status of `succeeded` or `failed`.
+Depending of the status of all checks, it calculate and set the status of the report from `pending` to `succeeded` of `failed`.
```
admd, line 63:
```suggestion
Depending of the status of all checks, it calculate and set the status of the report from `pending` to `succeeded` or `failed`.
```

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,245 @@
...
+
+- Apply a salt state on the VM
+  - send a nonce file
+  - request the attestation report
+- Store the data into the database
+  - Status: pending
+  - Store nonce and report data
+  - assign to a systemid and an execution date (create date of the report)
+- Run a Container which does the attestation
+  - search the DB for pending attestation
```
admd, line 179: This is more for an architect but should we not control this at Uyuni end? What I mean is that Uuyni should check if there is a pending attestation and then based on that it should  decide which container to forward the request for verification.

If we go above mentioned approach and we have multiple containers for all different checks, wouldn't they all be fighting and checking if something is there for them?

admd (commented):
Just a couple of questions but rest looks good. Thank you.

accepted/00000-coco-attestation.md
rjmateus: Controlling at n uyuni side do you mean by a API call to the container? I would say we should try to avoid develop more API's for this, and use the database for data transfer and control processing. They may be fighting for processing, but we need to get fine control on that.

rjmateus (commented):
Looks good to me, I would just change a bit the way we can control the processing

accepted/00000-coco-attestation.md
rjmateus: That also makes sense Abid ☺️
The difference this way is that we are making asynchronous processing and use a container to process it. With an API all should be done online

mcalmer:
> I've got some suggestion regarding tables in the database:
> 
> For example:
> 
> ```
> environment VARCHAR(128) NULL
>                       CONSTRAINT suse_srvcocoatt_rep_env_ck
>                         CHECK(environment IN ('kvm_amd_epyc_milan', 'kvm_amd_epyc_genoa')),
> ```
> 
> Shouldn't be better to create a new table with all the environment instead of having them hardcoded so that, when in the future new environments are added we do not change the schema itself but we add those into that new table? We could reference it a a FK to that new table.
> 
> Not a database expert but the idea came to my mind that might be helpful.

As we discussed in a meeting this will change and we will use a NUMERIC value and hold an enum in Java.
We cannot use the DB enum as this require creating a new type which might not work with database as a service in Clouds.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,245 @@
...
+The attestation is a process of multiple checks which can also depend on the hardware and used environment.
+SUSE Manager should provide a framework for executing such checks, store the results in the database and show the latest and historic results.
+The number of checks should be easy extendable in future versions.
+
+
+# Detailed design
+[design]: #detailed-design
+
+Confidential Compute Attestation cannot be performed on all systems.
+It depends on the used hardware and environment where the instance is running on.
```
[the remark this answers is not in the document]
mcalmer: I fear that we cannot find out if a system cab support it or not.
Or better only in a few cases we can find out.
But even a check for SLE15 SP6 is running as OS, is not a really good check.
We need `snpguest` tool, and I was able to compile it for SP5 . So if a customer provide this tool in a custom channel, also SLE 15 SP5 can be attested.

Checking the hardware is nearly impossible as KVM show a different string as CPU than the real one in the host.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,245 @@
...
+overall status of the report should also show `failed`.
+
+```sql
+CREATE TABLE suseServerCoCoAttestationReport
+(
+        id          NUMERIC NOT NULL
+                      CONSTRAINT suse_srvcocoatt_rep_id_pk PRIMARY KEY,
+        server_id   NUMERIC NOT NULL
+                      CONSTRAINT suse_srvcocoatt_rep_sid_fk REFERENCES rhnServer (id),
+        environment VARCHAR(128) NULL
```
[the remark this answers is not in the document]
mcalmer: It will become an NUMERIC value with an enum implementation in java.
ENUM in DB require to create a type which might be a problem in "database as a service" in some clouds.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,245 @@
...
+
+```sql
+CREATE TABLE suseServerCoCoAttestationResult
+(
+        id          NUMERIC NOT NULL
+                      CONSTRAINT suse_srvcocoatt_res_id_pk PRIMARY KEY,
+        report_id   NUMERIC NOT NULL
+                      CONSTRAINT suse_srvcocoatt_res_rid_fk
+                        REFERENCES suseServerCoCoAttestationReport (id),
+        result_type VARCHAR(32) NOT NULL
```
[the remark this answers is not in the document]
mcalmer: yes, solution will be the same

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,245 @@
...
+```
+
+The result table has the result for every check performed on the report.
+This is a many-to-one relation as many checks can be performed for one report.
+Every result has a type, status and summary which can be shown in a user interface.
+Optional it can have details if applicable.
+
+
+```sql
+CREATE TABLE suseServerCoCoAttestationResult
```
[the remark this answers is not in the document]
mcalmer: The problem is, that only the attestation container knows what checks should be performed.
My idea is, that the java code does not know about all of this. In best case a new environment or check would just result in a change in the state and the container. No changes to Java.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,245 @@
...
+It is made for AMD EPYC CPUs of type Milan and Genoa.
+The host and the hypervisor must support Confidential VMs to perform a successful attestation.
+
+The tool and the possible attestation workflows are described in the [README](https://github.com/virtee/snpguest/blob/main/README.md).
+
+The proposed workflow for attestation using SUSE Manager is:
+
+![SNP Guest Attestation](images/coco-attestation/CoCoAttestation.png)
+
+- Apply a salt state on the VM
```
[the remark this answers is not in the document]
mcalmer: asynchronous as all actions.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,245 @@
...
+  - send a nonce file
+  - request the attestation report
+- Store the data into the database
+  - Status: pending
+  - Store nonce and report data
+  - assign to a systemid and an execution date (create date of the report)
+- Run a Container which does the attestation
```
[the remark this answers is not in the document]
mcalmer: It was planned that a container should do all the check for 1 environment. 
But if 2 checks in one environment require quite different software it could make sense to have containers per check which needs to be performed.
But that would require that the java code knows about which environment requires which check.
Or we define this in a database table. I will think about this solution.

accepted/00000-coco-attestation.md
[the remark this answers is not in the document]
mcalmer: I will add a sentence which say that it is not possible to detect it

accepted/00000-coco-attestation.md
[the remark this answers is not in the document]
mcalmer: I will changed the RFC to have attestation containers for 1 or multiple result types instead of environments.
The mapping between environment types and required result types needs to be defined. Current suggestion is to define this in Java as it also need to create these entries.

accepted/00000-coco-attestation.md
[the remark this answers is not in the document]
mcalmer: I will change the RFC to follow this new way.

accepted/00000-coco-attestation.md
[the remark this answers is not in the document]
rjmateus: thank you for the explanation

rjmateus:
>As we discussed in a meeting this will change and we will use a NUMERIC value and hold an enum in Java.
We cannot use the DB enum as this require creating a new type which might not work with database as a service in Clouds.

@mcalmer in this regard, we can also have a "catalog" table in the database, where we could save the ID and the label (similar to what we have for architectures). Would that make sense?

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,245 @@
...
+                      CONSTRAINT suse_srvcocoatt_rep_id_pk PRIMARY KEY,
+        server_id   NUMERIC NOT NULL
+                      CONSTRAINT suse_srvcocoatt_rep_sid_fk REFERENCES rhnServer (id),
+        environment VARCHAR(128) NULL
+                      CONSTRAINT suse_srvcocoatt_rep_env_ck
+                        CHECK(environment IN ('kvm_amd_epyc_milan', 'kvm_amd_epyc_genoa')),
+        status      VARCHAR(32) NOT NULL
+                      CONSTRAINT suse_srvcocoatt_rep_st_ck
+                        CHECK(status IN ('pending', 'succeeded', 'failed')),
+        data        JSONB NOT NULL, -- machine readable data
```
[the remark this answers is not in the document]
mcalmer: I have changed it and we have now an `in_data` and `out_data`.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+- Apply a salt state on the VM
+  - create the entry on the report table with status `pending`
+  - assign to a systemid and an execution date (create date of the report)
+  - add the nonce to the entry in `in_data` and set it as pillar data
+  - request the attestation report
+- Store the state.apply result data into the database
+  - Store report data into `out_data`
+  - check the expected validation results and add all the entries into the result table with status `pending`
+- Run a Container which does the attestation
+  - search the DB for pending attestation
```
rjmateus, line 191:
```suggestion
  - search the DB for pending attestation of the container supported result types
```

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+A container can be enhanced and act for more result types if no other tooling is needed and to build another container would be overkill.
+
+Therefor the Container should contain the just the `snpguest` tool and the CA Certificates for the different Processor Models (ARK and ASK for Milan and Genoa).
+In case the VCEK is available, the verification can directly start and the container only needs access to the database.
+If the VCEK is not available, the container needs to fetch it from the AMD Key Distribution Center.
+This require internet access for the attestation container.
+
+The program which start the verification should regular check the DB for pending attestations.
+It should fork a process for a specific report result.
+It can use DB locking mechanisms to exclusively work on a specific report.
```
rjmateus, line 221: Should it lock the report or the results it can proccess?

rjmateus (commented):
Second round of review. Looks a great solution, and I just have a few questions.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+  - Azure Attestation
+  - etc.
+
+The information about the environment result in pillar data which help to select the matching states which should be executed on the minion for this task.
+
+The attestation is a 2 step process.
+First collect the data and second to verify them and write the result.
+
+### Step One
+The first step execute a state on the target system to collect the report data.
```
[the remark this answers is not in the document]
admd: Considering this would be an async call, it does make sense to have this additional status.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+# Summary
+[summary]: #summary
+
+SUSE Manager / Uyuni should help with Confidential Compute Attestation and being able to generate a report page for systems which should run in this mode.
+
+# Motivation
+[motivation]: #motivation
+
+More and more systems run as virtual machines or (public) cloud instances.
+Customers want to be sure that data of these VMs cannot be read by administrators or the cloud provider.
```
admd, line 13:
```suggestion
Customers want to be sure that the data of these VMs, which is loaded in memory, cannot be read by administrators or the cloud provider.
```

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+# Detailed design
+[design]: #detailed-design
+
+Confidential Compute Attestation cannot be performed on all systems.
+It depends on the used hardware and environment where the instance is running on.
+In most cases it is not possible to detect if an attestation can be performed or not.
+
+This requires a per system configuration which should contain at least two configuration options:
+
+- Enable/Disable Confidential Compute Attestation (Checkbox)
```
admd, line 34: If attestation can't be performed on a system, will we check that part at this time when user try to enable this checkbox or will that happen later? And if attestation is possible , Ideally I would only then mark this checkbox as enabled. But I am not sure how feasible is it to implement that way.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+- Environment selection
+  - KVM Guest Attestation on AMD EPYC Milan CPUs
+  - KVM Guest Attestation on AMD EPYC Genoa CPUs
+  - Azure Attestation
+  - etc.
```
admd, line 39: Should this depend on the previous configuration option and only comes into affect if attestation could be performed ? And will the appropriated selection be done automatically by SUSE Manager based on some criteria or will this be left to the user to select the correct Environment?

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+Also the expected mappings which environment type require which result type will life in Java.
+
+
+### Step two
+The second step is a process which run in a dedicated environment like a container which verify the result.
+
+Such an attestation container is responsible for one or more result types.
+The attestation container search regular for pending attestations of its result type.
+It read the data from the report table, performs all the needed checks and update the result table with the outcome
+and set the status to `succeeded` or `failed`.
```
admd, line 75: If some checks are failed and other are passed, would be great if we could also show this information to the user somewhere.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+We have 2 data columns. The `in_data` column is used for input data we need to provide via
+state.apply to the VM. The input data should be added as pillar data for that minion.
+
+The `out_data` column contains the results coming back from the state.apply.
+These data columns are of type JSONB to hold any kind of data in machine readable form.
+As different environment types perform different checks, multiple data structures are expected.
+
+The status is calculated from the different results. It should be set to `succeeded` when all
+results have the status `succeeded` as well. As soon as one result is set to `failed` the
+overall status of the report should also show `failed`.
```
admd, line 109: Same as above. Yes, overall status will be failed but still would be good show what tests were and what failed.

admd (commented):
I took another look. I have just left a couple of questions but rest looks good.

accepted/00000-coco-attestation.md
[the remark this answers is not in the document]
mcalmer: Remember that we also have an action. This hold that state. I think "pending" is enough here.
But we may need a pointer to the action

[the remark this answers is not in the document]
rjmateus: We need something to know that the report/Result is ready to be processed. It can be a new state in the report table or a pointer to the action table. 
The result attestation container(s) needs to know the action is already processed and data should be available.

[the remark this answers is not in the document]
mcalmer: The idea is, that we generate the "result" entries with status `pending` and the containers would just look on pending entries in the result table, not in the report table.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+
+```sql
+CREATE TABLE suseServerCoCoAttestationReport
+(
+        id          NUMERIC     NOT NULL
+                      CONSTRAINT suse_srvcocoatt_rep_id_pk PRIMARY KEY,
+        server_id   NUMERIC     NOT NULL
+                      CONSTRAINT suse_srvcocoatt_rep_sid_fk REFERENCES rhnServer (id),
+        env_type    NUMERIC     NOT NULL,
```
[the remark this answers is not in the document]
mcalmer: It is the same, but used differently. So adding a foreign key sounds wrong.
The mapping to some labels should happen only in Java and we would use the number always directly in queries.
So I do not see performance problems.
Maybe I just do not understand what you mean?

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+        id          NUMERIC NOT NULL
+                      CONSTRAINT suse_cocoatt_res_id_pk PRIMARY KEY,
+        report_id   NUMERIC NOT NULL
+                      CONSTRAINT suse_cocoatt_res_rid_fk
+                        REFERENCES suseServerCoCoAttestationReport (id),
+        result_type NUMERIC     NOT NULL,
+        status      VARCHAR(32) NOT NULL
+                      CONSTRAINT suse_cocoatt_res_st_ck
+                        CHECK(status IN ('pending', 'succeeded', 'failed')),
+        summary     VARCHAR(256) NOT NULL,
```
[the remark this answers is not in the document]
mcalmer: Maybe rename it to "title" or "description" ?
When developing a UI, and I have a checklist , this text should be used to describe what test was executed.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+![SNP Guest Attestation](images/coco-attestation/CoCoAttestation.png)
+
+- Apply a salt state on the VM
+  - create the entry on the report table with status `pending`
+  - assign to a systemid and an execution date (create date of the report)
+  - add the nonce to the entry in `in_data` and set it as pillar data
+  - request the attestation report
+- Store the state.apply result data into the database
+  - Store report data into `out_data`
+  - check the expected validation results and add all the entries into the result table with status `pending`
```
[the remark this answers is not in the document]
mcalmer: Currently there was not plan to let the user decide which check is important.
We would perform all the check "we" define for a special type.
What the user is doing with the results is up to him.

We could add this to the configuration.
After the user selected the environment type, we know the numbers of checks which could be executed for this type.
Now the user could mark some of them as "skip" and we could create the result directly with this status. 

I will add this as a possible extension.

accepted/00000-coco-attestation.md
```diff
@@ -0,0 +1,305 @@
...
+- Run a Container which does the attestation
+  - search the DB for pending attestation
+  - try to lock row (the job)
```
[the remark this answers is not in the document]
mcalmer: The idea is, to lock the row in the result table. The containers do not write in the report table. They just read from it.
Changing the status in the report table is done by a trigger. So this is independent from the containers

accepted/00000-coco-attestation.md
mcalmer: yes, the text needs to be adapted

accepted/00000-coco-attestation.md
mcalmer: Currently it seems that we have no chance to find out if an attestation can be done or not.
The administrator need to know that.

accepted/00000-coco-attestation.md
mcalmer: The Enable/Disable checkbox would show/hide this option.
You can only select an environment, when you want to perform an attestation.

As we cannot find out if this can be successful or not, the admin tell us what environment it is.

accepted/00000-coco-attestation.md
mcalmer: This is a general "do it or not" decision for the the admin.

accepted/00000-coco-attestation.md
mcalmer: The plan is to have a details page with a checklist.
This list show for every check if it was `succeeded` or `failed`.
The text behind the checkbox is the value `summary` from the result table. (maybe renamed to `title` or `description`

accepted/00000-coco-attestation.md
mcalmer: I explained the details view in the UI a bit better

accepted/00000-coco-attestation.md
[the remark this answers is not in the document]
rjmateus: Ok got your point, it's not needed to have it. My point about the performance was if we didn't add it to this table, and if we needed the value, we would load it from the table `suseServerCoCoAttestationConfig`.

accepted/00000-coco-attestation.md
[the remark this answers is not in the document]
rjmateus: Adding the "selected checks" to the config table looks as a good extension, from my point of view. thank you

accepted/00000-coco-attestation.md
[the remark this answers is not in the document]
mcalmer: The plan is to create them when salt return and we evaluate the result. 

`At the end of step one all the expected result checks are created` is part of this section now

admd (approved):
Thank you Michael. This looks good.

rjmateus (approved):
LGTM. Thank you Micheal

mikeletux (approved):

PR #87, opened 2024-02-20
Native AppStream support

cbbayburt:
This RFC proposes native support for AppStreams repositories in Uyuni.

See the [rendered version](https://github.com/uyuni-project/uyuni-rfc/blob/native-appstreams/accepted/00101-native-appstreams.md).


accepted/00101-native-appstreams.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Overview
+
+The design encompasses core functionality, new UI operations, modifications to the existing CLM feature, and testing infrastructure.
+
+## The core design
+
+The core design involves three key components for data collection and filtering of packages for client consuption:
+
+1. **Repo-sync**: A standalone Python module that extracts module metadata from source channels using `dnf`'s `libmodulemd` library during repo-sync, serving as the source of truth for the module information. As the data gathered during this process is critical to the feature, it must be executed as a single unit with the repo-sync ([See the PoC](https://github.com/cbbayburt/uyuni/commit/ed9391e8c6e0a66d1dd7cb0f3501332b0884f2f3)).
```
agraul, line 27: Can you expand on "as a single unit"? That sounds a bit like a transaction (succeed together or fail together), which we don't have in reposync.

The PoC script, as far as I can tell, works on already synced, imported, and linked packages. Is that correct? What are the downsides of doing it this way vs. doing it at a different time in the process?

accepted/00101-native-appstreams.md
```diff
@@ -0,0 +1,75 @@
...
+3. *Optional:* Run `mgr-libmod` for dependency resolution with real-time feedback.
+4. Schedule an Uyuni action to apply the selection on the client.
+5. The action calls `dnf` with the selection of module streams on the client via Salt.
+6. On job return, update the selection of modules in the database.
+
+To allow working on multiple clients at once, SSM will be used. In the SSM, the clients in the set will be first grouped by the modular channels they are assigned to. Then the user will select a group and makes a selection of modules that applies to the whole group.
+
+## Content lifecycle management
+
+The current CLM approach will be replaced with regular *ALLOW/DENY* logic in AppStream filters. During build, module filters will be translated to package filters that apply to all packages of the module. A module will only be included in the target repository if all its packages are visible.
```
rjmateus, line 58: This means we would not be able to filter packages and is version inside a module? Would not make sense to have the module available if at least one package from a module is present?
This looks to be to restrictive to me.

accepted/00101-native-appstreams.md
```diff
@@ -0,0 +1,75 @@
...
+A "dummy" modular repository will be built in OBS for comprehensive testing in the CI environment.
+
+Below is an example of a minimal test repository:
+
+![Test repository](images/00101-native-appstreams-test-repo.png)
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Migration plan for the typical user
```
Etheryte, line 75: While I know this is in the unresolved questions section, are we roughly sure we can migrate everything we currently support in a reasonable way?

accepted/00101-native-appstreams.md
cbbayburt: The new data we collect at this step is critical for modular repositories to work. If we somehow miss this data, Uyuni won't know if a package is really modular or not, and will treat it as a regular package. As a result, Uyuni will suggest wrong updates to existing packages (the initial problem we have with modularity).

Because of this, we have to make sure that module metadata is properly processed when syncing a repository. If not, the reposync for this channel should fail completely.

How do we currently make sure that the repos don't end up in an invalid state during reposync?

accepted/00101-native-appstreams.md
cbbayburt: The new approach is fundamentally different so we won't be able to offer an automated migration at all. So what I mean here is the migration of the customer's workflow.

We just need to make sure that the customers have a smooth way of switching without any disruptions with proper documentation. But in the end, the change is really just getting rid of many extra steps to achieve the same goal, so I'm sure the customers would be happy to go through the labor of migration.

accepted/00101-native-appstreams.md
cbbayburt: > How do we currently make sure that the repos don't end up in an invalid state during reposync?

It seems we do it using proper exception handling and DB transactions.

accepted/00101-native-appstreams.md
admd: While I acknowledge that customers may welcome the transition to the new approach, we cannot compel them to do so by breaking stuff; it should be done at their convenience. In other words, we must ensure that their existing clm project setups remain functional. 

Moreover, if it helps, we could consider applying this change solely to new CLM projects, potentially easing concerns regarding existing setups.

accepted/00101-native-appstreams.md
```diff
@@ -0,0 +1,75 @@
...
+
+The design encompasses core functionality, new UI operations, modifications to the existing CLM feature, and testing infrastructure.
+
+## The core design
+
+The core design involves three key components for data collection and filtering of packages for client consuption:
+
+1. **Repo-sync**: A standalone Python module that extracts module metadata from source channels using `dnf`'s `libmodulemd` library during repo-sync, serving as the source of truth for the module information. As the data gathered during this process is critical to the feature, it must be executed as a single unit with the repo-sync ([See the PoC](https://github.com/cbbayburt/uyuni/commit/ed9391e8c6e0a66d1dd7cb0f3501332b0884f2f3)).
+
+2. **Package profile update**: A Salt state that retrieves current module information from clients, storing the data in the database. The state calls `dnf module` commands to retrieve the list of enabled module streams together with their *name*, *stream*, *version*, *context*, and *arch* values ([see the PoC](https://github.com/cbbayburt/uyuni/commit/2c788f3144f5bfe8ddd904045e0a757a7a432923)).
```
admd, line 29: Can, I'm not fully informed on this topic. So please excuse my ignorance.

Could you clarify how we currently installing/removing/updating packages—are we using yum or dnf ?

If we're using yum, will we stick with it for package operations, or are we considering transitioning to DNF?

cbbayburt: Nowadays we're using the terms `yum` and `dnf` interchangibly, but since RH8, `yum` is just a wrapper around `dnf`. We still have a lot of mentions of `yum` in our python modules, but Salt is actually using `dnf`.

accepted/00101-native-appstreams.md
```diff
@@ -0,0 +1,75 @@
...
+
+## New UI operations
+
+The proposed UI includes a new page under the `System > Software` tab, visible only for clients with modular channels assigned. It displays available modules and their streams, allowing users to select and apply module changes to clients.
+
+The main workflow is as follows:
+
+1. Display modules and their available streams from all the assigned modular channels
+2. The user selects module streams to be enabled on the client
+3. *Optional:* Run `mgr-libmod` for dependency resolution with real-time feedback.
```
admd, line 49: Again, I lack knowledge here so another question. What kind of dependency issue would you expect? Do streams depend on other streams or are they self-contained?

Ideally, we should have mechanism in place, so user doesn't get into bad state. Just like we do with required/recommended channels. Pre-select the needed all the needed stuff when user select a module stream. But if that's asking for too much, this optional would be super helpful.

accepted/00101-native-appstreams.md
admd: thank you! So it's safe to say that for RHE and clones, we will be only DNF at the end. Thank you Can.

accepted/00101-native-appstreams.md
cbbayburt: Yes, a module may depend on other modules and they must be enabled together. If we miss this, dnf would complain about the missing dependencies. Other than that, mgr-libmod also checks for conflicting streams. If, for example the user tries to enable two streams of the same module, mgr-libmod would complain about this too.

accepted/00101-native-appstreams.md
cbbayburt: Sure, it's a bit tricky because we need to replace the old AppStream filters with the new functionality, but at least we can implement it so their existing projects won't be affected until they want to rebuild them.

accepted/00101-native-appstreams.md
rjmateus: I was thinking of scenarios where some packages in the module could be optional, and the user decided to exclude them in a filter. In this scenario, the module would never get available. Do you think a scenario like this one makes sense?

accepted/00101-native-appstreams.md
rjmateus: Would it be possible to implement it in a way that allows uses to continue using the flatten method and this one if they want?
Some customers may be familiar with the old one and may want to continue using it. Abid do you think such a case could exist?

cbbayburt: Flattening is only possible with the old-style AppStream filters. If we include both old and the new filters, it would be too confusing for users, and also they will keep using the old way for a longer time. That means more L3s and harder maintenance for us so I wouldn't recommend it.

accepted/00101-native-appstreams.md
cbbayburt: Yes, that's done by "Profiles" in AppStreams. In the example above, `postgresql` module usually has at least 2 profiles, one named `client` and one named `server`. Depending on which profile is activated, the list of actual installed packages changes. But this needs careful consideration because the user might want to enable different profiles for different clients from the same repository.

So maybe the best way is to give the user the freedom indeed. If `dnf` fails because of some missing package, then it's up to them to fix it in CLM.

accepted/00101-native-appstreams.md
rjmateus: Got it, and sounds good to me. :+1:

accepted/00101-native-appstreams.md
```diff
@@ -0,0 +1,75 @@
...
+
+1. **Repo-sync**: A standalone Python module that extracts module metadata from source channels using `dnf`'s `libmodulemd` library during repo-sync, serving as the source of truth for the module information. As the data gathered during this process is critical to the feature, it must be executed as a single unit with the repo-sync ([See the PoC](https://github.com/cbbayburt/uyuni/commit/ed9391e8c6e0a66d1dd7cb0f3501332b0884f2f3)).
+
+2. **Package profile update**: A Salt state that retrieves current module information from clients, storing the data in the database. The state calls `dnf module` commands to retrieve the list of enabled module streams together with their *name*, *stream*, *version*, *context*, and *arch* values ([see the PoC](https://github.com/cbbayburt/uyuni/commit/2c788f3144f5bfe8ddd904045e0a757a7a432923)).
+
+3. **Package cache**: Incorporates data from repo-sync and package profile update into existing queries, ensuring proper filtering of non-enabled modular packages. The existing tables, queries and stored procedures will be updated with additional joins and proper indexing to ensure minimum performance impact.
+
```
mcalmer, line 32: There is a 4th point needed. You need to generate new metadata for the Channels including new compiled modules.yaml from the database informations.

accepted/00101-native-appstreams.md
mcalmer: Maybe the module.yaml should be parsed first. I think during package import we call multiple times "commit" to solve problems with dependencies between packages which gets imported in parallel.

accepted/00101-native-appstreams.md
mcalmer: Possible problem: what happens if a customer want to allow only security updates and apply a filter like this?
Such an update do not match all all packages in a steam. We need to check what prevents to remove the whole stream incl. the security update.

accepted/00101-native-appstreams.md
cbbayburt: We have to do it in the end because we need package IDs for the relation.

accepted/00101-native-appstreams.md
cbbayburt: I also need to check how DNF behaves if some packages are missing in a module.

PR #89, opened 2024-04-12
Make Auditing CVEs possible

gabrielstein:
Make possible to audit CVEs per Hosts
Make possible to download PDF Reports sorted per CVE Numbers
Make possible to call reports via REST-API
Make possible to commit reports to a private git repository(so it would be possible to access historical data and make easy to understand the chronology from the cve patching)

### DIFF
--- a/accepted/00101-make_cve_reports_possible.md
+++ b/accepted/00101-make_cve_reports_possible.md
@@ -1,0 +1,60 @@
+- Feature Name: make_cve_reports_possible
+- Start Date: 2024-04-12 
+
+# Summary
+[summary]: #summary
+
+Create a meaninful CVE Audit that would ran as a report and it would be sorted per CVE-Number
+
+# Motivation
+[motivation]: #motivation
+
+- Why are we doing this? The CVE Audit is static and does not provide an overview from the hosts affected by a CVE.
+- What use cases does it support? Produce real-time reports sorted by CVE-Number for the auditors and security teams
+- What is the expected outcome? Have a table with the CVE Numbers and what hosts are affected. Have a report that could be called via REST-API and also could be integrated on every CI/CD workflows from operational teams.
+
+We have a static and not a understable overview from the hosts affected by a CVE. As today, one should open the CVE Audit option on the menu, search for a CVE number, add to the form and expect a list from hosts affected by only this CVE. If there is a higher number from CVEs would be a lot of working doing a manual search per CVE Number. This makes the life from infrastructure managers so difficult -it is not possible to have an overview that is always updated and showed in the dashboard -
+and it is also not possible to have a generated report per PDF from the hosts affected by a CVE(a report sorted per CVE-Number. It should be also provided a way to generate this reports per REST-API and make it possible to have it integrated in a CI/CD Workflow.
+
+# Detailed design
+[design]: #detailed-design
+
+Steps for the implementation:
+- Create a stored procedure/function on the reporting DB making it call on demand by the frontend/backend code.
+- Get some space on the main overview to place the results from this query
+- Create a view using the function and make possible via configs to setup the timeout and how often it would be called
+- Place the view results on the overview dashboard
+
+    Visual examples:
+
+    CVE-Number         How to fix                      Report Date: 12.04.2024 03:19PM
+    ----------------------------------------------------------------------------------
+    CVE-123456         
+       host1           Install update-123
+       host2           Install the latest fix
+       host3           Fixes are not delivered anymore 
+                                                                  Download cve-report-12.04.2024-03:19pm.pdf
+
+
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we **not** do this?
+
+  * If the timeout is not set it could case some data issues as congestion
+  * If the CVE Database is not updated(network issues, disk space, permissions) it could cause false positives
+  * Database
+  * No, because I would like to have develop it as a module - independent of other components and documented on how to remove it if needed
+
+# Alternatives
+[alternatives]: #alternatives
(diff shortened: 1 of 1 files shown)


accepted/00101-make_cve_reports_possible.md
```diff
@@ -0,0 +1,60 @@
...
+- What is the expected outcome? Have a table with the CVE Numbers and what hosts are affected. Have a report that could be called via REST-API and also could be integrated on every CI/CD workflows from operational teams.
+
+We have a static and not a understable overview from the hosts affected by a CVE. As today, one should open the CVE Audit option on the menu, search for a CVE number, add to the form and expect a list from hosts affected by only this CVE. If there is a higher number from CVEs would be a lot of working doing a manual search per CVE Number. This makes the life from infrastructure managers so difficult -it is not possible to have an overview that is always updated and showed in the dashboard -
+and it is also not possible to have a generated report per PDF from the hosts affected by a CVE(a report sorted per CVE-Number. It should be also provided a way to generate this reports per REST-API and make it possible to have it integrated in a CI/CD Workflow.
+
+# Detailed design
+[design]: #detailed-design
+
+Steps for the implementation:
+- Create a stored procedure/function on the reporting DB making it call on demand by the frontend/backend code.
```
rjmateus, line 23: We try to avoid store procedures as much as possible. The reason is because they are hard to maintain and debug.
Also, the report DB and "main" DB are separated in different schemas and can also be on different machines in the future.

rjmateus (commented):
Thank you for your contribution. I have one technical question that I have sent.
From the implementation point of view I think this RFC will benefit from the implementation of the OVAL data [1]. That feature will create and keep up to date a CVE table, and will run CVE audits to each minion. So, in this regard part of this will be implemented in there. The missing part will be provide good reporting.

[1] https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00098-cve-auditing-with-oval.md

accepted/00101-make_cve_reports_possible.md
```diff
@@ -0,0 +1,60 @@
...
+Create a meaninful CVE Audit that would ran as a report and it would be sorted per CVE-Number
+
+# Motivation
+[motivation]: #motivation
+
+- Why are we doing this? The CVE Audit is static and does not provide an overview from the hosts affected by a CVE.
+- What use cases does it support? Produce real-time reports sorted by CVE-Number for the auditors and security teams
+- What is the expected outcome? Have a table with the CVE Numbers and what hosts are affected. Have a report that could be called via REST-API and also could be integrated on every CI/CD workflows from operational teams.
+
+We have a static and not a understable overview from the hosts affected by a CVE. As today, one should open the CVE Audit option on the menu, search for a CVE number, add to the form and expect a list from hosts affected by only this CVE. If there is a higher number from CVEs would be a lot of working doing a manual search per CVE Number. This makes the life from infrastructure managers so difficult -it is not possible to have an overview that is always updated and showed in the dashboard -
```
admd, line 16: Can you please elaborate on what's not understandable from the current overview? 

How do you envision it from UI perspective? Available CVE number is huge.

Would you like to identify which system is affected by a particular CVE(s), or if any system is affected by any CVE at all? Are you solely interested in CVEs that affect systems or even in those CVEs which doesn't affect any of  the onboarded system?

Currently, we have limitation that we could only tell if a system is affected by CVE if assigned channels has patch which in turn fix that CVE. Depending on if that patch has been applied or not, we tell if system is affected or not. Based on this, we already show on system overview page if systems has some security updates or all good. 


Now if I summarize and IIUC, what you likely want is the ability to provide a list of CVEs and then receive results indicating whether systems are affected or not based on that list. Please correct me, if that wouldn't suffice.

accepted/00101-make_cve_reports_possible.md
```diff
@@ -0,0 +1,60 @@
...
+and it is also not possible to have a generated report per PDF from the hosts affected by a CVE(a report sorted per CVE-Number. It should be also provided a way to generate this reports per REST-API and make it possible to have it integrated in a CI/CD Workflow.
```
admd, line 17: We provide CSVs. Providing repost as PDFs is not something that we are gonna tackle for now.

accepted/00101-make_cve_reports_possible.md
gabrielstein: ok :)

gabrielstein: I would check the other RFC and I think that we can combine the work here.

accepted/00101-make_cve_reports_possible.md
gabrielstein: Hi @admd !

Can you please elaborate on what's not understandable from the current overview?
It is everything static. How a admin will know if there is a new CVE affecting their servers and what CVE number it is? I did a lot of demos for customers showing the CVE Audit and I always had to search for a CVE number and check if one of the hosts is affected. Similar to NeuVector[0] we could use the same databases which they use to do this CVE checks and show it the hosts from Uyuni are affected. 

It is nice that the CVEs are just show when the channel has a patch against it - but would not make more sense to display the real situation and what patches are being delivered? If it is just based on the channels from Uyuni it would give a false positive - well, my channels do not have the patch for CVE 12345, it means that my hosts are not affected. 

[0] - Example how CVE Databases works for NeuVector: https://open-docs.neuvector.com/scanning/updating

accepted/00101-make_cve_reports_possible.md
gabrielstein: @admd If we manage to have a real time dashboard the report in pdf would be easy.

PR #92, opened 2024-10-11
Add 'Improve MicroOS management' RFC

agraul:
Rendered: https://github.com/agraul/uyuni-rfc/blob/microos-management/accepted/0000-uyuni-microos-management.md


accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,142 @@
...
+
+- `virt.*`: About to be dropped
+- `bootloader`: TODO
+- `rebootifneeded` - The way this is written is incompatible with transactional systems
+
+### OS-unchanging operations
+
+These operations do not alter the system, i.e. they don't belong in a (new) snapshot. Uyuni applies them with `state.apply` for two reasons: structured output and no concurrency (`queue=True`).
+
+- `ansible.runplaybook`
```
m-czernek, line 43: Is `ansible.runplaybook` intended for SLM/MicroOS to act as a control node that executes the playbook against other minions? Because the playbook can be executed against the control node itself as well, which would then be OS-altering of course. 

Then again, if we make this a transactional action, that means the playbook files must be present on the transactional filesystem. We should probably make a note somewhere in the doc at least that we don't support Uyuni/SUMA server to alter itself with Ansible.

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,142 @@
...
+# Motivation
+[motivation]: #motivation
+
+MicroOS, or the downstream SLE Micro, is a different operating system than e.g. SLES. It is based on the same tools and packages, with few additions on top. These additions are few, but completely change the operating model.
+
+The main idea is: **All changes go into a new btrfs snapshot. This new snapshot is "pending" until a reboot activates it.**
+
+Until now, we tried to hide the differences in Uyuni and relied on Salt to do the *right thing*. This strategy was easy for us to use, but it did not work well. We need to change the approach we take with transactional systems.
+
+# Detailed design
```
m-czernek, line 18: Should we add info here about the changes that pertain to displaying packages? I understand that we want to also display what packages are in a new transaction, but not yet applied (or vice versa, what packages are removed, but still present in the currently-running snapshot). The details of this would go to a different squad, but they might be in scope for this RFC?

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,142 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+Why should we **not** do this?
+
+  * obscure corner cases
+  * will it impact performance?
+  * what other parts of the product will be affected?
+  * will the solution be hard to maintain in the future?
```
rjmateus, line 130: Is this section complete?

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,142 @@
...
+
+The main idea is: **All changes go into a new btrfs snapshot. This new snapshot is "pending" until a reboot activates it.**
+
+Until now, we tried to hide the differences in Uyuni and relied on Salt to do the *right thing*. This strategy was easy for us to use, but it did not work well. We need to change the approach we take with transactional systems.
+
+# Detailed design
+[design]: #detailed-design
+
+## Uyuni differentiates between OS-unchanging and OS-altering operations
+Instead of using the `transactional_update` executor, Uyuni calls either `state.apply` or `transactional_update.apply`. Uyuni is in full control of which states are applied in a new snapshot and which states are not. The SLS file is the smallest unit Salt can handle, there is no way to apply only parts of an SLS file inside a snapshot. Therefore, we split SLS files that currently mix OS-unchanging and OS-altering operations.
```
rjmateus, line 22: Just to clarify, for internal states (example `channels` state) the decision if we should run it using `state.apply` or `transactional_update.apply` is made by uyuni automatically, and user doesn't need to know anything about it?
How about High State? Should we add the option in the UI state scheduler for the user to select in which mode it should run? If that is the case, we should hide this option for non-transational systems.

accepted/0000-uyuni-microos-management.md
agraul: No, that's still the template.

accepted/0000-uyuni-microos-management.md
agraul: I think that's out of scope for now. I'd like to have it in the future, though

accepted/0000-uyuni-microos-management.md
agraul: Yes, Uyuni maps the internal states (see the mapping below) to either `state.apply` or `transactional_update.apply`. For those that can't be clearly mapped and custom states (which includes highstate of course), the user specifies it.

agraul:
I pushed the second version of the RFC, I think it's clearer now. Please review again and let me know of anything that's still unclear.

There are a few vague parts, especially around the backend implementation. That's on purpose as I don't have a good enough understanding to feel comfortable suggesting how to implement these ideas in Uyuni's Java backend. We can discuss them here and add them to the RFC.

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+calls `state.apply $list_of_sls_files` or `transactional_update.apply $list_of_sls_files`.
+
+All internal states that interact with the live system are applied with `state.apply.`. Internal states
+that change the operating system, e.g. by installing packages, are applied with
+`transactional_update.apply`. Today, many of our SLS files combine installing and using packages. 
+That does not fit the transactional model and we need to split these SLS files.
+
+Custom states give users a lot of flexibility managing their systems, which makes it hard for Uyuni
+to know how to apply custom states. To give some control to the user, we add a new config value:
+`java.salt_custom_states_use_transactional_update = True`. This config value controls whether
```
rjmateus: I'm not sure if this is enough. Customer can also have states that needs to run inside a transation and others that need to run outside. 
If they follow the same approach as we have for the internal states, they could split it into 2 sls files (states), but then don't have the ability to decide at a state (or configuration channel) level. 
How should a customer handle such use-case with just one global variable?

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+This is the mechanism used when users navigate to "Systems > Bootstrapping" in the WebUI.
+
+Uyuni targets the new system with Salt SSH and always uses `state.apply` to apply the `bootstrap`
+state. This is required because Uyuni relies on information collected on the client system to know
+what kind of system it is. This includes finding out if the new system is a transactional system.
+The `bootstrap`state install our Salt Bundle package correctly on both traditionally-managed and
+transactional systems.
+
+Triggering a reboot over Salt SSH has one problem: rebooting before Salt finishes loses the job data
+of the `bootstrap` state. This can be avoided by systemd feature called "Inhibitor Lock", which can
```
rjmateus: nice, I was not aware of this  "Inhibitor Lock"

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+delay a shutdown request while Salt SSH runs.
+
+#### Add Inhibitor Lock to Salt SSH
+Applications can set _inhibitor locks_ to block or delay system shutdown and sleep states. Salt SSH
+will be changed to set a _delay_ inhibitor lock to stop the system from rebooting immediately. Salt
+SSH has time to return job results back to the Salt Master, unless it takes longer than
+_InhibitDelayMaxSecs_. This config setting is specified in `logind.conf(5)` and can't be overridden
+by Salt SSH. The default is 5 seconds. When Salt SSH requires more than 5 seconds to return the job
```
rjmateus: I think it is a fair compromise. In theory reboot should be the last thing, so it should just be the time of retuning the event.

rjmateus (commented):
For the single state execution it looks good to me, just the custom states configuration and the high state execution.

I'm missing in this RFC is how a full high state will work? For the high state in a transactional system do we call the state.apply or the transactional-state.apply.
The high state also includes the custom states/configuration channels. How it will interact with flag `java.salt_custom_states_use_transactional_update`?

accepted/0000-uyuni-microos-management.md
admd: Agree! A config that applies globally wouldn't fly, we need to have a flexible way where the user can choose  between `state.apply` & `transactional_update.apply`, before applying a custom state.

admd:
> The high state also includes the custom states/configuration channels.

@rjmateus @agraul, this is more of a question as I lack some context. Do we really need to always make configuration channels part of high state? I personally don't like it but then I might be overlooking something there.

I would rather also evaluate the current state of highstate and rather keep it minimal. We do have recurrent states now so if it's rather safe to exclude config chnanles from high state, we should.

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+```
+
+### Internal States Application
+The Java backend maintains a list of SLS files in the `SaltParameters` class. This catalogue will be
+enhanced with a hard-coded mapping of SLS files to the respective Salt function (`state.apply` or
+`transactional_update.apply`).
+
+Some states need to be split into two SLS files, one for prerequisites and a second for the actual
+work. The Java backend code first trigger the prerequisites state. When that job returns
+successfully, Uyuni reacts with the "actual work" state without any further user interaction.
```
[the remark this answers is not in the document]
agraul: Activating a transaction cleanly is done via a reboot.

There is the `transactional-update apply` command that can activate a transaction by re-mounting parts of the file system. That is safe to do when all we do is add new software. It's not a great thing to do when the new transaction contains updates where files of running processes will be replaced.

We can look into making the `transactional_update` execution module smarter. When `activate_transaction` is set, it could look into the snapshot diff and trigger `transactional-update apply` or a reboot with something similar to
```sh
snapper status "${n1}..${n2}" | 
    awk '$2 ~ /\/usr\/lib\/sysimage\/rpm\/(Index|Packages).db/ {next};
         $1 !~ /^[+.]+/ {exit 1}'
```

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+- `util.mgr_switch_to_venv_minion`
+
+
+### Unsupported Internal States
+Not all internal states we have make sense on MicroOS / SUSE Linux Micro. These states are unsupported.
+
+- `appstreams.configure` - only useful for RHEL systems
+- `bootstrap.remove_traditional_stack` - traditional stack was never supported on these systems
+- `cocoattest.requestdata`- only supports SLES 15 SP6
+- `rebootifneeded` - The way this is written is incompatible with transactional systems
```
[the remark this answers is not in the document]
agraul: What's the use case for that on a transactional system? Or asked differently, where do we want to use it?

admd:
@cbbayburt @parlt91 @wweellddeerr I have added you as reviewer too as some part will go to Naica as well.

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+
+#### Request a reboot
+With a delay inhibitor lock taken, the `bootstrap` state requests a reboot from systemd. Systemd
+will accept the request immediately but delay the shutdown until the Salt SSH execution terminates
+and releases the inhibitor lock.
+
+### Bootstrap script
+The bootstrap bash script already handles rebooting the system after installing the Salt Bundle.
+
+### "Salt-initiated" bootstrap
```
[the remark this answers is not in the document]
agraul: It's mostly informational. I don't want us to forget the best practice for large-scale roll outs of new systems. As developers with our small environments we sometimes think the WebUI bootstrap method is the default

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+Why should we **not** do this?
+* More work for us maintaining SLS files with the new layout, as we need to think were to put different states
+* We don't know if all features need to work for transactional systems since they are managed differently.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Full control over custom states for the user. Requires changes to the WebUI / API and database
+  schema. This is something we could do later if the global config approach is not enough.
+- Support for two system states ("live" and "next") in the database, WebUI and API. This would allow
```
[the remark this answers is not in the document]
agraul: It would be relatively cheap to get the list of snapshots. `snapper` is fairly responsive for that, and it's already implemented.
<details><summary>snapper.list_snapshots</summary>
<p>

```
uyuni-server:/ # time salt 51-slmicro60.tf.local snapper.list_snapshots
51-slmicro60.tf.local:
    |_
      ----------
      cleanup:
      description:
          current
      id:
          0
      timestamp:
          1768489673
      type:
          single
      user:
          root
      userdata:
          ----------
    |_
      ----------
      cleanup:
      description:
          first root filesystem
      id:
          1
      timestamp:
          1768409855
      type:
          single
      user:
          root
      userdata:
          ----------
    |_
      ----------
      cleanup:
          number
      description:
          Snapshot Update of #1
      id:
          2
      timestamp:
          1768485478
      type:
          single
      user:
          root
      userdata:
          ----------
    |_
      ----------
      cleanup:
          number
      description:
          Snapshot Update of #2
      id:
          3
      timestamp:
          1768485513
      type:
          single
      user:
          root
      userdata:
          ----------
    |_
      ----------
      cleanup:
          number
      description:
          Snapshot Update of #3
      id:
          4
      timestamp:
          1768486092
      type:
          single
      user:
          root
      userdata:
          ----------
          transactional-update-in-progress:
              yes
    |_
      ----------
      cleanup:
          number
      description:
          Snapshot Update of #3
      id:
          5
      timestamp:
          1768489129
      type:
          single
      user:
          root
      userdata:
          ----------
    |_
      ----------
      cleanup:
          number
      description:
          Snapshot Update of #5
      id:
          6
      timestamp:
          1768489403
      type:
          single
      user:
          root
      userdata:
          ----------
          transactional-update-in-progress:
              yes

real	0m0.374s
user	0m0.204s
sys	0m0.055s
```

</p>
</details> 

The Salt module uses the Snapper D-Bus API, which apparently does not show the active/default snapshot indication that `snapper list` does. There are two D-Bus methods we can implement in the `snapper` module:
```
method GetDefaultSnapshot config-name -> bool number
method GetActiveSnapshot config-name -> bool number
```

With that we could replicate `snapper list` in the WebUI.

It will be harder to show what changed between snapshots. The computation is slow and we might need multiple API calls. Just listing changed files and the type of change for < 15 files can easily take 2.5s. This would need to happen in the background, similar to a hardwareprofile/packageprofile refresh.

<details><summary>snapper.status</summary>
<p>

```
uyuni-server:/ # time salt 51-slmicro60.tf.local snapper.status num_pre=5 num_post=6
51-slmicro60.tf.local:
    ----------
    /discardIfNoChange:
        ----------
        status:
            - created
    /etc/.updated:
        ----------
        status:
            - modified
    /etc/fstab:
        ----------
        status:
            - modified
    /etc/resolv.conf:
        ----------
        status:
            - modified
    /etc/selinux/targeted/contexts/files/file_contexts:
        ----------
        status:
            - extended attributes changed
    /etc/selinux/targeted/contexts/files/file_contexts.bin:
        ----------
        status:
            - extended attributes changed
    /etc/selinux/targeted/contexts/files/file_contexts.homedirs:
        ----------
        status:
            - extended attributes changed
    /etc/selinux/targeted/contexts/files/file_contexts.homedirs.bin:
        ----------
        status:
            - extended attributes changed
    /etc/selinux/targeted/policy/policy.33:
        ----------
        status:
            - extended attributes changed
    /etc/selinux/targeted/seusers:
        ----------
        status:
            - extended attributes changed
    /var/lib/rpm:
        ----------
        status:
            - created
    /var/run:
        ----------
        status:
            - created
    /var/spool:
        ----------
        status:
            - created

real	0m2.511s
user	0m0.237s
sys	0m0.056s
```
</p>
</details>

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+[drawbacks]: #drawbacks
+
+Why should we **not** do this?
+* More work for us maintaining SLS files with the new layout, as we need to think were to put different states
+* We don't know if all features need to work for transactional systems since they are managed differently.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Full control over custom states for the user. Requires changes to the WebUI / API and database
```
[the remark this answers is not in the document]
agraul: There are mainly two reasons why I put this in the "for later" bucket:
1. My takeaway from the discussion at the planning meeting was that we should start small and improve incrementally. Transactional systems are a bit different and might not need as many day-to-day changes as traditional deployments.
2. It's an order of magnitude more complex to decide how to apply individual SLS on a per-SLS basis and based on User input. I have some high-level ideas, but the implications for the Java backend and the UI design are outside of my areas of expertise. We'll need input from @uyuni-project/naica and @uyuni-project/java on how we can design fine grained control correctly.

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+### Unsupported Internal States
+Not all internal states we have make sense on MicroOS / SUSE Linux Micro. These states are unsupported.
+
+- `appstreams.configure` - only useful for RHEL systems
+- `bootstrap.remove_traditional_stack` - traditional stack was never supported on these systems
+- `cocoattest.requestdata`- only supports SLES 15 SP6
+- `rebootifneeded` - The way this is written is incompatible with transactional systems
+- `uptodate` - The way this is written is incompatible with transactional systems
+
+### Configurable States (`java.salt_custom_states_use_transactional_update`)
```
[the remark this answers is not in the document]
agraul: Let's continue this in the `#Alternatives` section

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,201 @@
...
+calls `state.apply $list_of_sls_files` or `transactional_update.apply $list_of_sls_files`.
+
+All internal states that interact with the live system are applied with `state.apply.`. Internal states
+that change the operating system, e.g. by installing packages, are applied with
+`transactional_update.apply`. Today, many of our SLS files combine installing and using packages. 
+That does not fit the transactional model and we need to split these SLS files.
+
+Custom states give users a lot of flexibility managing their systems, which makes it hard for Uyuni
+to know how to apply custom states. To give some control to the user, we add a new config value:
+`java.salt_custom_states_use_transactional_update = True`. This config value controls whether
```
agraul: > How should a customer handle such use-case with just one global variable?

The user can choose to have their SLS files applied outside of any transaction and explicitly open a transaction when required. 

Let's keep this discussion in the `#Alternatives` section where I listed more control in this area as a next step.

accepted/0000-uyuni-microos-management.md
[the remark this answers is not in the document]
rjmateus: Customers using micro on edge tend to disable the automatic reboot, and they try to avoid rebooting a machine as much as possible. Would be cool if we could have an action for "reboot if needed" and if a transaction was pending it would reboot, otherwise it would not.

accepted/00108-uyuni-microos-management.md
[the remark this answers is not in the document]
rjmateus: It would be cool to have at least the list. As a super enhancement, we could have a list of changes (could be computed on demand, instead of always have it) and also a "revert to snapshot". But both of those would be out of scope for this RFC. I think they can be added as future enhancements.

accepted/00108-uyuni-microos-management.md
cbbayburt: On top of the config flag, we can introduce a "[ ] Transactional" checkbox for custom states to override the default.

accepted/00108-uyuni-microos-management.md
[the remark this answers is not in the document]
admd: I believe this is a must-have to clearly indicate the state of things.

We should at minimum indicate:

1. Which snapshot is currently active
2. Which snapshot will be active after the next reboot
3. Other available snapshots (rollback to these can be considered optional for now)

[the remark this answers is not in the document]
agraul: @admd is the snapshot number enough for that? We could add a button to diff the snapshots on demand, but again, the execution time is slow

admd:
To kick off the UI design discussion, I put together a few mockups (painfully created via the browser console). These are by no means final or prescriptive; their sole purpose is to initiate discussion. They do not fully align one-to-one with the proposal and include changes based on alternative ideas mentioned in this PR.

**Change that could be made in init.sls file edit view of state channels**
1- 

Screenshot 2026-01-30 at 4 24 39 PM

2- 
Screenshot 2026-01-30 at 4 26 29 PM

**Change that we should make in apply highstate view**


Screenshot 2026-01-30 at 3 32 24 PM

**Change that we should be making in system event page**

When a high state event is clicked, it is expanded with the kind of following information

Screenshot 2026-01-30 at 4 10 49 PM

**Change that we should be making in system overview page**

Screenshot 2026-01-30 at 4 47 08 PM


These mock ups doesn't contain the snapshot pages that we have discussed above.

agraul:
It's a bit hard to discuss pictures without threads, but let me try to leave a few comments nonetheless.

### Pictures 1/2 (SLS view -> "Transactional System Strategy" section):
- To me it looks repetitive to have both a dropdown for "live" vs "staged" and a checkbox for reboot requirement, "staged" implies the need for a reboot
- It could be useful to have a toggle "request reboot as soon as possible" on the sls-level. Then, if any sls that's part of the list of states to apply has this set to true, Uyuni will trigger reboot with e.g. `activate_transaction=True`

### Picture 3 (apply highstate view)
- Two sections, one for "live" and one for "staged" makes sense. It should not be allowed to change the order of these two because a file in `/etc` could be modified in both phases. If the "staged" change happens first, the change to the same file in `/etc` in "live" will be discarded on reboot
- "Post reboot (deferred)": where do we want to draw the line between the "apply highstate view" and action chains?
- How do we handle a system that e.g. rolls-back after "staged" + reboot automatically? Is the "post reboot" state still applied?

### Picture 4 (state event)
- Is this how action-chain events look like? Or do we display them individually?
- When is this event displayed? After the first highstate (`state.apply`) or when everything is done?

### Picture 5 (system overview)
- Do we keep the warning on top of the "System Status" ?
- "System is up to date is wrong" -- this is a conceptual question. It is "up to date" as in "there are no more updates to install". Our DB thinks all the packages are installed.

### Not pictured: Recurring States
- I wonder how we can organize the list of states. Do we need two lists where some states can only be listed in the "live" and the other in the "staged" list of states?

parlt91:
Let me add a couple of remarks for the proposed UI. I am going to follow the same structure as Alex above.

### Picture 1/2:
Same remarks about redundancy of picker and radio buttons.

### Picture 3:
Proposed changes make sense to me.
Except: `Post reboot (deferred)` How is this going to work with action chains? AFAIK action chains still break if there is a reboot in the middle.

### Picture 4:
This looks nice on the surface, but is going to be a nightmare to implement. This is legacy UI. I'd rather avoid to do such an extensive change to this page to support displaying events like this. A redesign of the page in react might even be easier, but still out of scope.
Also I don't see a way how we would present this if the highstate was part of an action chain.

### Picture 5:
We shouldn't display the system as up to date if there's a snapshot pending. This also needs to be reflected on the systems list page. The rest looks reasonable and shouldn't be to hard to add provided we have the required snapshot info available.

Since Alex mentioned recurring actions. We'll have to review the list of internal states for transactional systems and decide how to present it to the user. The changes to the highstate page also need to be applied for the recurring highstate, but since they use the same component there might not be much to do.

Also I have concerns about scope and feasibility in time for 5.2 if we're also planning to add a dedicated snapshots page. Designing / adding new UI takes time and we have a shortage of dedicated frontend engineers.

I cannot say much about the work required to support these changes on the java backend since I lack knowledge about this part of the java codebase. AFAIK this was mostly on orion in the past.

rjmateus:
I was thinking about these options, andit  would make sense to have the ability to have a "next state on a salt state apply"?
I'm thinking about a new approach to action chains, where we could define a sequence of events, that would depend on the success of the previous one.
With that, we would be able to get better support for images 3 and 4, where we would execute the first stated, and add all the sequences as next steps. On image 4 it would show all the events.
If we have this infrastructure in place, we would be able to extend the recurrent actions in the future to have multiple steps (out of scope in here I know).

accepted/0000-uyuni-microos-management.md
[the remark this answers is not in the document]
agraul: I think there are two aspects here:
1. User changes something on a transactional system through Uyuni, which requires a reboot. This goes through `transactional_update.apply`, and we should set `activate_transaction=True`. That acts like a "reboot if needed". When this reboots is then determined by the reboot strategy. By default, this will be instant. But it can be configured to use `rebootmgr` with maintenance windows (outside of Uyuni).
2. User changes something on a transactional system outside of Uyuni and wants to reboot if there are pending changes. This can be achieved with two calls today: `salt '*' transactional_update.pending_transaction()` and `salt '*' transactional_update.reboot()`. We can't use that from within a transaction, but they should work from `state.apply`. I need to test it.

accepted/0000-uyuni-microos-management.md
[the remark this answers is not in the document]
agraul: There is a problem with `transactional-update apply` when we want to show snapshots: `snapper` does not know that `transactional-update apply` "applied default snapshot as new base for running system". As such, we can't correctly say which snapshot is active (which is accurate, as there is no single snapshot that's "equal" to the running system after `transactional-update apply`).

For example, the following is reported after running `transactional-update apply` on `51-slmicro62.smlm51.lab`:
```
51-slmicro62:~ # transactional-update apply
Checking for newer version.
transactional-update 5.0.7 started
Options: apply
Separate /var detected.
Syncing /etc state from old snapshot...
Using new snapshot - syncing from old parent "/etc"...
Deleted in new snapshot: "/.snapshots/10/snapshot/etc/./transactional-update.comparewith"
File changed: "/etc/./fstab"
Deleted in old snapshot: "/etc/./transactional-update.comparewith"
Processing files...
Copying "./fstab"

Using default snapshot 10 to replace running system...
Applying /usr...
Applying /etc...
Applying /boot...
Executing systemctl daemon-reexec...
Executing create_dirs_from_rpmdb...
Executing systemd-tmpfiles --create...
=> Applied default snapshot as new base for running system!
   Running processes will not be restarted automatically.
transactional-update finished

uyuni-server:/ # salt '*' --out=table snapper.list_snapshots
51-slmicro62.smlm51.lab:
----------
    --------------------------------------------------------------------------------------------------------------------------------------
    | Active* | Cleanup | Default* |      Description      | Id | Timestamp  |  Type  | User |                   Userdata                  |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |         |  False  |        current        | 0  | 1770988569 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |         |  False  | first root filesystem | 1  | 1758094726 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |  number |  False  | Snapshot Update of #1 | 2  | 1770905630 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |  number |  False  | Snapshot Update of #1 | 3  | 1770906268 | single | root | {'transactional-update-in-progress': 'yes'} |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |  number |  False  | Snapshot Update of #1 | 4  | 1770906548 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |         |  False  | Snapshot Update of #1 | 5  | 1770908087 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |  number |  False  | Snapshot Update of #5 | 6  | 1770912720 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |  number |  False  | Snapshot Update of #5 | 7  | 1770987133 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
    |  True  |  number |  False  | Snapshot Update of #5 | 8  | 1770987168 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |  number |  False  | Snapshot Update of #8 | 9  | 1770987297 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
    | False  |  number |   True  | Snapshot Update of #9 | 10 | 1770987371 | single | root |                      {}                     |
    --------------------------------------------------------------------------------------------------------------------------------------
```
\*: I patched `list_snapshots` to add the `Active` and `Default`

I'm not sure we should go for this, it will be confusing and it's not really recommended to use `transactional-update apply` generically.

accepted/0000-uyuni-microos-management.md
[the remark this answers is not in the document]
rjmateus: Ok got it. The main question in then if we should call the `transactional_update.apply` with the flag to active the transation automatically, or if we should call it with false, and let the user controls it from uyuni.
Would make sense to allow users to configure the behaviour? And then have a state on uyuni that would reboot if needed?

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,226 @@
...
+
+
+### Unsupported Internal States
+Not all internal states we have make sense on MicroOS / SUSE Linux Micro. These states are unsupported.
+
+- `appstreams.configure` - only useful for RHEL systems
+- `bootstrap.remove_traditional_stack` - traditional stack was never supported on these systems
+- `cocoattest.requestdata`- only supports SLES 15 SP6
+- `rebootifneeded` - The way this is written is incompatible with transactional systems
+- `uptodate` - The way this is written is incompatible with transactional systems
```
rjmateus: What would be the alternative for transactional systems if we want to install all the updates available?

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,313 @@
...
+```
+
+### Internal States Application
+The Java backend maintains a list of SLS files in the `SaltParameters` class. This catalog will be
+enhanced with a hard-coded mapping of SLS files to the respective Salt function (`state.apply` or
+`transactional_update.apply`).
+
+Some states need to be split into two SLS files, one for prerequisites and a second for the actual
+work. The Java backend code first trigger the prerequisites state. When that job returns
+successfully and the system has rebooted, Uyuni reacts with the "actual work" state without any further user interaction.
```
wweellddeerr, line 53: My understanding is that this entire flow (prereq -> reboot -> actual work) will be executed within the context of a single Action in Uyuni, is that correct? If so, a failure at any stage means that re-running the Action will execute the entire flow again from the beginning. Is that considered acceptable? Or the expectation is to implement a smarter mechanism for handling failures and retries?

For instance, if the Action fails after the reboot, re-running it will trigger the prereq again. Theoretically, this shouldn't require a new reboot since the transaction was already applied, allowing the actual work to then execute smoothly. However, the system has effectively been altered (a new snapshot booted and packages installed) even though the Action remains marked as 'Failed' in Uyuni. Similarly, if the failure happens in the reboot itself, we would end up with a pending snapshot marked as default for the next boot, despite the Action being in a failed state.

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,313 @@
...
+
+The flow is similar to _Action Chains_ that contain a reboot. Uyuni configures a Salt Reactor
+listening to `/salt/minion/*/start` events. The same mechanism could be used for applying the second
```
wweellddeerr, line 56: The start event can also be triggered by other reasons besides a real system reboot (e.g., a simple salt-minion service restart), right?

If so, wouldn't it be better to add an extra verification step here? Or is this considered a rare enough corner case that we can just accept and ignore for now?

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,313 @@
...
+
+#### Example Workflow with Salt CLI
+The following example first installs prerequisites for the `hardware.profileupdate` state, reboots
+the minion and then triggers the `hardware.profileupdate`.
+
+```sh
+% mgrctl exec -- salt slmicro61 transactional_update.apply hardware.prereq
+% rebootRequired=$(mgrctl exec -- salt --out=json slmicro61 transactional_update.pending_transaction | jq --exit-status '.["slmicro61"] == true')
+% if [[ $rebootRequired = "true"]]; then
+> mgrctl exec -it 'salt slmicro61 transactional_update.reboot && salt-run state.event salt/minion/slmicro61/start count=1'
```
wweellddeerr: What happens if the system already has a pending reboot before we even start applying our split states?

If I understood [Salt docs](https://docs.saltproject.io/en/3007/ref/modules/all/salt.modules.transactional_update.html) correctly, if we simply call `transactional_update.apply hardware.prereq` without explicitly managing the snapshot base, it defaults to branching from the currently running system. This means any pre-existing pending transaction will be silently orphaned and replaced as the default boot target, effectively dropping those previous changes. Conversely, if we were to force the use of the `--continue` flag to preserve that existing pending transaction, our automated reboot for the prereq step would inadvertently force those pre-existing changes to boot alongside ours.

Furthermore, because `transactional_update.pending_transaction` is a blind check, it will return `true` simply because a pending snapshot exists, making it impossible to distinguish if the reboot requirement was actually caused by our prereq execution or if it was already there.

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,313 @@
...
+### "Salt-initiated" bootstrap
+This method is used by users with large environments. They often have their own custom OS images and
+want to bootstrap new systems with little manual intervention.
+
+This bootstrap only needs to install the Salt Bundle when it is executed with the Salt Minion from
+the distribution. As part of the bootstrap procedure, a Salt highstate will replace the installed
+Salt Minion package with our Salt Bundle package. Users are advised to include the Salt Bundle into
+their OS images to avoid the reboot after switching from the distribution's Salt Minion to Salt Bundle. 
+
+## Expose System Snaphot information in Uyuni
```
wweellddeerr, line 306: What would be good triggers for gathering this information from the systems?

accepted/00108-uyuni-microos-management.md
agraul: > My understanding is that this entire flow (prereq -> reboot -> actual work) will be executed within the context of a single Action in Uyuni, is that correct?

From a user perspective, it's a single action (e.g. click on "Schedule Hardware Refresh), but internally this could be turned into multiple actions. From what I can tell, the current implementation of Action Chains does not do what we want: it puts everything into states, but here we need to control how states are applied. On a higher level what we want to do is like an action chain.

>  However, the system has effectively been altered (a new snapshot booted and packages installed) even though the Action remains marked as 'Failed' in Uyuni.

That's a good point. I think we'll need to pull the three steps apart a bit and show which parts were done and which parts were not reached due to a failure. The overall status of the action could still be "failed", as long as the user understands what has happened.

accepted/0000-uyuni-microos-management.md
agraul: `transactional_update.apply` uses `snapshot=continue`, i.e. we will add our changes to pending changes and activate the whole transaction. I think that's better than branching off the running state and dropping user-initiated changes or doing our changes and then rolling them back -- which causes two reboots and means we'll have to install our prerequisites again in the future.

> Furthermore, because transactional_update.pending_transaction is a blind check, it will return true simply because a pending snapshot exists, making it impossible to distinguish if the reboot requirement was actually caused by our prereq execution or if it was already there.

That's a good point. Maybe we should instead (or additionally) check whether `transactional_update.apply hardware.prereq` causes any `changes`. Something like the following:
```sh
% mgrctl exec -- salt --out=json slmicro61 transactional_update.apply hardware.prereq | jq 'all(.slmicro61[].changes; length == 0)'
```

accepted/00108-uyuni-microos-management.md
cbbayburt: > From a user perspective, it's a single action (e.g. click on "Schedule Hardware Refresh), but internally this could be turned into multiple actions. From what I can tell, the current implementation of Action Chains does not do what we want: it puts everything into states, but here we need to control how states are applied. On a higher level what we want to do is like an action chain.

I think the simplest would be to treat this as a single action in the backend as well. We can attach the necessary info about the current state of the execution to the action entity, and the backend can pick up where it left off in case of re-runs.

Otherwise, we'll need to define a 3rd flow apart from action chains and regular actions. Could be hard to maintain.

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,313 @@
...
+
+### Salt Formulas
+
+On non-transactional systems, Salt Formulas are applied as part of the Salt Highstate. Similar to
+_custom states_, Uyuni can't decide upfront if all formulas can be applied with
+`transactional_update.apply` or if `state.apply` is to be used.
+
+Salt Formulas are not part of the highstate for transactional systems.
+
+The `formulas` state is added to the list of states that can be selected for recurring actions with type "Custom state".
```
cbbayburt, line 151: So formulas are going to be excluded from the highstate, but can be included in recurring actions?

We need to communicate this clearly in both documentation and the UI.

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,313 @@
...
+Salt includes an execution module to interact with `snapper`, a snapshot management tool for btrfs
+snapshots. In particular, `snapper.list_snapshots` and `snapper.diff` expose information about
+available snapshots and how they differ.
+
+To store snapshot information, the database table`rhnServer` gets new columns:
+  ```sql
+  active_snapshot     NUMERIC,
+  default_snapshot    NUMERIC,
+  snapshots           NUMERIC[]
+  ```
```
cbbayburt, line 321: I would suggest normalizing these columns in a new table since these attributes are not fundamental to every server.

accepted/0000-uyuni-microos-management.md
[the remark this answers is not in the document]
cbbayburt: So is it decided that we put this out of scope for now and go back to it after the initial implementation?

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,313 @@
...
+* We don't know if all features need to work for transactional systems since they are managed differently.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Full control over custom states for the user. Requires changes to the WebUI / API and database
+  schema. This is something we could do later if the global config approach is not enough.
+- Support for two system states ("live" and "next") in the database, WebUI and API. This would allow
+  users to know not only what the current live system looks like, but also how it will look like
+  after a reboot. This is a lot of work, everything expects a system to be in a singular state all the time.
```
cbbayburt, line 350: Maybe we can do this with minimal intrusion by implementing an additional API namespace and some isolated UI views for this info, rather than embedding/modifying the existing stuff.

cbbayburt (approved):
Thanks for the updates, @agraul.

accepted/00108-uyuni-microos-management.md
wweellddeerr: I agree that keeping it as a single Action is probably the best approach. My main point was exactly to clarify if leaving the current 'all-or-nothing' execution as is would be acceptable for now, or if it will be necessary to implement this fine-grained control, both to guarantee user transparency and to ensure that a re-schedule can safely resume from the exact point of failure without blindly repeating the `prereq` step.

accepted/0000-uyuni-microos-management.md
```diff
@@ -0,0 +1,313 @@
...
+In Uyuni, we use the Highstate to ensure packages for different features are installed and that
+access tokens to Uyuni's repositories are kept up-to-date. Additionally, _custom States_ are part of
+Uyuni's highstate configuration. 
+
+Uyuni uses a [global variable](#configurable-states-javasalt_custom_states_use_transactional_update)
+to decide which function to use to apply _custom states_ for transactional systems. This variable is
+used by the Java code to decide which function to call, but this information
+(`transactional_update.apply` vs `state.apply`) is not available when our `mgr_master_tops.py` tops
+module composes the list of states for a highstate. Therefore, `mgr_master_tops.py` excludes custom
+states from the highstate for transactional systems.
```
rjmateus: Highstate will be applied with `transactional_update.apply` vs `state.apply`?

accepted/00108-uyuni-microos-management.md
rjmateus, line 151: I think we should add the option to run a single formula in the formulas UI. Maybe even give the option to select if should be executed with `transactional_update.apply` vs `state.apply`.
I know that you can run a single execution of recurrent action, but doing it to run just once the formula can be to cucumbersome.

accepted/0000-uyuni-microos-management.md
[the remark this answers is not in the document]
agraul: > So is it decided that we put this out of scope for now and go back to it after the initial implementation?

Yes, we will revisit this later.

accepted/0000-uyuni-microos-management.md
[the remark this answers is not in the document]
agraul: Conclusion:
- Don't reboot automatically
- Let users include `rebootifneeded` if they want automatic reboots
- When we improve `rebootmgr` integration, we can revisit this decision (with configurable reboot time windows)

accepted/00108-uyuni-microos-management.md
agraul: I'll look into adding this information in the start event. If we can include it, let's add the extra verification.

We don't want to actively ask the client if it was just rebooted, that's too expensive.

accepted/00108-uyuni-microos-management.md
agraul: - package list refresh should cause a refresh of this information
- TBD: how to detect changes manually introduced (analogous to pkgset beacon)

accepted/00108-uyuni-microos-management.md
agraul: It's probably a corner case that we won't hit that often. If we do, we can implement the `boot_info` grain check I added to the rfc text.

accepted/00108-uyuni-microos-management.md
agraul: Package list refresh is probably enough for now. If a beacon is required, we can look into it later.

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,364 @@
...
+
+All internal states that interact with the live system are applied with `state.apply`.
+Internal states that change the operating system, e.g. by installing packages, are applied with
+`transactional_update.apply`. Today, many of our SLS files combine installing and using packages. 
+That does not fit the transactional model and we need to split these SLS files.
+
+Custom states give users a lot of flexibility managing their systems, which makes it hard for Uyuni
+to know how to apply custom states. To give some control to the user, we add a new config value:
+`java.salt_custom_states_use_transactional_update = True`. This config value controls whether
+`transactional_update.apply` is used for all custom states applied from the Java backend. The
```
Martin-Weiss, line 29: How is this affecting customers that use GitFS for their salt states? Is the Java backend processing the states coming from /srv/salt and GitFS as well?

accepted/00108-uyuni-microos-management.md
```diff
@@ -0,0 +1,364 @@
...
+  users to know not only what the current live system looks like, but also how it will look like
+  after a reboot. This is a lot of work, everything expects a system to be in a singular state all the time.
+- Make sure a small subset of SLS files work on transactional systems and document the rest as unsupported.
+
+# Future Improvements
+[future]: #future
+- Directly apply Salt Formulas from the WebUI/API, without having to add it to recurring states. This will also benefit non-transactional system management that currently requires adding formulas to the highstate.
+- Full control over custom states for the user. Requires changes to the WebUI / API and database
+  schema. This is something we could if the global config approach is not enough.
+
```
Martin-Weiss, line 358: I am not sure if I understand the full picture but have the feeling that there are still some challenges and gaps.. especially on how users have to deal with custom states with `java.salt_custom_states_use_transactional_update = false`

From an expectations point of view - the user wants to do config management which is installing things, changing configs and restarting processes after changing settings. And often this things need to be changed with proper ordering and dealing with dependencies.

So we need to analyse states on "transactional or not" and then run then in the right order with automatic reboots based on maintenance windows.. 

So could we have some sort if intelligent loop until all states are applied like this:

- Any action required for transactional-update?

   Yes: 
         - Apply everything that needs transaction in transaction (needs a list of states for that) in the defined order
         - Reboot based on maintenance schedule
         - Do it again until no further transactional action is required

   No: 
         - Run non-transactional-actions

Challenges: 

- dependencies and tests - i.e. "after AD join install xyz".. so things that call external commands or scripts.
- detection of "continue" or "no continue" (maybe we can see this in the snapshots)

Martin-Weiss (commented):
Maybe I am missing some things.. but maybe this helps?!?

accepted/00108-uyuni-microos-management.md
rjmateus: Java is not processing the states coming from there. But you can create a configuration channel and import the desired salt state there.

Martin-Weiss: What do you mean by "import" - in case we use GitFS the idea is to use GitOps and no "extra import"... So if I get it right - if Java code is required for managing transactional systems it is basically against using GitOps and against the general salt concept - true?

rjmateus: If you want to use salt cmd direclty, you can control if the call will be executed inside or outside a transation (the all state you are applying).
I was just mentioning that users also have another option and can have the salt states in GitFS, and create a configuration channel that includes that salt state defined in gitfs. 
The states would still be managed in git, but its application can be managed with the configuration of a channel. It can be useful for automations with recurrent actions, for example.

Martin-Weiss: We do not want to use "salt cmd" - we want to use salt states - basically the same unmodified states against transactional and non-transactional systems and the salt component / module / agent should take care of the right things regardless of the system installation type.. 

But - maybe I miss a point - when creating a salt state in SMLM - how can I say "reboot automatically after installation of a package with or without continue in case it is a transactional system - but do not reboot in case it is not a transactional system"?  Can you share an example?

And how can the same be done if the state is in GitFS? (expectation is that there is no differnence)

rjmateus: > basically the same unmodified states against transactional and non-transactional systems and the salt component / module / agent should take care of the right things regardless of the system installation type..

That was exactly what we tried for several years, and it led to several problems. So we decided to follow the approach that MLM and it's users need to know they are managing a transactional system, have have it's states adapt to it.

> how can I say "reboot automatically after installation of a package with or without continue in case it is a transactional system - but do not reboot in case it is not a transactional system"

If you have a single state for everything, you cannot. Applying a state will apply everything. A change that would mean a huge re-implementation of the salt execution mechanism.

We are moving in the direction of having the states split into 2, one with the prerequisites (like installing packages) then users can reboot, and next apply the other steps of the state. With this RFC we are moving in that direction, but we still need to have the ability to chain several state apply.

Martin-Weiss: > > basically the same unmodified states against transactional and non-transactional systems and the salt component / module / agent should take care of the right things regardless of the system installation type..
> 
> That was exactly what we tried for several years, and it led to several problems. So we decided to follow the approach that MLM and it's users need to know they are managing a transactional system, have have it's states adapt to it.

Maybe I miss the reasoning and do not understand - but in any case the "logic" needs to be implemented and maintained / tested and supported somewhere. If this is not in the product and the customer / user has to do the coding and test / maintain the (salt) code - huge effort and cost is created and value on the product side might be limited.

> 
> > how can I say "reboot automatically after installation of a package with or without continue in case it is a transactional system - but do not reboot in case it is not a transactional system"
> 
> If you have a single state for everything, you cannot. Applying a state will apply everything. A change that would mean a huge re-implementation of the salt execution mechanism.
> 
> We are moving in the direction of having the states split into 2, one with the prerequisites (like installing packages) then users can reboot, and next apply the other steps of the state. With this RFC we are moving in that direction, but we still need to have the ability to chain several state apply.

So "SMLM" internal adjustments is one thing - but how customers/users need to deal with that is a total different one.. expecting lots of good examples and pre-delivered code that customers can just copy+paste.. and maybe a auto-conversion for existing states.

And - btw - the same requirements might exist for the Ansible part in SMLM as well.. not sure how much different or better transactional OS management will be with SMLM + Ansible playbooks..

vzhestkov: @Martin-Weiss `We do not want to use "salt cmd" - we want to use salt states - basically the same unmodified states against transactional and non-transactional systems` - this part is not really possible due to the nature of transactional systems, there are some actions which can be either performed just inside the transaction or makes sense only outside of the transaction, and the chances that the same state can be applied the same way inside and outside is almost 0%, in case of developing the state the author must know the internals of transactional system to make it working there. Within one `state.apply` process it's impossible to switch to apply it patially inside and outside.

To make it working more or less in the order of applying certain parts of the states sequentially with reboots in the middle the state should be splitted and the sequence of applyment shoud be controlled externally. For example one first part of the state is firing certain event to the master and on the master side the next `state.apply` is triggered for the second part maybe with some delay or waiting `salt/minion/.../start` event.

About using the states from GitFS, you can enable gitfs integration in salt, but Java is not taking care of it at all, the only thing I can suggest to use the states from GitFS with the web UI or API is to use `include` declaration: https://docs.saltproject.io/en/3006/ref/states/highstate.html#include-declaration

Martin-Weiss: So with that I can just say - does not fulfill expectations nor requirements and the solution will end up in a lot of "salt state code" will have to be developed/re-developed and modified in salt states in customer environments and this "customer code" then also needs to be maintained for a long period of time on the customer side of things.

Maybe will be the point in time to decide for switching to some other configuration management solution or to move away from transactional mode.

m-czernek: Do you have an example of a config management system that handles immutable OS management better? And could you in such case provided an example of what exactly is better in that context?

Martin-Weiss: No - do not have an example. I guess this is a unique added value SMLM with SL Micro is expected to deliver?

rjmateus: > I guess this is a unique added value SMLM with SL Micro is expected to deliver?

To be able to deliver that value the OS also needs to take into consideration how to manage it remotly at large scale.

As you state, no other tool has a good solution for transaction systems, because without the ability to activate changes without reboot we cannot implement in a way that doesn't force users to split the states between transactional and non-transactional.

vzhestkov: Even if we magically make `salt.state` working with some parts of the states outside and inside the transaction mixing them togather (from the technical point of view it could be possible to specify which parts of the states should go inside or outside, but will require huge development effort anyway), but the real problem is the transactional system here, you can't just jump in and out the transaction just because if anything has changed in the transaction, it should be applied first before become available on the running system (reboot, there is some hacky ways to apply the transaction on the fly, but it's not recommeded to use them in production by the developers of the transaction system), so the `state.apply` process should be made the way it should continue from the point where it stopped before reboot and continue after and there could be not just one reboot needed. And finally all these points togather push us back to the key point that state could be only made by the person who understand how transactional systems work and any attempt to unify the state for transactional and non-transactional case will lead to overcomplicated solution which will be hard to read, so better to keep one state for non-transactional systems and the other, or better to say others, for transactional.

rjmateus: From my point of view, we should move in the direction of having re-usable workflows (something like action chains that can be re-used), and then users could define the workflow and if each step needs to run inside or outside a transaction.
But this would be mid to long term solution, no something we can implement in the short term.

Martin-Weiss: I still believe we need some sort of automatic and user-visible queuing system for managing ordered changes driven by salt states..
And find some logic to automatically reboot when required and automatically continue after reboot etc.. 
This handling should be part of the system and not develop and testing effort for the customer.

PR #93, opened 2024-10-16
Provide RFC for Salt SSH for TU systems

m-czernek:
Related research: https://github.com/SUSE/spacewalk/issues/24809


accepted/uyuni-microos-salt-ssh.md
```diff
@@ -0,0 +1,85 @@
...
+[summary]: #summary
+
+Uyuni users may choose to manage minions by using Salt SSH, for example due to network constraints. However, currently, systems that are based on transactional update, i.e. openSUSE MicroOS or SUSE Linux Micro (TU minions), do not support management via Salt SSH.
+
+This RFC describes improvements for Salt SSH such that Uyuni can fully control TU minions by using Salt SSH.
+
+# Motivation
+[motivation]: #motivation
+
+Uyuni users want feature parity of TU minions to non-TU minions. Currently, Uyuni users can use the Salt, Salt SSH, or Salt SSH Proxy methods to control their deployed non-TU minions. However, for TU minions, SSH management does not work.
```
vzhestkov, line 14: I think naming here is not the same as we are using at least on the web UI. I suppose you mean the contact methods: `Salt` (minion) / `Push via SSH` (salt-ssh) / `Push via SSH tonnel` (salt-ssh with tonnel to https of the server).

accepted/uyuni-microos-salt-ssh.md
```diff
@@ -0,0 +1,85 @@
...
+
+TU minions are different from non-TU minions in that they mount some partitions as read-only, for example `/etc`. To modify such partitions, users use the `transactional-update` command.
+
+The `transactional-update` command temporarily mounts the partitions as read-write, and after the modifications are done, it creates a new BTRFS snapshot that is activated upon restart (or upon using `transactional-update apply`).
+
+Note that Salt SSH relies on both file systems for functions that modify the read-only file system:
+
+- The filesystem outside of a transaction (IAT) contains both the Salt client, and, more importantly, the `salt_state.tgz` file.
+- The filesystem inside of a transaction (OOT) requires both the Salt client and the `salt_state.tgz` because the execution must happen in a transaction.
```
vzhestkov, line 33: Looks like IAT/OOT mixed there, but I could be wrong, are they really using such acronyms?

m-czernek: Fixed. Not really used by upstream (we're the upstream for transactional systems, honestly), but I needed a short way to distinguish between in-transaction and out-of-transaction states.

accepted/uyuni-microos-salt-ssh.md
```diff
@@ -0,0 +1,85 @@
...
+This poses a problem: on TU systems, Salt needs to coordinate files between the IAT and OOT filesystems.
+
+## Solution
+
+There are two Salt modules that modify the filesystem:
+
+- `state` (e.g. `state.apply`)
+- `transactional_update` (e.g. `transactional_update.apply`)
+
+Users should use the `state` module modify OOT filesystem, and the `transactional_update` module to modify the IAT filesystem.
```
rjmateus, line 44: This means that they cannot be mixed together in the same state? So any OOT state should be defined in it's own sls file and executed with state.apply, and any states that needs to modify IAT should be in a different sls file and executed with transational_update.apply.
Do we have any state that modifies files/data in both locations?

rjmateus (commented):
Just one note from my side, other then that looks good to me. Lets also wait for ION squad team members review

accepted/uyuni-microos-salt-ssh.md
m-czernek: Correct. Users must specify whether they want to run a state inside, or outside of a transaction, and execute either `state.apply`, or `transactional_update.apply`.

>  Do we have any state that modifies files/data in both locations?

This is not possible, at least not without some significant modifications to Salt at a quite low level. Salt does not give us control at a more fine-grained level than the whole execution.

accepted/uyuni-microos-salt-ssh.md
```diff
@@ -0,0 +1,85 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+## Current State
+
+Currently, when Uyuni bootstraps a minion by using Salt SSH, Uyuni deploys a Salt client into `/var/tmp/venv-salt-minion`.
+When Salt SSH makes a `state` call to a client, for example `state.sls example-sls-filename`, it additionally compiles the `example-sls-filename` state (and all its dependencies, such as files used by the state) into a tar file, and deploys it to a generated location in `/var/tmp`, such as `/var/tmp/.root_12345_salt/salt_state.tgz` (together with other data, such as external modules, grain info, etc).
+
+TU minions are different from non-TU minions in that they mount some partitions as read-only, for example `/etc`. To modify such partitions, users use the `transactional-update` command.
```
aaannz, line 26: `/etc` is not read only. etc is an overlay and is read-write.

```suggestion
TU minions are different from non-TU minions in that they mount some partitions as read-only, for example `/usr`. To modify such partitions, users use the `transactional-update` command.
```

accepted/uyuni-microos-salt-ssh.md
```diff
@@ -0,0 +1,85 @@
...
+
+The `transactional-update` command temporarily mounts the partitions as read-write, and after the modifications are done, it creates a new BTRFS snapshot that is activated upon restart (or upon using `transactional-update apply`).
```
aaannz, line 28: This is technically not correct.
TU first creates a new read-write snapshot of currently running snapshot (or latest if -c is used) and at the end set this new snapshot as default one for next reboot.
It does not remount anything as rw even temporarily.

accepted/uyuni-microos-salt-ssh.md
```diff
@@ -0,0 +1,85 @@
...
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- Greater resource requirements: each execution of Salt SSH deploys a Salt client to the temporary `/var/cache` location, executes a new Salt process, and removing the temporary Salt client at the end of the execution. This means increased storage footprint (both RAM and disk), as well as added CPU cycles related to copying and cleaning up Salt.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Not supporting Salt SSH on TU systems.
```
aaannz, line 80: Instead of "not supporting" salt ssh on TU systems, could we instead suggest using `transactional_update` execution module? Or can we even provide a transactional_update state which would wrap TU execution module?

i.e. instead of `pkg.installed` suggest using  `transaction_pkg.installed`?

Of course this would open another can of worms like:
* users required to keep in mind mandatory reboot
* detect we are already running in the transaction via tu executor in normal salt

m-czernek: I'm not sure I understand here. This RFC is about requiring the users to use the `state` or `transactional_update` modules based on their needs:

> Users should use the `state` module modify OOT filesystem, and the `transactional_update` module to modify the IAT filesystem.

So if I'm not misunderstanding, we already suggest that user use the `transactional_update` when necessary. This leads us to changing how Uyuni/SUMA deals with TU minions as well.

Regarding:

> i.e. instead of pkg.installed suggest using transaction_pkg.installed?

It would work for some changes, e.g. changes that do not require other files, like the `pkg.installed` you mention. It would not work for changes that require anything on the filesystem, e.g.:

- modules that execute `sls` files
-  the `file` command that copies a file to a different location, etc.

This would require the same synchronization mechanism as we provide in this RFC. 

I can add it to alternative solutions with the caveat that we'd support only a subset of Uyuni's functionality (and of course, we'd also have to re-do how we treat TU minions on Uyuni's side since neither `state.apply` nor `transactional_update.apply` would work)

admd:
@m-czernek what's the state of this RFC? are we going to pick this up soon again?

m-czernek:
@admd the bulk of the research is done. This is basically a subtask of the overall TU minion support improvement story.

When revisiting the RFC, I'd like to verify a few things, e.g. if the `/var/cache` location on filesystem is still shared between TU and non-TU execution to confirm that the main deployment mechanism still works on the newest SLM.

We stopped with the research here because the TU story was deprioritized. We can pick it back up if we prioritize it again.

I believe that some of the confusion here also came from the fact that we were not yet in agreement on splitting the Salt states we have in Uyuni into parts and supporting explicitly executing states inside and outside of transactions. I believe we have already agreed on that in the main transactional update RFC, so finishing and implementing this one should be simpler at this point.

PR #94, opened 2024-11-04
How to package and upgrade to Salt 3008.x

meaksh:
Read the RFC [here](https://github.com/uyuni-project/uyuni-rfc/blob/master-salt-3008/accepted/00102-upgrade-to-salt-3008.md)

### DIFF
--- a/accepted/00102-upgrade-to-salt-3008.md
+++ b/accepted/00102-upgrade-to-salt-3008.md
@@ -41,6 +41,7 @@ There are a couple of Salt Extensions that we want to have integrated by default
 - transactional_update
 - openscap
 - docker / dockerng
+- virt (needed by virtualization-formula)
 - ...
 
 These are the minimum required extensions to be able to do basic operations in SUSE/openSUSE distributions and also allow basic operations on the context of SUSE Manager clients. We must maintain and support these builtin Salt Extensions.
@@ -63,7 +64,7 @@ The idea would be to include the releases for the different builtin extension di
 
 These builtin extensions will be then placed in `/usr/lib/python3.11/site-packages/saltext/` when installing `python311-salt` so they are available for Salt.
 
-**NOTE:** The usage of `git submodule` would make it a bit tricky when it comes to generate patches for the single unified Salt codebase, as apparentely you cannot get an unified diff for the main repo + submodules.
+**NOTE:** Depending on the maintaining workflow to use, the usage of `git submodule` would make it a bit tricky when it comes to generate patches (i.a. during bugfixing) for the single unified Salt codebase, as apparentely you cannot get an unified diff for the main repo + submodules.
 
 Alternatively, we could create separated packages for each of those default Salt Extensions and then make add them as `Requires` for the `python311-salt` package. In that case, an official Salt Extensions must be created (if not existing) for each extension we want to have a package, and then taking the sources from the official extension repository. One Salt Extension, one RPM package.
 
@@ -123,7 +124,9 @@ A testing build project can be found [here](https://build.opensuse.org/project/s
 
 ## Maintaining Salt
 
-TBD
+Given the irruption of Salt Extensions, we need to rethink the current workflow we use for maintaining Salt (and now also some of its extensions). In this regard, we want to improve our workflow, to make it simpler and adapted to the new Salt Extensions scenario and aligned with SUSE & OBS proposed workflows.
+
+As this section has itself enough significance, it will be covered in a dedicated RFC around "New SCM workflow for Salt", so discussion around this topic can happen separetely to the present RFC.
 
 # Drawbacks
 [drawbacks]: #drawbacks
--- a/accepted/00102-upgrade-to-salt-3008.md
+++ b/accepted/00102-upgrade-to-salt-3008.md
@@ -84,7 +84,7 @@ In case of packaging certain list of Salt Extensions, those will be available as
 
 As described by Salt [here](https://salt.tips/the-great-salt-module-migration/), starting with Salt 3008, both Salt Master and Minion can be extended by using `salt-pip` or `pip.installed` to install the Salt Extensions packages (`saltext-*`). Similar to the extending mechanism we currently have for the Salt Bundle. This makes it easy to integrate the customization of your minions with Uyuni an SUSE Manager, as based on a Salt state.
 
-If a Salt Extension is not existing yet (as most of the dropped modules are NOT yet migrated to Salt Extensions), they mention to provide them as custom modules in your Salt state tree (file_roots).
+If a Salt Extension is not existing yet (as most of the dropped modules are NOT yet migrated to Salt Extensions), they mention to provide them as custom modules in your Salt state tree (file_roots), i.a. `/srv/salt` as users alreayd do when extending Salt with their custom modules.
 
 As mentioned above, we would be probably providing a main Salt package containing some builtin extensions already there that at least covers basic operations.
 
--- a/accepted/00102-upgrade-to-salt-3008.md
+++ b/accepted/00102-upgrade-to-salt-3008.md
@@ -131,8 +131,8 @@ Since Salt 3008.x release has been delayed and it is not yet fully clear when it
 The alternative plan:
 - Prepare Salt 3006.0 package based on Python 3.11.
 - Drop Salt from Basesystem / Server Application modules for SP7 before feature cut-off. (ECO needed)
-- Create new "SLE-Module-Salt" module for SP4/5/6/7 based on "SLE-Module-Python3".
-- Deprecate Salt from Basesystem / Server Applicacion module in favor of "SLE-Module-Salt" for SP4/5/6. (ECO needed)
+- Create the new "SLE-Module-Salt" module for SP4/5/6/7 based on "SLE-Module-Python3", containing the current Salt 3006.0 based on Python 3.11.
+- Deprecate Salt from Basesystem / Server Applicacion module in favor of Salt coming from "SLE-Module-Salt" for SP4/5/6. (ECO needed)
 - Once 3008 is finally released, we can upgrade Salt in "SLE-Module-Salt" module.
 
 # Unresolved questions
--- a/accepted/00102-upgrade-to-salt-3008.md
+++ b/accepted/00102-upgrade-to-salt-3008.md
@@ -128,6 +128,13 @@ Since Salt 3008.x release has been delayed and it is not yet fully clear when it
 
 **IMPORTANT: Since we want to drop the Salt 3006.0 from the Basesystem and Server Application modules for SLE 15 SP7 (based on Python 3.6), in favor of a new Salt based on Python 3.11, even if 3008 is delayed we should NOT compromise the SLE 15 SP7 feature cutoff deadline to implement the new "SLE-Module-Salt" module even if it still provides 3006.0 version**
 
+The alternative plan:
+- Prepare Salt 3006.0 package based on Python 3.11.
+- Drop Salt from Basesystem / Server Application modules for SP7 before feature cut-off. (ECO needed)
+- Create new "SLE-Module-Salt" module for SP4/5/6/7 based on "SLE-Module-Python3".
+- Deprecate Salt from Basesystem / Server Applicacion module in favor of "SLE-Module-Salt" for SP4/5/6. (ECO needed)
+- Once 3008 is finally released, we can upgrade Salt in "SLE-Module-Salt" module.
+
 # Unresolved questions
 [unresolved]: #unresolved-questions
 
--- a/accepted/00102-upgrade-to-salt-3008.md
+++ b/accepted/00102-upgrade-to-salt-3008.md
@@ -11,7 +11,7 @@ This RFC describes the preparation and strategy to upgrade Salt to the upcoming 
 
 The upstream Salt project is aiming the future 3008.x release (LTS) for Autumn 2024. As we want Uyuni and SUSE Manager to align to the latest LTS version from Salt project, we need to prepare the strategy to drive this upgrade.
 
-Particularly, this new 3008.x LTS release is the first version after [the great module migration](https://salt.tips/the-great-salt-module-migration/), and lot of Salt modules have been purged from the main Salt package in favor of [Salt Extensions](https://github.com/salt-extensions), including for example, the `zypperpkg` and `transactional_update` modules.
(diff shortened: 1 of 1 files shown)


accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,124 @@
...
+Alternatively, we could reduce this list even more by just providing the `zypperpkg` and `transactional_update` modules and then provide the rest via `/usr/share/susemanager/salt/[_modules,_states,...]` in the Uyuni / SUSE Manager server.
+
+### What about Salt Extensions packages?
+
+Essentially there are two different approaches here:
+
+1. Creating separated packages for each one of the Salt Extensions (potentially hundreds) -> maybe too cumbersome if we go with all of them, and probably not really needed.
+2. Do not package Salt Extensions at all -> builtin extensions + manual customization.
+
+With option 1, we do package and do support all those extensions we decide to have a package for. On the other hand, with option 2, we do only take care of maintaing and supporting the reduced list of builin extensions (actually integrated in the main `python311-salt` package), and then users and customers would need to rely on extending with their desired extensions coming from the community.
```
agraul, line 53: There could be a third option, packaging a curated list of salt extensions. The benefit would be that Uyuni can mirror RPM repos, but not PyPI. Our current extension approach requires internet access or an independent PyPI mirror.

I don't think it's worthwhile to package all extensions, this is a good opportunity to focus our efforts on the Salt extensions that are popular and make sure we can test them as well.

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,124 @@
...
+
+There are a couple of Salt Extensions that we want to have integrated by default in our main `python311-salt` package. These are:
+- zypperpkg
+- dpkg
+- transactional_update
+- openscap
+- docker / dockerng
+- ...
+
+These are the minimum required extensions to be able to do basic operations in SUSE/openSUSE distributions and also allow basic operations on the context of SUSE Manager clients.
```
agraul, line 42: How do we package them? To we put build them separately as salt extensions (each with their own git repo to track the sources & specfile, following the new "The one and only one git workflow") and then use those as build requirements when building `salt`? Or do you want to pull them all into the same package git repo?

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,124 @@
...
+- The old Salt 3006.0 package in SLE15SP4/5/6 will be deprecated in favor of the new version coming in the new "SLE-Module-Salt".
+- The old Salt 3006.0 package will be dropped from SLE15SP7 Basesystem and Server Application modules before SLE15SP7 feature cutoff.
+
+#### Leap15:
+- Python 3.11 is already available in Leap.
+- The new Salt and any new dependencies will get into Leap from SLE.
+
+### New dependencies for Salt 3008.
+
+There are new dependencies that are not available as part of "SLE-Module-Python3", and we will need to take it from Factory or some other SLE/ALP source to include them in the new "SLE-Module-Salt" module.
```
agraul, line 81: Since these are libraries, they might also fit the Python3-Module. We should check with the SLE PMs

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,124 @@
...
+- `python-networkx` + its dependencies (python-matplotlib, python-pandas, python-scipy, python-FontTools)
+- `python-tornado6` 
+
+#### BuildTime:
+- `python-networkx` -> python-pydot, python-graphviz, graphviz
+- `python-meson-python` -> meson, patchelf
+- `python-pandas` -> python-versioneer-toml
+- `python-scipy` -> python-pythran
+
+A testing build project is here: https://build.suse.de/project/show/home:PSuarezHernandez:new-salt-version
```
agraul, line 93: Could we have this in OBS, so that everyone with access to this repo can take a look?

rjmateus:
@meaksh do you think it would make sense to also mention the upgrade strategy across SUMA versions? Should we upgrade suma 4.3 to salt 3008, no matter when it's released? How this relates to suma 4.3 LTS. Also I don't see any reference to salt-minion, so I'm assuming it's being updated at same time as the server (which can influence how and when we upgrade suma 4.3 to 3008).
Other then this, the RFC looks really good. Thank you for preparing it.

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,141 @@
...
+- transactional_update
+- openscap
+- docker / dockerng
+- ...
+
+These are the minimum required extensions to be able to do basic operations in SUSE/openSUSE distributions and also allow basic operations on the context of SUSE Manager clients. We must maintain and support these builtin Salt Extensions.
+
+This list could be reduced even more by just providing the `zypperpkg` and `transactional_update` modules and then providing the rest via `/usr/share/susemanager/salt/[_modules,_states,...]` in the Uyuni / SUSE Manager server.
+
+Builtin Salt Extensions will be integrated into our main "openSUSE/salt" GitHub repo codebase for 3008.x, and provided as content of the main `python311-salt` package.
```
agraul, line 50: How do we do that? Do we use git submodules? What happens to e.g. `zypperpkg` upstream? Shouldn't we take ownership and maintain it as an extension?

agraul (commented):
The "what" is pretty good, but I think we could think a bit more about the "how". SUSE proposed a "one and only" git-based packaging workflow for new products. While we don't have to go all the way there, it would be a good time to explore a better integrated git-packaging workflow that the one we currently have.

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,199 @@
...
+|   |-- saltext-dpkg
+|   |-- ...
+...
+```
+
+The idea would be to include the releases for the different builtin extension directly inside our GH repo, as it was previously done with "tornado". We should take this oportunity also to take the ownership, migrate and create the official `saltext-zypperpkg` extension and probably some others.
+
+These builtin extensions will be then placed in `/usr/lib/python3.11/site-packages/saltext/` when installing `python311-salt` so they are available for Salt.
+
+**NOTE:** The usage of `git submodule` would make it a bit tricky when it comes to generate patches for the single unified Salt codebase, as apparentely you cannot get an unified diff for the main repo + submodules.
```
agraul, line 66: Why do need to generate patches?

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,199 @@
...
+```
+salt
+|-- builin-extensions
+|   |-- saltext-zypperpkg
+|   |-- saltext-dpkg
+|   |-- ...
+...
+```
+
+The idea would be to include the releases for the different builtin extension directly inside our GH repo, as it was previously done with "tornado". We should take this oportunity also to take the ownership, migrate and create the official `saltext-zypperpkg` extension and probably some others.
```
agraul, line 62: How do we do scm on these directories? How do make sure we contribute our changes upstream and then pull them down? How do we publish extensions maintained by us to PyPI?

accepted/00102-upgrade-to-salt-3008.md
meaksh: I slightly modified the text, as generating patches would depend on the maintaining workflow we use.

accepted/00102-upgrade-to-salt-3008.md
meaksh: Yeah, all these questions must be still clarified and defined in a new maintaining worflow for Salt.

I added a note on the current RFC mentioning that we are covering the "new maintining workflow for Salt" in a separated RFC, so we can agree on the proposed plan here for the Salt 3008 upgrade, and start pushing it forward while we keep discussing and designing the new maintaining workflow.

accepted/00102-upgrade-to-salt-3008.md
meaksh: I've added some notes on this regard.

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,202 @@
...
+- A new custom "SLE-Module-Salt" module (inherited from SLE-Module-Python3) will be created for SLE15SP4/5/6/7.
+- The new Salt 3008.x package will be shipped via the new "SLE-Module-Salt" together with its any new dependency which is not part of the "SLE-Module-Python3" module.
+- The old Salt 3006.0 package in SLE15SP4/5/6 will be deprecated in favor of the new version coming in the new "SLE-Module-Salt".
+- The old Salt 3006.0 package will be dropped from SLE15SP7 Basesystem and Server Application modules before SLE15SP7 feature cutoff.
+
+#### Leap15:
+- Python 3.11 is already available in Leap.
+- The new Salt and any new dependencies will get into Leap from SLE.
+
+NOTE: SLMicro 6.0 works differently than SLE15, as it does contain Python 3.11 in the base channels.
```
mcalmer, line 107: You should say also something about SLES12 SP5 LTSS and maybe about other OSes when it comes to SUMA.
AFAIK venv-salt-minion is the only affected package on these platforms as we do not have master and proxy there?

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,202 @@
...
+[alternatives]: #alternatives
+
+Since Salt 3008.x release has been delayed and it is not yet fully clear when it would be, there are some risks that we cannot proceed with the upgrade to Salt 3008.x within some important deadlines when it comes to SUSE Manager and SLE.
+
+1. Stick to the current Salt 3006.0 forever -> no go.
+2. Stick to Salt 3006.0 until 3008 is released -> this will be the way to go if 3008.x is not released on time.
+
+**IMPORTANT: Since we want to drop the Salt 3006.0 from the Basesystem and Server Application modules for SLE 15 SP7 (based on Python 3.6), in favor of a new Salt based on Python 3.11, even if 3008 is delayed we should NOT compromise the SLE 15 SP7 feature cutoff deadline to implement the new "SLE-Module-Salt" module even if it still provides 3006.0 version**
+
+The alternative plan:
```
rjmateus, line 152: I think we should make clear that this is the plan we will be moving, since salt 3008 should not be on time.

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,202 @@
...
+  - SP5/6/7 gets 3008 (after June 2025) -> probably no-go as that mean SP5/SP6 managed minions will break for 4.3 LTS.
+  - 4.3 LTS cannot manage newer SPs with Salt 3008 (after June 2025).
+  - 5.0 gets 3008 at some point after the middle of its lifecycle (after June 2025).
+  - 5.1 is released with 3008 (June 2025).
+  - Salt Bundle (shared between 4.3 and 5.0) is never upgraded to 3008 before 4.3 LTS is EOL (June 2026).
+  - Salt Bundle (new client tools for 5.1) gets upgraded to 3008 (June 2025).
+  - Uyuni will get automatically 3008 from SLE into Leap (June 2025).
+  - Current upstream Salt 3006 EOL date is January 2026, but potentially extended if 3008 is delayed, so worst scenario we are 6 months in our own (probably less).
+
+b2) If Salt 3008 is NOT ready when SUMA 4.3 LTS begins (June 2025) -> We keep Salt 3006.0 for all SPs until LTS ends (June 2026):
```
rjmateus, line 184: if salt 3008 doesn't get released in time I think this would be the best approach, since it avoids the need to support 2 salt major versions at the same time.

rjmateus (approved):
looks good to me, just a few small comments. Thank you for the great contribution Pablo

accepted/00102-upgrade-to-salt-3008.md
meaksh: The classic Salt package is already EOL in SLE12, as it was part of the "Advanced Systems Management Module" which does not have LTSS. The version there was actually Salt 3000, as 3006.0 was not even supported.

For RES8 and Ubuntu 20.04 (client tools), we still ship a classic Salt package (3006.0), I'll add some notes to the RFC also on this regard.

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,202 @@
+- Feature Name: How to package and upgrade to Salt 3008.x
+- Start Date: 2024-10-31
+
+# Summary
+[summary]: #summary
+
+This RFC describes the preparation and strategy to upgrade Salt to the upcoming 3008.x release (LTS).
+
+# Motivation
```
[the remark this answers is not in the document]
meaksh: Makes sense, I'll add this note :+1:

accepted/00102-upgrade-to-salt-3008.md
```diff
@@ -0,0 +1,202 @@
...
+
+Alternatively, we could create separated packages for each of those default Salt Extensions and then make add them as `Requires` for the `python311-salt` package. In that case, an official Salt Extensions must be created (if not existing) for each extension we want to have a package, and then taking the sources from the official extension repository. One Salt Extension, one RPM package.
+
+### What about Salt Extensions packages?
+
+Essentially there are some different approaches here:
+
+1. Creating separated packages for each one of the Salt Extensions (potentially hundreds) -> maybe too cumbersome if we go with all of them, and probably not really needed.
+2. Do not package Salt Extensions at all -> builtin extensions + manual customization.
+3. Only package a reduced list of Salt Extensions -> we could consider some popular extensions that we should then maintain in the context of SLE.
```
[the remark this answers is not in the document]
meaksh: Having an RPM package for every potential Salt Extensions that a customer may use in their states is probably not realistic, so we definetely need a way for users and customers to be able to extend their available modules.

In this regard, in case a Salt Extension is not packaged as RPM, the users could use any of these possibilities:

- Installing extensions directly from PyPI using `salt-pip` or `pip.installed` state to install `saltext-*` packages. This would enable users to extend their minion by applying an state, so easy to integrate with Uyuni but requiring access to PyPI from the minion side.

- Providing Salt Extensions as custom modules for Salt via `/srv/salt/` from the Uyuni server.

accepted/00102-upgrade-to-salt-3008.md
meaksh: I just added some notes to cover the cases for other OSes.

accepted/00102-upgrade-to-salt-3008.md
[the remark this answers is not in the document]
rjmateus: Thank you for the insides. Option 1 is not feasible, since we cannot maintain a 1-to-1 compatibility to all extension.
I think we can move forward as is now in your proposal, but I'm expecting we will have some requirements for 5.2 to improve content consumption.

vzhestkov (approved):
LGTM :+1: hope we will not get something really unexpected on top of it

ycedres (approved):

PR #95, opened 2024-11-12
Role-Based Access Control (RBAC)

cbbayburt:
This RFC proposes Role-Based Access Control (RBAC) model in Uyuni.

See the [rendered version](https://github.com/uyuni-project/uyuni-rfc/blob/rbac/accepted/00000-rbac.md).


accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+
+This RFC proposes implementing role-based access control (RBAC) in Uyuni to meet users' needs for fine-grained access control and configuration.
+The solution is extensible, integrates with existing features of the codebase, and is independent of the ongoing efforts to migrate from the Struts framework.
+
+
+# Motivation
+[motivation]: #motivation
+
+Currently, access control rules are defined by static role information and controlled individually within each action.
+By default, each new feature must define its own access rules; if no rules are explicitly defined, the feature is available to all authenticated users.
```
Etheryte, line 18: Is this the default we want to go with? I'm not entirely sure which makes more sense, available to all by default or available to none by default. My gut feel is that available to none by default would be more sensible, but I think regardless of which one we go with, we should have a good reason for picking one over the other.

cbbayburt: No that's the current default. Proposed way is the opposite, mentioned down below:

> By default, any requested URI is denied access unless a match is found.

It's especially important because it'll also act as failsafe when a dev forgets to map an endpoint, they won't be able to access it when testing.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+
+- **`namespace`**: A label representing the namespace, expressed as a dot-separated string of components. Each component corresponds to a level of hierarchy for a task, with related tasks sharing the same components at higher levels.
+
+  *Example:*
+    - `clm.project.list`: Allows viewing the CLM projects list.
+    - `clm.project.details`: Allows viewing or modifying the details of a CLM project.
+
+- **`access_mode`**: Defines the access level for each namespace as either "View" (R) or "Modify" (W). Many namespaces will have separate entries for each mode, with different endpoints for each purpose.
+
+  For example, `clm.project.details` **[R]** allows viewing a project's details, where `clm.project.details` **[W]** allows modifying them.
```
Etheryte, line 94: A small detail: do we want to always grant R when the user/group has W, or is this not worth the hassle and the end user configuring this will have to grant both themselves? I doubt W withour R makes much sense?

accepted/00000-rbac.md
Etheryte: Ah okay, yes, I agree with that logic. I think I misunderstood which sentence talks about the old bits and which about new.

rjmateus:
> How do we plan to integrate all of this with the frontend?

@Etheryte In my opinion that is a good usability improvement but should be considered a second step. With this RFC we want to set the foundations in term of authentication and changing all the pipes to make it more robust and configurable. As next step I think we can improve a lot the UI, and make it more tailored to the end user.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+The code implementation is generic and relies on configurable access policy definitions, allowing new use cases and rule definitions to be addressed through configuration rather than code changes.
+The Uyuni Server deploys with sensible defaults and enables users to define custom access policies (user access groups) to meet their specific needs.
+
+The goal of access control is to answer the following question: "Can user U perform operation O on resource R?"
+
+Access control operates on three distinct levels, each with specific goals and constraints:
+
+1. **Authentication:** Identifies the user calling this action.
+2. **Authorization:** Determines if the user is allowed to call the endpoints for this action.
+3. **Accounting:** Assesses whether the user is permitted to operate on a specific entity.
```
aaannz, line 30: Maybe it is my bad English, but are you sure `Accounting` means this? Everywhere I looked, accounting is about measuring, logging etc.

From the sentence in the RFC I would assume accounting is some action which can block user access.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+3. Access control filters
+4. Management interface
+5. Development resources
+
+## 1. Access Control Data
+
+Every access point in Uyuni is mapped as an **endpoint** in the database and grouped into **namespaces**.
+Access control is enforced by defining rules for these namespaces, either per **user** or **access group**.
+
+Below is a simplified ER diagram illustrating the proposed structure for access information in the database:
```
aaannz, line 68: I know we are not discussing implementation details here, but seeing the ER diagram I can't help myself.

Please **NO NUMERIC IDs**, I can't stress this enough.

Use BIGINT with identity generated always. Exception being already existing IDs.
And btw. VARCHAR is the same as TEXT in postgres, so use TEXT everywhere.

aaannz (commented):
IIUC, this RBAC implementation will not help with something like proxy admin seeing only systems below that proxy, right?

Visibility will still be based only on organization level, correct?

accepted/00000-rbac.md
rjmateus: This is a concept in user access control called Authentication, Authorization, and Accounting (AAA)
Here some links talking about it:
- https://community.fs.com/encyclopedia/aaa.html
- https://info.support.huawei.com/info-finder/encyclopedia/en/AAA.html
- https://www.strongdm.com/blog/aaa-security

accepted/00000-rbac.md
cbbayburt: That's something I also spent some time to think about. In the end, I decided to leave it possible because in most cases people will use predefined defaults and in those defaults we will make sensible choices. And when we look at the whole feature set, we'll most likely discover some cases where this makes sense.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+
+Access to XML-RPC API methods is managed through the `BaseHandler` class, the base class for all XML-RPC handlers.
+Access is granted if the requested method's name and handler match the `class_method` field of an allowed endpoint.
+
+XML-RPC and HTTP APIs share the same endpoint entries in the database. However, while HTTP requests are validated against the `endpoint` field, XML-RPC requests are validated against the `class_method` field.
+
+### Performance
+
+Access control checks query the database on every request.
+To mitigate performance impact, a user's access rules can be preloaded into memory and stored in the user's session data during login, enabling faster access control validation.
```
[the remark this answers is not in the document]
cbbayburt: I agree. But when we decide to implement caching, I think safest and easiest solution would be to simply kick the user out and require a new login whenever their permissions change.

cbbayburt:
> How do we plan to integrate all of this with the frontend? If a user does not have W on a resource, it doesn't make sense to show them UI components that will just throw an error when they try to save changes. So we would also need to make this information available to the UI in some manner and enforce checks there as well for usability.

@Etheryte the new mechanism will provide means to check information regarding access rules to use on any page to show/hide/disable sections. You'll be able to check if the user has access to something and render that part conditionally.

The hardest part is to do this on all the existing pages. We have to keep this out of scope to make it feasible for 5.1. Until then, when the user navigates to some disallowed part, we'll simply return a branded 403 page.

Once the RBAC is done, we'll have to go page-by-page to add conditions according to the user's access rules.

With the initial implementation though, we'll at least hide the disallowed entries from the left menu bar.

accepted/00000-rbac.md
cbbayburt: Noted. Thanks for the tip.

accepted/00000-rbac.md
cbbayburt: I'll try to rephrase this to be more clear

cbbayburt: > IIUC, this RBAC implementation will not help with something like proxy admin seeing only systems below that proxy, right?
> 
> Visibility will still be based only on organization level, correct?

This actually refers to "Accounting". No, at this point we won't control access at this level.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+
+### Access group
+
+An **access group** is a collection of access rules that can be assigned to users, allowing administrators to manage permissions for multiple users at once.
+Access groups can be either predefined or created by administrators.
+Predefined groups serve as replacements for existing user roles on the Uyuni server.
+
+- **`label`**: The label of the access group.
+- **`description`**: The purpose or scope of the access group.
+- **`org_id`**: The ID of the organization to which the group belongs. This field will be `null` for predefined groups.
```
rjmateus, line 110: This "vendor" groups cannot be changed by the users.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+
+#### Existing roles
+
+To ensure a smooth transition, existing roles will remain until all JSP/Struts pages are fully removed. The new access control mechanism includes default user access groups replicating the following roles:
```
rjmateus, line 114: I'm not sure about this topic. We must create access groups mapping the existing roles, that is a must.
However, some of these roles need to stay around for a longer time, even if we remove struts, since they grant access to data and not just features: "Uyuni Administrator" and "Organization administrator" (not 100% sure about the remaining).
For the ones that don't grant access to data, we could try to remove them. The part that can be tricky is the struts ACL rules, which may stop working properly.

An alternative solution for the roles that don't grant data access, and avoid touching the struts ACL could be assign those roles to all users, since the access control to the features will happen in the next mechanism.

We also must develop a automatic migration mechanism from the old roles to the new user access groups.

accepted/00000-rbac.md
aaannz: Well not according to those links @rjmateus provided. Accounting is according to those links:

```
Accounting, the final process in the framework, is all about measuring what's happening within the network. As part of the protocol, it will collect and log data on user sessions, such as length of time, type of session, and resource usage. The value here is that it offers a clear audit trail for compliance and business purposes.
```

```
Accounting

Accounting records the used service type, start time, and data traffic to collect and record the network resource usage of the user for implementing time- or traffic-based accounting and network monitoring.
```

So to me this sound like Accounting it measuring and logging and no way it can decide user has access to this or that. It is simple passive step of logging.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+- Check if a user or group has access to a specific namespace
+- List all available namespaces for a particular feature
+
+
+## 5. Development Resources
+
+After the initial development and mapping of all existing endpoints to the new structure, ongoing effort is required to keep endpoint mappings and the namespace structure up-to-date. To reduce RBAC overhead when implementing new features, the following additional resources are proposed:
+
+- An extensive Wiki guide on adding necessary endpoints with proper namespace organization
+- Stored procedures and scripts to easily grant/revoke user access for development
```
rjmateus, line 250: I would prefer to avoid store procedures as much as possible. They are hard to test and are hidden inside the database.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+- CI checks to detect any unmapped endpoints when introducing new features
+- A failsafe mechanism to prevent access to unmapped endpoints
```
rjmateus, line 252: This would be achieved by design.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+
+- **Potential for configuration drift**: Over time, there is a risk that the RBAC namespace structure and actual endpoint mappings may drift apart. If new endpoints or features are not consistently mapped or if namespace groupings are mismanaged, this can lead to security gaps or unintended access permissions.
+
+- **Learning curve for administrators and developers**: Introducing a new RBAC model with detailed namespace mapping and access modes may require training for both developers and administrators. Adjusting to this new model and understanding the finer distinctions of access control may initially be challenging.
+
+# Unresolved Questions
+[unresolved]: #unresolved-questions
+
+- **Granularity of access control:** Should the RBAC model include finer-grained permissions, such as access by specific actions (create, update, delete) within a namespace, or is the proposed "View" vs. "Modify" access mode sufficient for most cases?
```
rjmateus, line 267: In my opinion, whenever its possible we should have as fine grain as possible, with a separation between the create, update, delete, for example.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+
+- **External integration needs:** Will the RBAC system require compatibility with external authentication or access control systems, such as LDAP, and if so, what integration points are required to support this?
```
rjmateus, line 269: That I would say is part of authentication. However, we could define a mapping between LDAP user groups and suma internal access groups, and automatically assign users to a group at login time.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+
+- **Additional access groups:** Should predefined access groups be expanded to cover specific use cases beyond those mirroring existing roles?
```
rjmateus, line 271: That is something we will find out, and we can react according to feedback from customers. One extra groups could be the read-only user that we have now for API only. that can be a new access group and stop relying on method names and a flag on user definition.

rjmateus (commented):
Looks really good Can. Just some small clarification, otherwise looks great to me.

accepted/00000-rbac.md
rjmateus: A now I'm trying to figure out from where I have read that statement :-|

rjmateus: @aaannz thank you for pointing it out. Resource access should be one step of the `Authorization` process. However, This RFC is focusing only at the part `...perform operation O...`. This would be already a big step, since the role/access control is spread across all the code (from API definition, to database access factories), and this is something we must clean.
Next step can be find a better way to perform resource access check. What do you think?

accepted/00000-rbac.md
```diff
@@ -0,0 +1,271 @@
...
+*[View in dbdiagram.io](https://dbdiagram.io/d/Uyuni-RBAC-672e18b3e9daa85acacbddd0)*
+
+### Endpoint
+
+An **endpoint** represents a specific access point that can be called by a client, such as a web page URL, an internal API endpoint used by a web page, or a public API endpoint.
+
+- **`endpoint`**: The accessible URI of the endpoint.
+- **`class_method`**: The Java class that handles incoming requests to the endpoint, such as a controller class for web endpoints or an API handler for API endpoints.
+- **`http_method`**: The HTTP method (e.g., GET, POST, PUT, DELETE) accepted by the endpoint. If multiple methods are supported, each should be defined as a separate endpoint since they typically serve different purposes.
+- **`scope`**: Indicates whether the endpoint is accessible through the web UI (including internal API calls) or the public API.
```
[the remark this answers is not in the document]
cbbayburt: I'll add a `auth_required` column to the RFC to exclude the URLs outside of authentication space (like login, about, API docs, etc.).

I'm not 100% sure about the second flag. It may or may not be needed, depending on the details of the implementation. So I'll leave it out for now.

accepted/00000-rbac.md
cbbayburt: What do you mean by data access?

> An alternative solution for the roles that don't grant data access, and avoid touching the struts ACL could be assign those roles to all users, since the access control to the features will happen in the next mechanism.

Yes, the problem is that the current access control with static roles is embedded everywhere (React as well) and trying to remove these is going to cost a lot of time. So instead, my plan is to leave them in, hide them from the UI (replaced by the new ones), and assign all these static roles to every existing and future user so they effectively won't do any access control.

However, I'm wondering if we should keep an unrestricted bypass mechanism in the filters for the Uyuni and Org Admin roles like:

```java
    public boolean canAccess(User user) {
        return (user.hasRole(RoleFactory.ORG_ADMIN) && user.getOrg().equals(getOrg())) ||
                user.hasRole(RoleFactory.SAT_ADMIN);
    }
``` 

or should we rely on RBAC only? Are there some areas that won't be covered by RBAC?

accepted/00000-rbac.md
cbbayburt: I agree. These will be just throwaway tools to help when we're implementing RFC. There's no plan to ship or persist them.

accepted/00000-rbac.md
Etheryte: Sounds good to me. 👍

accepted/00000-rbac.md
cbbayburt: Yes, I meant to just point to that.

accepted/00000-rbac.md
rjmateus: You can try in the existing implementation. If a user is part of organization and has not roles, and are not assigned to any systems groups, then the user cannot see any system. If the user has the role Org Admin, then it can see all the systems that are part of that organization.
So currently, the roles are not just to control access to pages but also access to data.

Etheryte:
Also a small ping for @thesp0nge, if you have the time to give this a read, any input you might have on this proactively would be much appreciated.

accepted/00000-rbac.md
cbbayburt: Oh ok now I get it. So those superuser logic must stay in the new implementation in any case. All the others can go once we're ready to strip their logic from the individual pages.

thesp0nge:
I'm sorry but for such kind of requests you really want to ask Wiktor, your security champion.

accepted/00000-rbac.md
cbbayburt: Sorry, it seems I mixed up all the terms. I rewrote the part so it's hopefully clearer now.

cbbayburt:
@Etheryte @aaannz @rjmateus, all the discussions should be addressed now. Please let me know if there's something I missed.

Etheryte (approved):
From my side this looks good, I think this is high level enough that most of the detailed discussions will happen in the specific implementation parts later.

cbbayburt:
> From my side this looks good, I think this is high level enough that most of the detailed discussions will happen in the specific implementation parts later.

Thanks for the review, Karl. Yeah I think this level of detail should be enough to get everyone on the same page and make the right decisions from the beginning so we won't have to scratch the whole design.

rjmateus (approved):
Looks good to me. Thank you for preparing it

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+After the initial development and mapping of all existing endpoints to the new structure, ongoing effort is required to keep endpoint mappings and the namespace structure up-to-date. To reduce RBAC overhead when implementing new features, the following additional resources are proposed:
+
+- An extensive Wiki guide on adding necessary endpoints with proper namespace organization
+- Stored procedures and scripts to easily grant/revoke user access for development
+- CI checks to detect any unmapped endpoints when introducing new features
+- A failsafe mechanism to prevent access to unmapped endpoints
+- Startup checks in the Uyuni Server to log any misconfigurations within a specific deployment
+
+# Drawbacks
+[drawbacks]: #drawbacks
```
szachovy, line 270: Spacing throughout the document is inconsistent, can you parse it through https://github.com/DavidAnson/markdownlint ?

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+![ER diagram](images/00000-rbac-er-diagram.png)
+*[View in dbdiagram.io](https://dbdiagram.io/d/Uyuni-RBAC-672e18b3e9daa85acacbddd0)*
+
+### Endpoint
+
+An **endpoint** represents a specific access point that can be called by a client, such as a web page URL, an internal API endpoint used by a web page, or a public API endpoint.
+
+- **`endpoint`**: The accessible URI of the endpoint.
+
+- **`class_method`**: The Java class that handles incoming requests to the endpoint, such as a controller class for web endpoints or an API handler for API endpoints.
```
admd, line 78: Will this be required field or optional? What will we have in case of struts actions ?

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+
+- **`http_method`**: The HTTP method (e.g., GET, POST, PUT, DELETE) accepted by the endpoint. If multiple methods are supported, each should be defined as a separate endpoint since they typically serve different purposes.
+
+- **`auth_required`**: Indicates if the endpoint requires authorization.
```
admd, line 82: If endpoint, will we have a separate namespace for such endpoints? What endpoints would usually fall in this category?

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+
+### Namespace
+
+A **namespace** is a logical grouping of endpoints that performs a specific task and defines the smallest unit of access control.
```
admd, line 86: How will we make sure that we don't miss endpoints which are not part of any namespace? my question is more going forward - will we have any tooling which identify this?

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+
+- **`namespace`**: A label representing the namespace, expressed as a dot-separated string of components. Each component corresponds to a level of hierarchy for a task, with related tasks sharing the same components at higher levels.
+
+  *Example:*
+    - `clm.project.list`: Allows viewing the CLM projects list.
+    - `clm.project.details`: Allows viewing or modifying the details of a CLM project.
+
+- **`access_mode`**: Defines the access level for each namespace as either "View" (R) or "Modify" (W). Many namespaces will have separate entries for each mode, with different endpoints for each purpose.
```
admd, line 94: I believe actions like export CSV will go under view(R) and delete under W, right?

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+
+- **`http_method`**: The HTTP method (e.g., GET, POST, PUT, DELETE) accepted by the endpoint. If multiple methods are supported, each should be defined as a separate endpoint since they typically serve different purposes.
+
+- **`auth_required`**: Indicates if the endpoint requires authorization.
+
+### Namespace
+
+A **namespace** is a logical grouping of endpoints that performs a specific task and defines the smallest unit of access control.
+
+- **`namespace`**: A label representing the namespace, expressed as a dot-separated string of components. Each component corresponds to a level of hierarchy for a task, with related tasks sharing the same components at higher levels.
```
admd, line 88: can an end point be part of more than 1 namespace?

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+
+- **`access_mode`**: Defines the access level for each namespace as either "View" (R) or "Modify" (W). Many namespaces will have separate entries for each mode, with different endpoints for each purpose.
+
+  For example, `clm.project.details` **[R]** allows viewing a project's details, where `clm.project.details` **[W]** allows modifying them.
+
+- **`description`**: A clear description of what a namespace grants access to.
+
+- **`scope`**: Indicates whether the endpoint is accessible through the web UI (including internal API calls) or the public API.
+
+#### Organization of namespaces
```
admd, line 102: We don't have any table to represent this in our sample database schema, right? will we have some kind of inheritance here that if I have access at the top level namespace, I get access to everything else. And if yes, can as an admin, still change it to be more granular.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+The other existing roles in Uyuni are as follows:
+
+- Activation Key Administrator
+- Configuration Administrator
+- Image Administrator
+- Channel Administrator
+- System Group Administrator
+- Read-only API User
+
+Each of these roles grants access to different features in Uyuni. This static access control is enforced individually within each feature. Stripping these control checks throughout the system requires significant effort. Therefore, in the initial phase, the logic of these static roles will coexist with their RBAC counterparts, but will be bypassed by assigning them to all existing and future users. This will effectively make RBAC the exclusive access control mechanism.
```
admd, line 137: This is not clear to me. Yes, we will keep the existing roles so we don't break the users setup when the migrate to 5.1.  But wow are we planning to get rid of them? Will we allow administrators to still play with old roles ? Having new and old permissions setup could be pretty confusing for the admin to understand.

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+
+
+## 2. User Access Information
+
+Users can be granted access either individually or through access groups.
+
+Users assigned to an access group inherit all permissions defined within that group.
+While administrators can grant individual users additional permissions beyond those provided by their assigned group, they cannot revoke permissions set by the group. Additionally, users can belong to multiple access groups and inherit all associated access rules.
```
admd, line 145: hmm, I didn't get this. Here additional permissions means which are not covered by groups or is it more in a sense that those permissions that user got through groups can be overridden by individually assigning to the user?

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+
+Predefined access groups are designed to meet the needs of most environments; however, administrators can create custom groups to address specific requirements as needed.
```
admd, line 147: Does this mean that namespaces will be static and administrator cannot change them or create new ones or will we provide the same flexibility(not considering here if that's good or bad) there as well that we are providing here in case of access groups?

accepted/00000-rbac.md
```diff
@@ -0,0 +1,285 @@
...
+- Grant view access to the entire images feature:
+
+  ```
+  Grant 'View' on 'cm.*' to 'Alice'
+  ```
+
+- Prevent access to sensitive image store information:
+
+  ```
+  Revoke 'View' on 'cm.store.details' from 'Alice'
```
admd, line 183: If Alice got this 'View' access through one accessgroup but then here Alice got that permissions revoked, I expect this would take precedence or?

admd (commented):
Only done have way through. I have Left some questions for my understanding.

accepted/00000-rbac.md
rjmateus: This is mandatory, especially for the XML-RPC API endpoint, where we use this field to control access. For Structs and spark-only endpoints this is not really needed, but we can populate it.

accepted/00000-rbac.md
rjmateus: If I got it right, the idea is to have this in an indirect way, by playing with the label name. For example:
namespace: admin
namespace: admin.createUser

accepted/00000-rbac.md
rjmateus: There are 2 roles that needs to stay for now, which are `satadmin` and `orgadmin`.
Other roles should be replaced with new user groups in the new system.
Internally, we may need to keep the old roles because of struts ACL configurations.

accepted/00000-rbac.md
cbbayburt: So when you're implementing a new URL for a feature, first thing you'll have to do is to add necessary entries (endpoint and namespace info) to the RBAC tables using schema migration. Otherwise you won't be able to visit the URL when testing yourself.

accepted/00000-rbac.md
cbbayburt: Correct. AFAICS this separation is clear enough for any action. Management UI will provide proper descriptions for each namespace and mode so it'll be clear to users.

Ideally, the UI should have free text search so for example you'll be able to search for the keyword CSV and see what namespaces are available for any CSV related action.

accepted/00000-rbac.md
rjmateus: The idea is to grant more permissions to the user. User will always have access to all endpoints granted in all the groups, plus all individual permissions grant to the user directly.

accepted/00000-rbac.md
cbbayburt: > If I got it right, the idea is to have this in an indirect way, by playing with the label name.

Correct. By default, granting access to a parent namespace doesn't automatically give access to the children. For that, you'll need to explicitly say "grant access to this namespace **and everything that falls below it**".

accepted/00000-rbac.md
rjmateus: Namespaces and access groups are just different levels of putting endpoints together. Namespaces are static and controled by us. Users can then create user groups linked to the endpoint in any shape or form, that will put together one or more namespaces.
On top of that, we will also have a few user groups controlled by us and the goal is that those groups match the existing roles.

accepted/00000-rbac.md
rjmateus: If it's removed from User direct grant but is still present in the group, then Alice still has access to the feature. All endpoints are revoked by default, and we are always adding/grant access in all levels we can configure.

ktsamis (commented):
So from what I understand with the new UI pages or API endpoints we would be using/testing it somehow like this, right? I put a few user stories with an example feature just to get an idea if I understand it correctly

User Story 1: Admin assigns access groups to users
As an administrator,
I want to assign a user to an access group,
So that they can access the features allowed by the group.

User Story 2: User accesses a feature within their namespace
As a user,
I want to access features and operations allowed by my role,
So that I can perform my job functions.

User Story 3: Unauthorized user attempts to access restricted resources
As a user,
I want the system to prevent me from accessing unauthorized features,
So that security and access rules are maintained.

User Story 4: Admin customizes access for a user
As an administrator,
I want to customize individual permissions for a user,
So that their access meets specific requirements.

User Story 5: Admin creates a new access group
As an administrator,
I want to create new access groups with specific permissions,
So that I can manage user access more efficiently.

User Story 6: User loses access when removed from a group
As an administrator,
I want users to lose permissions immediately when removed from an access group,
So that their access remains strictly controlled.

User Story 7: Permissions propagate correctly for multi-group users
As a user,
I want permissions from multiple groups to apply seamlessly,
So that my access covers all assigned roles.

```
Feature: Multi-group user access

Scenario: User belongs to multiple access groups with overlapping permissions
  Given a user "testuser" belongs to the access groups:
    | Group Name        | Namespace             | Access Mode |
    | Image Viewer      | cm.image.list         | View        |
    | Profile Manager   | cm.profile.*          | Modify      |
  When "testuser" logs in
  Then "testuser" should have the following combined permissions:
    | Namespace             | Access Mode |
    | cm.image.list         | View        |
    | cm.profile.*          | Modify      |
  And "testuser" should be able to modify image profiles
  And "testuser" should not be able to delete images
```

User Story 8: Audit access logs
As an administrator,
I want to track and log user actions,
So that I can audit access and detect misuse.

accepted/00000-rbac.md
srbarrios: I support the idea of Abid to have some kind of tool part of the unit tests that assure that we don't miss any. Let's not rely only to the side effect even if easy to see. But let's be sure that someone adding a new endpoint in a new Pull Request have feedback if that part is missing.

accepted/00000-rbac.md
rjmateus: From telemetry is not possible o see, we can only see that and endpoint is called. What we can do is we log all the called, be able to calculate a call trace and deduct what could have be the interaction path.

rjmateus:
@cbbayburt can this rfc be merged? Thank you

cbbayburt:
> @cbbayburt can this rfc be merged? Thank you

https://github.com/SUSE/spacewalk/issues/24196 is still open, and Welder is currently on it. I'll review and update the RFC before merging it after everything is done.

admd:
@cbbayburt can we merge this or is there still something pending?

PR #96, opened 2024-11-14
An improved workflow for maintaining Salt

meaksh:
Read the RFC [here](https://github.com/uyuni-project/uyuni-rfc/blob/master-new-maintaining-workflow-for-salt/accepted/0000-salt-new-maintaining-workflow.md)

### DIFF
--- a/accepted/0000-salt-new-maintaining-workflow.md
+++ b/accepted/0000-salt-new-maintaining-workflow.md
@@ -224,8 +224,12 @@ This "openSUSE/salt-extensions" repository will contain:
 When it comes to OBS, we will use the same SCM integration and OBS subprojects schema than the proposed for the main Salt codebase. Unique workflow for Salt and Salt Extensions.
 
 - PoC:
-  - https://github.com/meaksh/test-repo-1
-  - https://build.opensuse.org/package/show/home:PSuarezHernandez:tests/salt-extensions
+  * https://github.com/meaksh/test-repo-1
+  * https://build.opensuse.org/package/show/home:PSuarezHernandez:tests/salt-extensions
+
+- Example PR:
+  * https://github.com/meaksh/test-repo-1/pull/6
+  * https://build.opensuse.org/project/show/home:PSuarezHernandez:tests:github:CI:meaksh:test-repo-1:PR-6
 
 NOTE: We are using a single GitHub repo and single OBS package which provides all different salt-extensions RPM packages. This is preferred against having a separated GitHub repositories and OBS package for each Salt Extension, as it will reduce the number of submissions, maintenance incidents and resources needed.
 
--- a/accepted/0000-salt-new-maintaining-workflow.md
+++ b/accepted/0000-salt-new-maintaining-workflow.md
@@ -13,7 +13,7 @@ Our current workflow for maintaining Salt requires manual user intervention afte
 
 All these steps needs to be performed manually, with the help of some tooling, to eventually create a manual Submit Request to our Salt package in OBS.
 
-With Salt Extensions appearing now in the upcoming Salt 3008 release, we want to introduce a new workflow that suits better and solves some the limitations we currently have. 
+With Salt Extensions appearing now in the upcoming Salt 3008 release, we want to introduce a new workflow that suits better and solves some the limitations we currently have.
 
 The purpose of this RFC is:
 - Define an new workflow for Salt that reduces user intervention to zero after a given PR is merged in `openSUSE/salt` repository until getting the package ready to consume at OBS.
@@ -27,7 +27,7 @@ The purpose of this RFC is:
 In this new workflow the `openSUSE/salt` GitHub repository will become the unique source of trust, and it will contain:
 
 - Salt codebase
-- Packaging artifacts: spec file, changelog and extra sources
+- Packaging artifacts: spec file, changelogs, \_multibuild file and extra sources
 - OBS workflow file
 
 Taking inspiration from the [OBS/SCM integration guide](https://openbuildservice.org/help/manuals/obs-user-guide/cha-obs-scm-ci-workflow-integration), the new workflow will use OBS workflows and GitHub Webhooks to automate pulling the changes from GitHub to OBS.
@@ -55,7 +55,7 @@ For `systemsmanagement:saltstack:products` OBS target, it is not really necessar
 
 ### Packaging artifacts
 
-All current extra "Sources" files in our RPM package, together with spec file and changelog file will go now to a `pkg/suse/` directory in `openSUSE/salt`:
+All current extra "Sources" files in our RPM package will go now to a `pkg/suse/` directory in `openSUSE/salt`, together with the spec file, the different maintained changelogs files:
 
 ```
 pkg/suse/README.SUSE
@@ -63,25 +63,67 @@ pkg/suse/html.tar.bz2
 pkg/suse/salt-tmpfiles.d
 pkg/suse/transactional_update.conf
 pkg/suse/update-documentation.sh
+pkg/suse/mkchlog.sh
+pkg/suse/_multibuild
 pkg/suse/salt.spec
-pkg/suse/salt.changes
+pkg/suse/changelogs/factory/salt.changes
+pkg/suse/changelogs/sles15sp2/salt.changes
+pkg/suse/changelogs/sles15sp3/salt.changes
+pkg/suse/changelogs/sles15sp4/salt.changes
+pkg/suse/changelogs/sles15sp5/salt.changes
 ```
 
 This is the place now where all those files will be maintained.
 
-#### Salt RPM changelog
+#### Salt RPM changelogs
 
-As mentioned this is now at `pkg/suse/salt.changes` in `openSUSE/salt` GitHub repo.
+As mentioned, the changelog files are now maintained in the `openSUSE/salt` GitHub repo, under `pkg/suse/changelogs/` directory.
 
-When creating a PR to `openSUSE/salt` the user must also include the corresponding changes to the changes file, that can be generated as usual with `osc vc`.
+Our packaging artifacts will contain a `mkchlog.sh`, which is a helper script to generate a changelog entry to all maintained changelog in one shot. Something like this:
+
+```bash
+echo "Generating changelog entry for Salt package"
+if ! osc vc _temp.changes;
+then
+    exit 1;
+fi
+
+echo "Update changelog files"
+echo >> _temp.changes
+echo "$(cat _temp.changes salt.changes)" > salt.changes
+git add salt.changes
+
+for i in $(ls changelogs/*/salt.changes); do
+    echo "$(cat _temp.changes $i)" > $i
+    git add $i
+done
+
+rm _temp.changes
+```
+
+When creating a PR to `openSUSE/salt` the user must also include the corresponding changelog entry for all maintained changelog files.
 
 Similarly to the main Uyuni repository, we should add a GitHub action to warn the user in case no changelog entry is added in the PR.
+
+NOTE: I think it is better to decouple commit messages (focus on developers) from changelog entries (focus on users/customers), so I prefer to not use commit messages from "openSUSE/salt" to autogenerate the changelog entries but rather to manually write a meaningful changelog message to be included in your PR as part of your changes. Similarly to what we do in other Uyuni repositories.
 
 ### OBS project structure
 
 #### `systemsmanagement:saltstack:github/salt`
 
-This OBS package will only contain `_multibuild` file and a `_service` file that should look like: 
+This OBS package will be configured as "SCM managed", via Meta configuration, as the following:
+
+```
+<package name="salt" project="systemsmanagement:saltstack:github">
+  <title/>
+  <description/>
+  <scmsync>https://github.com/openSUSE/salt?subdir=pkg/suse/#openSUSE/release/xxxx</scmsync>
+</package>
+```
+
+The `_service` file together with the rest of packaging artifacts at `pkg/suse/` directory will be automatically pulled by `scmsync`.
+
+After this, the rest of the files will be automatically pulled by the services, as they are enabled inside the `_service` file that should look like: 
 
 ```
 <services>
@@ -91,7 +133,7 @@ This OBS package will only contain `_multibuild` file and a `_service` file that
     <param name="versionformat">@PARENT_TAG@</param>
     <param name="versionrewrite-pattern">v(.*)</param>
     <param name="revision">openSUSE/release/xxxx</param>
-    <param name="extract">pkg/suse/salt.*</param>
+    <param name="extract">pkg/suse/changelogs/factory/salt.changes</param>
   </service>
   <service name="set_version" />
   <service name="tar" mode="buildtime"/>
@@ -102,13 +144,13 @@ This OBS package will only contain `_multibuild` file and a `_service` file that
 </services>
 ```
 
-The rest of the files will be automatically pulled by the service, as they are enabled here. This package will be automatically refreshed by OBS at any new commit at `openSUSE/release/xxxx` branch.
+NOTE: This package will be automatically refreshed by OBS at any new commit at `openSUSE/release/xxxx` branch.
 
 #### `systemsmanagement:saltstack/salt`
 
-This is our ready-to-consume OBS package. It is linked to `systemsmanagement:saltstack:github/salt`, but here we disable the services and manually run them to get the spec file, changelog and obsinfo/obscpio files, so the package can be submitted to openSUSE or SLE.
+This is our ready-to-consume OBS package. We set the devel package (osc setdevelpackage) for this OBS package to `systemsmanagement:saltstack:github/salt`, but here we disable the services and manually run them to get the spec file, changelog and obsinfo/obscpio files, so the package can be submitted to openSUSE or SLE.
 
-The `_service` file should look like:
+The `_service` file here should look like:
 
 ```
 <services>
@@ -118,7 +160,9 @@ The `_service` file should look like:
     <param name="versionformat">@PARENT_TAG@</param>
     <param name="versionrewrite-pattern">v(.*)</param>
     <param name="revision">openSUSE/release/xxxx</param>
-    <param name="extract">pkg/suse/salt.*</param>
+    <param name="extract">pkg/suse/salt.spec</param>
+    <param name="extract">pkg/suse/_multibuild</param>
+    <param name="extract">pkg/suse/changelogs/factory/salt.changes</param>
   </service>
   <service name="set_version" mode="manual" />
   <service name="tar" mode="buildtime"/>
@@ -196,11 +240,11 @@ This way, we ensure `salt.spec` and `salt.changes` and obscpio/obsinfo files get
 
 I've implemented this structure and automation here:
 
-- GitHub repository: https://github.com/meaksh/salt/ (`openSUSE/devel/master` branch)
+- GitHub repository: https://github.com/meaksh/salt/ (`openSUSE/devel/master-version-2` branch)
 - OBS:
-  * https://build.opensuse.org/package/show/home:PSuarezHernandez:tests/salt
-  * https://build.opensuse.org/package/show/home:PSuarezHernandez:tests:github/salt
-  * https://build.opensuse.org/package/show/home:PSuarezHernandez:tests:github:CI:....:PR-XX/salt
+  * https://build.opensuse.org/package/show/home:PSuarezHernandez:tests/fake-package
+  * https://build.opensuse.org/package/show/home:PSuarezHernandez:tests:github/fake-package
+  * https://build.opensuse.org/package/show/home:PSuarezHernandez:tests:github:CI:....:PR-XX/fake-package
 
 - Example PR:
   * https://github.com/meaksh/salt/pull/10
@@ -247,8 +291,9 @@ NOTE: We are using a single GitHub repo and single OBS package which provides al
 [alternatives]: #alternatives
 
 1. Stick to our current workflow based on "salt-packaging" -> The workflow doesn't currently fit with Salt Extensions and we don't want to have different workflows between Salt and Salt Extensions.
-1. One dedicated GitHub repository and OBS package per each Salt Extension -> It won't save resources and will cause more submissions.
-2. The usage of "git submodules" as an alternative to adding the Salt Extensions sources manually would make it tricky to generate patches manually and also to integrate with "obs_scm".
+2. One dedicated GitHub repository and OBS package per each Salt Extension -> It won't save resources and will cause more submissions.
+3. The usage of "git submodules" as an alternative to adding the Salt Extensions sources manually would make it tricky to generate patches manually and also to integrate with "obs_scm".
+4. Use OBS `scmsync` integration -> while this allows to integrate even the `_service` file into the GitHub repo, it does not work well with SCM/CI workflow, therefore we cannot test the build of the package for an open PR because the branched package is not reflecting the actual changes on the PR.
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/accepted/0000-salt-new-maintaining-workflow.md
+++ b/accepted/0000-salt-new-maintaining-workflow.md
@@ -257,7 +257,7 @@ For the Salt Extensions that are packaged separately from the main Salt package,
 This "openSUSE/salt-extensions" repository will contain:
 - a common salt-extension spec file that will generate all RPM packages
 - The sources for each Salt Extension we package
-- A changelog file
+- The changelog files for each maintained codestream
 - OBS workflow file
 
 When it comes to OBS, we will use the same SCM integration and OBS subprojects schema than the proposed for the main Salt codebase. Unique workflow for Salt and Salt Extensions.
--- a/accepted/0000-salt-new-maintaining-workflow.md
+++ b/accepted/0000-salt-new-maintaining-workflow.md
@@ -113,7 +113,7 @@ The `_service` file should look like:
     <param name="scm">git</param>
     <param name="versionformat">@PARENT_TAG@</param>
     <param name="versionrewrite-pattern">v(.*)</param>
-    <param name="revision">openSUSE/devel/master</param>
+    <param name="revision">openSUSE/release/xxxx</param>
     <param name="extract">pkg/suse/salt.*</param>
   </service>
   <service name="set_version" mode="manual" />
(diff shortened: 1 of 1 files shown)


accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+pkg/suse/salt.changes
+```
+
+This is the place now where all those files will be maintained.
+
+#### Salt RPM changelog
+
+As mentioned this is now at `pkg/suse/salt.changes` in `openSUSE/salt` GitHub repo.
+
+When creating a PR to `openSUSE/salt` the user must also include the corresponding changes to the spec file, that can be generated as usual with `osc vc`.
```
vzhestkov, line 72: Not sure if I get this point right. Maybe it says `spec file` but the actual meaning is different, not sure, could you please clarify it?

meaksh: Woops, sorry. I meant "changelog" file :smile:
Let me fix this

vzhestkov: Do I understand it right, that in this case we will add changelog entries manually to the changelog file or manually but with `osc vc`, still not fully clear here. With the osc services there is way to use commit messages as a changlog items, don't we want to use it this way?

meaksh: I'm proposing here to manually add the changelog entry to `salt.changes` by using the `osc vc` command directly your Git tree, so the generated changelog entry (and header) can be included together with the code changes in the PR to `openSUSE/salt` repo.

This way a single PR to `openSUSE/salt` repo could contain all possible different bits: code changes, specfile changes, changelog entries, artifacts changes.

See this example PR: https://github.com/meaksh/salt/pull/10/files

Now that you mentioned the "osc services" for the changelog, I realized that the current proposal is not covering the fact that we need to maintain different changelogs depending on the targeted codestreams (as they are not aligned)

I'll have some thoughts and clarify this on the RFC text.

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+The `_service` file should look like:
+
+```
+<services>
+  <service name="obs_scm" mode="manual">
+    <param name="url">https://github.com/openSUSE/salt.git</param>
+    <param name="scm">git</param>
+    <param name="versionformat">@PARENT_TAG@</param>
+    <param name="versionrewrite-pattern">v(.*)</param>
+    <param name="revision">openSUSE/devel/master</param>
```
agraul, line 116: I was not aware we used an `openSUSE/devel/master` branch. Is this a new branch we will use?

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+
+- Example PR:
+  * https://github.com/meaksh/salt/pull/10
+  * https://build.opensuse.org/package/show/home:PSuarezHernandez:tests:github:CI:meaksh:salt:PR-10/salt
+
+Feel free to open new PRs against `openSUSE/devel/master` to see this in action.
+
+### Salt Extensions
+
+#### Builtin extensions
```
agraul, line 209: This has been part of the other RFC, but for me it's still not so clear that we can have "builtin extensions". How do we publish these to PyPI from the main repository? How do we get them to show up on https://extensions.saltproject.io/?

agraul (commented):
I generally like this design. I left a few questions in line but saved the big one for now: How do we package the Salt Bundle? Currently we have a split-brain problem where some of the sources are just in OBS. Can we integrate them into this workflow?

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+The sources for the builtin Salt Extensions will be located together with the main Salt codebase at the `openSUSE/salt` GitHub repository. No new packages or subpackages will be created for these extensions as they will be part of the main `python3*-salt` package.
+
+If a fix is needed for any of the builtin extensions, workflow would be the same as for a code fix in the main Salt package.
+
+#### Packaged Salt Extensions
```
rjmateus, line 214: Will this extra salt extension be installed by default on the suma container image? If not, how can users consume these extensions?

rjmateus (commented):
looks good to me, just a few questions to help me better understand the solution

accepted/0000-salt-new-maintaining-workflow.md
vzhestkov: I think in case of the server container with salt-master all of the extensions should be installed as there are some parts on the master side which could use even the modules which makes sense on the client side to do some calculations and for salt-ssh as well as one of the cases (the state rendering is performed on the master in this case)

accepted/0000-salt-new-maintaining-workflow.md
meaksh: I used `openSUSE/devel/master` here just as an example, it should point to the eventual `openSUSE/release/3008.x` branch.

Currently `openSUSE/devel/master` is just the devel branch I created with upstream `master` branch + our patches partially rebased on top (excluding patches to extensions).

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+This workflow will take care of:
+
+- Setting up a new subproject at `systemsmanagement:saltstack:github:CI:...:PR-XXXX/salt` for every incoming PR to build the Salt package according to the changes in the PR.
+- Triggering the services at `systemsmanagement:saltstack:github/salt` on any new push to `openSUSE/release/xxxx` to build the package accordingly.
+
+### Making OBS packages ready to be submitted to Maintenance
+
+Since the package at `systemsmanagement:saltstack:github/salt` has "services" enabled, and we cannot enable/disable services using OBS workflows, this means this package is not yet ready to be submitted to openSUSE or SLE, as they don't accept enabled services on their submissions. We must disable the services.
+
+In order to do this we will use an Jenkins job that monitors when a new build is done at `systemsmanagement:saltstack:github/salt` to trigger the following actions at the main `systemsmanagement:saltstack/salt` package:
```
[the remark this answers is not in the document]
meaksh: Oops, sorry, there is a mistake with the branch names here.

Both should be pointing to the same `openSUSE/release/xxxx` branch. I'll fix the RFC text.

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+   * branched and removed automatically from `systemsmanagement:saltstack:github/salt` by OBS workflow.
+
+The same OBS structure will apply to all our OBS targets: `products`, `products:testing` and `products:next`.
+
+### Packaging artifacts
+
+All current extra "Sources" files in our RPM package, together with spec file and changelog file will go now to a `pkg/suse/` directory in `openSUSE/salt`:
+
+```
+pkg/suse/README.SUSE
```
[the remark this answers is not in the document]
meaksh: Not really. We won't be carrying patch files anymore in our new OBS packages, as any code change will automatically inside the source tarball by the `obs_scm` integration.

The exception to this would be EMBARGOED bugs, where we cannot proceed publicly via GitHub, so we will put a patch file manually in IBS than will be removed the bug is public and we push the changes to GitHub.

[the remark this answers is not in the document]
m-czernek: A side question; currently, we have a lot of commits that are not in upstream, due to various reasons. With `salt-packaging`, we can easily check which patches are in upstream, and which patches are not. With removing the patch workflow, we're losing this insight.

Do you perceive this as a problem? Are we still sticking with the "fork & cherrypick commits" development style, or are we moving towards rebasing more frequently?

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+    <param name="compression">gz</param>
+  </service>
+</services>
+```
+
+The rest of the files will be automatically pulled by the service, as they are enabled here. This package will be automatically refreshed by OBS at any new commit at `openSUSE/release/xxxx` branch.
+
+#### `systemsmanagement:saltstack/salt`
+
+This OBS package is a branch from `systemsmanagement:saltstack:github/salt`, where we disable the services and manually run them to get the spec file, changelog and obsinfo/obscpio files, so the package can be submitted to openSUSE or SLE.
```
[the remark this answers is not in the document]
meaksh: I'll rephrase this

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+```
+
+This workflow will take care of:
+
+- Setting up a new subproject at `systemsmanagement:saltstack:github:CI:...:PR-XXXX/salt` for every incoming PR to build the Salt package according to the changes in the PR.
+- Triggering the services at `systemsmanagement:saltstack:github/salt` on any new push to `openSUSE/release/xxxx` to build the package accordingly.
+
+### Making OBS packages ready to be submitted to Maintenance
+
+Since the package at `systemsmanagement:saltstack:github/salt` has "services" enabled, and we cannot enable/disable services using OBS workflows, this means this package is not yet ready to be submitted to openSUSE or SLE, as they don't accept enabled services on their submissions. We must disable the services.
```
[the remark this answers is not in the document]
meaksh: Hmmm, interesting. I'll give it a try :+1:

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
meaksh: Fixed. I hope it does not create confusion now.

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
meaksh: This is a good point! We should probably want to have a way to easily identify what is upstreamed and what is not.

I'll elaborate this a bit on the RFC text. Thanks for the note :+1:

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
rjmateus: thank you for the clarification :)

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
rjmateus: thank you @m-czernek , that was exactly my concern when I made the comment. Commits that are not yet upstrem, and currently are maintained on salt-packaging project. 
We we start merging commit to our salt project that are not merge upstream yet, it can make it harder to integrate upstream version and know what is merge already or not.
@meaksh thank you for look into this topic.

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+The sources for the builtin Salt Extensions will be located together with the main Salt codebase at the `openSUSE/salt` GitHub repository. No new packages or subpackages will be created for these extensions as they will be part of the main `python3*-salt` package.
+
+If a fix is needed for any of the builtin extensions, workflow would be the same as for a code fix in the main Salt package.
+
+#### Packaged Salt Extensions
+
+For the Salt Extensions that are packaged separately from the main Salt package, we will create a separated GitHub repository where we will maintain these extensions.
+
+This "openSUSE/salt-extensions" repository will contain:
+- a common salt-extension spec file that will generate all RPM packages
```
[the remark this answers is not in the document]
rjmateus: ok, looks fine. But it has a high risk of growing a lot, depending on the number of extensions being packaged.

accepted/0000-salt-new-maintaining-workflow.md
rjmateus: thank for the clarification. I would upvote on include all the extension on the uyuni/suma container image.

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,256 @@
...
+
+- A bit more complex OBS structure than the current one. Including `obs_scm` service.
+- Having to still rely on Jenkins to get the packages ready to be released.
+
+# Alternatives
+[alternatives]: #alternatives
+
+1. Stick to our current workflow based on "salt-packaging" -> The workflow doesn't currently fit with Salt Extensions and we don't want to have different workflows between Salt and Salt Extensions.
+1. One dedicated GitHub repository and OBS package per each Salt Extension -> It won't save resources and will cause more submissions.
+2. The usage of "git submodules" as an alternative to adding the Salt Extensions sources manually would make it tricky to generate patches manually and also to integrate with "obs_scm".
```
agraul, line 251: Can you expand on those points a bit? In particular, what makes it tricky to generate patches when sources live in separate repositories opposed to separated sub-directories in a single repository? And why is that harder to integrate separate repositories?

I'm not saying we should go that way, I just want to understand the trade-offs.

accepted/0000-salt-new-maintaining-workflow.md
meaksh: I've added more text to the RFC on how to deal with different maintained changelogs

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,252 @@
...
+
+This is the place now where all those files will be maintained.
+
+#### Salt RPM changelog
+
+As mentioned this is now at `pkg/suse/salt.changes` in `openSUSE/salt` GitHub repo.
+
+When creating a PR to `openSUSE/salt` the user must also include the corresponding changes to the changes file, that can be generated as usual with `osc vc`.
+
+Similarly to the main Uyuni repository, we should add a GitHub action to warn the user in case no changelog entry is added in the PR.
```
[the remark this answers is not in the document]
meaksh: This should be now clarified on the RFC text

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
meaksh: Done

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
meaksh: I've added some notes about possible usage of SCMSYNC feature from OBS, but after my tests, it does not really work fine when it comes to integrate with the SCM/CI workflow, so the changes on a PR are not actually reflected in the `:github:CI:PR-XXXX` package as seems to be interfering with the SCMSYNC flag.

I gave it a try here:
- OBS package: https://build.opensuse.org/package/show/home:PSuarezHernandez:tests:github/fake-package
- Test PR: https://github.com/meaksh/salt/pull/12
- OBS PR package: https://build.opensuse.org/package/show/home:PSuarezHernandez:tests:github:CI:meaksh:salt:PR-12/fake-package

^^ notice how `salt.changes` on the OBS PR package is not reflecting the actual changes on the PR, and also missing `branch_request` file as we would expect.

I opted for not using `SCMSYNC` and just carrying the `_service` file out of the GitHub repo, which might be actually convenient as differnt OBS/IBS maintained project would has some particularities in its `_service` file, i.a. to point to the right changelog file, so maybe it is easier to just keep them in OBS instead of bringing all those particular files also to Git repo.

accepted/0000-salt-new-maintaining-workflow.md
meaksh: Let say we need to prepare a patch in OBS/IBS to fix some code in one extension or even in many of them in a single shot.

Assuming that we use "git submodules" in our Salt Extension repository and we have a sub-directory per extensions which is a "git submodule":

Then if you run `git format-patch` command on your Git repository root, then you won't get any diff for any of the "git submodules". You have to run `git format-patch` inside each "git submodule" directory to get a patch file which is actually not relative to your main Git repo but to the submodule repository, so we won't be able to apply that generated patch directly in our spec file but we would need to manually adjust it.

This is why I say that "git submodules" are not really straight forward when it come to generate patches.

When it comes to the `obs_scm` service (as we want to have a unified workflow for Salt and its extensions), I have not really tested how it behaves with "git submodules".

Hth!

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,256 @@
...
+If a fix is needed for any of the builtin extensions, workflow would be the same as for a code fix in the main Salt package.
+
+#### Packaged Salt Extensions
+
+For the Salt Extensions that are packaged separately from the main Salt package, we will create a separated GitHub repository where we will maintain these extensions.
+
+This "openSUSE/salt-extensions" repository will contain:
+- a common salt-extension spec file that will generate all RPM packages
+- The sources for each Salt Extension we package
+- A changelog file
```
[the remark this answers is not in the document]
meaksh: Actually, I just changed this as we would have multiple changelog files, one per maintained workflow.

Of course, merge conflicts could happen for PRs that are introducing different changelog entries at the same, then we would need to rebase the PR before merging it.

accepted/0000-salt-new-maintaining-workflow.md
meaksh: These "builtin extensions" I introduced it as an optional way for us to be able to provide extensions that do not have yet a proper Salt Extension package.

This could happen for example in these cases:

- Not officially published as Salt Extensions yet (sources would come from "salt-extensions-holding" repository) that we want to include but do not want to maintain upstream.
- Salt Extensions published but not packaged yet in OBS. (in this case we should probably go an package it p)
  might not be necessary but it is an option we have.

Builtin extenions would reduce, if necessary, the number of Salt Extensions packages we want to deal with.

It would be up to us to decide whether we want to have any builtin extensions or we just go with all needed extensions as separeted packages if they are already migrated to Salt Extensions.

accepted/0000-salt-new-maintaining-workflow.md
agraul: Thank you, I understand what you mean now. I agree that using sub-modules in that way is not a good approach. I don't really get the point about patches though, I thought we don't create them anymore with this RFC?

JFI, in the new "SUSE Packaging Git Workflow" git sub-modules are used in "project repositories". Each package that's inside the project is it's own git repository  (somewhere else) and they exist inside the project as sub-modules.

meaksh: For the general mainteance, we don't use patches but there are some cases patches are still needed, like the embargoed bugs, where the fixes are manually pushed to IBS, using a patch file, and not via GitHub repository until the bug is public.

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
meaksh: I've added a section in to RFC to clarify this.

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,300 @@
...
+
+This is the place now where all those files will be maintained.
+
+#### Tracking upstream and downstream patches
+
+This RFC proposes the usage of "obs_scm" service as the mechanism to pull the sources for the Salt package to build. This service will automatically produce a sources tarball according to a configured GitHub branch.
+
+For the regular Salt maintenance, this means it won't be needed anymore to manually produce patch files to add them to the spec file, as the tarball now contains the updated sources (with the exception of EMBARGOED bugs, where patches are still needed as we cannot push any fix to public GitHub repositories).
+
+To avoid losing the useful labeling of "PATCH-FIX_UPSTREAM" and "PATCH-FIX_OPENSUSE" (with a direct link to the origin PR on the spec file for each new patch we introduced into our Salt package), we will keep adding this information to the spec file on every new PR but this time without adding the patch itself, only the comment.
```
agraul, line 84: Won't this cause a lot of merge conflicts?

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
agraul: I can't find it, can you help me out with the line number or a link? With the list of "backported PRs" in the spec file, we might hit this issue in two places.

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,300 @@
...
+done
+
+rm _temp.changes
+```
+
+When creating a PR to `openSUSE/salt` the user must also include the corresponding changelog entry for all maintained changelog files.
+
+Similarly to the main Uyuni repository, we should add a GitHub action to warn the user in case no changelog entry is added in the PR.
+
+NOTE: I think it is better to decouple commit messages (focus on developers) from changelog entries (focus on users/customers), so I prefer to not use commit messages from "openSUSE/salt" to autogenerate the changelog entries but rather to manually write a meaningful changelog message to be included in your PR as part of your changes. Similarly to what we do in other Uyuni repositories.
```
agraul, line 114: Totally agree. In openSUSE/salt, we can't afford to lose meaningful commit messages

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,300 @@
...
+```
+_multibuild
+_service
+salt-xxxx.obscpio
+salt.changes
+salt.obsinfo
+salt.spec
+```
+
+Since services are disabled here, to allow submissions to openSUSE and SLE, this OBS package will be automatically synced with `openSUSE/release/xxxx` by a Jenkins job.
```
agraul, line 185: Something is unclear about the workflow to me. You write that we use `systemsmanagement:saltstack:github/salt` as the devel project. Does that mean the changes go "Github" -> `systemsmanagement:saltstack:github/salt` -> `systemsmanagement:saltstack/salt` -> `openSUSE:Factory`? What exactly is this Jenkins job doing?

accepted/0000-salt-new-maintaining-workflow.md
agraul: In the other RFC, the [list of built-in extension](https://github.com/uyuni-project/uyuni-rfc/blob/master/accepted/00102-upgrade-to-salt-3008.md#builtin-salt-extensions) includes `zypperpkg` and `transactional_update`. These are extensions *we* want to maintain, i.e. we will need to publish them on PyPI. What's the workflow for maintaining these built-in extensions? Do we have the same rules wrt. pull requests and what they have to contain? How do we publish them to PyPI? What's the plan here?

agraul (commented):
Thank you for the updates. For me there are two areas that are not very clear, it would be great if we could clarify them:

1. How do we maintain "our" salt extensions?
2. The current state of the RFC includes two places that will likely cause merge conflicts on every pull request.

accepted/0000-salt-new-maintaining-workflow.md
meaksh: > Does that mean the changes go "Github" -> `systemsmanagement:saltstack:github/salt` -> `systemsmanagement:saltstack/salt` -> `openSUSE:Factory`?

Yes, that's correct. Essentially:

- `systemsmanagement:saltstack:github/salt` -> services must be enabled here in order to automatically gets the changes from the corresponding GitHub branch, therefore we cannot use this source to submit to `Factory` (services MUST be disabled when submitting). (poc example:  https://build.opensuse.org/package/show/home:PSuarezHernandez:tests:github/salt)
- `systemsmanagement:saltstack/salt` -> this is expected to contain same codebase as the above package but in a way that is "ready-to-be-submitted". This means, with the services disabled, and containing the actual spec file, changelog and obsinfo/obscpio files. (poc example: https://build.opensuse.org/package/show/home:PSuarezHernandez:tests/salt)

> What exactly is this Jenkins job doing?

The jenkins jobs takes care of automatically pushing any new change that is added to `systemsmanagement:saltstack:github/salt` via OBS services to the package  "ready-to-be-submitted" package at `systemsmanagement:saltstack/salt`. It disables the services by adjusting the `_service`, then it manually runs the services to get the spec file, changelog and obsinfo/obscpio files. So after Jenkins runs, the package at `systemsmanagement:saltstack/salt` contains the same codebase than the `:github` package and it is ready to be submitted, i.a. to `openSUSE:Factory` or Maintenance.

I'll try to clarify this better on the RFC text.

agraul:
openSUSE:Factory started the move to the Git workflow for the devel projects: https://en.opensuse.org/openSUSE:OBS_to_Git

Before me merge this PR, we should make sure that our workflow is compatible with the new Git workflow. From what I can understand, we probably want to have something in src.opensuse.org

accepted/0000-salt-new-maintaining-workflow.md
agraul: We probably need a script that's a bit smarter and able to edit all the changelogs at once (e.g. to add a bsc number to the current entry). I have a POC, it's not too hard to have that and all the changelogs in the repo.

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
agraul: Using `git-notes` is pretty straight forward and does not rewrite history. I think we should go that route. The only downside is that Github does not display those notes, but proper git clients do

accepted/0000-salt-new-maintaining-workflow.md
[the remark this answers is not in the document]
agraul: To me this looks harder to maintain over time than separate packages. With `py2pack`, it's pretty easy to create new Python packages. Maybe we can it to create new rpms for Salt extensions?

accepted/0000-salt-new-maintaining-workflow.md
```diff
@@ -0,0 +1,300 @@
...
+- A bit more complex OBS structure than the current one. Including `obs_scm` service.
+- Having to still rely on Jenkins to get the packages ready to be released.
+
+# Alternatives
+[alternatives]: #alternatives
+
+1. Stick to our current workflow based on "salt-packaging" -> The workflow doesn't currently fit with Salt Extensions and we don't want to have different workflows between Salt and Salt Extensions.
+2. One dedicated GitHub repository and OBS package per each Salt Extension -> It won't save resources and will cause more submissions.
+3. The usage of "git submodules" as an alternative to adding the Salt Extensions sources manually would make it tricky to generate patches manually and also to integrate with "obs_scm".
+4. Use OBS `scmsync` integration -> while this allows to integrate even the `_service` file into the GitHub repo, it does not work well with SCM/CI workflow, therefore we cannot test the build of the package for an open PR because the branched package is not reflecting the actual changes on the PR.
```
agraul, line 295: The current approach uses a combination of Jenkins and the OBS<>Github integration, but we still rely on Jenkins as you point out in #Drawbacks. Do we win much by this setup? I wonder if a setup that does not use OBS' SCM/CI Github integration, one that just relies on Jenkins for triggering new builds, would be simpler (as in fewer moving parts). Long term, `scmsync` might be the way to go.

admd:
@meaksh what's the state of this RFC? what's left to be done and do you plan to finish it?

PR #98, opened 2024-12-05
RFC: Enhanced Ansible support

parlt91:
This RFC describes planned enhancements to the existing Ansible implementation in Uyuni

See the [rendered version](https://github.com/uyuni-project/uyuni-rfc/blob/enhanced-ansible-support/accepted/00104-enhanced-ansible-support.md).

### DIFF
--- a/accepted/00104-enhanced-ansible-support.md
+++ b/accepted/00104-enhanced-ansible-support.md
@@ -28,7 +28,7 @@ These points will be discussed one-by-one in detail in the next section.
 
 #### Raw output
 
-To show the raw output returned by the ansible command itself instead of the output returned by salt a change to the [runplaybook.sls](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/ansible/runplaybook.sls) will have to be made.
+To show the raw output returned by the Ansible command itself instead of the output returned by Salt a change to the [runplaybook.sls](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/ansible/runplaybook.sls) will have to be made.
 
 We're currently using the `state module`:
 ```
@@ -60,24 +60,51 @@ run_ansible_playbook:
 
 #### Displaying targeted systems
 
-Since the list of systems targeted by an ansible playbook can be very long, I propose to display the used inventory as a link in the event history details instead.
+In case the selected event details are of the playbook action type we will set an additional request parameter called `inventory` in `SystemHistoryEventAction.java` and render the list of systems.
 
-In case the selected event details are of the playbook action type we will set an additional request parameter called `inventory` in `SystemHistoryEventAction.java` and render an extra row that would look like:
 ```
-Inventory: <link_to_inventory>
+Inventory: <column_with_inventory_sytems>
 ```
 
-Clicking the link would then forward the user to the inventory file in `System` > `Ansible` > `Inventories`.
+There are two ways to acquire this list of sytems:
 
-The inventory then displays the list of systems registered in Uyuni that were targeted by the playbook execution.
+1) We will use the inotify beacon (see more info below), and add a new table (e.g. `suseAnsibleInventorySystems`) with a one-to-many relation to the inventory in suseAnsiblePath. In that table we would store all the systems we get from parsing the inventory file whenever there are changes to it. Columns would be:
 
-Since pre selecting a specific inventory based on the provided URL is not currently supported by the [ansible-path-content.tsx](https://github.com/uyuni-project/uyuni/blob/master/web/html/src/manager/minion/ansible/ansible-path-content.tsx) component, it will have to be adapted to support it.
-The URL to load a specific inventory would look like:
-```
-manager/systems/details/ansible/inventories?sid=<system_id>&iid=<inventory_id>
-```
-with `<system_id>` being the system id of the ansible control node and
-`<inventory_id` being the id of the inventory to load.
+    - inventory_id
+    - server_id
+
+2) Since we'll have the raw output of the Ansible playbook available in the event summary, we can parse this output to acquire the list of targeted systems. All the systems results are listed in `tasks` -> `hosts` of the output. See example output below.
+
+    ```
+    mgrcompat_|-run_ansible_playbook_|-run_ansible_playbook_|-module_run:
+        name:
+        - ansible.playbooks
+        changes:
+            ansible.playbooks:
+                custom_stats: {
+                    }
+                global_custom_stats: {
+                    }
+                plays:
+                -   play:
+                        duration:
+                            end: '2024-12-12T09:37:57.079186Z'
+                            start: '2024-12-12T09:37:51.668830Z'
+                        id: 52540032-6a07-cce6-30f6-000000000006
+                        name: Example Simple Playbook
+                    tasks:
+                    -   hosts:
+                            uyuni-minion-sles15sp4.tf.local:
+                                <output>
+                            <another_host>:
+                                <output>
+                            <yet_another_one>:
+                                .
+                                .
+                                .
+    ```
+    
+    This option has the upside that we don't need another table to store the list of systems and the systemlist can be parsed at runtime (when visiting the page).
 
 ## Add UI support to allow editing of variables defined in the playbook
 
@@ -112,20 +139,30 @@ For example
 The UI would generate input fields based on variable type (list, dict, string) that are pre-populated with the default variables set in the playbook (if any).
 Changing the values of the generated input fields will allow users to override the defaults and schedule the playbook using the variables set in said UI.
 
-Additionally we should add a `free-form` text field to allow users to define additonal variables that were not part of the `vars` section and thus don't have an input field generated. This input would have to be in JSON format.
+In the UI we already have yaml parsing and input field generation as part of formula with forms. During implementation we should investigate if this can be reused or extended to support our new use case.
 
-Based on the variables we will generate a JSON string called `extra_vars` that will be send to the backend along with other parameters needed to schedule the playbook.
+Additionally we should add a `free-form` text field to allow users to define additonal variables that were not part of the `vars` section and thus don't have an input field generated. This input would accept yaml or JSON. This input field could be hidden behind an advanced configuration option.
+
+In later iterations we can improve upon this and allow the user to set extra variables through a dynamically generated UI. The workflow would be the following:
+
+1) Click a button to add a new variable
+2) Select the variable type (list, dictionary, string)
+3) Based on the select we would generate input fields to allow the user to define new variables and set their values
+
+The set variables and `extra_vars` from the text field we will be sent to the backend and stored in the `rhnActionPlaybook` table, so they can be added to the event summary in a similar way then the list of targeted systems.
+
+Before sending the variables to Ansible we'll have to parse and merge the two generated strings. When merging one of the two strings needs to have priority in case there is variables defined in both of them. The current consensus is for the free-form defined variables to take priority.
 
 The workflow on the frontend would look like:
 1) Parse the selected playbook.
 2) Generate input fields based on the `vars` section of the playbook.
-3) Compile a JSON string from the inputs that will be send to the backend.
+3) Compile a yaml string from the inputs that will be send to the backend.
 
-The `ansible.playbooks` salt state we use to execute playbooks at the scheduled time supports a parameter called `extra_vars` and accepts a JSON string. Variables set through this parameter will take precedence over the ones defined in the playbook and thus allow us to override them.
+The `ansible.playbooks` Salt execution module function we use to execute playbooks at the scheduled time supports a parameter called `extra_vars` and accepts a JSON or yaml string. Variables set through this parameter will take precedence over the ones defined in the playbook and thus allow us to override them.
 
-To support scheduling playbooks using `extra_vars` on the backend we will have to to add a new column to the `rhnActionPlaybook` table and adapt the corresponding hibernate entity. The JSON string could either be stored as `varchar` or alternatively as `blob` if we expect them to be large.
+To support scheduling playbooks using `extra_vars` on the backend we will have to to add a new column to the `rhnActionPlaybook` table and adapt the corresponding hibernate entity. The yaml string could either be stored as `varchar` or alternatively as `blob` if we expect them to be large.
 
-During execution of the action the `extra_vars` will then be passed as pillar data to the `runplaybook.sls` state that then executes the `ansible.playbooks` state using the pillar data.
+During execution of the action the `extra_vars` will then be passed as pillar data to the `runplaybook.sls` state that then executes the `ansible.playbooks` execution module function using the pillar data.
 
 This state will have to be changed to look like:
 ```
@@ -163,7 +200,7 @@ Along with this we should also provide a reset button to restore the input field
 
 ## Support the recurrent scheduling of Ansible playbooks.
 
-To support the execution of Ansible playbooks using recurring actions we'll have to implement the follwoing steps:
+To support the execution of Ansible playbooks using recurring actions we'll have to implement the following steps:
 
 1) Create a new `RecurringActionType` called `PLAYBOOK`.
 2) Add a new database table called `suseRecurringPlaybook` with the following columns:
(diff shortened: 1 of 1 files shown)


accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+            /another/inventory/file:
+              mask:
+               - modify
+        - interval: 5
+        - disable_during_state_run: True
+    ```
+
+    This beacon would be updated whenever a inventory file is added/removed on Uyuni.
+
+    Along with the beacon we'll have to add a new `reactor` to the salt master that will react to events send by the beacons.
```
rjmateus, line 230: What would this reactor do? Shouldn't we just process the event? Save it on the database and let the salt event processor handle it?

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+    This way we'd only have to do one single salt execution per control node. We would then parse the output and compile it into a list of systems. This list will be used to add/remove the `ansible_managed` entitlement from systems registered in Uyuni.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+I don't currently see any drawback with the proposed implementation.
+
+# Future improvements
+- Make Ansible work with Uyunis system groups.
+- Allow adding Ansible playbooks to custom state recurring actions to support executing multiple playbooks in order.
```
rjmateus, line 288: Or have this feature directly on ansible playbook, and not integrated into the custom states. However, we need to be careful to not re-implement awx.

rjmateus (commented):
Overall looks good, but I have a few questions. Thank you for the great work

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+    # Integer variable
+    max_open_files: 1024
+
+    # Boolean variable
+    enable_firewall: true
+
+    # String variable
+    welcome_message: "Welcome to the server setup!"
+```
+The UI would generate input fields based on variable type (list, dict, string) that are pre-populated with the default variables set in the playbook (if any).
```
Etheryte, line 112: Since we already have a fair bit of code that does many similar things for formulas, do you think we could extend/reuse that, or do you think the underlying requirements are too different?

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+#### Displaying targeted systems
+
+Since the list of systems targeted by an ansible playbook can be very long, I propose to display the used inventory as a link in the event history details instead.
+
+In case the selected event details are of the playbook action type we will set an additional request parameter called `inventory` in `SystemHistoryEventAction.java` and render an extra row that would look like:
+```
+Inventory: <link_to_inventory>
+```
+
+Clicking the link would then forward the user to the inventory file in `System` > `Ansible` > `Inventories`.
```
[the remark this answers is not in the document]
admd: It's a tricky one. So far, we haven't been keeping track of any systems of inventory. Yes, currently, if inventory is removed from the control node, we will have an orphan inventory link in our database IIRC.

I would suggest to just make this targeted systems part of playbook action by parsing the output or inventory at schedule time. Keeping it linked with inventory will create the issue what you just brought up, Ricardo. Yes, it's understood that list could be long.

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+To show the raw output returned by the ansible command itself instead of the output returned by salt a change to the [runplaybook.sls](https://github.com/uyuni-project/uyuni/blob/master/susemanager-utils/susemanager-sls/salt/ansible/runplaybook.sls) will have to be made.
+
+We're currently using the `state module`:
+```
+run_ansible_playbook:
+  ansible.playbooks:
+     - .......
+```
+
+To get the raw output we will have to use module.run to call the `ansible.playbooks` function from the `execution module`:
```
admd, line 40: I would expect this would't have any performance downsides.

accepted/00104-enhanced-ansible-support.md
admd: Underlaying requirements aren't that different. If we could reuse some exiting components that would be ideal. My only concern is that formulas with forms code was pretty fragile and painful to maintain in the past, not sure if we want to build more stuff on top of it.

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+    # Boolean variable
+    enable_firewall: true
+
+    # String variable
+    welcome_message: "Welcome to the server setup!"
+```
+The UI would generate input fields based on variable type (list, dict, string) that are pre-populated with the default variables set in the playbook (if any).
+Changing the values of the generated input fields will allow users to override the defaults and schedule the playbook using the variables set in said UI.
+
+Additionally we should add a `free-form` text field to allow users to define additonal variables that were not part of the `vars` section and thus don't have an input field generated. This input would have to be in JSON format.
```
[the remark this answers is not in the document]
admd: Instead of free-form, would it be feasible to have some  UI where user can add the variables dynamically , we already know the possible types of the variables . So for user, it will be choosing the right type and then add the values as suitable/needed?

On a general note, in Ansible, YAML is far more commonly used than JSON so my preference would be always to have something that users are used if needed.

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+
+Based on the variables we will generate a JSON string called `extra_vars` that will be send to the backend along with other parameters needed to schedule the playbook.
```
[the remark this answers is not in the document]
admd: Agree! This goes in the same direction as systems of inventory so I would suggest fixing this problem together. When a playbook executed, we need to know exactly what variables were passed and which systems were targeted - this information should be possible to get at any time after the execution is done!

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+
+The workflow on the frontend would look like:
+1) Parse the selected playbook.
+2) Generate input fields based on the `vars` section of the playbook.
+3) Compile a JSON string from the inputs that will be send to the backend.
+
+The `ansible.playbooks` salt state we use to execute playbooks at the scheduled time supports a parameter called `extra_vars` and accepts a JSON string. Variables set through this parameter will take precedence over the ones defined in the playbook and thus allow us to override them.
+
```
admd, line 125: Ok so we will enable user to set the vars of the palybook and then if that user also set the same variable additionally, by mistake or so, under as part of our freeform idea,  which one will take precedence? or will we add some validation that if variable is already defined, it shouldn't be possible to define it again?

accepted/00104-enhanced-ansible-support.md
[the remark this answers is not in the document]
rjmateus: Abid I think that is a good option.
A possible alternative is to save the inventory link/name at schedule time, and then at execution time we save the file content in the database, for example as part of the execution result. 
I'm not sure what would be better, but one way or the other we should save the inventory list without link to the file.

The main good point of just saving in the result is for recurrent executions if users change the inventory file it will reflect be reflected in the next recurrent execution, and as user I would expect that to happen.

accepted/00104-enhanced-ansible-support.md
rjmateus: The front-end I think was not that bad, only the hide/show part was not obvious (but I think Karl/Richa have work or at least plans to improve it). The main problem is the formulas overall integration

accepted/00104-enhanced-ansible-support.md
rjmateus: I would add this extra variable in json/yaml/free text mode under an advance configuration option. Which one would have precedence I don't have a strong opinion. Maybe the advanced ones? over the fields?
Or maybe the input field one, which should make it easier to identify where the value is coming from.

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+This data would be stored whenever a playbook is scheduled from the UI.
+If there is already a stored config with matching `server_id`, `user_id` and `playbook_path` we would override the previous one.
+
+When loading the playbook scheduling page we would look up, if there is an existing config and merge the data into the generated input fields. Fields without a match (likely because the playbook has been edited by the user) would be omitted.
+
+Along with this we should also provide a reset button to restore the input fields to the playbook defined defaults.
+
+## Support the recurrent scheduling of Ansible playbooks.
+
+To support the execution of Ansible playbooks using recurring actions we'll have to implement the follwoing steps:
```
meaksh, line 166:
```suggestion
To support the execution of Ansible playbooks using recurring actions we'll have to implement the following steps:
```

accepted/00104-enhanced-ansible-support.md
meaksh: Yeah, I think it is better to use the `SaltReactor.java` here and not the Salt Master Reactors.

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+
+# Future improvements
+- Make Ansible work with Uyunis system groups.
+- Allow adding Ansible playbooks to custom state recurring actions to support executing multiple playbooks in order.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+The following points still need to be investigated in:
+- It looks like the `inotify beacon` requires `python3-inotify package` to be installed on every control node. Is this beacon already available by other means or do we have to make sure the package is installed when setting up a new control node in Uyuni?
```
meaksh, line 294: When we set a system as "Ansible Control Node" (in system properties), then this injects some bits to the system highstate, particularly it forces `ansible` package to be installed, so we can also make some other packages to be installed.

But an important thing here is that, beacons would be executed by `venv-salt-minion` (as in 5.0+ all registered clients are migrated to "venv-salt-minion"), so we need python3-inotify for the Salt Bundle. The good thing is that the Salt Bundle already contains `saltbundlepy-inotify`, meaning this will be available in all Salt Bundle supported systems.

Another thing worth to mention in this RFC that beacons won't work in Salt SSH, so this approach won't work for Ansible Control Nodes managed via Salt SSH.

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+- How to update the `inotify` beacon from the webUI when new inventories added/removed?
```
meaksh, line 295: The beacon definition in Uyuni is done via pillar data for the minions. So Uyuni provides this "beacon" configuration as part of the generated pillar data for a particular minion.

My understanding here is that, whenever you add/remove an inventory from the UI, this should update the generated pillar data for this minion to include the new  beacon configuration and then trigger 
 a `saltutil.refresh_pillar` and `saltutil.refresh_beacons` actions for the minion, so it refreshes the beacons on the running salt minion process.

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+- How will the results of the `ansible.targets` state triggered through above beacon be handled by the java backend?
```
meaksh, line 296: As clarified in previous comments, this would be done by `SaltReactor.java`.

meaksh (commented):
Thanks Pascal for this RFC! I've just added some comments and clarifications

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+Additionally we should add a `free-form` text field to allow users to define additonal variables that were not part of the `vars` section and thus don't have an input field generated. This input would have to be in JSON format.
+
+Based on the variables we will generate a JSON string called `extra_vars` that will be send to the backend along with other parameters needed to schedule the playbook.
+
+The workflow on the frontend would look like:
+1) Parse the selected playbook.
+2) Generate input fields based on the `vars` section of the playbook.
+3) Compile a JSON string from the inputs that will be send to the backend.
+
+The `ansible.playbooks` salt state we use to execute playbooks at the scheduled time supports a parameter called `extra_vars` and accepts a JSON string. Variables set through this parameter will take precedence over the ones defined in the playbook and thus allow us to override them.
```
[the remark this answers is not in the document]
meaksh: That's right, but then IIUC we should rephrase this to not really mention "salt state" here in this sentence. Maybe something like:

```suggestion
The `ansible.playbooks` salt execution module function we use to execute playbooks at the scheduled time supports a parameter called `extra_vars` and accepts a JSON string. Variables set through this parameter will take precedence over the ones defined in the playbook and thus allow us to override them.
```

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,296 @@
...
+
+To support scheduling playbooks using `extra_vars` on the backend we will have to to add a new column to the `rhnActionPlaybook` table and adapt the corresponding hibernate entity. The JSON string could either be stored as `varchar` or alternatively as `blob` if we expect them to be large.
+
+During execution of the action the `extra_vars` will then be passed as pillar data to the `runplaybook.sls` state that then executes the `ansible.playbooks` state using the pillar data.
```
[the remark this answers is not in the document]
meaksh:
```suggestion
During execution of the action the `extra_vars` will then be passed as pillar data to the `runplaybook.sls` state that then executes the `ansible.playbooks` execution module function using the pillar data.
```

accepted/00104-enhanced-ansible-support.md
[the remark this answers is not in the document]
rjmateus: I like option 2 more, it's more clean and direct.

accepted/00104-enhanced-ansible-support.md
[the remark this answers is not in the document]
rjmateus: @parlt91 yaml extra extra_args (the ones from free text) only need to be parsed and merged when passing them to ansbile for execution. You don't want to merge then on the front-end and pass all together to the backend, otherwise, you lose track of what is on the fields, and what is on the free form.

cbbayburt (approved):
Thanks, @parlt91. Looks good to me. All my concerns are already discussed under other reviews.

Here's a couple of remarks:

1. I think it's important that users have the exact list of the clients on which the playbook is executed. IIUC we'll store this in action details. This should be a hard requirement.

2. We should explore the formulas with forms UI to reuse as much as we can. Formulas UI also uses yaml to define the forms, so there's some opportunity there.

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,332 @@
...
+
+    The beacons will be executed by `venv-salt-minion`. To add support for the `inotify` beacon to them we'll have to make sure the `saltbundlepy-inotify` package is installed along the `ansible` package whenever a new control node is added. Already existing control nodes will have to run a highstate to get the package installed.
+
+    The beacon would be updated whenever a inventory file is added/removed on Uyuni. To support this we will add the beacon to the minions pillar data and use the `MinionPillarManager.generatePillar(...)` function to update it whenever needed. In addition we will have to run `saltutil.refresh_pillar` and `saltutil.refresh_beacons` to refresh data on the running minion process
+
+    On the java backend we will add support for the beacon to the `SaltReactor.java` and trigger an update of the ansible managed systems whenever there were changes to the inventory files.
+
+    Additionally we'll trigger a manual update of the minion list whenever inventories are added/removed.
+
+    One thing to mention is that this approach (using the beacon) will `NOT` work with ansible control notes managed through Salt SSH.
```
meaksh, line 271: Small typo:
```suggestion
    One thing to mention is that this approach (using the beacon) will `NOT` work with ansible control nodes managed through Salt SSH.
```

accepted/00104-enhanced-ansible-support.md
```diff
@@ -0,0 +1,332 @@
...
+              mask:
+                - modify
+            /another/inventory/file:
+              mask:
+               - modify
+        - interval: 5
+        - disable_during_state_run: True
+    ```
+
+    The beacons will be executed by `venv-salt-minion`. To add support for the `inotify` beacon to them we'll have to make sure the `saltbundlepy-inotify` package is installed along the `ansible` package whenever a new control node is added. Already existing control nodes will have to run a highstate to get the package installed.
```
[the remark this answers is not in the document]
meaksh: No, already in 5.0, the registered clients are automatically migrated to `venv-salt-minion`, if not using it yet, as part of the highstate.

rjmateus (approved):
LGTM

admd (approved):
Thank you Pascal

meaksh:
Hey @HVSharma12, thanks for notes and feedback about `ansible-runner`!

As I mentioned in Slack, we didn't know about this way of operating Ansible, neither that users are frequently using this, i.a. in AWX. I see the benefits of enabling Salt's ansible modules to handle the Ansible executions via `ansible-runner` instead of the normal CLI interface.

This is definetely something we should explore and probably adopt, and I think it deserves its own RFC, something like `Adding support for ansible-runner in Uyuni and Salt`. This will require development efforts on the Salt side.

For now, the present RFC about the "Enhacement of Ansible support" is mostly around exposing more functionality to Uyuni from what we have already implemented, without much intervention on the Salt side.

I'd vote for preparing a separated RFC, describe the design, so hopefully we can agree on a proposal and go with the implementation it at some point.

rjmateus:
thank you for your contribution @HVSharma12 
I agree with Pablo, the current RFC is in the scope of giving more visibility and option to interact with ansible. But you proposal looks promising and I think we should research it after the current development effort.
@admd any comment on this?

admd:
I have created an issue https://github.com/uyuni-project/uyuni/issues/11802 for evaluating support for ansible-runner in Uyuni and Salt.

@parlt91 can we merge this RFC or is there still something pending?

PR #100, opened 2024-12-13
RFC: OpenAPI Adoption for Describing Uyuni HTTP API

wweellddeerr:
See [rendered version](https://github.com/uyuni-project/uyuni-rfc/blob/d5c640779acdd794da92a443ef97160706f54594/accepted/00000-openapi.md).

### DIFF
--- a/accepted/00000-openapi.md
+++ b/accepted/00000-openapi.md
@@ -1,0 +1,184 @@
+## Feature Name: OpenAPI Adoption for Describing Uyuni HTTP API
+**Start Date:** 13/12/2024
+
+### Summary
+Introduce the OpenAPI specification to describe and document Uyuni's HTTP API, replacing the existing static documentation. It aims to enhance user and developer experiences by providing interactive, standardized, and more accessible API documentation.
+
+### Motivation
+The adoption of OpenAPI aims to enhance both user and developer experiences by modernizing API documentation, making it more accessible, interactive, and aligned with industry standards. The current static API documentation lacks interactivity, making it difficult for users to explore and use API endpoints effectively. Integrating OpenAPI-generated interactive documentation, such as Swagger UI, directly into Uyuni’s Web UI allows users to explore, experiment with, and test API endpoints in real-time. This eliminates the need for additional programming effort and helps users achieve their goals more efficiently.
+
+OpenAPI is a widely recognized standard supported by a robust ecosystem of tools, including API validators, client generators, and testing frameworks. Transitioning to OpenAPI enables developers and users to leverage these tools, streamlining workflows and boosting productivity. Additionally, OpenAPI annotations integrate seamlessly with Java’s syntax and provide strong typing, reducing errors and simplifying maintenance compared to the current Javadoc-based approach, which relies on non-standard tags. This shift ensures consistent, standardized documentation while minimizing the effort required to maintain it.
+
+The expected outcome of this integration is to deliver Uyuni's API documentation as a valid OpenAPI JSON specification, complemented by an interactive Swagger UI integrated into Uyuni’s web interface, allowing users to explore and experiment with the API.
+
+
+### Detailed Design
+
+#### A. Describing the API
+Endpoints will be described using Swagger annotations, replacing the current Javadoc comments. Each API method will be annotated with Swagger elements such as `@Operation`, `@Parameter`, `@ApiResponse`, and `@Schema` to define essential details, including summaries, parameters, request bodies, and responses. For example:
+
+**Current documentation:**
+```
+    /*
+ 	* @apidoc.doc Update Content Environment with given label
+ 	* @apidoc.param #param_desc("string", "projectLabel", "Content Project label")
+ 	* @apidoc.param #param_desc("string", "envLabel", "Content Environment label")
+ 	* @apidoc.param
+ 	*  #struct_begin("props")
+ 	*  	#prop_desc("string", "name", "Content Environment name")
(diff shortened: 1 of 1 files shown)


accepted/00000-openapi.md
```diff
@@ -0,0 +1,184 @@
...
+- Initial Development Complexity
+  - Creating the one-time scanner tool to automate the migration of existing Javadoc-based documentation adds complexity. While this tool will simplify the process, its development and validation represent a time-intensive task.
+
+
+### Alternatives
+
+- Keep the documentation as it is, since it is stable and functional.
+  - The existing documentation, while static and less interactive, is stable and functional. Retaining it avoids the need for migration and eliminates the development and maintenance effort associated with Swagger integration. However, this approach would not address the limitations of static documentation, such as lack of interactivity and adherence to industry standards.
+
+- Keep the format for describing the documentation using JavaDocs and write a new Java Doclet to generate the OpenAPI specification by parsing the definitions in JavaDocs.
```
cbbayburt, line 176: I imagine we might still be obliged to have a non-interactive online documentation in parallel.
In this case, we should consider this alternative.

Can we generate such documentation from openAPI?

accepted/00000-openapi.md
```diff
@@ -0,0 +1,184 @@
...
+ - `io.swagger.core.v3.swagger-annotations`
+
+##### A.2. Existing API Endpoints
+To facilitate the transition, a one-time scanner tool will be developed to automate the conversion of existing Javadoc-based documentation into Swagger annotations. This tool will be implemented as a Java Doclet that systematically scans the codebase to identify predefined patterns in the Javadoc comments. Once identified, the tool will annotate the corresponding methods with the appropriate `@Operation` annotations.
+
+Additionally, the scanner will detect the use of API serializers that describe the return types of endpoints. It will parse the associated Javadoc comments of these serializers to extract the necessary details and incorporate them into the `@Operation` definitions.
+
+##### A.3. XML-RPC API code examples
+
+For XML-RPC endpoints, it is possible to provide a specific section in the web UI containing sample scripts. In addition to this, the `description` and `examples` fields of Swagger operation documentation can be used to aid understanding.
```
cbbayburt, line 102: Can openAPI auto-generate code snippets as usage examples in any way?

How do we differentiate between XMLRPC and HTTP in the docs UI?

accepted/00000-openapi.md
```diff
@@ -0,0 +1,184 @@
...
+    properties = {
+        @StringToClassMapItem(key = "label", value = String.class),
+        @StringToClassMapItem(key = "name", value = String.class),
+        @StringToClassMapItem(key = "description", value = String.class)
+    }
+)
+
+```
+
+For migrating the documentation of existing endpoints, this approach—defining the request body as an anonymous model—will be used.
```
rjmateus, line 132: Agree. It will make the migration simple, and avoid change the method definition, wich can lead to more changes or break the backward compatibility

accepted/00000-openapi.md
rjmateus: That is a great question Can. We may need to have a static API documentation to integrate in the SUSE Documentation portal. @wweellddeerr can we make it with openAPI/swagger UI in build time?

accepted/00000-openapi.md
```diff
@@ -0,0 +1,184 @@
...
+
+OpenAPI is a widely recognized standard supported by a robust ecosystem of tools, including API validators, client generators, and testing frameworks. Transitioning to OpenAPI enables developers and users to leverage these tools, streamlining workflows and boosting productivity. Additionally, OpenAPI annotations integrate seamlessly with Java’s syntax and provide strong typing, reducing errors and simplifying maintenance compared to the current Javadoc-based approach, which relies on non-standard tags. This shift ensures consistent, standardized documentation while minimizing the effort required to maintain it.
+
+The expected outcome of this integration is to deliver Uyuni's API documentation as a valid OpenAPI JSON specification, complemented by an interactive Swagger UI integrated into Uyuni’s web interface, allowing users to explore and experiment with the API.
+
+
+### Detailed Design
```
rjmateus, line 15: In this design you are only mentioning the XML-RPC API which is also exposed through the JSON over HTTP API through the spark framework.
How about the API used by the Front-end? Should we also document it? Should it be documentation only internally or to end-users also?

accepted/00000-openapi.md
```diff
@@ -0,0 +1,184 @@
...
+ 	* @apidoc.returntype $ContentEnvironmentSerializer
+ 	*/
+```
+
+**Proposed Swagger Documentation:**
+```
+    @Operation(
+        summary = "Update Content Environment",
+        tags = {"contentmanagement"},
+        description = "Update Content Environment with given label",
```
rjmateus, line 40: During the migration process, we should try to generate the data to provide more information on how to use it on XML-RPC API also, so we can replace the existing static documentation and have more information to users.

rjmateus (commented):
Thank you for the great contribution. I have just a few small questions, other then that looks great to me.

accepted/00000-openapi.md
Etheryte: Generating the JSON spec at build time would also have other benefits, namely we could generate Typescript types for the endpoints and use those in future code. This is outside the scope of this RFC, but just something positive for the future.

stdevel:
Would love to see this coming. This would make it so much easier to integrate Uyuni in other tools such as Terraform an Ansible.

PR #101, opened 2025-02-25
Adjust Salt Singlespec RFC to latest design

meaksh:
This PR just update the RFC about Salt singlespec to align it with the actual design we implemented.


vzhestkov (approved):
Looks good :+1: thanks for taking care of it

PR #103, opened 2025-03-14
Aligin disconnected design with implementation

ycedres:
After first iteration on the Disconnected Setup implementation we decided to remove LogCLI component. Also we used the Infinity Datasource Plugin for Grafana instead of Prometheus, as this plugin provides a better fit for this use case.


accepted/00101-health-check-tool.md
```diff
@@ -237,7 +242,7 @@ When the tool has access to an Uyuni server, it can run either "standalone" or v
 
 
 # Security considerations
-In the current design, we are exposing metrics via Prometheus and making them available in Grafana, and more importantly we are exposing log messages via Loki/LogCLI to CLI and Grafana users. It is important to notice that after running this tool, and until related containers are destroyed, the Grafana Dashboards (and other components like Prometheus and Loki/LogCLI) are exposing metrics and log messages that may contain sensitive data and information to any non-root user on the system or to anyone that have access to this host in the network.
+In the current design, we are exposing metrics via Prometheus and making them available in Grafana, and more importantly we are exposing log messages via Loki to CLI and Grafana users. It is important to notice that after running this tool, and until related containers are destroyed, the Grafana Dashboards (and other components like Prometheus and Loki) are exposing metrics and log messages that may contain sensitive data and information to any non-root user on the system or to anyone that have access to this host in the network.
```
meaksh, line 245:
```suggestion
In the current design, we are exposing metrics via Prometheus and making them available in Grafana, and more importantly we are exposing log messages via Loki to Grafana users. It is important to notice that after running this tool, and until related containers are destroyed, the Grafana Dashboards (and other components like Prometheus and Loki) are exposing metrics and log messages that may contain sensitive data and information to any non-root user on the system or to anyone that have access to this host in the network.
```

meaksh (commented):
Added a small comment. I would also consider using white background on the new images, so you can see the black fonts when using "dark mode" in GitHub.

meaksh (approved):

PR #109, opened 2025-10-07
RFC: Multi-file Autoinstallation Support

rjpmestre:
This RFC proposes extending the autoinstallation profile mechanism to support multiple files per profile and storing them in the database (instead of relying on Cobbler).

See the [rendered version](https://github.com/uyuni-project/uyuni-rfc/blob/7d1118ab0a1e409ccc0dce7618a2b03e5a9b80e2/accepted/00000-multi-file_autoinstallation_support.md).


accepted/00000-multi-file_autoinstallation_support.md
```diff
@@ -0,0 +1,95 @@
...
+
+- Introduces new tables and data management, increasing complexity.
+- Must maintain backward compatibility with single-file profiles.
+- Validation and compatibility checks may introduce overhead.
+- Full templating/snippet handling still depends on Cobbler at this stage.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- Keep only one file per profile - too restrictive for complex setups.
```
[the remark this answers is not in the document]
rjpmestre: I can see how supporting a compressed file upload could be convenient. But I’d still prefer to keep the individual file contents. The major reason being that one of the future steps we outlined in the document is turning this into a templating mechanism. And for that we need to be able to edit, parameterize, and validate each file. Do you agree @aaannz ?

accepted/00000-multi-file_autoinstallation_support.md
```diff
@@ -0,0 +1,95 @@
...
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should we limit:
+  - Number of files in a profile?
+  - Number of files per type?
```
rjmateus: I'm not sure in this one, but we should follow what the OS's supports. Does sle16 support more than one agama file? Does it support more than one ignition, and so on.

accepted/00000-multi-file_autoinstallation_support.md
```diff
@@ -0,0 +1,95 @@
...
+  - File size?
```
rjmateus: We must have some limit for it. These are text files normally, so should not be to big.

rjmateus (commented):
Overall looks good, just a few improvements and clarification that I think would help the overall understanding.

accepted/00000-multi-file_autoinstallation_support.md
```diff
@@ -0,0 +1,95 @@
...
+- Different configuration formats must coexist in the same context (eg: AutoYaST or ignition + custom scripts).
+
+## Source of truth
+Files will be stored locally (in the database), making the system the primary source of truth, rather than relying on external services like Cobbler.
+
+## Validation
+Ensure that uploaded file types are compatible with the selected OS or distribution (eg, Agama files for SLE15SP6).
+  
+## Benefits
+To sum it, supporting multiple files would:
```
aaannz:
```suggestion
```

accepted/00000-multi-file_autoinstallation_support.md
```diff
@@ -0,0 +1,95 @@
...
+- Ensure better reliability and traceability;
+- Reduce dependency on Cobbler
+
+# Detailed design
+[design]: #detailed-design
+
+## Database
+- New tables:
+  - suseAutoInstallFileType — id, type, description, timestamp
+  - suseAutoInstallFile — id, type FK, content, timestamp
```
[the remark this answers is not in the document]
aaannz: What @rjmateus said. Plus we have two timestamps - creation and modification time.

Additionally, where will we add the format and OS compatibility. Will this business logic be in the Java code or we'll have a reference table in a database?

Generally I think that support for new OS will not need additional code, just assigning correct format with particular OS, adding it to the database data is better. However currently AFAIK we do not have OS information available in the database during the build time, only once synced with SCC and/or common channels.

accepted/00000-multi-file_autoinstallation_support.md
aaannz: As far as I know currently two files are for ignition/combustion combo. But that are two different types. So for now I think it is just one file per type, but allowing multiple types.

If we want to have a limit, then the suseAutoInstallFileType database schema should be extended to cover this information.

accepted/00000-multi-file_autoinstallation_support.md
aaannz: I think this is for us to arbitrarily decide. I did not see any maximum size validation anywhere and those autoinstallation files can contain binary data as well.

From security POV, I think this is fairly irrelevant because as soon as attacker can provide their own file, it is trivial to generate OOM condition if that is the goal here even with small sizes.

accepted/00000-multi-file_autoinstallation_support.md
```diff
@@ -0,0 +1,95 @@
...
+  - Number of files per type?
+  - File size?
+
+- Should list the allowed file types (in the UI)?
+- Should we restrain a known file type list or allow a "any" type? If so, how would it impact the previous questions?
+- UI/UX design for multi-file edit/upload - the provided screen update is a suggestion. UI team should review it so we make sure were following FE rules and usability constraints are met.
+
+## Future steps
+- Should files be validated for content according to type?
+- Templating/snippet support should be covered by another RFC?
```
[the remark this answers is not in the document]
aaannz: Depends if we want powerful templating like cheetah in the cobbler, or will plain string replacement suffice, maybe with some trivial tests like if not set use default.

PR #112, opened 2026-01-05
Add proposal for Kubernetes server deployment

cbosdo:


Pablogoliva (approved):

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+
+- What use cases does it support?
+
+The typical use case is where the user already has Kubernetes skills and needs to deploy Uyuni to manage servers for applications that are not yet cloud-native ready.
+In most of the cases the user will have a separate team administering the Kubernetes cluster and the Uyuni administrators would just be users on it.
+
+- What is the expected outcome?
+
+The expected outcome is a helm chart to deploy the Uyuni server and a set of documentation and examples to use it.
+The rationale behind this is that it is impossible to craft one helm to rule all the possible components combinations.
```
rjmateus, line 37: Can we make one single Helm chart that, by default, only deploys the required services (like server and DB), but has the other deployments as conditional based on a value flag? I think about helm flow control (I never tested it): https://helm.sh/docs/chart_template_guide/control_structures/#ifelse

This way we could have one helm to rule them all and avoid users to extend our.

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+The optional components could be deployed by a helm chart embedding the Uyuni one.
+To make this easier and for documentation purpose too, a Git repository with example helm charts would be prepared.
```
rjmateus, line 39: If we have one single Helm with conditional Deployments/Services, it would make our user life much easier and avoid the need for examples repo, and we could just have documentation with the available values.

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+
+# Detailed design
+[design]: #detailed-design
+
+## Global design
+
+Here is a diagram showing what components would be deployed.
+
+![components design chart](images/server-kubernetes/design.svg)
```
rjmateus, line 48: In the diagram, would DB also depend on SSL certificates?

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+
+* The red items are expected to be deployed by the user and won't be created by the server helm chart.
+* The database could optionally be deployed by the user outside the server namespace: the `db` service will need to be configured with appropriate IP or FQDN in such a case.
```
rjmateus, line 51: As a default from the diagram, should we consider that we deploy the database with our Helm chart? From the diagram looks like it's only considering deploying a service and not the deployment application.

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+Possible solutions are:
+
+* Setup traefik or nginx to route those ports. *This requires a configuration file to be dropped on a cluster node.*
+* Use metalLB to route those ports.
+* Use a gateway API implementation routing those ports.
+
+## Making it easier
+
+There can be multiple deployment possibilities and the idea is to not limit them.
+In order to make it easier for the user to deploy the server helm chart, a Git repository of example helm charts using the server ones will be provided.
```
rjmateus, line 71: Do you think we should provide a helm chart to deploy the server or a helm chart to deploy the "uyuni application" which would take care of deploying any component needed to run uyuni (of course, excluding the ones you have put in red in the diagram)

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+
+* Using [cert-manager](https://cert-manager.io/) to generate SSL certificates.
+  Multiple setups can be considered: self-signed CA, CA from an [OpenBao](https://openbao.org/) vault or step-ca / ACME / let's encrypt.
+
+* Hard-coded third-party SSL CA encrypted using [sealed-secrets](https://github.com/bitnami-labs/sealed-secrets) or at least documentation on how to do it.
+
+Those cases would come with a simple database container like the one deployed in the podman setup.
+Other database deployments may be added to that repository at a later stage.
+
+An other step to make the installation easier would be to provide [UI configuration for rancher](https://ranchermanager.docs.rancher.com/how-to-guides/new-user-guides/helm-charts-in-rancher/create-apps).
```
rjmateus, line 82: I would move it to next-steps/future work,

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+## Sharing scripts
+
+All the setup scripts are currently generated by `mgradm` itself.
+This would mean to duplicate those scripts in the helm chart and would be error prone.
+In order to avoid such a situation, the scripts generated by `mgradm` will be moved into container images like the server one, or specific images if needed.
+
+## Proxy SSL certificates
+
+When creating a proxy configuration, the user needs to provide either the CA key, certificate and password or third-party certificate matching the proxy FQDN.
+In the case of a Kubernetes deployment, only the later would be possible.
```
rjmateus, line 97: We could also think about an option to not receive the private key, and it would be placed directly in the proxy (with any certificate manager the customer has).
I think we would need to receive a least the public key because of proxy certificate chain.
Would also be a good idea to have the option to have the certificates outside the configuration YAML file, and have a specific naming convention to where to look at.

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+
+See [issue uyuni-project/uyuni-tools#419](https://github.com/uyuni-project/uyuni-tools/issues/419) for a solution to this problem for podman.
+
+In the Kubernetes case, the proxy certificate generation will be left to the PKI tool: the secret would then be created directly on the proxy.
+This could be handled with cert-manager on the proxy cluster using the same step-ca issuers than on the server, or getting the CA from a shared vault.
+
+## Test suite changes
+
+Since the test matrix will blow up and the number of persons taking care of the CI won't increase accordingly, the test suite will need some changes to test Kubernetes deployments.
+There is no need to run the full test suite on Kubernetes, just deploy a server, a proxy and a client and verify the connectivity and that all components are up and running.
```
rjmateus, line 113: Agree. We may also need to add some pxe boot testing to make sure Cobbler ports are fine, especially the UDP ones.

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+Such a smoke test would run much faster and could be run an different setups more easily.
+
+**The assumption here is that the features would be tested on a podman deployment and wouldn't differ on Kubernetes.**
+
+Some benchmarks would also be needed to test and validate various storage or network back-ends more easily.
+Those may or may not be included in the test suite, but still need to be provided for users to validate their back-end.
+
+## Support tools
```
rjmateus, line 121: This is a super important topic, since if we don't have a support config we are basically blind in case an L3.

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+There are two reasons for this:
+
+- this would be a second and different delivery mechanism.
+- the test matrix of the Kubernetes components can explode very fast.
+  Just for RKE2 there are four possible network stacks and several storage drivers.
+
+Of course, a second deployment mechanism means more tests to perform, more possible CI failures to look into and more hardware needed.
+
+Another important problem will be handling the bug reports: figuring out if the problem lies in Uyuni or in the Kubernetes setup will be tough.
+This requires new tools to help diagnosing, but also that all developers and testers get up to speed with the Kubernetes concepts.
```
rjmateus, line 140: We can also try to contact ECM support team and see if they could give some guidance/training to our team and also to MLM supporters.

accepted/00110-server-kubernetes-deployment.md
cbosdo: There are too many possible external dependencies for this. Examples would be much better to avoid ending up with a gigantic helm chart that no one understands any more. Here is the examples repo I have at the moment with only the self-signed CA using cert-manager. I would see at least an example with cert-manager and other issuers for openbao and ACME.

accepted/00110-server-kubernetes-deployment.md
cbosdo: This is simplifying one thing to add complexity to the other.

accepted/00110-server-kubernetes-deployment.md
cbosdo: In that diagram, the DB is in a limbo state: ATM we ship a container for it, and the services are ready to be altered to accept databases setup outside of our chart. The DB deployed by our chart expects a secret for its SSL certificate and a configmap for the corresponding CA. cert-manager can even generate this as shown in https://github.com/cbosdo/uyuni-charts/blob/main/selfsigned/templates/certificates.yaml#L123

accepted/00110-server-kubernetes-deployment.md
cbosdo: explained above: the diagram is half a step ahead already and I was too lazy to change this from the previous version as it will come anyway

accepted/00110-server-kubernetes-deployment.md
cbosdo: I don't get the question, the server has a broad meaning here: https://github.com/uyuni-project/uyuni/pull/10973 deploys the pieces not in red and the DB already and I'll add coco, hub api and Saline as optional components to toggle with values.

accepted/00110-server-kubernetes-deployment.md
rjmateus: I was not talking about external tooling like cert-manager. Those I think it should not be part of our helm chart, and I think the examples repo makes sense.
I was thinking about our internal optional components, like saline and HUB XML RPC API. These ones I think should be part of our helm chart and can use a conditional deployment with flow control based on some flag value.

accepted/00110-server-kubernetes-deployment.md
cbosdo: The RFC doesn't define any milestone / time line. Adding it here doesn't harm at all.

accepted/00110-server-kubernetes-deployment.md
rjmateus: We can pass some of the complexity to our users/customers, but we need to be careful not to pass all the complexity to them, or it becomes unusable.

accepted/00110-server-kubernetes-deployment.md
cbosdo: UDP is for TFTP only, but yes, the more we test the better.

As an example, here is how I manually tested the DB port with SSL:

```
$ kubectl get cm -n uyuni db-ca -o "jsonpath={.data.ca\.crt}" >db-ca.crt
$ podman run --rm -ti -v $PWD/db-ca.crt:/root/.postgresql/root.crt:ro -e PGSSLMODE=verify-ca registry.suse.com/suse/postgres:16 /usr/bin/psql -h $FQDN -U pythia -W reportdb
```

And how I tested the salt ports:

```
$ cat "master: $FQDN" >master.conf
$ podman run --rm -ti -v $PWD/master.conf:/etc/venv-salt-minion/minion.d/master.conf ghcr.io/uyuni-project/uyuni/ci-test-opensuse-minion:master venv-salt-minion
```

accepted/00110-server-kubernetes-deployment.md
rjmateus: Ok. At this stage in the diagram, I would not let the DB be in a limbo and put it explicitly as deployed with our helm chart. 
Deviation from that, like customers deploying the DB outside or manually in the cluster can be seen as deviation and be documented latter (since you are already handling it in the code for further extension)

accepted/00110-server-kubernetes-deployment.md
cbosdo: I should probably remove this from the RFC as this is very MLM-specific…

accepted/00110-server-kubernetes-deployment.md
cbosdo: This is an MLM internal topic that probably should not be discussed here…

accepted/00110-server-kubernetes-deployment.md
rjmateus: ok, got it. That was what I was looking for. I think I got confused with the part above where you talked about users being able to extend uyuni helm to include optional parts. In there, look like you are considering as optinal cert-manager (for example) and I thought you were talking about the "optional" components on uyuni, like coco.

cbosdo: No, the optional parts are what we cannot control because there will be a million ways to do it

accepted/00110-server-kubernetes-deployment.md
cbosdo: Of course on Kubernetes this needs to be even bypassed. The proxy cert pair needs to be deployed using an external PKI like cert-manager. How to do it could also be covered by docs / example charts.

accepted/00110-server-kubernetes-deployment.md
rjmateus: Got it. Would be possible to make it more clear in the RFC?

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,156 @@
...
+## Proxy SSL certificates
+
+When creating a proxy configuration, the user needs to provide either the CA key, certificate and password or third-party certificate matching the proxy FQDN.
+In the case of a Kubernetes deployment, only the later would be possible.
+Even for podman deployments, the server container should not be the one responsible for certificates generation.
+
+See [issue uyuni-project/uyuni-tools#419](https://github.com/uyuni-project/uyuni-tools/issues/419) for a solution to this problem for podman.
+
+In the Kubernetes case, the proxy certificate generation will be left to the PKI tool: the secret would then be created directly on the proxy.
+This could be handled with cert-manager on the proxy cluster using the same step-ca issuers than on the server, or getting the CA from a shared vault.
```
Pablogoliva, line 103: With the proxy cluster do you mean the proxy pod?.

cbosdo: I mean the cluster where the proxy pod will be running: proxies are pretty likely to run in smaller cluster deployed at the edge.

accepted/00110-server-kubernetes-deployment.md
cbosdo: I updated the chart to make this clearer

rjmateus (approved):

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Test suite changes
+
+Since the test matrix will blow up and the number of persons taking care of the CI won't increase accordingly, the test suite will need some changes to test Kubernetes deployments.
+There is no need to run the full test suite on Kubernetes, just deploy a server, a proxy and a client and verify the connectivity and that all components are up and running.
```
Bischoff, line 113: In fact this would make our life more complicated, IMHO.

It is easier for us to run just yet another normal test suite with no changes. The only changes needed would be to adapt sumaform for a different deployment.

Also, if we were to create a new test suite with only a minimal smoke test, we would be at risk of missing problems that could arise from the complex interaction between containers and network components.

An example that immediately jumps to my mind is Retail.

Bischoff (commented):
Commented on the test suite part.

accepted/00000-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,157 @@
...
+Moving to Kubernetes is necessary to support users with the internal requirement to run all container-based applications on Kubernetes or who operate in mixed environments.
+As a bonus an Uyuni server deployed on a Kubernetes cluster, using network volumes would bring easy recovery.
+
+With the current state of the server containers, scalability won't be a goal for the Kubernetes deployments.
+However, this will make it easier after future refactorings.
+
+- What use cases does it support?
+
+The typical use case is where the user already has Kubernetes skills and needs to deploy Uyuni to manage servers for applications that are not yet cloud-native ready.
+In most of the cases the user will have a separate team administering the Kubernetes cluster and the Uyuni administrators would just be users on it.
```
Bischoff: "User" is used for two different things in this sentence. I suggest:
```
In most of the cases the user will have a separate team administering
the Kubernetes cluster and the Uyuni administrators would just be users on it.
```
=>
```
In most of the cases there will be a separate team administering
the Kubernetes cluster and the Uyuni administrators would just be users on it.
```

accepted/00110-server-kubernetes-deployment.md
cbosdo: Already done

accepted/00110-server-kubernetes-deployment.md
cbosdo: The RFC is not stating what is supported or not, and this is not it's goal. I'll leave it as is for now: we ship the internal DB, and when possible will add the bits for third party DB. There will be a proper statement of what is supported or not for SUSE customers, and any community member willing to help getting the third party DB is more that welcomed to implement and test it. Those are different things that shouldn't be mixed here.

accepted/00110-server-kubernetes-deployment.md
cbosdo: I'll leave that kind of decision to you in the end. I have the feeling that running the full testsuite will not be manageable if we later need to expand the test matrix to more that one CSI and/or CNI backend or Kubernetes distribution.

accepted/00110-server-kubernetes-deployment.md
rjmateus: You are contradicting yourself in this part. 
The diagram in this RFC presents DB as if it were placed outside, as if it were to be deployed by a third party.
The current approach (according to your last comment) is that helm chart will deploy the database, but the service is already ready to be deployed outside. The support for 3th party DB deployments would be added later.

I'm not saying that is support from a SUSE customers perpective. But Uyuni also has tested (or supported, or advised) deployment scenarios, and that was what I was pointing to, which 3th party DB is not tested and in the diagram it shows as if was the default deployment method.

cbosdo: I am not saying the external DB is anywhere near ready, just that where possible we are planning with this in mind.

rjmateus: [This diagram ](https://github.com/uyuni-project/uyuni-rfc/pull/112/files#diff-758c90fc08e0c278ecd5232ee2c259ec6138142ae0594f53e4977aa7257a122a) presents the database as an external component. 
If that is something for the long term and not part of the initial implementation, that must be said in the RFC.

accepted/00110-server-kubernetes-deployment.md
Pablogoliva: Something to be taken into account @Bischoff  is that the clients (minions) are going to be common machines (VMS or linux servers installed on bare-metal machines) and the communication with them is going to be through salt, not trough kubernetes. At the same time, the problem is the environment where the product is going to be installed, not the usability of the product by itself. At least in the beginning.

accepted/00110-server-kubernetes-deployment.md
```diff
@@ -0,0 +1,157 @@
...
+
+- `mgradm` using the Kubernetes API: this is what is currently implemented in `mgradm`.
+  This approach introduces a dependency on the Kubernetes API in `mgradm` and requires more synchronization with the latest Go version and Kubernetes API changes.
+
+Not deploying the server on Kubernetes would not have any impact on current users since it only provides an alternative way to install Uyuni.
+However this would rule out users who are already cloud-native.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
```
srbarrios, line 166: ## Remote commands on the server container 

I would like to propose adding an item to the Unresolved questions section regarding the interaction with the server container.

Currently, the workflow often relies on mgrctl to execute commands inside the server. However, installing the mgrctl package usually requires access to the Client Tools channels, which in turn necessitates registering the admin's machine as a Salt Minion against the server.

This creates a "chicken-and-egg" situation for new Kubernetes deployments: to manage the server, I need mgrctl, but to get mgrctl, I need a registered client (which requires a running server).

Suggestion: For Kubernetes deployments, we should consider dropping the requirement for mgrctl entirely and rely on native Kubernetes tooling. The documentation and tooling (if any) should guide users to use kubectl exec directly to run commands into the server container. This aligns with the "Motivation" section of supporting users with Kubernetes skills and eliminates the external RPM dependency on the operator's machine.

accepted/00110-server-kubernetes-deployment.md
srbarrios: During the initial development cycle I think it gonna be easier to just focus on covering installation/configuration of server and proxy on independent clusters. And this will already cover the onboarding of a minion (the proxy itself).

Later on, I think we need to consider how to properly design a test suite that covers only the features that could be affected by the fact of being in a Kubernetes cluster, not covering those features that can be covered in the podman flavour, and will not have any different behavor in this one.
For instance, will make sense to cover retail (different network setup) or migration/backups/etc (volumes setup), but it might be fine to don't cover management of user/organization/ak/channels/lock packages/ etcetera.

It might be as easier as use a new Cucumber Tag for this purpose, as we have for non-containerized server.

accepted/00110-server-kubernetes-deployment.md
cbosdo: `mgrctl` is not strictly needed as you can achieve the same with more or less complicated `kubectl` commands. What could be more problematic is that `mgradm` has some commands that would be useful for a server running on Kubernetes: copying a distribution or manage GPG keys… and that tool is not built for clients.

cbosdo:
I just updated the PR with the latest innovations on the setup and update so it's not outdated too fast

PR #113, opened 2026-03-25
RFC: Confidential Compute Attestation for IBM Z series

CDellaGiusta:
Confidential Compute Attestation for IBM Z series in Multi Linux Manager

EPIC: https://github.com/SUSE/spacewalk/issues/28357
ISSUE: https://github.com/SUSE/spacewalk/issues/30113

[Rendered version](https://github.com/CDellaGiusta/uyuni-rfc/blob/master/accepted/00109-ibm-coco-attestation.md)


accepted/00000-ibm-coco-attestation.md
```diff
@@ -0,0 +1,452 @@
...
+**Store those additional inputs where the other inputs are**, that is in table `suseServerCoCoAttestationConfig`
+(object `ServerCoCoAttestationConfig`)
+
+We should expand the table, adding the following 3 fields
+
+- **in_data jsonb NOT NULL**
+
+>This field will contain data coming from settings input  (base 64 encoded, as json)  
+>  - hostKeyDocument  
+>  - secureExtensionHeader
```
mcalmer: in_data is ok for enhancement for the config table. Maybe we can find a different name for this field where you can store anything depending on the selected type.

accepted/00000-ibm-coco-attestation.md
```diff
@@ -0,0 +1,452 @@
...
+
+>This field will contain the result of creating the attestation request from the input data (base 64 encoded, as json)
+> - attestation measurement request (binary file, about 500 bytes)
+> - attestation_protection_key (key, 32 bytes)
+
+- **status character varying(32) COLLATE pg_catalog."default" NOT NULL**
+> The status field should be one of:
+> - PENDING when new and waiting for the attestation request creation
+> - SUCCEEDED if the creation of the attestation request is ok
+> - FAILED if for some reason (wrong or insufficient in_data) the attestation request creation fails
```
mcalmer: I think these 2 fields have nothing to do with configuration values. They are part of the attestation request.
We have field for them already in `suseServerCoCoAttestationReport`.
When an attestation action should be executed we need to generate these data and put them into the `in_data` field. This table also has a `status` field already.
In case we need an additional step to let the container prepare the data (generating a new attestation request) , we can enhance the status by a "PREPARE" state.

accepted/00000-ibm-coco-attestation.md
```diff
@@ -0,0 +1,452 @@
...
+- Data output back from the execution from the minion to the MLM server: json result of the application of
+the salt state
+
+There are 2 main modifications to be done, they can be both be related to the behaviour of the class
+CoCoAttestationAction
+
+### Modification A: different salt states for each environment
+The method `CoCoAttestationAction.getSaltCalls` must call `State.apply` with a different salt request data file
+depending on the type of attestation. In fact, depending on AMD or IBM environment, different input data,
+tool requests and outputs are needed.
```
mcalmer: Hmm, this is not needed when you follow your "Implemenation Proposal".
It is a bit unclear what you want here.

accepted/00000-ibm-coco-attestation.md
```diff
@@ -0,0 +1,452 @@
...
+For a proof of concept of this solution, see https://github.com/uyuni-project/uyuni/pull/11656/commits 
+(proof of concept on how to partition CoCoAttestationRequestData (see CoCoAttestationRequestDataCarlo and
+CoCoAttestationRequestDataCarlo2)
+
+
+
+### New service to create attestation requests, to be added to coco-attestation microservice 
+If we decide to create the requests on the coco-attestation microservice in previously mentioned
+Architectural choice n. 2, we need to implement the creation of the attestation request in the coco-attestation
+microservice. If not, this paragraph can be ignored.
```
mcalmer: Just a note: if we do this, we can also use it for AMD. For AMD we "only" generate the NONCE, but this is also something which could be done from the container. This would streamline the whole process:
- general configuration
- request to perform an attestation
- generate "input data" in the attestation container made for the attestation type
- make the salt state apply to get the data from the minion
- process the result from the minion and perform the attestation

accepted/00000-ibm-coco-attestation.md
CDellaGiusta: I chose the same names (`in_data`, `out_data`) used in table `suseServerCoCoAttestationReport`
Any name suggestions are welcome, of course!

accepted/00000-ibm-coco-attestation.md
CDellaGiusta: We have first to decide where the attestation request (binary file) is computed, and whether it is reused at each attestation measurement. Your solution is the one to go to if we decide to create the attestation request every time in MLM (in that case the `out_data` field is not even necessary). More on that when we'll discuss this point

accepted/00000-ibm-coco-attestation.md
CDellaGiusta: Yes, this has to be reviewed.
The idea is that depending on the test we're doing, the salt state apply should run different commands on the minion. But this can be done in the same state file, as I tried to do here: https://github.com/uyuni-project/uyuni/pull/11656/changes/837f1a064c93aa78fd070a9e92e4ecb918deef0b#diff-b86adc5262189788b149dcc05af06416b5f9f64ab79c8dbfa4fba092b4fc26df

accepted/00109-ibm-coco-attestation.md
```diff
@@ -0,0 +1,452 @@
...
+## The attestation process for IBM Z series 
+
+IBM provides a script named `pvattest` to perform all confidential computing operations needed.  
+This command is inside the package `s390-tools` (https://github.com/ibm-s390-linux/s390-tools 
+and https://build.opensuse.org/package/show/openSUSE:Factory/s390-tools).  
+**Important**: the package `s390-tools` is currently provided only on x86_64 and s390x platforms.  
+
+
+## Step 0: checking the `pvattest` script
+Be aware:  in order to run all the preparation and verification commands (`pvattest create`, `pvattest verify`)
```
rjmateus, line 76: We should not install it on the server image. It should be installed on the CoCo-specific images.

accepted/00000-ibm-coco-attestation.md
```diff
@@ -0,0 +1,452 @@
...
+>-  we can provide earlier feedback to the customer if the host-key document certificate is invalid
+(expired or the wrong one)
+>
+>cons:
+>- the two intermediate certificates for validation must be obtained (`DigiCertCA.crt` and `ibm-z-host-key-signing.crt`)
+>- their expiry date has to be monitored and new ones must be obtained if the case
+>- the more arguments you put, the more the chances the command to fail for some reason
+
+#### Final architectural choice 1):
+TBD
```
rjmateus: The recommendation from IBM is to use it. In my opinion, we must follow the vendor specification and verify the full chain; we would be responsible if anything goes wrong.

accepted/00000-ibm-coco-attestation.md
```diff
@@ -0,0 +1,452 @@
...
+With the microservice not activated, since the attestation verification cannot be done anyway, 
+do we really bother that the attestation measurement/execution is not doable?
+
+#### Unsafe: Compute the attestation request on the IBM Z minion
+>- Although s390-tools package dependency is installed anyway, we have assume the guest/minion as “untrusted” 
+(this is the whole point of the attestation).   
+Therefore it is not safe to compute the request on the guest/minion itself
+
+
+#### Final architectural choice 2):
```
rjmateus: In my opinion, we should compute it on the coco micro-service and avoid polluting the already to big server image.
I don't think it's a problem not being able to run the verification if the service is not enabled.

accepted/00000-ibm-coco-attestation.md
```diff
@@ -0,0 +1,452 @@
...
+does protect us from the reply attacks, according to IBM.  
+However, they never used this nonce data in this way.  
+Then, although recognizing the solution is working, IBM recommends to re-create a new attestation request every time,
+to avoid reply attacks (they don’t want to commit on our proposed solution).
+
+#### 3B) Compute every time we want to run an attestation request
+>This solution could be more complex, implementation is depending on where the request is computed
+(result of architectural choice n.2).
+
+#### Final architectural choice 3):
```
rjmateus: Request must be computed on the server side. If we need to create for each attestation, then we will have a hard link to the coco micro-service. The UI/API call to perform the attestation would trigger something on the CoCo micro-service, and then when the request is generated, something should be triggered on the server side to performe the attestation.
Create in the request on each attestation is what IBM suggests, so we should consider having it.

accepted/00109-ibm-coco-attestation.md
```diff
@@ -0,0 +1,540 @@
...
+
+
+
+
+### Architectural choice n. 2:  Where to compute the attestation request?
+The `pvattest` tool must be installed in the place where the attestation request is computed.
+The `pvattest create` command creates two files as outputs (the attestation request binary file and the protection key)
+that must be stored in the MLM server to be used for the attestation.
+
+- The attestation request could be computed on the minion guest/minion, since the s390-tools package dependency
```
rjmateus, line 161:
```suggestion
- The attestation request could technically be computed on the minion guest/minion, since the s390-tools package dependency
```

accepted/00109-ibm-coco-attestation.md
```diff
@@ -0,0 +1,540 @@
...
+is installed there anyway.  
+However, we have to assume the guest/minion as “untrusted” (this is the whole point of the attestation),
```
rjmateus, line 163:
```suggestion
However, guest/minion is “untrusted” (this is the whole point of the attestation), 
```

accepted/00109-ibm-coco-attestation.md
```diff
@@ -0,0 +1,540 @@
...
+The `pvattest create` command creates two files as outputs (the attestation request binary file and the protection key)
+that must be stored in the MLM server to be used for the attestation.
+
+- The attestation request could be computed on the minion guest/minion, since the s390-tools package dependency
+is installed there anyway.  
+However, we have to assume the guest/minion as “untrusted” (this is the whole point of the attestation), 
+so it is not safe to compute the request on the guest/minion itself.  
+
+
+- The attestation request could be computed directly on the MLM server: this is a simple solution,
```
rjmateus, line 167:
```suggestion
- The attestation request could be computed directly on the MLM server container: this is a simple solution,
```

The server itself is composed by several containers, the all group makes it the MLM server. The server container is the big fat container with all the services that don't have a specific container to run.

accepted/00000-ibm-coco-attestation.md
```diff
@@ -0,0 +1,540 @@
...
+
+## Step 2: Run the attestation measurement to get the attestation report
+Once the attestation request has been created, an attestation measurement must be done on the minion to get an
+attestation report.
+
+**Where**: on the minion/secguest (this needs `/dev/uv`)  
+
+**Inputs**:
+- **attestation_request.bin** (the generated attestation measurement request)
+- **random_user_nonce.txt** (file with random user generated nonce)
```
rjmateus: Since we are generating the attestation request at each execution, should we keep the usage of the random user nonce?

rjmateus (approved):
Overall looks good to me. Just some small clarifications and I think we are good to go.

rjmateus:
@CDellaGiusta, is this RFC already in the final version for release? I see you have made some changes but the diagram still shows the taskomatic task that is not needed anymore.
Thank you

CDellaGiusta:
> @CDellaGiusta, is this RFC already in the final version for release?

@rjmateus Yes, it should be.

> I see you have made some changes but the diagram still shows the taskomatic task that is not needed anymore. 

Actually it's "taskomatic", and not the old "taskomatic task". That was meant to express 4B, the point in time where Taskomatic schedules the action (where you see the shape with the `@Schedule time` label) in an asynchronous way with regards to the CoCo microservice

rjmateus:
ok, Thank you for the clarification. I will review it now

rjmateus (approved):

PR #116, opened 2026-06-03
RFC: native LDAP authentication

SURYAS1306:
**Rendered version** -> [RFC](https://github.com/SURYAS1306/uyuni-rfc/blob/rfc-native-ldap-authentication/accepted/00000-native-ldap-authentication.md)
## Summary

Native Java LDAP/AD authentication for Uyuni with optional JIT user provisioning and LDAP group → Uyuni role mapping via revived `rhnUserExtGroup` tables.

## Motivation

Replace host-level-only LDAP (PAM/SSSD) and proxy-only external authentication with in-product configuration, testing, and group-aware RBAC mapping.

## Proof of work (community bonding)

- Auth flow mapped: `LoginController` → `LoginHelper` → `UserManager` → `UserImpl`
- LDAP prototype validated using UnboundID SDK against local OpenLDAP fixtures
- Deployment workflow validated on containerized Uyuni (`mvn package`, `mgrctl cp`, Tomcat restart)
- Design aligned with mentor feedback (login order, failure modes, table reuse, RBAC scope)

## Open design questions for reviewers

See **Unresolved questions** in the RFC (six design-level items). Feedback especially welcome on secret and trust-material handling, scoping external groups to a server, superuser reachability, and JIT org assignment.

## Related links

- GSoC proposal: Native LDAP authentication provider
- Mentor references: RBAC RFC #95, Spacewalk/IPA wiki, RBAC Endpoint Mapping wiki, Satellite 6 LDAP references

### DIFF
--- a/accepted/00000-native-ldap-authentication.md
+++ b/accepted/00000-native-ldap-authentication.md
@@ -1,0 +1,269 @@
+- Feature Name: native_ldap_authentication
+- Start Date: 2026-06-04
+
+# Summary
+[summary]: #summary
+
+Add a native, Java-based LDAP/Active Directory authentication provider to Uyuni's web application. Administrators configure the directory connection from Uyuni itself instead of relying on the host operating system (SSSD/PAM) or a reverse proxy. A successful LDAP login can optionally create or update the Uyuni user, and LDAP group membership is mapped to Uyuni roles by reusing the existing external group mapping tables (`rhnUserExtGroup` / `rhnUserExtGroupMapping`) and the temporary-role lifecycle already used for header-based external authentication.
+
+# Motivation
+[motivation]: #motivation
+
+Today Uyuni can authenticate against a directory only indirectly:
+
+- **PAM/SSSD**: the host OS is configured for LDAP and Uyuni delegates the password check to PAM (`WEB_PAM_AUTH_SERVICE` plus a per-user `usePamAuthentication` flag). The directory configuration lives in `sssd.conf`, edited over the CLI, outside Uyuni.
+- **Header-based external auth** (`REMOTE_USER`): a frontend web server or IdP authenticates the user and passes identity and group headers, which Uyuni trusts (`LoginHelper.checkExternalAuthentication`).
+
+In both cases the directory integration is invisible to Uyuni: an administrator cannot configure, test, or troubleshoot it from the product, and there is no managed concept of an "LDAP group" inside Uyuni. The external group string is consumed transiently during a `REMOTE_USER` login and then discarded.
+
+This feature brings directory integration into Uyuni so that an administrator can point Uyuni at an Active Directory, FreeIPA, or OpenLDAP server, optionally enable just-in-time user creation, map a directory group such as `uyuni-admins` to the Org Admin role, and have those users log in with the right roles without touching `sssd.conf`. Local Uyuni users continue to work unchanged.
+
+# Detailed design
+[design]: #detailed-design
+
+## Existing authentication flow
+
+```text
+Browser / API client
+  -> LoginController.performLogin()
+     -> LoginHelper.checkExternalAuthentication()        (REMOTE_USER header path)
+     -> if no external user:
+        -> UserManager.loginUser(login, password)
+           -> UserFactory.lookupByLogin(login)
+           -> UserImpl.authenticate(password)
+              -> PAM   (if WEB_PAM_AUTH_SERVICE set and user.usePamAuthentication)
+              -> local DB password check otherwise
+     -> LoginHelper.successfulLogin()
+```
+
+The design reuses two existing patterns rather than reinventing them:
+
+- The **PAM service/factory shape** (`PamServiceFactory` → `DefaultPamServiceFactory` → `PamServiceWrapper`) is the structural template for isolating an external credential check behind a factory.
+- The **external-auth lifecycle** in `LoginHelper` already performs just-in-time user creation (`CreateUserCommand`), profile updates (`UpdateUserCommand`), external group to role mapping (`UserGroupFactory.lookupExtGroupByLabel`), and temporary-role assignment. This is the model for the LDAP user and role lifecycle.
+
+A standalone proof of concept (UnboundID LDAP SDK 7.0.4, Java 17) was run against a seeded OpenLDAP fixture to confirm the operations the design depends on: service-account bind, pooled connections, user search, group resolution by `(member=<userDN>)` without a `memberOf` overlay, a successful credential bind, and rejection of invalid credentials.
+
+## Proposed architecture
+
+LDAP authentication is implemented as a service behind a factory (mirroring PAM), with the user lifecycle handled at the login orchestration layer so just-in-time provisioning and role mapping are possible:
+
+```text
+LdapServiceFactory            (selects the configured implementation; testable seam)
+  -> LdapAuthenticationService
+       -> LdapConnectionProvider   (UnboundID LDAPConnectionPool, TLS, timeouts, failover)
+       -> LdapUserResolver         (bind/search/bind, attribute extraction)
+       -> LdapGroupResolver        (member= search and/or memberOf, normalization)
+  -> LdapLoginAdapter            (maps the LDAP result onto CreateUserCommand /
+                                  UpdateUserCommand / UserManager.resetTemporaryRoles)
+```
+
+Class names are indicative; the responsibilities are the design contract: connection handling, user lookup, group lookup, and mapping onto the Uyuni user lifecycle.
+
+```mermaid
+sequenceDiagram
+    participant B as Browser/API
+    participant UM as UserManager.loginUser
+    participant LS as LdapAuthenticationService
+    participant DIR as LDAP/AD server
+    participant LA as LdapLoginAdapter
+    participant DB as PostgreSQL
+
+    B->>UM: login + password
+    UM->>UM: lookupByLogin(login)
+    alt local user and valid local password
+        UM->>DB: local DB password check -> success
+    else native LDAP
+        UM->>LS: authenticate(login, password)
+        LS->>DIR: service bind, search user, bind as user DN, resolve groups
+        LS-->>UM: identity + attributes + group labels
+        UM->>LA: provision/update user, map groups
+        LA->>DB: web_contact + rhnUserGroupMembers (temporary='Y')
+    else PAM
+        UM->>UM: UserImpl PAM branch
+    end
+    UM->>B: successfulLogin(user)
+```
+
+## Login ordering
+
+The credential resolution order for username/password login is **Local DB → Native LDAP → PAM**. Concretely, in `UserManager.loginUser` (after the `REMOTE_USER` external-auth check that already runs first in `LoginController`):
+
+1. Look up the user by login. If a local user exists, has a usable local password, and that password verifies, log in locally.
+2. Otherwise attempt native LDAP (bind/search/bind). On success, provision or update the user, map groups to temporary roles, and log in.
+3. Otherwise attempt PAM (unchanged).
+4. Otherwise fail with a single generic message (see Failure modes).
+
+This ordering is why the LDAP step lives at the orchestration layer and not inside `UserImpl.authenticate()`: step 2 may need to create a user that does not yet exist locally, while `UserImpl.authenticate()` runs only after `UserFactory.lookupByLogin()` has found one.
+
+## LDAP authentication flow
+
+The provider uses the standard bind/search/bind sequence:
+
+1. Bind with the configured service account (or anonymous bind if explicitly enabled).
+2. Search for the user under the configured base DN using the configured filter; require **exactly one** match.
+3. Bind as the discovered user DN with the password supplied at login.
+4. Resolve the user's groups.
+5. Return the normalized login, profile attributes, and group labels to the lifecycle adapter.
+
+User input is never concatenated into a filter; values are escaped (the POC uses `Filter.createEqualityFilter`, which escapes automatically). Selecting the server type pre-fills attribute and filter defaults (AD vs. IPA/POSIX), so administrators rarely type raw attribute names:
+
+```text
+ldap.server_type        = ACTIVE_DIRECTORY        # or FREE_IPA / OPENLDAP / POSIX
+ldap.url                = ldaps://ad.example.com:636
+ldap.bind_dn            = CN=uyuni-reader,OU=Service,DC=example,DC=com
+ldap.user_base_dn       = OU=Users,DC=example,DC=com
+ldap.user_filter        = (&(objectClass=user)(sAMAccountName={login}))
+ldap.login_attribute    = sAMAccountName           # uid for FreeIPA/OpenLDAP/POSIX
+ldap.group_base_dn      = OU=Groups,DC=example,DC=com
+ldap.group_filter       = (&(objectClass=group)(member={userDn}))
+ldap.group_name_attribute = cn
+ldap.provisioning_mode  = JIT                      # default; EXISTING_ONLY opt-in
+ldap.default_org_id     = 1
+```
+
+## User provisioning
+
+Two modes, controlled by `provisioning_mode`:
+
+- **JIT** (default): a successful LDAP login creates the Uyuni user if absent, reusing `CreateUserCommand` exactly as `LoginHelper.newRemoteUser` does today (login set, non-usable placeholder password, attributes, org, temporary roles).
+- **EXISTING_ONLY**: LDAP authenticates only users that already exist in `web_contact`.
+
+JIT is the default because the project goal and the Satellite/Foreman precedent both assume directory users can log in without a pre-created account, and Uyuni already implements this lifecycle for external auth. When JIT is enabled, the UI requires a valid `default_org_id`, mirroring the existing `EXTAUTH_DEFAULT_ORGID` behavior. Attribute- or group-based org mapping is a later enhancement. `EXISTING_ONLY` remains available for deployments that require manual user creation first.
+
+On each successful login, selected profile fields (first name, last name, email) are refreshed from LDAP via `UpdateUserCommand`, as `updateRemoteUser` does; missing LDAP attributes do not overwrite existing Uyuni values.
+
+## Group resolution and role mapping
+
+The portable baseline is a group-side search, which the POC confirmed works without a server-side `memberOf` overlay:
+
+```text
+(&(objectClass=groupOfNames)(member=<userDN>))     # OpenLDAP / generic
+(&(objectClass=group)(member=<userDN>))            # Active Directory
+```
+
+`memberOf` on the user entry is supported as an optional optimization where the directory exposes it. Group labels are normalized to a configured attribute (default `cn`), with full-DN matching available where simple names are ambiguous. Nested groups are AD-specific (matching-rule OID `1.2.840.113556.1.4.1941`) and are deferred to a later iteration.
+
+Role mapping reuses the existing external group mechanism rather than introducing a new model. `rhnUserGroup` remains the role-assignment model; `rhnUserExtGroup` / `rhnUserExtGroupMapping` are dormant but reusable; `access.accessGroup` is the authorization layer for new endpoints, not the LDAP role target.
+
+```text
+LDAP groups
+  -> normalize to external group labels
+  -> UserGroupFactory.lookupExtGroupByLabel(label)   (rhnUserExtGroup -> rhnUserExtGroupMapping)
+  -> resolve org-scoped rhnUserGroup rows (the roles)
+  -> assign as temporary roles (CreateUserCommand/UpdateUserCommand.setTemporaryRoles)
+  -> on each login, UserManager.resetTemporaryRoles recomputes the set
+```
+
+Manually assigned permanent roles are never removed by LDAP login; only temporary (LDAP-derived) roles are recomputed, so a user removed from a directory group loses the corresponding Uyuni role on next login. The existing `EXTAUTH_KEEP_TEMPROLES` switch (`UserExternalHandler.setKeepTemporaryRoles`) governs whether temporary roles survive a later non-LDAP login.
+
+This also promotes the LDAP group to a first-class, managed entity: the `rhnUserExtGroup` row becomes the persisted representation of a known LDAP group, optionally scoped to a specific server (see Database design), managed from the Admin UI/API. The existing `REMOTE_USER` path is left untouched in v1; the LDAP adapter calls the same `LoginHelper.getRolesFromExtGroups()` resolution, and integration tests cover these previously dormant paths.
+
+## RBAC integration
+
+The new RBAC model (`access.namespace`, `access.endpoint`, `access.accessGroup`) is used only to protect the new LDAP configuration endpoints: the setup pages and API methods are registered in the RBAC endpoint mapping[^3] so only privileged administrators can read or change LDAP configuration. LDAP-derived user roles continue to flow through `rhnUserGroup`. Automatic population of `access.accessGroup` from LDAP groups is out of scope for v1.
+
+## Database design
+
+Reused tables:
+
+| Table | Use |
+| --- | --- |
+| `web_contact` | user identity; JIT users get a non-usable placeholder password |
+| `rhnUserGroup` / `rhnUserGroupType` | org-scoped role instances (unchanged) |
+| `rhnUserGroupMembers` | role membership; `temporary='Y'` for LDAP-derived roles |
+| `rhnUserExtGroup` / `rhnUserExtGroupMapping` | external group to role mapping (revived) |
+| `suseCredentials` | bind password, via a new `ldap` credential type (see below) |
+
+One new entity (working name `suseLdapAuthServer`, one row per configured directory) holds the connection and mapping configuration. Rather than fix the exact columns, types, and constraints now, the design captures the data the entity must hold, grouped by purpose; the concrete schema is settled during implementation:
+
+- **Connection** - label, enabled flag, server type, host, port, transport, and a connect/response timeout. Server type (`ACTIVE_DIRECTORY` / `FREE_IPA` / `OPENLDAP` / `POSIX`) and transport (`LDAPS` / `STARTTLS` / `PLAIN`) are modeled as Java enums, as CoCo attestation models `env_type`[^6].
+- **Bind account** - the service-account bind DN (null meaning anonymous bind) and a reference to the stored bind password (see below).
+- **User lookup** - user base DN, user search filter, login attribute, and optional first-name, last-name, and email attributes.
+- **Group lookup** - group base DN, group filter, group-name attribute (default `cn`), a `memberOf` toggle, and an optional pre-authentication scope filter.
+- **Provisioning** - provisioning mode (`JIT` / `EXISTING_ONLY`) and the default org applied to JIT-created users.
+
+This entity is the single source of truth for a directory, and the enum-backed fields keep server-type and transport handling type-safe in Java rather than relying on free-form strings.
+
+To make external groups first-class and optionally LDAP-scoped, the design adds a nullable reference from `rhnUserExtGroup` to the new directory entity. A null value preserves today's server-agnostic behavior (used by `REMOTE_USER`); a set value scopes the mapping to one directory. Modeling this as a nullable link on the existing table, rather than a parallel new table, avoids duplicating the established external-group mechanism.
+
+The **bind password** is stored in `suseCredentials` under a new `ldap` credential type, reusing the `PasswordBasedCredentials` pattern already used for SCC, registry, and cloud-RMT credentials (Base64-encoded at rest). In practice this adds an `ldap` type to the credential model and a corresponding credentials entity; the password is write-only in the UI/API, as SCC credentials are. Whether Base64-at-rest is sufficient or stronger encryption is required is left open below.
+
+## Transport security
+
+Production deployments must use encrypted transport (`LDAPS` or `STARTTLS`); `PLAIN` is allowed only for explicit dev/test. Trust material is validated against the JVM trust store. Because the server container already mounts CA anchors at `/etc/pki/trust/anchors/`, administrators can add the directory CA there or upload it through the UI. The exact integration is left open below.
+
+## Connection handling
+
+A pooled service connection (`LDAPConnectionPool`, validated in the POC) is used for searches, and a short-lived dedicated connection is used for each user credential bind. A connect/response timeout of 3–5s ensures an unreachable directory fails fast. Multiple servers can be combined via UnboundID's `FailoverServerSet` for HA in a later iteration.
+
+## Failure modes
+
+| Situation | Behavior |
+| --- | --- |
+| Active session, LDAP goes down | Unaffected; LDAP is checked only at login, never mid-session |
+| Local user logs in, LDAP down | Unaffected; local auth is first and independent |
+| LDAP user logs in, LDAP unreachable | Fail after the timeout; show "invalid credentials or directory unreachable, try again later" |
+| LDAP user, wrong password | Reject (invalid credentials) |
+| LDAP user also has a local password | No silent fallback to local after LDAP failure |
+| User search returns 0 or >1 entries | Treat as authentication failure (ambiguous) |
+| Group lookup fails after successful bind | Authenticate, assign no LDAP-derived roles, log the failure |
+| User maps to no Uyuni roles | Login succeeds with no temporary roles |
+
+The generic failure message intentionally does not distinguish "user unknown" from "wrong password" from "server down" to avoid user enumeration; details go to the server log.
+
+## User interface and API
+
+A new **Admin -> Setup -> LDAP** area follows Satellite's proven four-tab layout: **Server** (host, port, transport, type), **Account** (bind DN/password, base DNs, filters, provisioning mode, default org), **Attribute mappings** (login/name/email/group attributes, pre-filled from server type), and **User Groups** (external group to role bindings, extending the existing `ExtGroupDetailAction` UI). Three test actions give immediate feedback before LDAP is enabled and map directly onto POC operations: test connection (service bind), test user lookup (search), and test group resolution. A freshly provisioned user with no mapped groups logs in with no permissions until a mapping matches.
+
+Both API interfaces are provided, following the HTTP API RFC convention: the existing `user.external` XML-RPC namespace (`UserExternalHandler`) is extended for group-to-role mappings, a new `auth.ldap` namespace handles server CRUD and the test operations, and the same handlers are auto-exposed over JSON via the Spark route wrapper.
+
+## Phasing
+
+1. **Backend authentication** - `LdapServiceFactory` + `LdapAuthenticationService`, bind/search/bind, Local -> LDAP -> PAM ordering, minimal config, unit + integration tests against the fixture.
+2. **Provisioning + roles** - JIT user creation, profile sync, group-to-role mapping via `rhnUserExtGroup`, temporary-role reset.
+3. **UI + API** - four-tab Admin UI, test actions, XML-RPC + JSON endpoints, RBAC registration.
+4. **Advanced** - auth-source filter, multiple servers/failover, `memberOf` and AD nested groups.
+
+## Testing
+
+A local OpenLDAP fixture (seeded with `alice`/`bob` and `uyuni-admins`/`uyuni-users`) backs integration tests, seeded from the existing POC. Unit tests cover filter construction/escaping, attribute mapping, group normalization, role mapping, and temporary-role reset. Regression tests confirm local and PAM logins and the ordering are unaffected. Note: the core test tree currently has pre-existing compile breakage (`RhnMockHttpServletHttpServletResponse`, `SimpleTestingResponse`); whether it is fixed as part of this work is for the maintainers to decide.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- It adds code to a security-sensitive path; mistakes in ordering, fallback, filter escaping, or mapping could cause auth/authz bugs.
+- It introduces a runtime dependency on an external directory; timeouts and failure modes must be disciplined.
+- It likely adds a new third-party Java dependency (UnboundID LDAP SDK) that must clear licensing/packaging review.
+- It revives dormant `rhnUserExtGroup` code that may need repair and test coverage.
+- LDAP/AD deployments vary widely; v1 must choose clear defaults and document its limits rather than support every schema.
+
+# Alternatives
+[alternatives]: #alternatives
+
+- **Stay on PAM/SSSD only** - no new code, but no in-product configuration, testing, profile sync, or group-to-role mapping. Does not meet the goal.
+- **Improve only header-based `REMOTE_USER`** - keeps authentication outside Uyuni and still requires a proxy or IdP, with no in-product visibility. The Spacewalk/IPA integration[^4] is the precedent for this approach and its limits.
+- **LDAP only inside `UserImpl.authenticate()`** - minimal and mirrors the PAM branch, but cannot do first-login provisioning because user lookup precedes authentication; viable only for an `EXISTING_ONLY` build.
+- **Implementation-level choices** - for the LDAP client library, JNDI avoids a third-party dependency but is lower-level with harder pooling, StartTLS, and testability, while Apache Directory LDAP API and Spring LDAP are further candidates; the UnboundID SDK[^9] is the baseline pending licensing review. For bind-password storage, a plain column on `suseLdapAuthServer` or a container-level secret were considered but rejected in favor of the existing `suseCredentials` pattern for consistency with how Uyuni already stores SCC and registry secrets.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+1. **Secret and trust material** - is Base64-at-rest (as for SCC/registry credentials) acceptable for the bind password, or is encryption required, and where should the directory CA live: the JVM trust store, the container's `/etc/pki/trust/anchors/`, or an upload-through-UI flow?
+2. **Scoping external groups to a server** - is the nullable `rhnUserExtGroup.ldap_server_id` FK the right approach, or should mappings stay server-agnostic?
+3. **Superuser reachability** - may an LDAP group grant the top-level admin role, or should the highest privileges remain local-only?
+4. **Org assignment for JIT users** - is a single configured default org enough for v1, or is attribute-/group-based mapping needed early?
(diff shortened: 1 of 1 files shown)


SURYAS1306:
Validation update:

I refined the draft based on the POC and a second source-reading pass.

What was validated:

- OpenLDAP POC confirms service bind, user search, credential bind, group-side group resolution, wrong-password rejection, and validates the proposed JIT/existing-user flow.
- Source tracing confirms LDAP needs to live at the login orchestration layer because `UserManager.loginUser` performs `lookupByLogin` before `UserImpl.authenticate`.
- Source tracing confirms `rhnUserExtGroup` / `rhnUserExtGroupMapping` are live and already used by header-based external authentication; mapping is external group → role type, with org scoping via `rhnUserExtGroup.org_id`.
- Source tracing confirms the reuse assumptions around `PasswordBasedCredentials`, `CredentialsType`, `PamServiceFactory`, and `UserManager.resetTemporaryRoles`.

Changes made to the RFC:

- Added a scope-of-validation note so the draft is clear about what the POC did and did not cover.
- Added the validation findings and refined the role-mapping details based on the source analysis.
- Added the `LoginHelper` extraction and temporary-role/RBAC migration risks.
- Reduced unresolved questions to a focused set of design-level decisions for reviewer input.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,275 @@
...
+        UM->>LA: provision/update user, map groups
+        LA->>DB: web_contact + rhnUserGroupMembers (temporary='Y')
+    else PAM
+        UM->>UM: UserImpl PAM branch
+    end
+    UM->>B: successfulLogin(user)
+```
+
+## Login ordering
+
```
admd, line 135: We don't really need ordering or fallback options. That's not very robust and also not very pefromative.

based on what you have described and here I am assuming something like ldap_server_id will be added in web_contact table so it would be something like this



- A user with ldap_server_id = null and usePamAuthentication = false → local DB. Full stop.
- A user with ldap_server_id = null and usePamAuthentication = true → UserManager.loginUser() → UserImpl.authenticate() → PAM branch. Full stop.
- A user with ldap_server_id set → LdapAuthService directly. Full stop.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,275 @@
...
+  -> LdapLoginAdapter            (maps the LDAP result onto CreateUserCommand /
+                                  UpdateUserCommand / UserManager.resetTemporaryRoles)
+```
+
+Class names are indicative; the responsibilities are the design contract: connection handling, user lookup, group lookup, and mapping onto the Uyuni user lifecycle.
+
+```mermaid
+sequenceDiagram
+    participant B as Browser/API
+    participant UM as UserManager.loginUser
```
admd, line 77: I would suggest handling most of the part in LoginController , pariculary look into checkExternalAuthentication before really directly going into UserManager.loginUser() that will make things simple.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,275 @@
...
+
+**Scope of validation.** To be precise about what this evidence does and does not cover: the POC ran against **OpenLDAP only**, over **plain LDAP** (no TLS), as a **standalone simulation** of the orchestration logic. It did not run against Active Directory or FreeIPA, did not exercise LDAPS/StartTLS or trust-store validation, did not test anonymous bind or the `memberOf` path (only group-side `member=` search), and was not wired into the real `UserManager.loginUser` flow or `CreateUserCommand` against a live database. Those paths are designed from code reading and precedent, not yet executed; the unresolved questions below call out the ones that carry risk.
```
aaannz: This is not related to the existing flow topic. Its a nitpick, but add `## Proof of concept` section.

accepted/00000-native-ldap-authentication.md
admd: Ok reading this again, I can see that ldap_server_id might create confusion here as that would indicate per user an ldap_server_id. So instead of that, maybe add auth_type in rhnUserInfo where auth_type could be ('LOCAL', 'PAM', 'LDAP').  But main idea is the same that we should already have the enough information to decide where should we send the authentication request.

And if user doesn't exist yet anywhere and lookup fail, there we will need to go check and authenticate with our ldap servers and if that fails too, throw the final login exception.

To have multiple LDAP servers then, we can then define some priority and try one by one.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,275 @@
...
+## Failure modes
+
+| Situation | Behavior |
+| --- | --- |
+| Active session, LDAP goes down | Unaffected; LDAP is checked only at login, never mid-session |
+| Local user logs in, LDAP down | Unaffected; local auth is first and independent |
+| LDAP user logs in, LDAP unreachable | Fail after the timeout; show "invalid credentials or directory unreachable, try again later" |
+| LDAP user, wrong password | Reject (invalid credentials) |
+| LDAP user also has a local password | No silent fallback to local after LDAP failure |
+| User search returns 0 or >1 entries | Treat as authentication failure (ambiguous) |
```
aaannz: This should be also logged with details, so admin can correct it. But just to the log. No details presented to the user except invalid credentials or user unknown.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,275 @@
...
+| Group lookup fails after successful bind | Authenticate, assign no LDAP-derived roles, log the failure |
+| User maps to no Uyuni roles | Login succeeds with no temporary roles |
+
+The generic failure message intentionally does not distinguish "user unknown" from "wrong password" from "server down" to avoid user enumeration; details go to the server log.
+
+## User interface and API
+
+A new **Admin -> Setup -> LDAP** area follows Satellite's proven four-tab layout: **Server** (host, port, transport, type), **Account** (bind DN/password, base DNs, filters, provisioning mode, default org), **Attribute mappings** (login/name/email/group attributes, pre-filled from server type), and **User Groups** (external group to role bindings, extending the existing `ExtGroupDetailAction` UI). Three test actions give immediate feedback before LDAP is enabled and map directly onto POC operations: test connection (service bind), test user lookup (search), and test group resolution. A freshly provisioned user with no mapped groups logs in with no permissions until a mapping matches.
```
aaannz: What do you mean by `proven four-tab layout`?

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,275 @@
...
+4. **Advanced** - auth-source filter, multiple servers/failover, the `memberOf` path, AD nested groups, and AD large-group range retrieval (`member;range=`).
+
+## Testing
+
+A local OpenLDAP fixture (seeded with `alice`/`bob` and `uyuni-admins`/`uyuni-users`) backs integration tests, seeded from the existing POC. Unit tests cover filter construction/escaping, attribute mapping, group normalization, role mapping, and temporary-role reset. Regression tests confirm local and PAM logins and the ordering are unaffected. The primary validation target for the v1 test matrix and documentation is still to be confirmed with maintainers (Active Directory, FreeIPA, or OpenLDAP); other directories are best-effort. Note: the core test tree currently has pre-existing compile breakage (`RhnMockHttpServletHttpServletResponse`, `SimpleTestingResponse`); whether it is fixed as part of this work is for the maintainers to decide.
+
+# Drawbacks
+[drawbacks]: #drawbacks
+
+- It adds code to a security-sensitive path; mistakes in ordering, fallback, filter escaping, or mapping could cause auth/authz bugs.
```
aaannz: This can be partly mitigated by what Abid suggested: No ordering and no fallback.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,275 @@
...
+- **LDAP only inside `UserImpl.authenticate()`** - minimal and mirrors the PAM branch, but cannot do first-login provisioning because user lookup precedes authentication; viable only for an `EXISTING_ONLY` build.
+- **Implementation-level choices** - for the LDAP client library, JNDI avoids a third-party dependency but is lower-level with harder pooling, StartTLS, and testability, while Apache Directory LDAP API and Spring LDAP are further candidates; the UnboundID SDK[^9] is the baseline pending licensing review. For bind-password storage, a plain column on `suseLdapAuthServer` or a container-level secret were considered but rejected in favor of the existing `suseCredentials` pattern for consistency with how Uyuni already stores SCC and registry secrets.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+1. **Secret storage and trust material** - is Base64-at-rest (as for SCC/registry credentials) acceptable for the bind password, or is encryption required, and where should the directory CA live: the JVM trust store, the container's `/etc/pki/trust/anchors/`, or an upload-through-UI flow?
```
aaannz: 1) changing the secret storage is a different task. Let's use method we currently have and different effort can be prepared to migrate to salted credentials. Interface should remain the same so should be orthogonal to this task.

2) With Hub design we moved to CA upload via Web/API. See Admin -> Hub Configuration -> Add Peripheral.
Unfortunately HUB ISS was not implemented using generic CA storage, and we AFAIK do not have CA handling services. So I am for Web/API upload, but the implementation will be LDAP specific at this time.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,393 @@
...
+
+Step 2 is a dispatch, not a cascade: each login hits exactly one credential backend. `checkLdapAuthentication` mirrors the external-auth lifecycle (`newRemoteUser` / `updateRemoteUser` / `getRolesFromExtGroups`).
+
+**Known user:** lookup succeeds → read `auth_type` → dispatch to exactly one backend. Wrong password on that backend fails immediately; no fallback to another backend.
+
+**Unknown user (lookup fails):** try configured LDAP servers in administrator-defined priority order. On success, JIT-provision the user (if `provisioning_mode = JIT`) with `auth_type = LDAP`. If all servers fail, reject with a generic login error.
+
+**Migration:** existing users with `use_pam_authentication = Y` become `auth_type = PAM`; all others default to `LOCAL`. JIT-created LDAP users get `auth_type = LDAP`.
+
+This is why LDAP does not belong inside `UserImpl.authenticate()`: provisioning and group mapping require the `LoginHelper` lifecycle, and `UserManager.loginUser` assumes the user already exists.
```
cbbayburt: While this works with Web UI and HTTP API auth, XML-RPC API still uses the alternative path (`UserManager.loginUser -> UserImpl.authenticate`). If we don't adapt/replace this path, LDAP users won't be able to login to XML-RPC API.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,393 @@
...
+
+- **Connection** — label, enabled, server type, host, port, transport, timeout, priority (for multi-server probe order).
+- **Bind account** — reference to `suseCredentials` (bind DN as `username`; no row = anonymous bind).
+- **User lookup** — user base DN, filter, login and profile attribute names.
+- **Group lookup** — group base DN, filter, group-name attribute, optional `memberOf` toggle.
+- **Provisioning** — mode (`JIT` / `EXISTING_ONLY`) and default org for JIT users.
+
+A nullable directory FK on `rhnUserExtGroup` optionally scopes external-group mappings to one server (null = server-agnostic, as today for `REMOTE_USER`).
+
+Bind passwords reuse the existing `PasswordBasedCredentials` / `suseCredentials` pattern as-is (bind DN stored as credential `username`, Base64-at-rest like SCC/registry credentials). A future migration to salted or encrypted credential storage is orthogonal to this feature and should not change the LDAP integration interface. Schema constraints on `suseCredentials` need a new `ldap` credential type.
```
cbbayburt: `suseCredentials.username` is currently defined as `VARCHAR(64)`, so it probably won't fit some DNs. We should either extend this field, or store the DN in `suseLdapAuthServer` instead.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,393 @@
...
+  -> mapped role types (rhnUserGroupType); unmapped labels skipped
+  -> temporary roles via CreateUserCommand / UpdateUserCommand
+  -> UserManager.resetTemporaryRoles on each login
+```
+
+Manually assigned permanent roles are never removed. Only temporary (LDAP-derived) roles are recomputed. The `private` methods in `LoginHelper` (`getRolesFromExtGroups`, `newRemoteUser`, `updateRemoteUser`) must be extracted into a shared component before the LDAP path can call them.
+
+## RBAC integration
+
+The new RBAC model (`access.namespace`, `access.endpoint`, `access.accessGroup`) protects only the new LDAP configuration endpoints[^3]. LDAP-derived user roles continue through `rhnUserGroup`. Automatic `access.accessGroup` population from LDAP is out of scope for v1.
```
cbbayburt: The group mapping should primarily target RBAC access groups, not just the legacy ones (`rhnUserGroupType`).

Except for the administrative roles, (`satellite_admin` and `org_admin`), we've already migrated to RBAC access groups, so I think this mapping needs to be done from the get-go.

OTOH, administrative roles have some special handling in the backend, so it's going to be a while until we can completely migrate away from those, so `satellite_admin` and `org_admin` in `rhnUserGroupType` need to be considered as well.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,393 @@
...
+- **Implementation choices** — JNDI vs. UnboundID SDK[^9]; bind password in `suseCredentials` vs. plain column (latter rejected).
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+1. **LDAP server per user** — should `rhnUserInfo` also store which directory server an LDAP user belongs to (`ldap_server_id`), or probe all servers in priority order on every LDAP login?
+2. **Scoping external groups to a server** — nullable directory FK on `rhnUserExtGroup` (extending `(label, org_id)` uniqueness), or server-agnostic mappings only?
+3. **Superuser reachability** — may an LDAP group grant the top-level admin role, or remain local-only?
+4. **Org assignment for JIT users** — single default org enough for v1, or attribute/group-based mapping needed early?
+5. **Temporary roles under RBAC migration** — target legacy temporary roles for v1 and adapt later, or align with RBAC migration now?
```
cbbayburt: Also mentioned above, I think we should align with RBAC from the start.

accepted/00000-native-ldap-authentication.md
cbbayburt: With RBAC, all new users are automatically joined to the access group: `regular_user` (in `createUserCommand`). We may want to have the option to override this for LDAP.

cbbayburt (commented):
Thanks for the PR, @SURYAS1306.

Below are my comments, mostly from the RBAC perspective.

SURYAS1306:
### Review feedback update:

Updated the draft to address @cbbayburt's comments.

### What was addressed:

- **XML-RPC path:** `UserManager.loginUser` bypasses `LoginController`, so the LDAP credential check now lives in a shared service it delegates to; provisioning/mapping stay in `LoginHelper`.

- **Credential storage:** `suseCredentials.username` is `VARCHAR(64)` (too short for DNs); bind DN now stored on `suseLdapAuthServer`, password only in `suseCredentials`.

- **RBAC mapping:** LDAP groups map to RBAC access groups by default; only `org_admin`/`satellite_admin` stay on the legacy path.

- **Provisioning:** added a per-server `auto_join_regular_user` toggle.

### Remaining open question:

How to mark LDAP-derived RBAC access-group membership, since RBAC has no `temporary` flag today.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+
+Add a native, Java-based LDAP/Active Directory authentication provider to Uyuni's web application. Administrators configure the directory connection from Uyuni itself instead of relying on the host operating system (SSSD/PAM) or a reverse proxy. A successful LDAP login can optionally create or update the Uyuni user, and LDAP group membership is mapped to Uyuni roles by reusing the existing external group mapping tables (`rhnUserExtGroup` / `rhnUserExtGroupMapping`) and the temporary-role lifecycle already used for header-based external authentication.
+
+## At a glance
+
+- **Admin value** — configure, test, and troubleshoot LDAP from Uyuni; map directory groups to Uyuni roles.
+- **Unchanged** — local DB users, per-user PAM users, and header-based `REMOTE_USER` auth.
+- **Login orchestration** — `LoginController` → `LoginHelper` (same layer as `checkExternalAuthentication`), not inside `UserManager.loginUser`.
+- **Per-user routing** — `auth_type` on `rhnUserInfo` (`LOCAL` / `PAM` / `LDAP`); no cascade between backends for known users.
+- **JIT** — unknown logins probe configured LDAP servers (priority order); on success, create the user with `auth_type = LDAP`.
```
rjmateus: Should we try to login on any LDAP server plus the pam? Adding also pam to this Logic would allow legacy users to star using this capability (which Im not sure if is good. If we keep this only on LDAP they would have an incentive to migrate to the new mechanism).

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+- **Role mapping** — LDAP groups map to RBAC access groups by default; `org_admin`/`satellite_admin` stay on legacy roles until they migrate.
```
rjmateus: Roles on LDAP must have a specific prefix, like uyuni, and must map only those (removing the prefix when saving on uyuni dB)

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+     -> if no external user:
+        -> UserManager.loginUser(login, password)
+           -> UserFactory.lookupByLogin(login)
+           -> UserImpl.authenticate(password)
+              -> PAM   (if WEB_PAM_AUTH_SERVICE set and user.usePamAuthentication)
+              -> local DB password check otherwise
+     -> LoginHelper.successfulLogin()
+```
+
+Native LDAP extends this at the same orchestration layer: after `checkExternalAuthentication`, a new `checkLdapAuthentication` step handles directory users and JIT provisioning before `UserManager.loginUser` is called for `LOCAL` / `PAM` users only (see Login routing).
```
rjmateus, line 49: Which one is called would depend on the login method defined for the user, or Im getting it wrong?

rjmateus (commented):
First round

accepted/00000-native-ldap-authentication.md
SURYAS1306: Agreed - there will be no LDAP→PAM cascade. Each known user authenticates through a single backend selected by `auth_type` (`LOCAL` / `PAM` / `LDAP`). Existing PAM users remain `PAM` until migrated. Only unknown users probe the configured LDAP servers for JIT provisioning; PAM is never used as a fallback.

accepted/00000-native-ldap-authentication.md
SURYAS1306: Makes sense - I'll add a configurable group prefix (default: `uyuni`). Only LDAP groups matching that prefix will be considered, and the prefix will be stripped before looking up the role mapping (e.g. `uyuni-admins` → `admins` in `rhnUserExtGroup`). The filtering will happen before `lookupExtGroupByLabel`.

One question: should the prefix be configurable per LDAP server, or should we keep a single global prefix (default `uyuni`)?

accepted/00000-native-ldap-authentication.md
SURYAS1306: Yes - for known users, `auth_type` determines the authentication backend (`LOCAL` / `PAM` / `LDAP`). `REMOTE_USER` authentication remains unchanged and is checked first. Unknown users probe the configured LDAP servers for JIT provisioning; if no match is found, authentication fails normally.

accepted/00000-native-ldap-authentication.md
rjmateus: AS a first step I think it should be safe to keep the default prefix `uyuni_`, and not let it to be configurable.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+**Unknown user (lookup fails):** try configured LDAP servers in administrator-defined priority order. On success, JIT-provision the user (if `provisioning_mode = JIT`) with `auth_type = LDAP`. If all servers fail, reject with a generic login error.
+
+**Migration:** existing users with `use_pam_authentication = Y` become `auth_type = PAM`; all others default to `LOCAL`. JIT-created LDAP users get `auth_type = LDAP`.
+
+### Two authentication entry points (Web UI/HTTP API vs. XML-RPC)
+
+Uyuni has two credential-checking entry points, and LDAP must be reachable from both:
+
+- **Interactive path** — `LoginController.performLogin()` → `LoginHelper`. Runs the full lifecycle: credential check, JIT provisioning, profile sync, and group-to-role mapping.
+- **Programmatic path** — `auth.login` (`AuthHandler`) → `UserManager.loginReadOnlyUser` → `UserManager.loginUser(login, password)` → `UserImpl.authenticate(password)`. This path does **not** pass through `LoginController`/`LoginHelper`; it is what `spacecmd` and XML-RPC API clients use.
```
rjmateus, line 145: @cbbayburt do you know if have already migrate the read-only API user implementation to be based on RBAC?
If we have that in place maybe we can delegate the XML-RPC API login directly on the `LoginController`/`LoginHelper` (or have it integrated in a single execution path somehow).

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+
+Because credential verification lives in the standalone `LdapAuthenticationService`, both paths call the same service. `UserManager.loginUser(login, password)` becomes `auth_type`-aware: for an `auth_type = LDAP` user it verifies the password through `LdapAuthenticationService` instead of the local/PAM `UserImpl.authenticate` branch, so an already-provisioned LDAP user can authenticate over XML-RPC.
+
+The **provisioning and group-to-role lifecycle stays in the `LoginHelper` path** (it needs the request-scoped external-group context). A first-time LDAP user who has never logged in interactively therefore does not yet exist in `web_contact`; JIT provisioning happens on their first Web UI/HTTP API login, after which the XML-RPC path works normally. (Whether JIT should also be driven from the XML-RPC path in v1 is an open question for review.)
```
rjmateus: It should be fine to force at least one UI login. The only problem is for users that may be XML-RPC API only, who may not have access to the web UI.
For those cases the uyuni admin can also manually create the user. 
However, if we integrate the authentication mechanisms into the same path, the behavior can be the same.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+```
+
+## User provisioning
+
+Two modes, controlled per LDAP server (`provisioning_mode`):
+
+- **JIT** (default): a successful LDAP login creates the Uyuni user if absent, reusing `CreateUserCommand` as `LoginHelper.newRemoteUser` does today. The new user is stored with `auth_type = LDAP`.
+- **EXISTING_ONLY**: LDAP authenticates only users that already exist in `web_contact` with `auth_type = LDAP`.
+
+When JIT is enabled, the UI requires a valid `default_org_id`, mirroring `EXTAUTH_DEFAULT_ORGID`. On each successful login, profile fields (first name, last name, email) are refreshed from LDAP via `UpdateUserCommand`; missing LDAP attributes do not overwrite existing Uyuni values.
```
rjmateus: Should we allow the creation org to be controlled from the LDAP side, with some customer OU group?
For example, a group named uyuni_org_<ID>  or a custom field named `uyuni_org` (I'm just thinking louder; I'm not telling you to implement it this way). 
If we implement it like this, we can still have a `default_org_id`, to map to a default organization if none is defined for that user.
We would need to be careful not to allow org change, if the value changes in the LDAP side. It can be complicated to change the organization a user is assigned to after it's creating, since a lot of data access depends on that.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+(&(objectClass=groupOfNames)(member=<userDN>))     # OpenLDAP / generic
+(&(objectClass=group)(member=<userDN>))            # Active Directory
+```
+
+`memberOf` is an optional optimization. Nested groups are deferred.
+
+Role mapping reuses the `rhnUserExtGroup` / `rhnUserExtGroupMapping` machinery (live today for `REMOTE_USER`), but the mapping **target is dual**, reflecting the in-progress RBAC migration (reviewer input):
+
+- **RBAC access groups (default target).** Most roles have already migrated to RBAC access groups — `channel_admin`, `config_admin`, `system_group_admin`, `activation_key_admin`, `image_admin`, and `regular_user` (all present in `AccessGroupFactory`). LDAP groups map to these via `user.addToGroup(AccessGroup)` / `removeFromGroup`.
+- **Legacy roles (`rhnUserGroupType`).** `org_admin` and `satellite_admin` are **not** yet migrated (absent from `AccessGroupFactory`, still in `RoleFactory`) because of special backend handling. LDAP groups that grant these two administrative roles continue to map through the legacy `rhnUserGroup` / temporary-role mechanism.
```
rjmateus: Should we allow these special super users to be controlled from LDAP? I would not mind that in a first implementation, we would only allow to assign this speciall groups directly in uyuni side. @admd @cbbayburt what do you think?

rjmateus (commented):
Another round. Later today I will finish the review

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+| RBAC access groups (`access.accessGroup` + membership) | default target for LDAP-derived roles; membership recompute is the open design point above |
+| `rhnUserExtGroup` / `rhnUserExtGroupMapping` | external group to role/access-group mapping |
+| `suseCredentials` | bind password via new `ldap` credential type |
+
+One new entity (`suseLdapAuthServer`, one row per directory) holds connection and mapping configuration:
+
+- **Connection** — label, enabled, server type, host, port, transport, timeout, priority (for multi-server probe order).
+- **Bind account** — bind DN stored **on `suseLdapAuthServer`** (its own column); password referenced from `suseCredentials` (no bind DN + no credential = anonymous bind).
+- **User lookup** — user base DN, filter, login and profile attribute names.
+- **Group lookup** — group base DN, filter, group-name attribute, optional `memberOf` toggle.
```
rjmateus: Is User and group lookup only one table field with all this data? Like, in one field in the database to hold  base DN, filter, group-name, etc?

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+
+Bind passwords reuse the existing `PasswordBasedCredentials` / `suseCredentials` pattern (Base64-at-rest like SCC/registry credentials), needing a new `ldap` credential type. The **bind DN is not stored in `suseCredentials.username`** because that column is `VARCHAR(64)` (verified) and cannot hold long directory DNs; the DN lives on `suseLdapAuthServer` instead (reviewer input), leaving `suseCredentials` to hold only the secret. A future migration to salted or encrypted credential storage is orthogonal to this feature and should not change the LDAP integration interface.
+
+## Transport security
+
+Production deployments use `LDAPS` or `STARTTLS`; `PLAIN` only for explicit dev/test. Directory CA certificates are uploaded through the LDAP setup UI/API (write-only upload, same interaction pattern as Hub peripheral CA upload in **Admin → Hub Configuration → Add Peripheral**). Uyuni does not yet provide a generic CA-storage service, so v1 implements LDAP-specific CA handling rather than reusing a shared component. Uploaded CAs are installed into the server trust store used for LDAP connections.
```
rjmateus: Where should we collect it? If the connection of ldaps is starting from JAVA side we may need to save the CA on uyuni to be used when configuring the connection. I don't think Tomcat has access to java `cacert` to add the certificate (and we don't want users to mess around with that manually).

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+| --- | --- |
+| Active session, LDAP goes down | Unaffected; LDAP checked only at login |
+| `LOCAL` user, LDAP down | Unaffected; routed to local DB only |
+| `LDAP` user, LDAP unreachable | Fail after timeout; generic error message |
+| `LDAP` user, wrong password | Reject; no fallback to local or PAM |
+| Unknown user, all LDAP servers fail | Generic login failure |
+| User search returns 0 or >1 entries | Authentication failure; **log details** (filter, base DN, result count) for the administrator |
+| Group lookup fails after successful bind | Authenticate; no LDAP-derived roles; **log failure with details** |
+| User maps to no Uyuni roles | Login succeeds with no LDAP-derived roles |
+| Same login exists as `LOCAL` user | LDAP JIT blocked; `auth_type` on existing row wins |
```
rjmateus: Not only local users, but any user with any login method. If the `username` (login) already exists in the database with any other method, we need to block the JIT. We cannot have conflicting users in the database.
Another question is, what if more than one LDAP server is configured and the same username exists in more than one LDAP server? Should we consider that the first to log in gets the access, and the next one should get an authentication error.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+- **LDAP handling *only* inside `UserImpl.authenticate()`** — cannot JIT-provision or map groups; rejected in favor of `LoginController` orchestration. (The credential *check* is still delegated from `UserManager.loginUser`/`UserImpl.authenticate` so XML-RPC login works; provisioning/mapping stay in `LoginHelper`.)
+- **Local → LDAP → PAM cascade** — tries multiple backends per login; rejected as neither robust nor performant.
+- **Implementation choices** — JNDI vs. UnboundID SDK[^9]; bind password in `suseCredentials` vs. plain column (latter rejected).
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+1. **Marking LDAP-derived RBAC access-group membership** — RBAC access-group membership has no temporary/externally-managed flag today, unlike legacy roles (`temporary='Y'` on `rhnUserGroupMembers`). To recompute LDAP-derived access groups on each login without removing manual grants, do we (a) add a temporary/externally-managed flag to access-group membership, or (b) track LDAP-derived memberships via the ext-group mapping origin and reconcile from that? (Raised by reviewer; RBAC-owner decision.)
```
rjmateus: That is one question that I had in my mind when I was reading the RFC. Should we give an option to control all group assignments on Uyuni side and LDAP side, and merge them together? Or should we allow managing from one side only (and allow configuring it).
Meaning, if Uyuni is configured to manage groups locally, we ignore the LDAP ones. If we configure it to manage groups from LDAP we ignore the local ones (or block them), and all group assignment is managed from LDAP.
This would avoid merging groups from both sides.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+2. **JIT over the XML-RPC path** — an LDAP user who has only ever used the XML-RPC API is not yet provisioned (provisioning runs in the `LoginHelper` lifecycle). Is "must log in once via Web UI/HTTP API to be provisioned" acceptable for v1, or should JIT also run from `UserManager.loginUser`?
+3. **Superuser reachability** — may an LDAP group grant `satellite_admin` (top-level admin), or remain local-only?
```
rjmateus: I think we should keep sat_admin and org_admin as special local users only.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,419 @@
...
+# Alternatives
+[alternatives]: #alternatives
+
+- **Stay on PAM/SSSD only** — no in-product configuration or group mapping. Does not meet the goal.
+- **Improve only `REMOTE_USER`** — auth stays outside Uyuni; no in-product visibility[^4].
+- **LDAP handling *only* inside `UserImpl.authenticate()`** — cannot JIT-provision or map groups; rejected in favor of `LoginController` orchestration. (The credential *check* is still delegated from `UserManager.loginUser`/`UserImpl.authenticate` so XML-RPC login works; provisioning/mapping stay in `LoginHelper`.)
+- **Local → LDAP → PAM cascade** — tries multiple backends per login; rejected as neither robust nor performant.
+- **Implementation choices** — JNDI vs. UnboundID SDK[^9]; bind password in `suseCredentials` vs. plain column (latter rejected).
+
+# Unresolved questions
```
rjmateus, line 398: I think we are missing one important topic in this RFC. AD users normally authenticate using kerberos, and I don't see any reference to it's authentication integration in here.

rjmateus (commented):
last step of the review

accepted/00000-native-ldap-authentication.md
SURYAS1306: Thanks. Leaving the RBAC / read-only API side for @cbbayburt as tagged.

From the LDAP side, the open question for me is: for v1, should XML-RPC login go through the same orchestration as the Web path (so behavior, including JIT provisioning, stays consistent), or is "provision via the Web/admin first, then XML-RPC only authenticates existing users" sufficient?

Open to updating the RFC accordingly.

accepted/00000-native-ldap-authentication.md
SURYAS1306: makes sense

For API-only users I see two options in the RFC: (1) admin creates the user first, or (2) the same login orchestration runs on the XML-RPC path so JIT/group sync also apply there.

For v1, would you prefer sticking with "at least one UI login / admin create" and treat the unified XML-RPC path as a future improvement?

accepted/00000-native-ldap-authentication.md
SURYAS1306: Thanks - the warning about not changing org after creation is clear (lots of access depends on it).

Open for discussion: is v1 only `default_org_id`, with LDAP-driven org (e.g. `uyuni_org` / attribute) later? And do we want the RFC to state explicitly that org assignment happens only at create time and is never refreshed from LDAP afterward?

accepted/00000-native-ldap-authentication.md
SURYAS1306: For now, it sounds reasonable to keep `org_admin` / `satellite_admin` managed only on the Uyuni side for the first implementation. I'll wait for @admd and @cbbayburt's feedback before updating the RFC.

accepted/00000-native-ldap-authentication.md
SURYAS1306: Clarification: the intent is separate columns on `suseLdapAuthServer` (user base DN, filter, login attribute, group base DN, filter, group-name attribute, etc.), not one blob field. I can make that more explicit in the DB section if that helps.

accepted/00000-native-ldap-authentication.md
SURYAS1306: Good point on where we collect the CA and how Java uses it without writing the JVM `cacerts` (and without admins editing trust stores by hand).

I'd like your input before locking the RFC — any preferred pattern on the Uyuni side?

accepted/00000-native-ldap-authentication.md
SURYAS1306: On the first point: blocking JIT when the login already exists under any auth method (not only LOCAL) looks correct—we shouldn’t create conflicting users. I’ll update the RFC wording that way once we’re aligned.

For the same username across multiple LDAP servers, this is still open. Options I see:

- First successful login "owns" the Uyuni account; later logins via another server fail.
- Bind permanently to the first matching server by priority.
- Reject if more than one server matches the same login.

Any preference?

accepted/00000-native-ldap-authentication.md
SURYAS1306: Makes sense.
Open decision: keep temporary LDAP-derived roles alongside permanent local roles (closer to today’s REMOTE_USER model), or have a mode where groups are either fully local-managed or fully LDAP-managed?

Open to updating the RFC based on the preferred direction.

accepted/00000-native-ldap-authentication.md
rjmateus: I think that for the first version is fair to have the JIT on the webUI and then XML-RPC only authenticates, as you have suggested. Thank you

accepted/00000-native-ldap-authentication.md
SURYAS1306: Makes sense as a first-implementation direction. I’ll wait for any further input from @admd / @cbbayburt before updating the RFC.

accepted/00000-native-ldap-authentication.md
SURYAS1306: Kerberos isn't covered in the RFC yet.

Before updating the RFC, I'd like to confirm whether Kerberos/SSO should be considered part of this feature or remain out of scope.

accepted/00000-native-ldap-authentication.md
rjmateus: makes sense to me. Thank you

accepted/00000-native-ldap-authentication.md
rjmateus: >v1 only default_org_id, with LDAP-driven org (e.g. uyuni_org / attribute) later?

makes sense to me. I think the enhancement can go to a section for future work, so we can keep track of them.

accepted/00000-native-ldap-authentication.md
rjmateus: ok, I think would be better to have some clarification on it. Thank you

accepted/00000-native-ldap-authentication.md
rjmateus: I think that at the RFC we can state that the public CA is collected and stored on the uyuni database (maybe on the LDAP connection table).
Then how to use it can be an implementation detail

accepted/00000-native-ldap-authentication.md
SURYAS1306: Thanks - updated the RFC to a fixed `uyuni_` prefix for v1 (not configurable).

accepted/00000-native-ldap-authentication.md
SURYAS1306: Makes sense. RFC now says JIT stays on the Web UI path for v1, and XML-RPC only authenticates users that already exist.

accepted/00000-native-ldap-authentication.md
SURYAS1306: Agreed. Left v1 on `default_org_id`, org only at create time, and moved LDAP-driven org into Future work.

accepted/00000-native-ldap-authentication.md
SURYAS1306: Clarified in the DB section - separate columns on `suseLdapAuthServer`, not one field.

accepted/00000-native-ldap-authentication.md
SURYAS1306: Updated that part: store the public CA in the LDAP connection table in the DB, and leave how Java uses it as an implementation detail.

accepted/00000-native-ldap-authentication.md
admd: @rjmateus I am not sure about this point. Can you please elaborate it a bit? Why do we need to strict this to specific roles prefixes? I mean an LDAP server admin is free to choose any roles group and he should be free to map that to any role at uyuni side? what am i missing here?

accepted/00000-native-ldap-authentication.md
admd: I get the hesitation but we need to solve this problem differently like in case of lockouts how to get user back.

Adding this exception creats needless security compliance risk. So we should definitely allow `org_admin` to be controlled from LDAP. 

BTW Ricardo, do we even allow multiple satellite_admin ? I am not sure if we do, but if it's possible, maybe we should keep only user as a backup which could be used in special cases to recover but rest of the users should be allowed to be mapped from LDAP.

accepted/00000-native-ldap-authentication.md
admd: We should do this a bit differently. If a role can be assigned to multiple users e.g org-admins, we could kinda allow to have one local user (more as a backup to be used in exceptional situations and even there I would prefer to limit this to only super users like sat_admin) but then we should allow to control rest of the users from LDAP.  Related comment https://github.com/uyuni-project/uyuni-rfc/pull/116#discussion_r3747626412

accepted/00000-native-ldap-authentication.md
admd: Good point Ricardo but that will go beyond the scope of this project. We will need to handle it but maybe as a separate step.

accepted/00000-native-ldap-authentication.md
```diff
@@ -0,0 +1,439 @@
...
+2. **Superuser reachability** — may an LDAP group grant `org_admin` / `satellite_admin`, or remain local-only for v1? (Reviewer preference: local-only; waiting on tagged mentors.)
+3. **Same username on multiple LDAP servers** — if more than one configured server can authenticate the same login, should the first successful login own the Uyuni account (later servers fail), bind permanently to the first matching server by priority, or reject when more than one server matches?
+4. **Kerberos / SSO for Active Directory** — AD environments often use Kerberos. Should Kerberos/SSO be in scope for this feature, or remain out of scope for v1 (password bind only)?
+
+**Settled assumptions (reviewer input):**
+- **LDAP server per user** — `rhnUserInfo` stores `ldap_server_id` (which directory authenticated/provisioned the user). *Confirmed by mentor.*
```
admd, line 407: @SURYAS1306 did you consider this point https://github.com/uyuni-project/uyuni-rfc/pull/116#discussion_r3413708421

accepted/00000-native-ldap-authentication.md
rjmateus: My idea with this was to levarage automatic mapping. You would not want to automatically map any group, since LDPA admin may end up with mappings that were not wanted and go unnoticed.
If we always force users to go to UYUNI and manyally map the LDAP group to a UYUNI group, then this prefix would not be needed at all (maybe just for filtering the groups presented).

accepted/00000-native-ldap-authentication.md
rjmateus: I'm fine on allowing org_admin from LDAP, but remember that it will not only give access to endpoints but also to content.
We can have multiple sat_admin users (it's not super clear on how to add then, by the way :) )
The satd_admin we can also map from the LDAP. But we need to make sure that we have at least one local SAT_ADMIN user, or the customer has a high risk of losing access to the UYUNI in case LDAP explodes.

accepted/00000-native-ldap-authentication.md
rjmateus: I think we should allow the user to decide where the users are coming from. If they want to miss local and ldap ones, it's up to them. 
In my opinion, we should focus on making sure they don't lose complete access to UYUNI server. And for that, I think we must have a rule in place that forces the existance of at least one local sat_admin user.

accepted/00000-native-ldap-authentication.md
rjmateus: Got it Abid.

PR #117, opened 2026-06-08
RFC: Feature Toggle system

denyskon:
## Summary

This RFC proposes a configurable **Feature Toggle** system for Uyuni with 16 global toggles that control sidebar menu visibility, backend URL accessibility, and API endpoints.

## Key Design Points

- **RBAC-based approach** using `java.disabled_namespaces` (comma-separated namespace prefixes) in `rhn.conf`
- **Reuses existing RBAC infrastructure** for all three enforcement layers:
  1. `Access.aclAuthorizedFor()` — checks `DisabledNamespaces.isDisabled()` before SAT_ADMIN bypass (menu visibility)
  2. `UserManager.verifyRoleBasedAccess()` — checks `DisabledNamespaces.isDisabled()` before SAT_ADMIN bypass (API/backend)
  3. `FeatureFilter` servlet filter — blocks URLs matching disabled features via URL patterns
- **16 features** mapped to RBAC namespaces:
  - Autoinstallation → `systems.autoinstallation`
  - Images → `cm`
  - SSM → `systems.ssm`
  - Content Management → `clm`
  - Configuration → `config`
  - Audit → `audit`
  - Patches → `patches`
  - VHMs → `systems.vhms`
  - Hub → `admin.hub`
  - Ansible → `systems.ansible`
  - Paid Channels → `software.paid_channels`
  - Product Migration → `systems.software.migration`
  - PTF → `systems.software.ptf`
  - Subscription Matching → `clm.subscription_matching`
  - Remote Commands → `salt.remote_commands`
  - Support → `systems.details.support`

## Implementation

- `DisabledNamespaces` utility — parses `java.disabled_namespaces`, caches prefixes, checks via hierarchical matching
- `FeatureNamespace` enum — maps features to RBAC namespace prefixes
- `FeatureTogglesAction` + `featuretoggles.jsp` — admin UI writing comma-separated string to `rhn.conf`
- `MenuTree.java` and nav XML files — simplified, no longer use `is(java.disable_*)` ACL checks
- `FeatureFilterTest.java` — tests the filter layer

## Files Changed

- `accepted/00000-feature-toggles.md` (to be renumbered before merge)

## Related Work

- Feature namespace enum: `com.suse.manager.webui.menu.FeatureNamespace`
- RBAC utility: `com.redhat.rhn.domain.access.DisabledNamespaces`
- Filter: `com.redhat.rhn.frontend.servlets.FeatureFilter`
- Admin action: `com.redhat.rhn.frontend.action.satellite.FeatureTogglesAction`

## Review Notes

This updated RFC reflects the RBAC-based approach suggested by @cbbayburt in review. The original standalone approach has been replaced with one that leverages the existing RBAC infrastructure, covering both UI and API endpoints automatically.

Please review the detailed design, drawbacks, and alternatives sections. Feedback welcome!

### DIFF
--- a/accepted/00000-feature-toggles.md
+++ b/accepted/00000-feature-toggles.md
@@ -230,11 +230,11 @@ admins view and modify feature states.
    without a namespace, it cannot be toggled. This requires discipline when
    adding new RBAC namespaces.
 
-3. **SAT_ADMIN bypass change** — Placing the disabled check **before** the
-   SAT_ADMIN bypass changes the traditional behavior where system admins could
-   access everything. This is intentional (disabled features are truly disabled)
-   but may surprise operators who expect admin access to override all
-   restrictions.
+3. **SAT_ADMIN bypass change** — System administrators traditionally bypass
+   all RBAC checks. By placing the disabled check **before** the SAT_ADMIN bypass,
+   we ensure that disabled features are truly unavailable to everyone. This is
+   intentional (a disabled feature is a hard system-wide off-switch) but may
+   surprise operators who expect admin access to override all restrictions.
 
 4. **Toggle accumulation** — Each new feature added to Uyuni should ideally get
    a toggle. Without enforcement (e.g., a CI check), toggles will gradually
@@ -255,26 +255,22 @@ admins view and modify feature states.
    - *Verdict*: Rejected — the reviewer pointed out the significant overlap
      with existing RBAC infrastructure
 
-2. **Positive-config naming** (`java.enable_*` instead of `java.disable_*`)
-   - *Pros*: More intuitive
-   - *Cons*: Inconsistent with existing toggles; would require migration
-   - *Verdict*: Rejected — we use `java.disabled_namespaces` which is
-     semantically clear
-
-3. **Database-backed toggles** (new `suseFeatureToggles` table)
+2. **Database-backed toggles** (new `suseFeatureToggles` table)
    - *Pros*: Easier to query programmatically; could support per-org granularity
    - *Cons*: Adds schema migration overhead; `rhn.conf` is simpler
    - *Verdict*: Rejected — `rhn.conf` is the established admin-config pattern
 
-4. **Environment variables** (`UYUNI_DISABLE_IMAGES=1`)
-   - *Pros*: Works well for containerized deployments
-   - *Cons*: Requires restart; no admin UI
-   - *Verdict*: Rejected — restart requirement is unacceptable
+3. **Per-organization granularity** (DB table with `org_id`)
+   - *Pros*: Different orgs can have different features enabled
+   - *Cons*: Significantly more complex UI and backend; most deployments
+     want global toggles
+   - *Verdict*: Deferred — may be added as a future enhancement if needed
 
-5. **Feature branches** (Git branch per feature)
-   - *Pros*: Clean separation
-   - *Cons*: Does not solve runtime configuration
-   - *Verdict*: Rejected — branching is a development practice, not runtime config
+4. **Environment variables** (`UYUNI_DISABLE_FEATURES=images,patches`)
+   - *Pros*: Works well for containerized deployments
+   - *Cons*: Requires restart; no admin UI; inconsistent with existing config
+   - *Verdict*: Rejected — restart requirement is unacceptable for a toggle
+     system meant to be changed at runtime
 
 # Unresolved questions
 [unresolved]: #unresolved-questions
--- a/accepted/00000-feature-toggles.md
+++ b/accepted/00000-feature-toggles.md
@@ -1,0 +1,296 @@
+- Feature Name: feature_toggles
+- Start Date: 2026-06-08
+- RFC PR:
+
+# Summary
+[summary]: #summary
+
+This RFC proposes a configurable **Feature Toggle** system for Uyuni that allows
+satellite administrators to selectively disable UI features and their associated
+backend endpoints globally. Features are controlled via a single
+`java.disabled_namespaces` configuration key in `rhn.conf` that lists RBAC
+namespace prefixes to disable. The system leverages the existing RBAC
+infrastructure for enforcement: the `DisabledNamespaces` utility checks each
+namespace at the existing RBAC gate points (`Access.aclAuthorizedFor()`,
+`UserManager.verifyRoleBasedAccess()`), and the `FeatureFilter` servlet filter
+blocks URLs matching disabled features. The system is managed via an
+administrative UI page that persists changes through
+`ConfigureSatelliteCommand`.
+
+# Motivation
+[motivation]: #motivation
+
+Uyuni supports a broad and growing set of capabilities — content lifecycle
+management, system provisioning, auditing, hub federation, Ansible automation,
+virtual host management, and many more. Not all deployments need or want every
+feature:
+
+- **SLE-only sites** have no use for patch management on non-SUSE systems and
+  may want to hide the "Patches" menu entirely.
+- **Minimal-footprint deployments** (e.g., edge gateways) never use Kickstart or
+  PTF (Program Temporary Fix) workflows and prefer a cleaner UI.
+- **Multi-tenant or peripheral servers** acting as Uyuni Hub clients should not
+  expose Hub Configuration or Subscription Matching menus locally.
+- **Security-hardened environments** want to reduce attack surface by completely
+  disabling endpoints rather than merely hiding menu items.
+
+Today, Uyuni already has *ad-hoc* toggles (`java.disable_remote_commands_from_ui`,
+`java.disable_supportdata_upload`), but they are scattered, inconsistently
+enforced, and lack a unified management surface. Administrators must edit
+`rhn.conf` manually, restart services, and hope the change took effect.
+
+A **unified Feature Toggle system** gives operators explicit, reversible control
+over which features are exposed. It reduces UI clutter, lowers the attack
+surface, and makes per-deployment customization a first-class concern rather than
+a configuration hack.
+
+# Detailed design
+[design]: #detailed-design
+
+## Overview
+
+The system uses a **single configuration key** `java.disabled_namespaces` in
+`rhn.conf` that holds a comma-separated list of RBAC namespace prefixes. When a
+namespace prefix is listed, **all** namespaces that match or start with that
+prefix are disabled. For example, `java.disabled_namespaces = cm,patches`
+disables `cm`, `cm.image.list`, `cm.build`, `patches`, and `patches.security`.
+
+Enforcement happens in **three layers** that reuse the existing RBAC
+infrastructure:
+
+1. **URL blocking** (`FeatureFilter` servlet filter) — intercepts every request
+   to a disabled feature's URL patterns and returns HTTP 403 before any
+   controller or handler runs.
+2. **Menu visibility** (`MenuTree.java` via `isUserAuthorizedFor()`) — calls
+   `Access.aclAuthorizedFor()` which now checks `DisabledNamespaces.isDisabled()`
+   **before** the SAT_ADMIN bypass. Disabled items are hidden from the sidebar.
+3. **API/backend blocking** (`UserManager.verifyRoleBasedAccess()`) — checks
+   `DisabledNamespaces.isDisabled()` **before** the SAT_ADMIN bypass. API calls
+   to disabled features are rejected.
+
+```
+┌─────────────────────────────────────────────────────────────────┐
+│                        User Request                              │
+└────────────────────────┬────────────────────────────────────────┘
+                         │
+                         ▼
+┌─────────────────────────────────────────────────────────────────┐
+│  1. FeatureFilter (servlet filter)                               │
+│     • Checks request path against disabled feature URL patterns   │
+│     • If match → HTTP 403 (before auth / controller)            │
+│     • Excludes login, logout, and admin toggle page             │
+└────────────────────────┬────────────────────────────────────────┘
+                         │ (passthrough)
+                         ▼
+┌─────────────────────────────────────────────────────────────────┐
+│  2. RBAC Gate Points (reused from existing RBAC)                │
+│     • Access.aclAuthorizedFor() → checks DisabledNamespaces     │
+│     • UserManager.verifyRoleBasedAccess() → checks              │
+│       DisabledNamespaces                                        │
+│     • Both run BEFORE the SAT_ADMIN bypass                      │
+│     • Disabled features are blocked for ALL users               │
+└────────────────────────┬────────────────────────────────────────┘
+                         │ (rendered page / API response)
+                         ▼
+┌─────────────────────────────────────────────────────────────────┐
+│  3. Admin UI (FeatureTogglesAction)                              │
+│     • Satellite admin toggles checkboxes → form POST              │
+│     • ConfigureSatelliteCommand → writes java.disabled_namespaces │
+│     • Updates /etc/rhn/rhn.conf in the running container         │
+└─────────────────────────────────────────────────────────────────┘
+```
+
+## Feature → RBAC Namespace Mapping
+
+A `FeatureNamespace` enum maps each feature to its RBAC namespace prefix:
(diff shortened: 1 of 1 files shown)


denyskon:
This RFC and the corresponding tested implementation was designed by OpenCode/GLM-5.1 according to a feature request from my team. As someone without uyuni development experience I looked through the implementation extensively, however I have only a limited chance of really understanding all the consequences. Extensive feedback is thus welcome.

denyskon:
Corresponding feature request: https://github.com/uyuni-project/uyuni/issues/11974

cbbayburt:
Hi @denyskon, thank you for the well-written RFC.

At first glance, I notice some significant overlap with the recent RBAC feature in the backend design.

Namely;

- A servlet filter that filters endpoints and returns 403s
- Conditional rendering of the sidebar and arbitrary pages/sections using ACLs

See [RFC: Role Based Access Control](https://github.com/uyuni-project/uyuni-rfc/blob/rbac/accepted/00000-rbac.md)

From that perspective, the idea of this implementation is identical to RBAC, but system-wide instead of per user.

Another point is that the design does not cover the APIs. Almost every feature has an API module that provides the same functionality, and these must be filtered as well.

I think a good pragmatic approach would be to utilize the existing RBAC backend instead, and add the "feature toggle" UI on top of it.

RBAC already groups the endpoints per feature as RBAC namespaces. In theory, you could revoke access to `images.*` namespace for every user to effectively toggle this feature off. The "feature toggle" UI can operate on this logic to provide system-wide toggles for many features, including their API extensions.

The simplest example would be to add a "list of namespaces to be disabled" in `rhn.conf` and override the RBAC state to hide and reject access to the endpoints that are part of this list.

This approach would help reduce both the amount of work required and the amount of redundant backend logic.

However, there are a few caveats with this approach:

 - Even though the backend for conditional rendering (sidebar MenuTree, ACL for JSP pages) is implemented, it takes a significant effort to review every page (or individual sections of a page) and write ACLs for them, so this is still in progress. Nevertheless, I think it's much more efficient to continue this effort instead of implementing a parallel rendering logic for the system-wide toggles.
  - RBAC is completely bypassed for "system administrator" so if we want to disable the feature completely including for system admins, we'll need to make exceptions to this logic.

denyskon:
@cbbayburt I asked OpenCode to take a look at the RBAC and try to implement is using the existing infrastructure, which seems doable without major issues. Updates the RFC. I would appreciate another look :)

denyskon:
@cbbayburt Another look would be really helpful as I currently have some more capacity to actually implement the proposed plan.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+| `PATCHES` | `patches` | Patches |
+| `VIRTUAL_HOST_MANAGERS` | `systems.vhms` | Virtual Host Managers |
+| `HUB` | `admin.hub` | Hub |
+| `ANSIBLE` | `systems.ansible` | Ansible |
+| `PAID_CHANNELS` | `software.paid_channels` | Paid Software Channels |
+| `PRODUCT_MIGRATION` | `systems.software.migration` | Product Migration |
+| `PTF` | `systems.software.ptf` | PTF |
+| `SUBSCRIPTION_MATCHING` | `clm.subscription_matching` | Subscription Matching |
+| `REMOTE_COMMANDS` | `salt.remote_commands` | Remote Commands |
+| `SUPPORT` | `systems.details.support` | Support Data Upload |
```
cbbayburt, line 124: Most of these selected features are isolated and can be disabled individually, except:

- System Set Manager: which is intertwined with general systems management and other features, so this one cannot be disabled individually
- Patches: which is one of the core features of Uyuni, and might introduce unexpected consequences when disabled

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+
+## DisabledNamespaces utility
+
+`com.redhat.rhn.domain.access.DisabledNamespaces` is a thread-safe utility class:
+
+- Parses `java.disabled_namespaces` config (comma-separated list)
+- Caches the parsed `Set<String>` in a `volatile` field for performance
+- `isDisabled(String namespace)` — returns `true` if the namespace exactly
+  matches or starts with any disabled prefix
+- `refresh()` — re-parses the config after runtime changes
```
cbbayburt, line 137: Missing detail:

The `rhn.conf` is parsed by [com.redhat.rhn.common.conf.ConfigDefaults.java](https://github.com/uyuni-project/uyuni/blob/eed5deeeecd51a4e2a9715c44879c245be53e546/java/core/src/main/java/com/redhat/rhn/common/conf/ConfigDefaults.java). We should keep the parsing logic there.

However, this class doesn't support hot-reloading config values, so it needs a bit more research. Nevertheless, this class should be the only source of configuration.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+
+1. **Toggle expiration** — Should feature toggles have a sunset date? A toggle
+   for a brand-new feature might be removed once the feature is stable.
+
+2. **Audit logging** — Should changes to `java.disabled_namespaces` be logged
+   in the audit trail (who changed what, when)?
+
+3. **Toggle dependencies** — Some features logically depend on others. For
+   example, disabling "Patches" should arguably also disable "System Set
+   Manager → Patches". Currently, toggles are independent.
```
cbbayburt, line 286: Some features have rather obscure dependency chains. I'd keep the feature set to only isolated features to prevent unexpected behavior.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+
+4. **Per-organization granularity** — Should toggles be global or per-org?
+   A future extension could add `org_id` to the toggle state.
+
+5. **API exposure** — Should the toggle state be readable via the XML-RPC or
+   HTTP APIs so that external automation can query which features are enabled?
```
cbbayburt, line 292: I'd keep this out of scope for the initial implementation.

accepted/00000-feature-toggles.md
cbbayburt: Most of these features have at least one corresponding API namespace (`api.*`). Features must be toggled together with their API endpoints. To that end, this map should be 1-Feature-to-N-Namespaces

In relation to this, since you already define a mapping from "feature name" to "list of namespaces", you might want to consider configuring feature toggles in `rhn.conf` by their "feature names" rather than "namespaces". This would make the configuration simpler and prevent arbitrary, potentially nonsensical configurations.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+     waiting for RBAC namespace coverage
+   - *Cons*: Duplicates RBAC logic, doesn't cover APIs, more maintenance
+     burden, inconsistent with project architecture
+   - *Verdict*: Rejected — the reviewer pointed out the significant overlap
+     with existing RBAC infrastructure
+
+2. **Database-backed toggles** (new `suseFeatureToggles` table)
+   - *Pros*: Easier to query programmatically; could support per-org granularity
+   - *Cons*: Adds schema migration overhead; `rhn.conf` is simpler
+   - *Verdict*: Rejected — `rhn.conf` is the established admin-config pattern
```
cbbayburt, line 261: I personally don't have a strong opinion either way, but I think this is still a valid alternative that shouldn't be ruled out. For instance, this approach would allow hot-reload without any extra effort.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+# Motivation
+[motivation]: #motivation
+
+Uyuni supports a broad and growing set of capabilities — content lifecycle
+management, system provisioning, auditing, hub federation, Ansible automation,
+virtual host management, and many more. Not all deployments need or want every
+feature:
+
+- **SLE-only sites** have no use for patch management on non-SUSE systems and
+  may want to hide the "Patches" menu entirely.
```
cbbayburt, line 29: Correction: SUSE systems also provide patch information, so SLE-only sites would require the Patching feature as well.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+- **Minimal-footprint deployments** (e.g., edge gateways) never use Kickstart or
+  PTF (Program Temporary Fix) workflows and prefer a cleaner UI.
+- **Multi-tenant or peripheral servers** acting as Uyuni Hub clients should not
+  expose Hub Configuration or Subscription Matching menus locally.
```
cbbayburt, line 33: I would make the case for "Single-server" deployments (where there is no hub architecture) to disable the Hub feature.

cbbayburt (changes requested):
Thanks for the updates, @denyskon. Below are my next round of comments.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+over which features are exposed. It reduces UI clutter, lowers the attack
+surface, and makes per-deployment customization a first-class concern rather than
+a configuration hack.
+
+# Detailed design
+[design]: #detailed-design
+
+## Overview
+
+The system uses a **single configuration key** `java.disabled_namespaces` in
```
rjmateus, line 52: We want to try to move away from making this kind of changes in files on the server. We should consider have this configurations on the database.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+prefix are disabled. For example, `java.disabled_namespaces = cm,patches`
+disables `cm`, `cm.image.list`, `cm.build`, `patches`, and `patches.security`.
+
+Enforcement happens in **three layers** that reuse the existing RBAC
+infrastructure:
+
+1. **URL blocking** (`FeatureFilter` servlet filter) — intercepts every request
+   to a disabled feature's URL patterns and returns HTTP 403 before any
+   controller or handler runs.
+2. **Menu visibility** (`MenuTree.java` via `isUserAuthorizedFor()`) — calls
```
rjmateus, line 64: This can also be implemented to filter out the features a user don't have access to.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+   `Access.aclAuthorizedFor()` which now checks `DisabledNamespaces.isDisabled()`
+   **before** the SAT_ADMIN bypass. Disabled items are hidden from the sidebar.
+3. **API/backend blocking** (`UserManager.verifyRoleBasedAccess()`) — checks
```
rjmateus, line 67: This can also be controlled already with the RBAC implementation, and we don't need to fully disable the feature.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+│     • Excludes login, logout, and admin toggle page             │
+└────────────────────────┬────────────────────────────────────────┘
+                         │ (passthrough)
+                         ▼
+┌─────────────────────────────────────────────────────────────────┐
+│  2. RBAC Gate Points (reused from existing RBAC)                │
+│     • Access.aclAuthorizedFor() → checks DisabledNamespaces     │
+│     • UserManager.verifyRoleBasedAccess() → checks              │
+│       DisabledNamespaces                                        │
+│     • Both run BEFORE the SAT_ADMIN bypass                      │
```
rjmateus, line 90: So this mean the SAT_ADMIN would also be blocked by this feature disable.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+│     • Disabled features are blocked for ALL users               │
+└────────────────────────┬────────────────────────────────────────┘
+                         │ (rendered page / API response)
+                         ▼
+┌─────────────────────────────────────────────────────────────────┐
+│  3. Admin UI (FeatureTogglesAction)                              │
+│     • Satellite admin toggles checkboxes → form POST              │
+│     • ConfigureSatelliteCommand → writes java.disabled_namespaces │
```
rjmateus, line 98: We are avoiding that the web application to make changes to the rhn.conf. That would be a no-go. Did you consider save this data on the database?

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+`DisabledNamespaces`, disabled features are automatically hidden from the menu.
+
+The old `checkAcl(user, "not is(java.disable_*)")` calls have been removed from
+`MenuTree.java` and the nav XML files (`system_detail.xml`, `ssm.xml`,
+`activation_key.xml`). The menu visibility is now controlled purely by the RBAC
+`isUserAuthorizedFor()` mechanism.
+
+## Admin UI and persistence
+
+A dedicated admin page at `/rhn/admin/config/FeatureToggles.do` lets satellite
```
rjmateus, line 195: We will not implement new features using the old struts approach. New pages must be implemented with react.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+[drawbacks]: #drawbacks
+
+1. **Testing combinatorics** — With 16 toggles, the state space is 2¹⁶ = 65,536
+   combinations. Manual QA of every combination is intractable. We rely on
+   the fact that each toggle is independent (no toggle depends on another) and
+   that the filter, menu, and API layers are tested individually.
+
+2. **RBAC namespace coverage** — The feature toggle system only works for
+   features that have corresponding RBAC namespaces. If a new feature is added
+   without a namespace, it cannot be toggled. This requires discipline when
```
rjmateus, line 230: All features must have a corresponding RBAC namespace or it's unusable.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+
+1. **Standalone feature toggle system** (original approach)
+   - *Pros*: Independent of RBAC rollout state, works immediately without
+     waiting for RBAC namespace coverage
+   - *Cons*: Duplicates RBAC logic, doesn't cover APIs, more maintenance
+     burden, inconsistent with project architecture
+   - *Verdict*: Rejected — the reviewer pointed out the significant overlap
+     with existing RBAC infrastructure
+
+2. **Database-backed toggles** (new `suseFeatureToggles` table)
```
rjmateus, line 258: Saving in the database should be the way to go. Update the rhn.conf is much harder and error prone.

accepted/00000-feature-toggles.md
```diff
@@ -0,0 +1,292 @@
...
+4. **Environment variables** (`UYUNI_DISABLE_FEATURES=images,patches`)
+   - *Pros*: Works well for containerized deployments
+   - *Cons*: Requires restart; no admin UI; inconsistent with existing config
+   - *Verdict*: Rejected — restart requirement is unacceptable for a toggle
+     system meant to be changed at runtime
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+1. **Toggle expiration** — Should feature toggles have a sunset date? A toggle
```
rjmateus, line 278: If the feature is not stable we have 2 options: Do not release it, or release it as tech Preview, and in that case we want to make it available for users to provide feedback

rjmateus (commented):
Overall, the only real change this RFC bring is hiding features from sat_admin user. For regular users that is already possible through the RBAC features.
I don't see a use case where we want to hide/disable features to the super user sat_admin, and for that reason I don't think we should move forward with this changes.

denyskon:
@rjmateus The goal of this feature is to declutter the UI for end users.

- I don't know how the situation is for large organisations, but we have 3-4 people who actually have access to Uyuni and all of them have the same administrative permissions. There is no practical reason to restrict the roles of these users.
- Lots of features are SUSE-only while Uyuni nowadays supports a large set of operating systems. For our Debian-only fleet, it's impractical to have all of them available.
- RBAC is really verbose and hard to setup, while a simple toggle for enabling/disabling a feature would simplify it.

For our use case there is a substantial UX improvement in this feature.

rjmateus:
Thank you for explaining you use case. But I would prefer to see the UI evolve in a direction to make it simpler to all users instead of hiding portions of it.

PR #119, opened 2026-06-14
RFC: MQTT event publisher for reactor events

geetxnshgoyal:
Rendered version -> [RFC](https://github.com/geetxnshgoyal/uyuni-rfc/blob/mqtt-event-publisher/accepted/00000-mqtt-event-publisher.md)

## Summary

Adds an MQTT event publisher to the Uyuni reactor pipeline. Uyuni events are published as JSON to a Mosquitto broker, so external tools (Node-RED, Grafana, custom scripts) can subscribe and react in real time without polling the API.

## Motivation

Right now there is no way for external tools to get notified when something happens in Uyuni (minion registers, highstate fails, batch kicks off). The only option is polling the XML-RPC / HTTP API, which adds latency and unnecessary server load. MQTT gives us a lightweight pub/sub channel that decouples consumers from the server entirely.

## Status

Nine event types are published, covering both Salt reactor events and Java-native application events:

| Topic suffix | Source |
|---|---|
| `systems/registered` | `RegisterMinionEventMessage` |
| `jobs/returned` | `JobReturnEventMessage` |
| `states/applied` | `ApplyStatesEventMessage` |
| `images/deployed` | `ImageDeployedEventMessage` |
| `batches/started` | `BatchStartedEventMessage` |
| `users/created` | `UserFactory` |
| `orgs/created` | `OrgFactory` |
| `clm/build_started` | `ContentManager` |
| `clm/build_completed` | `ContentManager` |

Each event is its own class implementing an `MqttEvent` interface that declares only `getTopicSuffix()` and `getPayload()`, so adding a topic means adding a class rather than extending a dispatcher that would otherwise grow without bound. This addresses the review comment about avoiding a single large handler class.

Events published from application code are deferred until the surrounding database transaction commits, so a rollback never leaves an event announcing a change that was discarded.

## Verification

The pipeline has been deployed and exercised end to end against a running Uyuni instance:

- A real Salt job (`test.ping` on a registered minion) produced a `jobs/returned` event published 0.112 s after the Salt result.
- A real Content Lifecycle build produced `clm/build_started`, and correctly produced no completion event for an asynchronous build, where the channel alignment has only been scheduled at that point.
- Broker access control confirmed: anonymous connections refused, wrong passwords refused, a subscriber's attempt to publish blocked, and a publisher unable to read events back.

## Changes since the first review round

All review threads have been resolved and the RFC text updated accordingly:

- **Event set configurability**: now filterable through `uyuni.mqtt.events.enabled`.
- **Buffering vs skip-and-log**: skip-and-log kept, as agreed, to avoid unbounded memory growth. Documented in Failure Modes.
- **Topic prefix**: replaced the static `events` segment with the server FQDN, so several instances can share a broker and consumers can subscribe per instance or across all of them.
- **Authentication**: username and password support added, along with a broker image that disables anonymous access and enforces separate write-only and read-only roles. A new "Broker side" section documents this.
- **Handler class growth**: addressed by the per-event class design described above.
- **QoS and queue bounds**: both configurable, with a bounded queue that discards oldest and logs on exhaustion.

The `users/created` payload now carries `userId` rather than the `creator` field, which was never populated in practice. TLS between publisher and broker remains deferred to a future enhancement.

Feedback welcome on anything that still looks wrong.


accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,551 @@
...
+  and adds load to the database.
+- **Apache Kafka** — Much heavier infrastructure requirement. MQTT is
+  sufficient for the expected event throughput and is simpler to deploy.
+
+# Unresolved questions
+[unresolved]: #unresolved-questions
+
+- Should the set of published event types be configurable (e.g. via a
+  system property or database setting), or is the hardcoded set of five
+  sufficient for the initial release?
```
aaannz: There must be a filter what messages to publish to the broker to prevent overloading it in larger installations.

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,551 @@
...
+- Should the publisher buffer a bounded number of events when the broker
+  is temporarily unavailable, or is the current skip-and-log behavior
+  acceptable?
```
aaannz: Skip and log is fine.

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,551 @@
...
+- Should the MQTT topic prefix (`uyuni/events/`) be configurable to
+  support multi-server environments where each Uyuni instance publishes
+  to a shared broker?
+- Authentication and TLS: should the first version support
+  username/password or TLS client certificates for broker connections,
+  or is unauthenticated localhost-only access sufficient initially?
```
aaannz: Assumption is that broker will not be local. At least username/password should be supported.

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,551 @@
...
+
+- Should the set of published event types be configurable (e.g. via a
+  system property or database setting), or is the hardcoded set of five
+  sufficient for the initial release?
+- Should the publisher buffer a bounded number of events when the broker
+  is temporarily unavailable, or is the current skip-and-log behavior
+  acceptable?
+- Should the MQTT topic prefix (`uyuni/events/`) be configurable to
+  support multi-server environments where each Uyuni instance publishes
+  to a shared broker?
```
aaannz: Personally I would drop the `events` and replace it with the instance fqdn.
So like `uyuni/example.com/`.

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,551 @@
...
+    batch -- No --> log_debug["LOG.debug: Unhandled event type"]
+
+    handle_reg --> publish["MqttPublisherService.publish()"]
+    handle_job --> publish
+    handle_apply --> publish
+    handle_image --> publish
+    handle_batch --> publish
+```
+
+## SaltReactor Integration
```
aaannz, line 366: How about instead having this in the SaltReactor, we handle this in the [RhnServletListener](https://github.com/uyuni-project/uyuni/blob/master/java/core/src/main/java/com/redhat/rhn/webapp/RhnServletListener.java#L124) directly?

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,551 @@
...
+
+This is the same pattern every other handler in `SaltReactor` uses.
+The MQTT action runs alongside the existing handlers — nothing
+changes for current event processing.
+
+## Topic Hierarchy
+
+```mermaid
+graph TD
+    root["uyuni/"] --> events["events/"]
```
aaannz: I suggest to drop `events` and replace it by instance fqdn.

That way we can subscribe to different instances or all via wildcard.

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,551 @@
...
+| `RegisterMinionEventMessage`    | `uyuni/events/systems/registered` | New minion bootstrapped    |
+| `JobReturnEventMessage`         | `uyuni/events/jobs/returned`      | Salt job completed         |
+| `ApplyStatesEventMessage`       | `uyuni/events/states/applied`     | Highstate / state.apply    |
+| `ImageDeployedEventMessage`     | `uyuni/events/images/deployed`    | OS image deployed          |
+| `BatchStartedEventMessage`      | `uyuni/events/batches/started`    | Batch operation started    |
+
+**Payload extraction:** Each handler picks out only the fields that
+make sense for external consumers. For example, `handleRegisterMinion`
+publishes `minionId`, `machineId`, `saltbootInitrd`, and
+`managementKey` — we do not dump the entire event message object.
```
aaannz: In the initial implementation this is fine. But I think this will need to be later refactored to prevent thousands line Java class with different handlers in one class.

Consider that once this matures, we might have tens of different topics.

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,571 @@
+- Feature Name: mqtt_event_publisher
+- Start Date: 2026-06-11
+
+# Summary
+[summary]: #summary
+
+This document describes the addition of an MQTT-based event publishing
+mechanism to the Uyuni server. The reactor pipeline will publish selected
+Salt event types as structured JSON messages to a local Mosquitto MQTT
```
admd: I believe you extended the scope of it and it's not limited to salt event type, right?

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,571 @@
...
+broker, making them available for external consumption by automation tools
+such as Node-RED, Grafana alerting pipelines, or custom scripts.
+
+The first iteration covers five event types that represent the most
```
admd: Is this still the case?

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,571 @@
...
+
+- MQTT is an OASIS standard with broad tooling support.
+- Mosquitto is lightweight (< 1 MB RSS), runs as a single process,
+  and is already available in openSUSE and SLES package repositories.
+- The publish/subscribe model is a natural fit: Uyuni publishes events
+  and any number of consumers can subscribe without coupling.
+- QoS levels (0, 1, 2) allow consumers to choose their delivery
+  guarantee. QoS 1 ("at least once") provides a good balance between
+  reliability and performance for operational events.
+- Eclipse Paho, the reference Java MQTT client, is mature, small, and
```
admd: Add some references to all such points like QoS levels, Eclipse Paho etc. so user can ready more about them easily if needed.

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,571 @@
...
+
+**Key design decisions:**
+
+1. **Asynchronous client.** The reactor processes Salt events on a hot
+   path. A synchronous publish call would block the reactor thread
+   while waiting for a broker ACK. `MqttAsyncClient` returns immediately
+   and handles delivery acknowledgement in the background.
+
+2. **Single-thread executor.** All publish operations are submitted to a
+   single-thread executor. This guarantees message ordering and avoids
```
admd: Whit it guarantee the ordering, I believe this singe-thread might struggle in case of many events getting generated at a single time.

Ordering is important but mostly in the scope of minion so events for the same minion must i stay ordered but overall it should be ok not have a strict ordering across the fleet. So generally, we should be able to relax this ordering requirement a bit.

In anycase, it would be good to add some words about the usecase where this single-thread executre might be slow to react to the batch operations like mass registration events, or a large state.apply where we could have huge number of event. How do we cope with that?

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,571 @@
...
+   contention on the Paho client. The thread is created as a daemon
+   thread so it does not prevent JVM shutdown.
+
+3. **Configurable broker URL.** The default broker URL is
+   `tcp://mosquitto:1883`, matching the container name in the Uyuni
+   containerized deployment. It can be overridden via the JVM system
+   property `uyuni.mqtt.broker.url`.
+
+4. **Graceful degradation.** If the broker is unavailable at startup or
+   disconnects at runtime, events are logged and skipped. The reactor
```
admd, line 264: And how is it going to be when process is just slow, will we even consider queue as full and then rest of events will be ignored? because if we don't have a mechanism to handle that, we might end up filling up the whole memory.

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,571 @@
...
+```
+
+## Why MQTT?
+
+- MQTT is an OASIS standard with broad tooling support.
+- Mosquitto is lightweight (< 1 MB RSS), runs as a single process,
+  and is already available in openSUSE and SLES package repositories.
+- The publish/subscribe model is a natural fit: Uyuni publishes events
+  and any number of consumers can subscribe without coupling.
+- QoS levels (0, 1, 2) allow consumers to choose their delivery
```
admd: Would it possible to make this configurable with good default of course? will we be able to handle it?

accepted/00000-mqtt-event-publisher.md
```diff
@@ -0,0 +1,571 @@
...
+| `stateNames`              | String[] | Names of states applied            |
+| `forcePackageListRefresh` | Boolean  | Whether package list refresh forced |
+| `directCall`              | Boolean  | Whether this was a direct call      |
+
+**`uyuni/<fqdn>/images/deployed`**
+
+| Field       | Type   | Description                        |
+|-------------|--------|------------------------------------|
+| `machineId` | String | Target machine ID                  |
+| `grains`    | Object | Salt grains from the deployed host |
```
admd, line 497: This could be pretty big payload, are we find with whatever payload comes in? Serialization could take quite a bit.

accepted/00000-mqtt-event-publisher.md
admd: +1. This one is critical.

accepted/00000-mqtt-event-publisher.md
geetxnshgoyal: Correct! We have expanded the scope to publish both Salt-based reactor events and Java-native application events (user creation, organization creation, and Content Lifecycle Management build started/completed)

accepted/00000-mqtt-event-publisher.md
geetxnshgoyal: Resolved. I have implemented username and password authentication via the JVM properties `uyuni.mqtt.broker.username` and `uyuni.mqtt.broker.password` (falling back to host/container environment variables `UYUNI_MQTT_BROKER_USERNAME` and `UYUNI_MQTT_BROKER_PASSWORD`). The RFC has been updated to document these parameters.

accepted/00000-mqtt-event-publisher.md
geetxnshgoyal: No, we have expanded the scope to cover nine event types: five Salt Reactor events and four native Java application events (user creation, organization creation, and CLM build started/completed). The RFC summary, topic hierarchy, and payload details have been updated to reflect all nine events

accepted/00000-mqtt-event-publisher.md
geetxnshgoyal: Done. Added direct reference links to the MQTT specification, Eclipse Paho client documentation, Mosquitto broker homepage, and details about MQTT QoS levels in the "Why MQTT?" section of the RFC.

accepted/00000-mqtt-event-publisher.md
geetxnshgoyal: Great point. i've updated the design rationale section in the RFC. The single-thread executor uses an in-memory queue to process publishes asynchronously (< 1 ms per call over local connection) to prevent blocking the SaltReactor.

I also noted that if strict global ordering needs to be relaxed for scaling in the future, I can partition task submissions using a thread pool based on the minion ID, maintaining order within each minion while processing different minions concurrently

accepted/00000-mqtt-event-publisher.md
geetxnshgoyal: Agreed. I have added a task card for this to the project board. I'll add configuration properties for both the QoS level (`uyuni.mqtt.qos`) and a bounded queue limit (`uyuni.mqtt.queue.limit` with an oldest-discard rejection policy) to resolve this and protect host memory.

accepted/00000-mqtt-event-publisher.md
geetxnshgoyal: Since image deployments are relatively rare operational lifecycle events, the transient CPU/memory overhead of serializing the full grains map is negligible.

I have added a performance note about this in the RFC payload description, highlighting that administrators can disable these events via uyuni.mqtt.events.enabled if needed.

accepted/00000-mqtt-event-publisher.md
geetxnshgoyal: Very true, I'll configure a bounded queue (`uyuni.mqtt.queue.limit` defaulting to 10000) that discards the oldest message and logs a warning on exhaustion to prevent out-of-memory errors under broker backpressure.

accepted/00000-mqtt-event-publisher.md
admd: > I can partition task submissions using a thread pool based on the minion ID, maintaining order within each minion while processing different minions concurrently

👍 That would be great.

PR #126, opened 2026-09-15
Add RFC for exploring Generative UI in Uyuni

jordimassaguerpla:
This RFC proposes exploring Generative UI (GenUI) for AI-driven interactions in Uyuni, focusing on a Controlled Component Catalog to dynamically compose UIs from multiple MCP tool outputs. It outlines motivations, design considerations, potential drawbacks, and next steps for implementation.


jordimassaguerpla:
As a reference, here the poster from the undergrad thesis and a video demo:

PosterGenUI-20260910

https://github.com/user-attachments/assets/f7c43436-56d3-4316-80eb-6106b32cf9e7

and the code

https://github.com/jordimassaguerpla/tfg-mcp-ui

cbbayburt (commented):
Thank you Jordi, here's my take on the different options:

1. MCP Apps: This is a cool new way of packaging together views, presentations and visualizations together with the data and operations in the MCP server. In our environment, however, the MCP server sits on the data/operation layer and is only responsible for Uyuni operations/data retrieval. We delegate the whole frontend and presentation layer to Uyuni Web UI. Following a strict separation of concerns, this is probably not a good fit for us.

2. CopilotKit: This could cover a lot of heavylifting for us, however, I realized that our current React dependency is a few major versions behind what CopilotKit supports (16 vs 18).

3. AG-UI: We can benefit from this protocol not just for presentation, but also for event handling of a conversation. I've already established AG-UI as the frontend protocol in the [AI Assistant RFC](https://github.com/uyuni-project/uyuni-rfc/pull/127).

I think overall, this RFC should drill down on AG-UI into more detail, whereas mentioning CopilotKit and other libraries as possible additional options.

We could maybe even add some implementation plans on how to translate AG-UI entities to rendered frontend components in a way suitable in the Uyuni codebase.

Also important to note: Advanced UI functionality and components would be introduced at a quite later stage of development, so we can evaluate different GenUI frameworks as time goes on.

jordimassaguerpla:
Thanks @cbbayburt  for your comments.

I agree that AG-UI should be the protocol to use.

I would propose starting with a limited set of React components, kind of a component catalog, and selecting the UI deterministically from the agent using a predefined set of rules. This would basically be option 1 for the simpler and predictable cases.

Once we have that in place, we could explore a more flexible approach using the same component catalog, but letting the LLM decide the representation. This could cover more complex cases where we cannot predict the UI in advance, for example when combining the results of multiple tools.

We could also evaluate hybrid approaches later, based on what we learn from these experiments.

jordimassaguerpla:
@cbbayburt after reading the rfc on the ui assistant, I have reviewed this rfc to focus on the visuals selections.

